Issue #55, opened 2021-12-02
Component improvements in event block


cpvalente:
We are planning some work on restructuring some of the components. Mostly focusing on the usability of the app rather than adding new functionality

I thought it would be good to start a documentation on this, focusing for now on Event block which is next in line in worklist

- [x] Improve time input workflow, lots of good stuff at #50 and #46
- [ ] Show a small progressbar in running event when collapsed 
- [ ] Add playbutton to event block
- [ ] Add skip action #172 
- [ ] Allow more text to be added in notes (note field is text area) #59 
- [ ] Add a note event: event with a single text field and time without duration. eg: 08:30 mic check.
- [ ] Ability to link event times
- [ ] Visually diferentiate past and upcoming events #125 

and some aesthetic stuff
- [x] inconsistent border radius, see #53 
- [x] use nore space in event list, see #57 

will also have to match styles to the other blocks, this should be mostly styling, but while we are here
- [ ] delay field is maxed at 60, however this correction only happens on submit and not onBlur 
- [ ] When creating an event with the sortcut Alt+E, the option "Start time is last end" is not respected #171 

### Show a small progressbar in running event when collapsed 
User request. In half window mode the app shows only the eventlist. Some people like using this and triggering events remotely. It would be nice to have a clearer indication of which component is running and its progress

### Add playbutton to event block
User request. It would be nice to be able to trigger an event straight from its definition. I am assuming this is for people who use ontime as a stage timer application rather than a rundown list

### Add a note event
User request. One of the oldest things in the list. Some workflows would benefit from having a simpler event type for things without a duration. eg: 08:30 Mic check Michael Fassbender.

### Ability to link event times
We are looking at a possible workflow where the time start of an event could be linked to the end of another.
eg; event two would have as start time a link to its previous event and a duration of 30min. 
This could significantly complicate the usability and clutter the interface with advanced features if not approached correctly

xztraz:
add the text "delay" to delay-box.  so it says    "$$ minutes delay" 

![image](https://user-images.githubusercontent.com/25690871/146647325-7f412bbc-8612-4efe-8b0c-13dc847a877f.png)

xztraz:
Ability to link times, could in gui be as simple as just having a default end-time as the text "next" or a minus sign or alike. and if pushing it it assumes you want to enter something. no clutter.
![image](https://user-images.githubusercontent.com/25690871/146647672-87caea2b-2624-49ed-9694-d6daf7014427.png)

[...]

PR #186, opened 2022-07-30
Feat/161

cpvalente:
This should offer an MVP for showing times in 12 hour format.
The user can decide this in the settings menu via the dropdown or override locally in the viewers by adding format=12 or format=24 to the URL parameters

time entry and in app times are all in 24 hours. will not redesign since the components are under review with #55


client/src/features/viewers/countdown/Countdown.jsx
```diff
@@ -13,19 +14,20 @@ import { fetchTimerData, sanitiseTitle, timerMessages } from './countdown.helper
...
 
 export default function Countdown(props) {
-  const [searchParams] = useSearchParams();
-  const { backstageEvents, time, selectedId } = props;
+  const { backstageEvents, time, selectedId, settings } = props;
   const [follow, setFollow] = useState(null);
   const [runningTimer, setRunningTimer] = useState(0);
   const [runningMessage, setRunningMessage] = useState('');
   const [delay, setDelay] = useState(0);
+  const [localTimeFormat, setLocalTimeFormat] = useState(null);
```
alexrohleder, line 22: I think this can be removed. The `clock`, `endTime` and `startTime` seems too verbose. Maybe we can rewrite it like this?

```js
const timeFormat = searchParams.get("format") || settings.timeFormat;

const clock = formatDate(time.clock, timeFormat);
const startTime = formatDate(follow.timeStart + delay, timeFormat);
const endTime = formatDate(follow.timeEnd + delay, timeFormat);
```

No need for hooks

client/src/features/viewers/foh/Public.jsx
```diff
@@ -11,19 +12,47 @@ import { titleVariants } from '../common/animation';
 import style from './Public.module.scss';
 
 export default function Public(props) {
-  const { publ, publicTitle, time, events, publicSelectedId, general } = props;
+  const { publ, publicTitle, time, events, publicSelectedId, general, settings } = props;
   const [pageNumber, setPageNumber] = useState(0);
   const [currentPage, setCurrentPage] = useState(0);
+  const [localTimeFormat, setLocalTimeFormat] = useState(null);
```
alexrohleder, line 18: The same as for the `Clock.tsx` component.

client/src/features/viewers/foh/Public.jsx
```diff
@@ -92,6 +121,7 @@ export default function Public(props) {
           isBackstage
           setCurrentPage={setCurrentPage}
           setPageNumber={setPageNumber}
+          format12={format12}
```
alexrohleder, line 124: Maybe instead of passing `format12` down as prop, and doing many `if` statements down the line, we could set up `settings.timeFormat` taking in account the query string too?

client/src/features/viewers/ViewWrapper.jsx
```diff
@@ -222,8 +237,10 @@ const withSocket = (Component) => {
       ...timer,
       finished: playback === 'start' && timer.isNegative && timer.startedAt,
       clock: stringFromMillis(timer.clock),
+      clock12: format(new Date(timer.clock - hoursToMilliseconds(1)), 'hh:mm:ss a'),
       clockMs: timer.clock,
       clockNoSeconds: stringFromMillis(timer.clock, false),
+      clock12NoSeconds:  format(new Date(timer.clock - hoursToMilliseconds(1)), 'hh:mm'),
```
alexrohleder, line 243: I think we should not have this here. We should pass down only the `timer` and format it at the very point it is used in a given format.

alexrohleder (commented):
In general, I think we should aim in having a smaller centralized solution for this problem. It seems we're solving it multiple times across the app. We can discuss ways of doing that, please let me know if you have any doubt.

client/src/features/viewers/foh/Public.jsx
cpvalente: I am not entirely sure what do you mean here, perhaps we could take a look on our next chat

client/src/features/modals/AppSettingsModal.jsx
```diff
@@ -95,33 +105,27 @@ export default function AppSettingsModal() {
...
-      doStarTimeIsLastEnd,
-      emitError,
-      emitWarning,
-      formData,
-      refetch,
-      setDefaultPublic,
-      setShowQuickEntry,
-      setStarTimeIsLastEnd,
-    ]
+    [data.pinCode, data.timeFormat, doDefaultPublic, doShowQuickEntry, doStarTimeIsLastEnd, emitError, emitWarning, formData, refetch, setDefaultPublic, setShowQuickEntry, setStarTimeIsLastEnd]
```
alexrohleder, line 128: There is no need for wrapping this event handler in a `useCallback` because you're passing it to an HTML element, therefore it will not be expensive if it needs to "rerender".

I guess in this case the complexity is not worth the gains.

alexrohleder: The same goes for the other event handlers in this component.

`handleContextChange` for example, is being called inside an arrow function that is re-created on every render. Therefore having `handleContextChange` inside a `useCallback` does nothing. The same goes for `handleChange`.

The `revert` is passed to a component that re-renders if this is not a stable reference, but I think the component is so simple it makes no difference performance-wise.

client/src/features/modals/AppSettingsModal.jsx
```diff
@@ -95,33 +105,27 @@ export default function AppSettingsModal() {
...
       }
+
+      // set fields with error
+      if (!e.status) {
+        emitError(`Invalid Input: ${e.message}`);
+      } else {
+        await postSettings(formData);
+        await refetch();
+        emitWarning(e.message);
+        setChanged(false);
```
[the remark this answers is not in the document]
alexrohleder: And there are simpler and more reliable ways of checking if a form changed or not than controlling it via React state but via native DOM APIs. I can show you how that would be later today :)

client/src/features/table/TableHeader.jsx
```diff
@@ -3,20 +3,21 @@ import { Tooltip } from '@chakra-ui/tooltip';
...
 import { formatDisplay } from '../../common/utils/dateConfig';
-import { stringFromMillis } from '../../common/utils/time';
+import { formatTime } from '../../common/utils/time';
 
 import PlaybackIcon from './tableElements/PlaybackIcon';
 
 import style from './Table.module.scss';
 
-export default function TableHeader() {
+export default function TableHeader({ timeFormat = '24' }) {
```
alexrohleder, line 20: I think we can access this `timeFormat` in different ways than passing it down everywhere. It is inside your query client; we could access it from the `formatTime` function and potentially add the logic to consider query strings there.

cpvalente:
Hey Alex, really good input here.
I have made some changes to the state and centralised the time formatting.

The PR is becoming pretty big so I believe its best to leave some of the refactorings for separate PRs. 
For now I will create branches for refactoring forms, we will get to the rest eventually

Either way, it would be great to have some feedback on the work here. Thank you again, good stuff

client/src/features/editors/list/EventList.jsx
```diff
@@ -19,21 +20,24 @@ export default function EventList(props) {
...
-    } else {
-      const previousEvent = events[cursor];
-      const nextEvent = events[cursor + 1];
-      if (type === 'event') {
-        eventsHandler('add', { type: type, after: previousEvent.id });
-      } else if (previousEvent?.type !== type && nextEvent?.type !== type) {
-        eventsHandler('add', { type: type, after: previousEvent.id });
+  const showQuickEntry = useAtomValue(showQuickEntryAtom);
+
+  const insertAtCursor = useCallback(
```
alexrohleder, line 25: I know this is not part of the PR, but it seems we don't need `useCallback` in this component.

alexrohleder (approved):

client/src/features/editors/list/EventList.jsx
cpvalente: Removing the callback here would mean changing the code structure so not to break the next callback along `handleKeyPress`
This component will come under scrutiny in #55 and I will revise then

PR #188, opened 2022-08-02
Prefeat/skip

cpvalente:
This PR is setting up ground for skipping events
It refactors the logic around mutation of events and adds parsing to skipped events from both excel and json

No features are exposed in UI just yet (upcoming with #55 )


server/src/utils/parser.js
```diff
@@ -279,6 +295,7 @@ export const validateEvent_v1 = (eventArgs) => {
       timeType: 'start-end',
       duration: validateDuration(start, end),
       isPublic: e.isPublic != null && typeof e.isPublic === 'boolean' ? e.isPublic : d.isPublic,
+      skip: e.skip != null && typeof e.skip === 'boolean' ? e.skip : d.skip,
```
alexrohleder, line 298: if the `typeof e.skip` is boolean then it will always be different than null.

server/src/classes/__tests__/eventtimer.test.js
```diff
@@ -68,8 +68,8 @@ test('object instantiates correctly', async () => {
   expect(t.nextEventId).toBeNull();
   expect(t.selectedPublicEventId).toBeNull();
   expect(t.nextPublicEventId).toBeNull();
-  expect(t.numEvents).toBe(0);
```
alexrohleder, line 71: Should we also remove it from the source code if we're not testing `numEvents` anymore? I don't see other mentions of this in the PR.

cpvalente: We no longer expose numEvents as a class property, but it is useful to know this for testing some of the behaviours.
Unless I am misunderstanding the suggestion, I believe this is a good assertion

PR #193, opened 2022-08-14
Refact/feature sockets

cpvalente:
This PR is setting ground for upcoming features in the interface.
It should help with separation of concerns between features and to resolve issues with multiple listeners on sockets.


client/src/common/components/errorBoundary/ErrorBoundary.jsx
```diff
@@ -54,7 +54,7 @@ class ErrorBoundary extends React.Component {
             <p
               className={style.report}
               onClick={() => {
-                if (window.process.type === 'renderer') {
+                if (window?.process?.type === 'renderer') {
```
alexrohleder, line 57: I don't think we need these `?`

client/src/common/hooks/useSocketProvider.js
```diff
@@ -0,0 +1,239 @@
...
+  TIMER,
+} from '../api/apiConstants';
+import { useSocket } from '../context/socketContext';
+
+export const useEventListProvider = () => {
+  const { data } = useQuery(FEAT_EVENTLIST, () => undefined, {
+    cacheTime: Infinity,
+    staleTime: Infinity,
+  });
+  const placeholder = useMemo(
```
alexrohleder, line 19: if you want to have the same object always, you could just declare the placeholder outside of the hook. No need for `useMemo` here.

client/src/common/hooks/useSocketProvider.js
```diff
@@ -0,0 +1,239 @@
...
+  return data ?? placeholder;
+};
+
+export const useMessageControlProvider = () => {
+  const socket = useSocket();
+  const { data } = useQuery(FEAT_MESSAGECONTROL, () => undefined, {
+    cacheTime: Infinity,
+    staleTime: Infinity,
+  });
+  const placeholder = useMemo(
```
alexrohleder, line 35: I think we can just declare this placeholder outside of the hook.

client/src/features/editors/list/EventList.jsx
```diff
@@ -15,25 +15,26 @@ export default function EventList(props) {
...
-    } else {
-      const previousEvent = events[cursor];
-      const nextEvent = events[cursor + 1];
-      if (type === 'event') {
-        eventsHandler('add', { type: type, after: previousEvent.id });
-      } else if (previousEvent?.type !== type && nextEvent?.type !== type) {
-        eventsHandler('add', { type: type, after: previousEvent.id });
+  const data = useEventListProvider();
+
+  const insertAtCursor = useCallback(
```
alexrohleder, line 22: likely no need to `useCallback` here

alexrohleder:
Code-wise what we achieved here is that we removed the socket logic from the components and abstracted it out inside this socket provider hook. 

As I understood we call this hook at the root component, which will subscribe to socket messages and store their values inside React Query store. That way we can use the React Query APIs to fetch and mutate the data, **for the user there will be no distinction API-wise on what is REST and what is WebSocket**.

I'm not the biggest fan of that hook being called at the root component, as we discussed it don't seem to belong there because of the namings. The solution would be to have more components composing the root. But that's not critical.

I think this is good.

alexrohleder (approved):

PR #197, opened 2022-08-30
Feat/table part1

cpvalente:
This feature adds ability to export a CSV from the table component.
There is also some cleanup here in the user interface. the relevant code is in
`client/src/features/table/TableWrapper.jsx`
and 
`client/src/features/table/utils.js`


client/src/common/hooks/useFullscreen.js
```diff
@@ -0,0 +1,31 @@
...
+
+  const toggleFullScreen = useCallback(() => {
+    if (!document.fullscreenElement) {
+      document.documentElement.requestFullscreen();
+    } else {
+      if (document.exitFullscreen) {
+        document.exitFullscreen();
+      }
+    }
+    setFullScreen(document?.fullscreenElement);
```
alexrohleder, line 27: I don't think we need the `?.` here. On line 20 we access this property without it.

client/src/common/hooks/useFullscreen.js
```diff
@@ -0,0 +1,31 @@
...
+
+export default function useFullscreen() {
+  const [isFullScreen, setFullScreen] = useState(document?.fullscreenElement);
+
+  useEffect(() => {
+    const handleChange = () => {
+      setFullScreen(document?.fullscreenElement);
+    };
+    document.addEventListener('fullscreenchange', handleChange, false);
+    document.addEventListener('resize', handleChange, false);
```
alexrohleder, line 12: Do we need the third argument here? False is the default value. Maybe you wanted to use `{ passive: true }` ?

alexrohleder (approved):

client/src/common/hooks/useFullscreen.js
cpvalente: thank you alex. reflex from electron

PR #198, opened 2022-09-02
Feat/132

cpvalente:
This PR enables basic features for a CSS override.
The CSS override file is exposed with express and added at runtime  with a hook

It could be tricky for the users to find the right place to add the css file since this location is platform dependend and buried in the package. I will sort this out separately, potentially a upload function

There is a flash of the page on rendering, this is because we defer rendering until we get a stylesheet (or not)
This is pretty unpleasant and also caused issues with some stuff in <StudioClock>, perhaps it is best not to defer this rendering? do you have any opinions here?


alexrohleder (approved):
There are many files on this PR. I'll approve based on the `useRuntimeStylesheet` hook we discussed. It seems to handle all the cases and deals with the scheduling issue. Very good 😁

alexrohleder:
* We cannot have a skeleton of the UI because we don't know the UI
* We cannot know the background before receiving the CSS, but we can assume it will be black

We send the index.html with a black background inlined in the body style. We defer the app's rendering until we get the stylesheet and parse it.

PR #267, opened 2022-12-07
V2 sentry

cpvalente:
Adding instrumentation and error reporting ahead of v2 beta testing

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -53,6 +53,7 @@
     ]
   },
   "devDependencies": {
+    "@sentry/vite-plugin": "^0.2.4",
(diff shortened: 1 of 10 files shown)


alexrohleder:
> Hi @alexrohleder , v2 is being tested by a few users. getting ready to send to a second group. Have you got experience with sentry?

Hey @cpvalente, I use it in a production project right now. Besides your implementation we have also added a [WebPack plugin that auto-configures the releases](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/webpack/). On your case you could use the [Vite version](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/vite/).

Issue #279, opened 2023-01-05
Minimal View


tcconway:
It looks like the "Minimal Timer" doesn't adhere to the End Message text. It only counts in the negative.

Steps to repeat:

1. Put some text in "End Message" in Settings.
2. Run a timer, allowing it to go past 0
3. Notice that the End Message shows up in the default timer display and not the "Minimal View" display, which shows it counting in the negative.

Thanks.

cpvalente:
Hi @tcconway 
<Minimal Timer> attempts to not be opinionated on how it runs.

From the users I have contact with, it is often used in production rather than on a stage / speaker environment.
Hence the smaller feature set.

A way forward would be to allow the configuration of the view, perhaps a URL parameter that could enable the end message.
Something like `http://localhost:4001/minimal?showEndMessage=true`

There is also an argument that the current default view <Stage/Presenter Timer> is too opinionated and users might prefer the <Minimal Timer> as a speaker timer. In which case we should be looking at a different set of solutions

What do you think?
What is your use case for <Minimal Timer>?

tcconway:
Looks amazing in v2

PR #286, opened 2023-02-16
Fix: #279 End Message in Minimal Timer

kellhogs:
I've Implemented the end message in the minimal timer, following the behavior from the normal timer. If it exist's it will be shown upon passing 0. Also it is possible to disable the end message using the `?hideendmessage=true` url parameter

### DIFF
--- a/apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
+++ b/apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
@@ -1,5 +1,6 @@
 import { useEffect } from 'react';
 import { useSearchParams } from 'react-router-dom';
+import { EventDataType } from 'common/models/EventData.type';
 import { useAtom } from 'jotai';
 
 import { overrideStylesURL } from '../../../common/api/apiConstants';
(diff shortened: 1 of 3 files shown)


apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
```diff
@@ -153,27 +158,27 @@ export default function MinimalTimer(props: MinimalTimerProps) {
     >
       <NavigationMenu />
       {!hideMessagesOverlay && (
+        <div className={showOverlay ? 'message-overlay message-overlay--active' : 'message-overlay'}>
+          <div className='message'>{pres.text}</div>
+        </div>
+      )}
+      {time.finished && general.endMessage && !hideEndMessage ? (
```
cpvalente, line 165: super nitpick:
I would prefer having "complex" logic away from the presentation.
This would mean extracting this to a variable before the render method, next to the other stuff
like 
`const showEndMessage = time.current < 0 && general.endMessage && !hideEndMessage ;`

Also, note I am using `time.current < 0 `instead of `time.finished`.
`time.finished `was a lazy shortcut from the early days that I came to regret and would like to get rid of it :D

cpvalente (changes requested):
Good work, in general this would be good to merge.
I did leave a small comment, mostly related to a bad pattern in the code I have written prior.

Again, great stuff!

cpvalente (approved):
Great stuff!

PR #289, opened 2023-02-17
Feature/#276/qrcodes

kellhogs:
Ontime settings, Event data:
Removed: Event URL,
Added: Public URL, Backstage URL

handeled database saving on server side.

modified public and backstage view to represent new data

### DIFF
--- a/apps/client/src/features/modals/EventSettingsModal.jsx
+++ b/apps/client/src/features/modals/EventSettingsModal.jsx
@@ -113,7 +113,7 @@ export default function SettingsModal() {
             <Input
               {...inputProps}
               name='publicUrl'
-              placeholder='www.onsite.no'
+              placeholder='www.getontime.no'
               value={formData.publicUrl}
               onChange={(event) => handleChange('publicUrl', event.target.value)}
             />
@@ -145,7 +145,7 @@ export default function SettingsModal() {
             <Input
               {...inputProps}
               name='backstageUrl'
-              placeholder='www.onsite.no'
+              placeholder='www.getontime.no'
               value={formData.backstageUrl}
               onChange={(event) => handleChange('backstageUrl', event.target.value)}
             />
--- a/apps/client/src/features/modals/EventSettingsModal.jsx
+++ b/apps/client/src/features/modals/EventSettingsModal.jsx
@@ -27,8 +27,9 @@ export default function SettingsModal() {
 
     setFormData({
       title: data.title,
-      url: data.url,
+      publicUrl: data.publicUrl,
       publicInfo: data.publicInfo,
+      backstageUrl: data.backstageUrl,
       backstageInfo: data.backstageInfo,
       endMessage: data.endMessage,
     });
@@ -145,7 +146,7 @@ export default function SettingsModal() {
               {...inputProps}
               name='backstageUrl'
               placeholder='www.onsite.no'
-              value={formData.backstageurl}
+              value={formData.backstageUrl}
               onChange={(event) => handleChange('backstageUrl', event.target.value)}
             />
           </div>
--- a/apps/client/src/features/modals/EventSettingsModal.jsx
+++ b/apps/client/src/features/modals/EventSettingsModal.jsx
@@ -103,7 +103,7 @@ export default function SettingsModal() {
           </div>
           <div className={style.hSeparator}>Additional Screen Info</div>
           <div className={style.spacedEntry}>
-            <FormLabel htmlFor='pubUrl'>
+            <FormLabel htmlFor='publicUrl'>
               Public URL
               <span className={style.labelNote}>
                 <br />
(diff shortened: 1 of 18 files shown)


apps/client/src/features/modals/EventSettingsModal.jsx
```diff
@@ -130,6 +134,22 @@ export default function SettingsModal() {
               onChange={(event) => handleChange('publicInfo', event.target.value)}
             />
           </div>
+          <div className={style.spacedEntry}>
+            <FormLabel htmlFor='backstagekUrl'>
```
cpvalente, line 138: typo in backstageUrl

apps/client/src/features/modals/EventSettingsModal.jsx
```diff
@@ -99,19 +103,19 @@ export default function SettingsModal() {
           </div>
           <div className={style.hSeparator}>Additional Screen Info</div>
           <div className={style.spacedEntry}>
-            <FormLabel htmlFor='url'>
-              Event URL
+            <FormLabel htmlFor='pubUrl'>
```
cpvalente, line 106: Nitpick nr 1: I think we should stick to the same naming convention in each entry. So `publicUrl` here and in the name of the input

apps/client/src/features/viewers/backstage/Backstage.tsx
```diff
@@ -146,30 +141,23 @@ export default function Backstage(props) {
...
+      <div className={showPublicMessage ? 'public-container' : 'public-container public-container--hidden'}>
         <div className='label'>Public message</div>
         <div className='message'>{publ.text}</div>
       </div>
 
       <div className='info'>
         <div className='qr'>
-          {general.url != null && general.url !== '' && (
-            <QRCode value={general.url} size={qrSize} level='L' />
+          {general.backstageUrl != null && general.backstageUrl !== '' && (
```
cpvalente, line 156: now that we have typescript we can be more confident about the data type, so

`{general.backstage && (`

apps/client/src/features/viewers/public/Public.tsx
```diff
@@ -106,29 +105,23 @@ export default function Public(props) {
...
+      <div className={showPublicMessage ? 'public-container' : 'public-container public-container--hidden'}>
         <div className='label'>Public message</div>
         <div className='message'>{publ.text}</div>
       </div>
 
       <div className='info'>
         <div className='qr'>
-          {general.url != null && general.url !== '' && (
-            <QRCode value={general.url} size={qrSize} level='L' />
+          {general.publicUrl != null && general.publicUrl !== '' && (
```
cpvalente, line 120: same here as in the other comment
`{general.publicUrl && (`

cpvalente (changes requested):
You did a great job at tracking the data flow, good stuff! I am impressed

I don't have any major comments, just a few cleanups and nitpicks

apps/client/src/features/modals/EventSettingsModal.jsx
```diff
@@ -99,19 +103,19 @@ export default function SettingsModal() {
...
                 <br />
-                Shown as a QR code in some views
+                QR code to be shown on public screens
               </span>
             </FormLabel>
             <Input
               {...inputProps}
-              name='url'
+              name='publicUrl'
               placeholder='www.onsite.no'
```
cpvalente, line 116: Wow. This is some old code...

Can you please update this to redirect to the ontime website?

Www.getontime.no

Make sure to update all references

apps/client/src/features/table/utils.js
```diff
@@ -39,7 +39,7 @@ export const makeTable = (headerData, tableData, userFields) => {
   const data = [
     ['Ontime · Schedule Template'],
     ['Event Name', headerData?.title || ''],
-    ['Event URL', headerData?.url || ''],
+    ['Public URL', headerData?.publicUrl || ''],
```
cpvalente, line 42: Are we not missing a backstage url here?

apps/client/src/features/viewers/studio/StudioClock.jsx
```diff
@@ -46,7 +46,7 @@ export default function StudioClock(props) {
   const [searchParams] = useSearchParams();
   const showSeconds = searchParams.get('seconds');
   formatOptions.showSeconds = Boolean(showSeconds);
-  formatOptions.format = `hh:mm${formatOptions.showSeconds ? 'mm' : ''}`;
+  formatOptions.format = `hh:mm${formatOptions.showSeconds ? ':ss' : ''}`;
```
cpvalente, line 49: Great catch!

cpvalente (changes requested):
I am recommending some small changes for now , I will build and run the feature later today. Good stuff

apps/client/src/features/table/utils.js
kellhogs: i have thought about adding it there but decided against it. Maybe the export is shared with third parties that may not have access to the backstage url.

cpvalente: We don't really protect the data in the app. The assumption is that anyone that has access to the page should be entitled to the data

kellhogs: alright

cpvalente (approved):

PR #290, opened 2023-02-19
Feat/countdown style

kellhogs:


apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
```diff
@@ -133,14 +133,18 @@ export default function MinimalTimer(props: MinimalTimerProps) {
   const isNegative = (time.current ?? 0) < 0;
   const showFinished = isNegative && !userOptions?.hideOvertime;
   const showEndMessage = time.current < 0 && general.endMessage && !hideEndMessage;
+  const countUp =
+    time.duration - time.current > time.duration ? time.duration * -1 + time.current : time.duration - time.current;
+  const timer =
+    time.countdownStyle === 'count-down' ? time.current : time.countdownStyle === 'count-up' ? countUp : time.clock;
```
cpvalente, line 139: it looks like we make this calculation in several places in the code.
The initial idea was to keep the code local, so people making their own integrations would have some examples without having to dig too far into the app.

I am not sure I agree with my old self on that anymore
I suggest we make a function **deriveTimerValues()** somewhere in a utils file (**src/common/utils/viewerUtils.ts** ?) which receives the timer object and returns an object with all these things we need.
We would apply it for now in **Timer.tsx** and **MinimalTimer.tsx**

This would also mean that we can easily make some tests on the logic here

cpvalente, line 139: I would prefer we skip the nested ternaries
Once we have our function, you could extract this into an utility function that receives the countdown style and the object of timers and returns the appropriate timer

apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
```diff
@@ -133,14 +133,18 @@ export default function MinimalTimer(props: MinimalTimerProps) {
   const isNegative = (time.current ?? 0) < 0;
   const showFinished = isNegative && !userOptions?.hideOvertime;
   const showEndMessage = time.current < 0 && general.endMessage && !hideEndMessage;
+  const countUp =
+    time.duration - time.current > time.duration ? time.duration * -1 + time.current : time.duration - time.current;
```
cpvalente, line 137: Not really sure about the logic here, should count up not simply be the time since the timer started?

cpvalente (changes requested):
I was unable to test this, for some reason having a counter as count up and count down did not seem to change the view. Perhaps I did not test it correctly

I have left some comments below that will help us build this feature.
Apart from that
- let's rename **countdownStyle** to **timerType** (just **timer type** in the UI labels)
- let's remove lint changes in otherwise untouched files
- what happens on count up when the timer is negative?

apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
kellhogs: i changed it up to use time.elapsed (stupid me hasn't seen that this exist) apart from that this logic negates the time when it goes over the duration. I do that to trigger the overtime class from the timer component wich doesn't show the negative sign before the time.

apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
cpvalente: Sounds reasonable. I will leave the organisation to your judgement. 

However I would say that, if we deem that it has enough complexity, the logic that calculates this derived states should be extracted to a file and tested

apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
cpvalente: Actually don't really know what the behaviour should be once we go overtime. Maybe keep counting but show in red?

apps/client/src/features/viewers/common/viewerUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+export function getTimerByType(timerObject?: TimerTypeParams): string | number | null {
+  let timer = null;
+  if (!timerObject) {
+    return timer;
+  }
+
+  if (timerObject.timerType === TimerType.CountDown) {
+    timer = timerObject.current;
+  } else if (timerObject.timerType === TimerType.CountUp) {
+    timer = timerObject?.elapsed ?? 0;
```
cpvalente, line 22: Small bug here. if the timer is count-up and there is added time. the timer appears to count down.
I am under the impression that the bug is in the backend. Would you mind taking a look=

apps/client/src/features/viewers/timer/Timer.tsx
```diff
@@ -61,29 +63,62 @@ export default function Timer(props) {
...
       <div className={showOverlay ? 'message-overlay message-overlay--active' : 'message-overlay'}>
         <div className='message'>{pres.text}</div>
       </div>
 
-      <div className='clock-container'>
+      <div
+        className='clock-container'
+        style={{
+          opacity: `${showClock ? 1 : 0}`,
+        }}
```
cpvalente, line 104: I would prefer that we do this with a classname

something like
className={`clock-container ${showClock ? '' : 'hidden'}`}

Your solution is fine, and the performance impact is negligible. But currently we define all the styles in the CSS, and it is best to maintain some consistency to help our future selves with maintenance

apps/client/src/features/viewers/timer/Timer.tsx
```diff
@@ -61,29 +63,62 @@ export default function Timer(props) {
...
+        className='timer-container'
+        style={{
+          fontSize: `${100 / stageTimerCharacters}vw`,
+        }}
+      >
+        {showEndMessage ? (
+          <div className='end-message'>{general.endMessage}</div>
         ) : (
-          <TimerDisplay time={normalisedTime} hideZeroHours className={isPlaying ? 'timer' : 'timer--paused'} />
+          <div className={`timer ${!isPlaying ? 'timer--paused' : ''} ${showFinished ? 'timer--finished' : ''}`}>
```
cpvalente, line 119: I guess this is the way to go, few suggestions:
- lets make the calculation for fontsize in the body of the function, before the return
- same with the classname, lets do it up on the body

apps/client/src/common/components/timer-display/TimerDisplay.tsx
```diff
@@ -19,12 +19,18 @@ interface TimerDisplayProps {
...
-      : formatDisplay(millisToSeconds(time), hideZeroHours);
+  let display = '';
+
+  if (typeof time === 'string') {
+    display = time;
+  } else if (time === null || typeof time === 'undefined' || isNaN(time)) {
+    display = '-- : -- : --';
+  } else {
+    display = formatDisplay(millisToSeconds(time), hideZeroHours);
+  }
```
cpvalente, line 30: this logic seems to be repeated in TimerDisplay, Timer and MinimalTimer.
Could we make it a function in the viewerUtils?

cpvalente (changes requested):
Great work, I think we are on the right track and can soon merge this 

There is a small bug which I believe might not be related to your changes, it would be great if you could check it out for me either way

I also left some small suggestions for changes and pushed a commit to improve typing, remember to fetch

apps/client/src/common/components/timer-display/TimerDisplay.tsx
```diff
@@ -19,12 +19,18 @@ interface TimerDisplayProps {
 const TimerDisplay = (props: TimerDisplayProps) => {
```
cpvalente, line 19: Looks like this component is now only used in PlaybackTimer, perhaps we have a good chance to either move the logic there and delete this or tighten the implementation to only have what is necessary. No need to think about reusability if we decide not to reuse it

apps/client/src/features/viewers/common/viewerUtils.ts
kellhogs: i think the issue is that the 'elapsed' time is calculated against the 'duration' of the timer. Adding time to the timer let's 'elapsed' go negative wich makes it count down in frontend i think.

kellhogs: i've fixed the issue in frontend for now by setting elapsed to 0 if it's smaler than 0. Although i think we should takle that issue in backend. I just dont wanna mess with the calculations for now. Maybe you have some thoughts on that.

cpvalente (approved):

PR #292, opened 2023-02-20
wip: code suggestions

cpvalente:
few suggestions to the code changes, you don't need to merge, can copy paste the changes you want to keep or take it as inspiration and do your own thing


apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
```diff
@@ -133,23 +134,13 @@ export default function MinimalTimer(props: MinimalTimerProps) {
   const isNegative = (time.current ?? 0) < 0;
   const showFinished = isNegative && !userOptions?.hideOvertime;
   const showEndMessage = time.current < 0 && general.endMessage && !hideEndMessage;
```
cpvalente, line 136: remember that we will still need to check this. end message should never appear on a count up (I guess)

PR #297, opened 2023-02-24
Add arm platforms to docker build

kellhogs:
This change should add the arm images needed for the Raspberry Pi to the automated build process

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -179,6 +179,8 @@ jobs:
       # Prepare builder
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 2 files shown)


cpvalente:
This looks fine to me. Should we coordinate to make a release and test the docker image?

cpvalente (changes requested):
Not sure what happened here. But we are already in 1.9.8
The pipeline will fail if the version already exists. 

Since this is a new feature, we should probably be bumping to 1.10.0

kellhogs:
mhh weird .... maybe i didn't sync master before updatting the package.json

cpvalente (approved):

PR #303, opened 2023-02-26
Feat/update check

kellhogs:
settings now show the latest version fetched from github with clickable link to the release

image


apps/client/src/common/api/ontimeApi.ts
```diff
@@ -151,3 +151,15 @@ export const uploadData = async (file: string, setProgress: (value: number) => v
...
+
+/**
+ * @description HTTP request to get the latest version and url from github
+ * @return {Promise}
+ */
+export async function getLatestVersion(): Promise<object> {
+  const res = await axios.get(`${githubURL}`);
+  const url = res.data.html_url;
+  const version = res.data.tag_name;
+  // console.log({ url: url, version: version });
```
cpvalente, line 163: Remember to delete debug code

apps/client/src/common/api/ontimeApi.ts
```diff
@@ -151,3 +151,15 @@ export const uploadData = async (file: string, setProgress: (value: number) => v
...
 };
+
+/**
+ * @description HTTP request to get the latest version and url from github
+ * @return {Promise}
+ */
+export async function getLatestVersion(): Promise<object> {
+  const res = await axios.get(`${githubURL}`);
+  const url = res.data.html_url;
+  const version = res.data.tag_name;
```
cpvalente, line 162: we should probably have some sort of error handling here

apps/client/src/common/api/ontimeApi.ts
```diff
@@ -151,3 +151,15 @@ export const uploadData = async (file: string, setProgress: (value: number) => v
...
+  // console.log({ url: url, version: version });
+  return { url: url, version: version };
```
cpvalente, line 164: nice idea in providing the URL

apps/client/src/features/modals/AppSettingsModal.jsx
```diff
@@ -49,6 +51,9 @@ export default function AppSettingsModal() {
       pinCode: data.pinCode,
       timeFormat: data.timeFormat,
     });
+    getLatestVersion().then((data) => {
+      setVersionData(data);
+    });
```
cpvalente, line 56: What if we only fetch the data if the user requests it. This request goes online, and I would think that the user should consent to this by triggering the action themselves
As you know, many video networks are behind offline or near-offline networks

If you think that the we should fetch the data either way, add a new useEffect for it with an empty dependency array
If we do in useEffect, we also need to add a cleanup function

so code would look something like

```
useEffect(() => {

    // we use abort controller to be able to cancel request
    const controller = new AbortController()
    const fetchData = async () => {
        axios.get(url, {signal: controller} )
        // handle stuff here
    }

    // cleanup function tells the component to cancel the fetch if component unmounts
    return(() => {
        controller.abort()
    })

// empty dependency array means useEffect runs in mount / unmount, only once
}, [])
```

apps/client/src/features/modals/AppSettingsModal.jsx
```diff
@@ -128,6 +133,24 @@ export default function AppSettingsModal() {
...
+      const text = `Running latest version: ${installedVersion}`;
+      return text;
+    } else {
+      const text = (
+        <a href={versionData.url} target='_blank' rel='noreferrer'>{`New version available: ${latestVersion}`}</a>
+      );
+      return text;
+    }
+  };
+
```
cpvalente, line 153: we can simplify this a little by making some assumptions.
if the version in github is v1.1.1 and the local is 1.1.1, I know that for them to be the same, the github one has to contain the local

Also, 
- redeclaring constants in collocated scopes can cause hard-to-debug issues in future, it would be simpler and more dummy proof to immediately return
- we probably want to make sure to always return the same data type. By Calling a function named versionString() I would expect to get a string representation of my app version, not a component

cpvalente (changes requested):
I think you did a great job here, there are some react gotchas to keep in mind.
Otherwise we could improve the code slightly with some suggestions

See a small demo I made in the code sandbox for some inspiration
https://codesandbox.io/s/version-check-u4ql71?file=/src/VersionCheck.tsx

Alltogether, great work here, let me know if I can help

PR #304, opened 2023-02-27
V2 fix timer

cpvalente:
This aims to simplify the logic around calculating the time basics. it will end up being important
it also makes the decision that user-added times are not included in the elapsed timers. elapsed timer is purely the time past since we started a timer


kellhogs (approved):
Looks good to me. Haven't found anything obvius that could be wrong. Good Work.

PR #308, opened 2023-03-11
Feat/end action

kellhogs:
implement functions for choosing what happens when the timer hit's zero

Continue: go in overtime like it is right now
Pause: pauses the running timer when it reaches zero
Stop: stop the timer when it reaches zero 
Next: load next timer when current timer reaches zero


apps/server/src/services/PlaybackService.js
```diff
@@ -127,6 +127,29 @@ export class PlaybackService {
     }
   }
 
+  /**
+   * Starts playback on next event
+   * @param {string} [fallbackAction] - 'stop', 'pause', or null
+   */
+  static startNext(fallbackAction) {
```
cpvalente, line 134: remember also to type typescript
`static startNext(fallbackAction?: 'stop' | 'pause' ) {`

apps/server/src/services/PlaybackService.js
```diff
@@ -127,6 +127,29 @@ export class PlaybackService {
...
+    const nextEvent = eventLoader.findNext();
+    if (nextEvent) {
+      const success = PlaybackService.loadEvent(nextEvent);
+      if (success) {
+        socketProvider.info('PLAYBACK', `Loaded event with ID ${nextEvent.id}`);
+        PlaybackService.start();
```
cpvalente, line 140: We also need to provide feedback for the play

apps/server/src/services/PlaybackService.js
```diff
@@ -127,6 +127,29 @@ export class PlaybackService {
...
+      }
+    } else if (fallbackAction === 'stop') {
+      socketProvider.info('PLAYBACK', `No next event found! Stopping playback`);
+      PlaybackService.stop();
+    } else if (fallbackAction === 'pause') {
+      socketProvider.info('PLAYBACK', `No next event found! Pausing playback`);
+      PlaybackService.pause('pause');
+    } else {
+      socketProvider.info('PLAYBACK', `No next event found! Continuing playback`);
+    }
```
cpvalente, line 150: nice, I like how you left the default behaviour as a fallback. very elegant

apps/server/src/services/TimerService.ts
```diff
@@ -302,6 +311,37 @@ export class TimerService {
...
+        if (
+          this.playback === 'play' &&
+          this.timer.endAction === EndAction.Next &&
+          this.timer.current <= 0 &&
+          this.timer.finishedAt === null
+        ) {
+          this.timer.finishedAt = this.timer.clock;
+          PlaybackService.startNext();
+        }
+
```
cpvalente, line 344: I still think that all of this should be handled as a side effect in `_onFinish()`

packages/types/src/definitions/EndAction.type.ts
```diff
@@ -0,0 +1,6 @@
+export enum EndAction {
+  Continue = 'continue',
+  Pause = 'pause',
+  Stop = 'stop',
+  Next = 'next'
+}
```
cpvalente, line 6: enums FTY 😄

cpvalente (changes requested):

apps/server/src/services/PlaybackService.js
kellhogs: the file isn't a typescript file yet. Thats the reason why i left it out. do you want me ton convert it?

apps/server/src/services/PlaybackService.js
kellhogs: the play feedback comes from the start function

apps/server/src/services/PlaybackService.js
kellhogs: thanks

packages/types/src/definitions/EndAction.type.ts
kellhogs: what do you think of the implementation in the EventEditor? (dont know how to reference the code in this comment )

apps/server/src/services/TimerService.ts
```diff
@@ -302,6 +311,38 @@ export class TimerService {
...
+        if (
+          this.playback === 'play' &&
+          this.timer.endAction === EndAction.Next &&
+          this.timer.current <= 0 &&
+          this.timer.finishedAt === null
+        ) {
+          this.timer.finishedAt = this.timer.clock;
+          PlaybackService.loadNext();
```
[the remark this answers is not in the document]
kellhogs: resolved in a later commit

apps/server/src/services/TimerService.ts
kellhogs: by moving the logic to the `_onFinish()` function it causes the timer to go negative before my logic get's handled
e.g by choosing pause it pauses at -00:00:01

apps/server/src/services/PlaybackService.js
cpvalente: Yes please. Let me know if you need help

apps/server/src/services/PlaybackService.js
cpvalente: Nice. Good catch

apps/server/src/services/TimerService.ts
cpvalente: Hmm. That sounds plausible.
The timer updates once a second and assumes a finished timer has gone past the 0

So this means we might be reading the timer at -300ms which gets rounded to -1s

The only other way is to have a 0 seconds and a -0 seconds which would be worse I guess

I short: the logic is doing what is supposed to do, although the presentation looks strange

We still need to go through the function to handle other side effects like automation sync.

What do you think?
If we think that the -1 hurts the confidence of the user to the timer,  my instinct would be to cut the pause functionality, potentially replace with a play next - load next

apps/client/src/features/event-editor/EventEditor.tsx
```diff
@@ -169,17 +169,18 @@ export default function EventEditor() {
...
+            onChange={(event) => handleChange('endAction', event.target.value)}
           >
-            <option value='start-end'>Start to end</option>
-            <option value='duration'>Duration</option>
-            <option value='follow-previous'>Follow previous</option>
-            <option value='start-only'>Start only</option>
+            {Object.keys(EndAction).map((key) => (
+              <option key={key} value={EndAction[key as keyof typeof EndAction]}>
+                {key}
+              </option>
```
cpvalente, line 182: Good effort.
I wonder if the typing here is a little complicated 

Since we are dealing with 3-4 options I wonder if being explicit about those options has better observability. It also makes the code less magic

packages/types/src/definitions/EndAction.type.ts
cpvalente: For me it sounds a little complicated.

The brain needs to work a bit to read it, even though it is straightforward 

I wonder 
- are there easier ways to write that?
- is the code abstraction worth for such a simple case?

As long as it works. I will leave it up to you

packages/types/src/index.ts
```diff
@@ -15,6 +15,7 @@ import { Playback } from './definitions/runtime/Playback.type.js';
 import { Settings } from './definitions/core/Settings.type.js';
 import { TimerLifeCycle } from './definitions/core/TimerLifecycle.type.js';
 import { TimerType } from './definitions/TimerType.type.js';
+import { EndAction } from './definitions/EndAction.type.js';
```
cpvalente, line 18: Can I super nitpick and ask you to keep the alphabetical order?

cpvalente (changes requested):
Good stuff. I made some small comments, but the code looks good

apps/client/src/features/event-editor/EventEditor.tsx
kellhogs: i wanted to archive this solution because i could't imediately figure out how to do this. But i will agree it seems to be a little overcomplicated for this application. I willr emove it

apps/server/src/services/TimerService.ts
kellhogs: I think for the end user it looks strange if the timer doesn't stop at 0. I agree with your suggestion and will remove the pause functionality and also expand the functionality to load-next and play-next

apps/server/src/services/PlaybackService.ts
```diff
@@ -105,14 +106,28 @@ export class PlaybackService {
 
   /**
    * Loads event after currently selected
+   * @param {string} [fallbackAction] - 'stop', 'pause', or null
+   * @return {boolean} success
    */
-  static loadNext() {
+  static loadNext(fallbackAction?: 'stop' | 'pause' | null): boolean {
```
cpvalente, line 112: No need for null here

apps/server/src/services/PlaybackService.ts
```diff
@@ -127,6 +142,17 @@ export class PlaybackService {
     }
   }
 
+  /**
+   * Starts playback on next event
+   * @param {string} [fallbackAction] - 'stop', 'pause', or null
+   */
+  static startNext(fallbackAction?: 'stop' | 'pause' | null) {
```
cpvalente, line 149: Same here, no need for null

apps/server/src/services/TimerService.ts
```diff
@@ -316,6 +321,15 @@ export class TimerService {
   _onFinish() {
     eventStore.set('timer', this.timer);
     integrationService.dispatch(TimerLifeCycle.onFinish);
+    if (this.timer.endAction === EndAction.Stop) {
+      PlaybackService.stop();
+    } else if (this.timer.endAction === EndAction.LoadNext) {
+      setTimeout(() => {
+        PlaybackService.loadNext();
+      }, 0);
```
cpvalente, line 329: It is nice to document when you need to add a weird workaround.
Our future selves will not remember what was this all about

cpvalente (changes requested):
I haven't been able to test, but it looks good. Just some small cleanup 

Good stuff!

cpvalente:
This looks good to me, I have merged the master into it to help with the merge conflicts and gave it some testing

I hope you can check it in and see that I didn't destroy any of your work.
I also noticed that the behaviour with the timer going negative is a bit strange. I wonder if we can think of a way to improve??

We leave that for next time

kellhogs:
Looks to me like evreything still works fine. So i would say we are good to merge.

PR #312, opened 2023-03-20
V2 docker

kellhogs:
Add new docker workflow for V2
Add new docker-compose template for V2

Auto populate override.css in docker image


.github/workflows/build.yml
cpvalente, line 1: What about the rest?
This file also builds the distributable for all the other platforms...

.github/workflows/docker_v2_ci.yml
```diff
@@ -0,0 +1,70 @@
...
+        # Build's context is the set of files located in the specified PATH or URL
+        context: .
+        # Path to the Dockerfile
+        file: ./Dockerfile
+        # List of target platforms for build
+        platforms: linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v6
+        # Push is a shorthand for --output=type=registry
+        push: true
+        # List of tags
+        tags: ${{ secrets.DOCKERHUB_USERNAME }}/ontime:${{ env.RELEASE_VERSION  }} , ${{ secrets.DOCKERHUB_USERNAME }}/ontime:v2_latest
```
cpvalente, line 69: The choice of the tag is pretty interesting.

The `:latest` tag, as controversial as it might be, indicates to the user that this was the last thing we pushed (which might or might not be the most advanced version of the app)

It is a well-established pattern, and is expected that any repo would have something with `latest`
However, we are publishing something that will go through beta first. So maybe `:latest`, should be kept to the v1 versions until we release the first stable v2 release.
In this case, your `v2_latest` makes sense

But if that is the issue to solve, that our tag wants to disconnect itself from "the latest major version" would a simpler tag like `beta` be more explicit?

What do you think?

Dockerfile
```diff
@@ -1,21 +1,27 @@
 FROM node:16-alpine
 
-WORKDIR /apps/server
+# Set environment variables
+# Environment Variable to signal that we are running production
+ENV NODE_ENV=production
+# Ontime Data path
+ENV ONTIME_DATA=/external/
+
+WORKDIR /Resources/extraResources/
```
cpvalente, line 9: A-ha, this is what you meant with /Resources/extraResources annoying dir.

I agree. This is something related to electron, and we shouldn't put our docker users through it if we can avoid it

Dockerfile
```diff
@@ -1,21 +1,27 @@
...
 
 # Prepare UI
-COPY /apps/client/build ../client/build
+COPY /apps/client/build ./client/
 
 # Prepare Backend
-COPY /apps/server ./
+COPY /apps/server/dist/ ./server/
+COPY /demo-db/ ./preloaded-db/
+COPY /apps/server/src/external/ ./external/
```
cpvalente, line 17: great stuff

Dockerfile
```diff
@@ -1,21 +1,27 @@
...
+EXPOSE 4001/tcp 8888/udp 9999/udp
 
-CMD ["npm", "start:headless"]
+CMD ["node", "server/docker.cjs"]
 
 # Build and run commands
 # !!! Note that this command needs pre-build versions of the UI and server apps
 # docker build -t getontime/ontime .
-# docker run -p 4001:4001 -p 10.0.0.12:8888:8888/udp --mount type=bind,source="$(pwd)/ontime-db",target=/server/preloaded-db getontime/ontime
+# docker run -p 4001:4001 -p 8888:8888/udp -p 9999:9999/udp -v ./ontime-db:/external/preloaded-db/ -v ./ontime-styles:/external/external/styles/ getontime/ontime
```
cpvalente, line 27: hmm, I am unsure I understand the reason for the external directory that would make `/external/external/styles`, am I missing something?

At least in terms of semantic, the preloaded db is not external. External attempts referencing to files that the user might want to add to the distribution (hinting forwards to being able to make your own views and add here)

Perhaps the folders can be at the root like
```
app.js
docker.js
/external-styles/override.css
/db/db.json
```

apps/server/package.json
```diff
@@ -49,6 +49,7 @@
     "setdb": "shx cp ../../demo-db/db.json src/preloaded-db/db.json",
     "prebuild": "pnpm setdb",
     "build": "pnpm prebuild && esbuild src/app.ts --log-level=error --platform=node --format=cjs --bundle --minify --outfile=dist/index.cjs",
+    "build:docker": "pnpm prebuild && esbuild src/app.ts --log-level=error --platform=node --format=cjs --bundle --minify --outfile=dist/app.cjs && esbuild src/index.ts --log-level=error --platform=node --format=cjs --bundle --minify --outfile=dist/docker.cjs",
```
cpvalente, line 52: Would it not be simpler to create the `docker.cjs` ourselves?
it is just a few lines of code that we manually maintain in electron anyway...

apps/server/src/app.ts
```diff
@@ -118,6 +119,7 @@ const checkStart = (currentState: OntimeStartOrder) => {
 export const startDb = async () => {
   checkStart(OntimeStartOrder.InitDB);
   await dbLoadingProcess;
+  populateStylesFile;
```
cpvalente, line 122: does the function not need to be called?

I am also conflicted about the idea of using a function called `startDb` to do this.
Perhaps we change the function name to `init`?

apps/server/src/modules/loadStyles.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @description ensures directories exist and populates database
+ * @return {string} - path to db file
+ */
+const populateStyles = () => {
+  const stylesInDisk = resolveStylesPath;
+  ensureDirectory(resolveStylesDirectory);
+
+  // if dbInDisk doesn't exist we want to use startup db
```
cpvalente, line 14: Looks good, just a few copy-paste artifacts
`path to db file`
`if dbInDisk doesn't exist we want to use startup db`

cpvalente (changes requested):
Great work. my comments are mostly very nitpicky and aiming for our future selves

apps/server/src/app.ts
kellhogs: im unsure about why this works ... but it works 😅. I guess it is because we are exporting a const in loadStyle?

kellhogs: we could move this call to the startServer function. it doesen't really matter when in the start process it get's called

apps/server/package.json
kellhogs: how would we do this?

.github/workflows/build.yml
kellhogs: deleting this wasn't intenional. small oversight by me. deleted it in my repo to stop it from running. Havent looked at these build steps at all since i was focusing on the docker stuff

.github/workflows/docker_v2_ci.yml
kellhogs: maybe beta_v2 for the "latest" version of v2

Dockerfile
kellhogs: the docker users shouldn't even notice it beeing like that the files they need to acces can be put anywhere by the ONTIME_DATA variable

.github/workflows/build.yml
cpvalente: No worries, do you think you could merge your changes in the existing file? otherwise we can always recreate the pipeline later

Dockerfile
cpvalente: sure, is the reason that it makes it easier for us to predict the directory across platforms? resources/extraResources is quite a mouthful 😅

apps/server/package.json
cpvalente: we would simply create a file in a separate directory containing the code to start up the app.
you can see this code in the main.js inside electron

we would then use the dockerfile to move that docker.js into the built directory. we can have a look together if you would like

apps/server/src/app.ts
cpvalente: That is right, that seems to be a copy-paste artifact.
populateStylesFile is the value of the execution of populateStyles()
`export const populateStylesFile = populateStyles();`

This was necessary in the dbLoading because of the await, but not here.
Would it be more explicit to call populateStyles function directly?

When it comes to moving it, that is an excellent question.
The express server needs the files ready when it spins up. so I am surprised you haven't found any issues.
This populating function should run at the top of the app

.github/workflows/docker_v2_ci.yml
cpvalente: sounds good

.github/workflows/build.yml
kellhogs: i thought maybe it's wise to keep the docker build in a seperate pipeline. There are many action's in the docker pipeline wich makes the file very long and harder to read when combined with the rest.

Dockerfile
kellhogs: i will try to propose a more sensible path solution when we figured out the rest of your mentions

apps/server/src/app.ts
kellhogs: the express server get's initialized in the startServer function wich runs after startDB. So maybe we could populate the styles file at thhe top of that function?

.github/workflows/build.yml
cpvalente: sounds reasonable. in that case lets just rename the file to be explicit to the fact that this only relates to docker.
maybe just build-docker.yml is enough

apps/server/package.json
cpvalente: lets take this on slack.
The short version is that the code in index.ts is very small and we could write it again, but there is nothing fundamentally wrong with what you've done, it sounds overcomplicated to me to get a build tool to write 10 lines of very straightforward code

apps/server/src/app.ts
cpvalente: thats right. I would then suggest that we continue using the startDb function but rename it to init or initAssets, we would also need to rename the step in the enum

Issue #324, opened 2023-04-04
Feat: Localization of web views


kev-ac:
It would be great to be able to translate web views to enhance the experience for guests. 
If others feel the need to have this feature and could possibly be merged in I can work on a PR. 

Proposed solution: 
- Library: react-i18next
- Set desired language with an URL query param

cpvalente:
Hi @kev-ac, thank you for reaching out and for offering to help.

I completely agree this would be a good improvement and don't see any downsides to having a translation in place
Few thoughts:
- translations should only happen to views available to guests. This would mean cuesheet and editor should not be translated.
- with this in mind, there is very little to be translated, and I wonder if i18 might be overkill. Many users consume the views on less powerful devices such as TVs and tablets. Keeping the bundle sizes for the views low is important. I imagine that a simple solution using a translation map should be sufficient and not too cumbersome (?)
- this PR should be targeted for v2

What do you think?

cpvalente:
From the app side, this would be a great addition. I would welcome a PR.

If you have availability to make a proposal, it would be fantastic. 

I have documented the [development process ](https://github.com/cpvalente/ontime/blob/v2/DEVELOPMENT.md)in v2. Please do get in touch if I can be of help mail@getontime.no

cpvalente:
Hey, good work. I dont think I would have done it much differently myself.
Few questions:
- what does the common prefix symbolise? I hope I guessed correctly below
- would it be simpler to use a javascript object?
- can you please add the new option to `OverridableOptions`? I will eventually get around to make a menu to change these

Either way, these are the strings in the views:
```
export const langEn: TranslationMap = new Map([
  ['common.time_now', 'Time now'],
  ['common.now', 'Now'],
  ['common.next', 'Next'],
  ['common.started_at', 'Started At'],
  ['common.expected_finish', 'Expected Finish'],
  ['common.stage_timer', 'Stage Timer'], // is the name good to begin with?
  ['countdown.select_event', 'Select an event to follow'],
  ['countdown.to_start', 'Time to start'],
  ['countdown.waiting', 'Waiting for event start'],
  ['countdown.running', 'Event running'],
  ['countdown.ended', 'Event ended at'],
  ['common.start_time', 'Start Time'],
  ['common.end_time', 'End Time'],
]);
```

kev-ac:
> what does the common prefix symbolise? I hope I guessed correctly below

The common prefix was chosen for strings which are shared across multiple views. 
[...]

PR #325, opened 2023-04-04
Add translations for views

kev-ac:
Relates to #324

### DIFF
--- a/apps/client/src/translation/TranslationProvider.jsx
+++ b/apps/client/src/translation/TranslationProvider.jsx
@@ -1,0 +1,44 @@
+import React, { createContext, useContext, useState } from 'react';
+
+import { langDe } from '@/translation/languages/de';
+import { langEn } from '@/translation/languages/en';
+
+const ALLOWED_LANGUAGES = ['en', 'de'];
+
+const translationsMap = new Map([
+  ['en', langEn],
+  ['de', langDe],
+]);
+
+export const TranslationContext = createContext(undefined);
+
+export const TranslationProvider = ({ children }) => {
+  // Default language
+  const [language, setLanguageState] = useState('en');
+
+  const getString = (key) => {
+    return translationsMap.get(language).get(key);
+  };
+
+  const setLanguage = (language) => {
+    language = language.toLowerCase();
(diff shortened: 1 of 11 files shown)


pnpm-lock.yaml
cpvalente, line 1: What happened here? Would you mind checking out from v2 and re-running pnpm i?

apps/client/src/translation/TranslationProvider.jsx
```diff
@@ -0,0 +1,51 @@
...
+    getString,
+    setLanguage,
+  };
+
+  return <TranslationContext.Provider value={contextValue}>{children}</TranslationContext.Provider>;
+};
+
+export const useTranslation = () => {
+  const { getString, setLanguage } = useContext(TranslationContext);
+  return { getString, setLanguage };
```
cpvalente, line 50: I wonder if we could use something less generic like `getLocalisedString`
What do you think?

cpvalente (changes requested):
This looks great to me. I made some small nitpicky comments. Mostly wondering about the pnpm-lock

Also, would you mind running the linter in the changed files? otherwise, it will pollute further commits
Context in routers sounds good. It shouldn't affect anything else

Would you mind implementing the other views or would you prefer me to?

pnpm-lock.yaml
kev-ac: Oops, this file shouldn't have been committed, I was too fast there. Will revert it.

apps/client/src/translation/TranslationProvider.jsx
kev-ac: Yes, sounds like a good thing to do.

kev-ac:
> Also, would you mind running the linter in the changed files? otherwise, it will pollute further commits

Eslint runs in my IDE automagically when I save a file, so there shouldn't be any issues here. Also a manual run doesn't give any recommendations. 

> Would you mind implementing the other views or would you prefer me to?

I can do it, but I could take some time. Doing this in some free time in-between other things, surely how you also do it :-) 
So feel free if you have some spare time.

cpvalente:
I am working on some other small fixes and can do that as well. 

Do you think we can merge this? I am happy to approve of you remove the draft flag

PR #326, opened 2023-04-05
V2 beta1

cpvalente:
Small fixes and tweaks based on first open beta round
- Studio Clock: Bug fix countdown
- Event block: show alert on time input
- Rundown: create all events with 0 duration
- Rundown: do not remove blocks when applying delays
- Blocks: add title field
- Style: add overrideable styles in progress bars
- Event Editor: show character count in event fields
- Several small style and behaviour tweaks


apps/client/src/features/rundown/common/TitleEditor.tsx
kellhogs, line 1: Maybe `TitleEditor `is a bit to general here?
I had to think a moment and read the code to realise that this is about the `BlockTitles`. 
Maybe `BlockTitleEditor `would be a better name?

kellhogs (commented):
Great work. I don't have much to say here

PR #327, opened 2023-04-05
feat: notify ws clients on rundown changes

cpvalente:
Broadcast ws message when the rundown changes.

So far message looks like
```
{
    type: 'ontime-refetch',
    payload: null,
};
```

The payload could include a patch object, but we would need to define behaviour flags for updating single/multiple items. One for the future I guess

What do you think about the name **refetch**? We will be stuck with it once we implement, not too sure it is communicative

### DIFF
--- a/apps/server/src/adapters/websocketAux.ts
+++ b/apps/server/src/adapters/websocketAux.ts
@@ -1,0 +1,12 @@
+import { socket } from './WebsocketAdapter.js';
+
+/**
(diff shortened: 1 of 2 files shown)


kellhogs (approved):
Code looks good to me. For the name i couldn't come up with a better one. 
Also the function of this message is to notify external programms that there is new data to fetch so i think refetch is fine here.

PR #328, opened 2023-04-05
Add translations for views (#325)

cpvalente:
Adds localised strings for 

* Add foundation for views translation

* WIP: Move hard-coded strings to translation map

* Add foundation for views translation

* WIP: Translations

* Use flat Javascript object for translation mappings

* Revert lock file back to original state

* Rename getString() to getLocalizedString()

### DIFF
--- a/apps/client/src/features/viewers/backstage/Backstage.tsx
+++ b/apps/client/src/features/viewers/backstage/Backstage.tsx
@@ -15,11 +15,11 @@ import { TimeManagerType } from '../../../common/models/TimeManager.type';
 import { formatDisplay, millisToSeconds } from '../../../common/utils/dateConfig';
 import { getEventsWithDelay } from '../../../common/utils/eventsManager';
 import { formatTime } from '../../../common/utils/time';
+import { useTranslation } from '../../../translation/TranslationProvider';
 import { titleVariants } from '../common/animation';
 import { TitleManager } from '../ViewWrapper';
 
 import './Backstage.scss';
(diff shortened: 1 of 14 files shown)


cpvalente:
@kev-ac 

Sorry for rushing into it, I was hoping to get this lined up before going on holidays 
I have implemented the remaining strings and created placeholders in the german string object. What do you think?

Is it just a matter of updating the german strings?

apps/client/src/features/viewers/backstage/Backstage.tsx
kev-ac, line 1: This file still references getString() not getLocalizedString()

kev-ac:
> I have implemented the remaining strings and created placeholders in the german string object. What do you think?

Placeholders for the german string object wouldn't be necessary because the lookup function falls back to the default language (EN) if it does not find a matching string in the requested language.

That being said: I cannot find the placeholders you've created in this PR. Am I missing something?

apps/client/src/translation/languages/de.ts
kev-ac, line 1: I would suggest removing the german placeholders before merging it into v2. This way the current behavior keeps the same. I will push the translations once I get to it.

kev-ac (changes requested):
LGTM. See comment below

apps/client/src/translation/languages/de.ts
cpvalente: perfect, I have made the change and will merge.
Thank you again

PR #337, opened 2023-04-13
Fix/docker port

kellhogs:


cpvalente:
I wonder if this is a case of fixing the symptom rather than the problem.
I found several threads with similar issues in running websocket inside docker. Somewhere it mentioned

> You can also try running the Docker container with the --network=host option, which will use the host network interface instead of creating a new network namespace.

Either way, we can merge this so we have a working solution for now
What do you think? Is there something we can try quickly to see if there is progress?

PR #398, opened 2023-05-19
fix: reset rundown on upload

cpvalente:
Fixes issue where uploading a database left stale data in the runtime store


kellhogs (approved):
Very nice and simple solution

PR #401, opened 2023-05-20
add endAction and timerType validation to excel upload

bunday:
Introduced reading the `End Action` and `Timer Type` on the Excel file used for import


cpvalente:
Hi, thank you for your work.
As I mentioned before, this part of the code is quite verbose and not the easiest to get your head around as the first try, good work!

I will probably write some e2e tests for this since it is one of the app's core features. But meanwhile, we have some unit tests in parser.tests.js 
Would you mind dropping in a happy path testing? You could just append the new data that you made to the existing test

bunday:
Hi, thanks for the nice words, I sure will update the new data with the existing test

cpvalente (approved):
Great work! I look forwards to seeing more of your contributions

PR #404, opened 2023-05-20
fix: prevent css collision

cpvalente:
Fixes an issue where the multi progress bar had classes with the same name as the progress bar, so weird stuff was happening.

Easy fix, tricky to debug

@kellhogs Can you check that I did not break anything?


kellhogs (approved):
Seems like everything is working as it should 👍🏻

Issue #349, opened 2023-04-19
Change API timer.current to hh:mm:ss & add a new one for miliseconds?


tcconway:
Based on the great [Websocket API](https://cpvalente.gitbook.io/ontime/control-and-feedback/websocket-api)...

I'd like to request `timer.current` to return a human-readable result such as 14:34:56 -- in typical hh:mm:ss format. With no timer running, it should probably return 00:00:00

Then have `timer.miliseconds` (suggested) return miliseconds? With no timer running, it should probably return 0.

Another naming convention could be
`timer.currentTime`
`timer.currentMils`

PR #405, opened 2023-05-20
Websocket/human readable time

kellhogs:
Add's timmer.currentTime as human readable timestring to the Timer RuntimeStore to be accesed by the websocket api #349

### DIFF
--- a/apps/server/src/utils/time.ts
+++ b/apps/server/src/utils/time.ts
@@ -11,7 +11,7 @@ export const DAY_TO_MS = 86400000;
  * @param {string} string - time string "23:00:12"
  * @returns {boolean} string represents time
  */
-export const isTimeString = (string) => {
+export const isTimeString = (string: string): boolean => {
   // ^                   # Start of string
   // (?:                 # Try to match...
   //  (?:                #  Try to match...
@@ -31,7 +31,8 @@ export const isTimeString = (string) => {
  * @argument {string} date - excel string date
  * @returns {number} - time in milliseconds
  */
-export const dateToMillis = (date) => {
+
+export const dateToMillis = (date: Date): number => {
   const h = date.getHours();
   const m = date.getMinutes();
   const s = date.getSeconds();
@@ -44,7 +45,7 @@ export const dateToMillis = (date) => {
  * @param valueAsString
  * @return {number}
  */
-const parse = (valueAsString) => {
+const parse = (valueAsString: string): number => {
   const parsed = parseInt(valueAsString, 10);
   if (isNaN(parsed)) {
     return 0;
@@ -58,7 +59,7 @@ const parse = (valueAsString) => {
  * @param {boolean} fillLeft - autofill left = hours / right = seconds
  * @returns {number} - time string in millis
  */
-export const forgivingStringToMillis = (value, fillLeft = true) => {
+export const forgivingStringToMillis = (value: string, fillLeft = true): number => {
   let millis = 0;
 
   // split string at known separators    : , .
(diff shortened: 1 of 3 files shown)


apps/server/src/utils/time.ts
```diff
@@ -31,7 +31,8 @@ export const isTimeString = (string) => {
  * @argument {string} date - excel string date
  * @returns {number} - time in milliseconds
  */
-export const dateToMillis = (date) => {
+
+export const dateToMillis = (date: Date): number => {
```
kellhogs, line 35: @cpvalente did i type this correctly? jsdocs states date to be a string but on a string `getHours()` cant be called so i made it a `Date` type

cpvalente: Yes, that is correct. The dateToMillis() seems to be called in parseExcelDate with a date object. good stuff!

apps/server/src/utils/time.ts
```diff
@@ -106,13 +107,48 @@ export const forgivingStringToMillis = (value, fillLeft = true) => {
...
+  if (hideZero && hours < 1) return [minutes, s % 60].map(format).join(':');
+  return [hours, minutes, s % 60].map(format).join(':');
+}
+
+/**
+ * @description Converts milliseconds to seconds -- Copied from client code
+ * @param {number | null} millis - time in seconds
+ * @returns {number} Amount in seconds
+ */
+export const millisToSeconds = (millis: number | null): number => {
```
cpvalente, line 149: are this copies of the ones in the client? should we move them to the shared code instead?
Also, maybe we should leverage luxon for the parsing instead?
I can see people start asking for AM PM stuff next

cpvalente:
I should have given some more directions here. Sorry for not being more specific on the implementation details.

I feel we shouldn't be adding this to the timer service. See below:

We should avoid derived data in our store, it risks bloating the object and makes the services unperformant and spammy
This is why we deleted isNegative
The consumers of the runtime data (eg. companion) will have some capacity to calculate the derived data, even if slightly inconvenient.

For the integrations feature on the other hand, most consuming devices (eg. vMix, d3...) are unable to parse data. So we would need it to have a friendly format. This is where the aliases come in handy

Going back to the feature, I believe this behaviour should be the responsibility of the integrationService.

Implementation could look something like

we create a new object definition with some aliases we find handy. maybe like
```
{
  alias.runningTimer
  alias.startTime
  alias.endTime
  alias.expectEnd
  alias.duration
  alias.clock
}
```

Our `parseTemplateNested` function could check for the key `alias` which is not part of the store and have a small lookup for resolving the values

Hope this helps

cpvalente (commented):
Few notes on implementation. Thank you for your work

cpvalente:
@kellhogs , it doesnt look like I can request your review since this is your PR. but would you mind checking my code?
Lots of shortcuts...

cpvalente:
Hi @kellhogs , can I please ask you for some sanity checking here?
The documentation on how to use the feature is here https://ontime.gitbook.io/v2/control-and-feedback/integrations#human-readable-data

kellhogs (commented):
Code looks good to me. But if im not mistaken we are not publishing this on the websocket yet. the originial issue was asking for this to be published on the websocket

cpvalente:
> Code looks good to me. But if im not mistaken we are not publishing this on the websocket yet. The originial issue #349 was asking for this to be published on the websocket

Great catch, I have renamed the PR to reflect the code.
We don't have a websocket integration just yet and will not have one until we have proven the stability of the OSC integration.

Thank you again

apps/server/src/services/integration-service/integrationUtils.ts
```diff
@@ -1,34 +1,62 @@
...
-    const variableName = match[1];
-    if (Object.hasOwn(state, variableName)) {
-      parsedTemplate = parsedTemplate.replace(match[0], state[variableName]);
-    }
-  }
+function formatDisplayFromString(value: string, hideZero = false): string {
+  let valueInNumber = null;
 
-  return parsedTemplate;
+  if (value !== 'null') {
```
hackertronix, line 9: this if block feels redundant and makes the code less readable... you should probably just flip the check and return nulll if there is no value instead of complicating this :)

PR #406, opened 2023-05-20
Timer/blinking and blackout buttons

kellhogs:
add buttons in the editor to let the timer text blink and to blackout the timer


cpvalente:
You can use rebase to fix this. 
The easiest is probably to cherry pick your commits into a new branch. If you are uncomfortable with the terminal, maybe some tool would help?

I think vscode has some nice plugins which would help with this

I can take another look once I have a clearer picture of the changes, but as of now:
- this feature should be behind an option. I am personally very much against aggressive styling and would not want anything flashing
- there shouldn't be a need to change anything in the timer service. We should be skeptical of the need for changes here since it is one of the most performance critical part of the app

kellhogs:
The timer service is in there because i messed up. Basically just the last commit is relevant to this

kellhogs:
Maybe you have more insight on why this test fails @cpvalente from the looks of my manual testing it should get through. Running the test locally results in more test failing for me 😕

apps/client/src/theme/_viewerDefs.scss
```diff
@@ -26,3 +26,16 @@ $element-border-radius: 8px;
...
+
+.blinking {
+  animation: blinking $timer-blinking-time linear infinite;
+}
+
+@keyframes blinking {
+  50% {
+    opacity: 0;
+  }
+}
```
cpvalente, line 39: not sure the shadowing here is a good idea.
how about the class just be called blink?

apps/client/src/features/control/message/MessageControl.tsx
```diff
@@ -21,6 +25,30 @@ export default function MessageControl() {
...
+        <label className={style.blackoutLabel}>Blackout timer screens</label>
+        <Button
+          className={style.blackoutButton}
+          variant={data.timerMessage.timerBlackout ? 'ontime-filled' : 'ontime-subtle'}
+          leftIcon={data.timerMessage.timerBlackout ? <IoEyeOffOutline size='24px' /> : <IoEye size='24px' />}
+          onClick={() => setMessage.timerBlackout(!data.timerMessage.timerBlackout)}
+          data-testid='toggle timer blackout'
+        >
+          {/* {data?.onAir ? 'Ontime is On Air' : 'Ontime is Off Air'} */}
+        </Button>
```
cpvalente, line 50: More nitpicking:

- can you remove commented code? you can then also use the self closing version of the elements like <Button />
- I think the blink button should also blink, to be SUPER annoying so that the operator doesn't forget it on
- How about the icon called `IoSunny` and `IoSunnyOutline` to make it blink?
- How about the icon called `IoMoon` for the blackout?

apps/client/src/features/control/message/MessageControl.module.scss
```diff
@@ -13,6 +13,36 @@
...
+    font-size: $inner-section-text-size;
+  color: $label-gray;
+}
+
+.blackoutButton {
+  grid-area: 2 / 2 / 3 / 3;
+  width: 100%;
+  justify-self: end;
+}
+
```
cpvalente, line 45: Mega nitpick list:

- Could you use grid-template-areas here. I find it so hard to think about this way of defining grid
- We should either a) decrease space between first and second row or b) increase space between timer stuff and public stuff. I am inclined towards a) as a better option
- mind indentation

packages/types/src/definitions/runtime/MessageControl.type.ts
```diff
@@ -2,3 +2,10 @@ export type Message = {
   text: string;
   visible: boolean;
 };
+
+export type TimerMessage = {
+  text: string;
+  visible: boolean;
+  timerBlink: boolean;
+  timerBlackout: boolean;
+}
```
cpvalente, line 11: we should tie these two definitions together so that TimerMessage is an extension of Message, like

```
export type TimerMessage = Message & {
  timerBlink: boolean;
  timerBlackout: boolean;
}
```

cpvalente (commented):
Lots of nitpicking. Great work here

apps/client/src/features/control/message/MessageControl.tsx
kellhogs: on the sunny symbol i can aggree, for the blackout i actually like the eye symbol more if you dont mind

cpvalente: I don't. Thank you for sticking up for what you think it's best

cpvalente:
Hi @kellhogs , can you please double-check my changes are aligned with your ideas?
I made mostly small style tweaks

apps/client/src/theme/_v2Styles.scss
```diff
@@ -60,6 +60,6 @@ $text-body-size: 15px;
 
 @keyframes blink {
   50% {
-    opacity: 0;
+    opacity: 40%;
```
kellhogs, line 63: I think having the opacity only go down to 40% here makes it to subtle here.
I think this feature should intenionally be super annoying since it's purpose is to rise awarness from the speake to the timer

cpvalente: That is a good point.
The reason I took it up is because the button disappears in the interface, which I find a bit strange.

Do you think there should be different settings for editor and timer? Or would 20% everywhere work?

kellhogs: In my oppinion 20% everywhere would work fine

kellhogs:
Your changes look very good to me @cpvalente 
Moving the timer message and the buttons down was a good call.

and thank you for finishing my work

PR #407, opened 2023-05-21
Seconds for progressbar input

kellhogs:


apps/client/src/features/modals/settings-modal/TimeInput.tsx
kellhogs, line 1: This component is supposed to format the input like 00:00 and the take an input in this format. If the user just put's in 70 for example we assume it's seconds wich would result in 01:10

i couldn't figure out with my `OnBlur()` and` OnChange() ` functions how to properly integrate this into the form

cpvalente:
Hi @kellhogs , would you be available to pick this up?

Issue #403, opened 2023-05-20
Swap event data


cpvalente:
Given a rundown with two events, we are currently able to swap the position of the event using the drag-and-drop feature.
However, there is no way to swap two events, while keeping the schedule. ie:

```
Event A > 09:00 - 09:30
Event B > 09:30 - 10:00
Event C > 10:00 - 10:30
```

If Event B were to run late, we would like to swap its place with Event C, keeping the schedule unchanged so that the rundown would read:
```
Event A > 09:00 - 09:30
Event C > 09:30 - 10:00
Event B > 10:00 - 10:30
```

### Requirements
- The user should be able to use the context menu to swap the data between two <EventBlocks>
- The timer (start, end, duration) data should remain as before, and only the remaining information should be swapped
- We would create a REST endpoint specifically for this action

### UI UX guidelines
We do not have a clear picture in UI as of now. Few guidelines:

On right-clicking an event block, we would have a new action: "Add to swap", which would add the `eventId` to the state of the context
If we right click an EventBlock, and we have a previously added an `eventId`, we would now see an added action `Swap with $eventId`

This would trigger the new endpoint created with both event Ids (to and from)

cpvalente:
@asharonbaltazar 
I have made some revisions here, hoping to get your feedback. 

We would like our `ContextMenuContext` to remain generic, so it would not be ideal to create implementation-specific logic in there.
I wonder if a good way forwards is to create a small Zustand store that uses the `ContextMenuContext`

asharonbaltazar:
If you are referring to the feature where a user can select events to swap using the context menu and ontime needs a way to track the selections, I think a Zustand store makes sense. Also, if I had known you were using Zustand, I'd probably written `ContextMenuContext` to use it as well.

asharonbaltazar:
> If Event B were to run late, we would like to swap its place with Event C, keeping the schedule unchanged so that the rundown would read:
```ts
Event A > 09:00 - 09:30
Event C > 09:30 - 10:00
Event A > 10:00 - 10:30
```

Is there a typo here? Did you mean to write:
```ts
Event A > 09:00 - 09:30
Event C > 09:30 - 10:00
Event B > 10:00 - 10:30
```

Also, what did you mean by this?
> The user should be able to use drag-and-drop to swap the data between two 

Did you mean the present DND functionality should remain intact, and this new feature should be added to the context menu? Or did you mean changing the current DND functionality to swap the time between the event blocks and add the same functionality to the context menu?

[...]

Issue #388, opened 2023-05-18
Event Editor: fields show stale data


cpvalente:
The Info field in the event editor does not seem to reset when moving across events

- Enable edit mode on the <MenuBar> (left menu)
- Select an event in rundown. The event editor (bottom drawer) should now open
- Edit the event info field eg: "test info"
- Select another event in rundown without closing the event Editor

You will now see that "test info" remains in the text area. we would have expected this to be cleared
We should also see if there is a potential for the same to happen with other fields

asharonbaltazar:
Assign me, please

asharonbaltazar:
I see three fields – Title, Presenter, and Subtitle (and Note to the right). 

So if I edit one of these inputs and quickly switch to another event, the edited input doesn't reset. This is the bug?

cpvalente:
That is right. When you move from editing an event to another, the input fields data should change to reflect the new field
This is very obviously seen in the note field and in some cases the title - presenter and subtitle

I would suspect that the remaining inputs (colour and times) could also be vulnerable

Issue #358, opened 2023-04-23
Help needed: Change server port during runtime


cpvalente:
Currently, we use port 4001 for the [ontime express server](https://github.com/cpvalente/ontime/blob/master/apps/server/src/app.ts#L137).
The express server is responsible for serving the react app bundle and hosting most of the APIs (REST, WebSocket)

The React app expects to be able to find these resources at that port.

The challenge here is that, were the user able to change the runtime port, we would need to communicate this to front-end during runtime. Something like:
- Using the UI, the user changes the port for ontime
- The nodejs server would restart using the new port for the express app
- The react client would then connect to the new port

If any of this fails, both the server and the react app need to fallback to the last working port

We are looking for suggestions on how to best achieve this behaviour. PRs on this are welcome

Refs #262

cpvalente:
Hi @asharonbaltazar, first of all thank you for the initiative.
This is something that we really want as part of the application but I do not have a clear picture on what it would entail.


### Adding some more pointers to the description of the issue
We currently piggyback from the express app to serve our REST API and Websocket.
Ontime uses REST to distribute its persisted data (project file) and WebSocket to distribute the runtime data. These changes in WebSockets happen often and are initialised by the server

The [express server is locked to port 4001](https://github.com/cpvalente/ontime/blob/master/apps/server/src/app.ts#L134).
Changing the port here should be trivial, we also have [a field in the project file](https://github.com/cpvalente/ontime/blob/master/demo-db/db.json#L41) and [a form in the UI](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/modals/settings-modal/AppSettings.tsx#L68) ready for letting the user defining the port. We would also need to [redefine our constants](https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/api/apiConstants.ts#L1)

### The (maybe) challenge
The UI needs to know which port to reach to get the react app, since the built static files are served by express.

It potentially is just a matter of the user changing the port in the URL like `localhost:4001` -> `localhost:9999`

**Some (maybe) edge cases**
- Most users consume the [app through electron](https://github.com/cpvalente/ontime/blob/master/apps/electron/electron.config.js#L7). we would need to know in electron which port is being used at start
- We would also need to make sure the [docker distribution](https://github.com/cpvalente/ontime/blob/master/Dockerfile#L20) remains working

In short, I am unsure on what challenges this feature actually poses. There are several moving targets to look after
It might well "just work"

Hope this helps

[...]

Issue #100, opened 2022-02-27
Over/Undertime


Keksstar:
Would be nice to have the total time of all events somewhere to see if you´re running short or too long... Useful for Streamingevents where you need to stay in defined times.

cpvalente:
Hey @Keksstar, thank you for your suggestion.
It sounds good to me, could you please describe the feature a bit more in depth?
- Would it be just a sum of all existing timers?
- How would it handle the current delay feature?
- Is this related to roll mode only?
- What is the format of the information and in what context should it appear (in the app? in some of the screens?)

Thank you again, there is some other work planned related to timers and I can probably fit this in there

Keksstar:
Hi @cpvalente , thanks for the fast reply!

In general I see this as an additonal mode.

Lets say we have an Event with Start 8pm, End 10pm given with four speakers.
Speaker 1 has 30min
Speaker 2 has 30min
Speaker 3 has 30min
Speaker 4 has 30min

We start timer for speaker 1, he´s really fast and only needs 20min. This creates an undertime of 10min. - This is something the director needs to know and also it´s nice to know for the speaker that he doesn´t need to hurry. So both should see it. (and all others in the backoffice) - But it´s not an public information. (should be displayed as "-10min")
Speaker 2 loves his work, and talks full 60min. This creates an overtime of +20min as we already had 10min undertime and he wasted this time completly....
So only 40min left for Speaker 3+4.
and so on, I think you get the point.

The beginning time of the next event should automatically adjusted.
Let´s say Speaker 1 starts at 8pm and has 30min, Speaker 2 start at 8:30pm... but now Speaker 1 needs 35min, so the new starting time for Speaker 2 adjusts itself to 8:30+5=8:35pm
Or vice versa when he´s faster.
Also times needs to be adjusted if you skip or delete an entry.


So to answer your question:
- nope, see above
- Delay is just an "break", it´s like any other entry. It just get´s summed up with all other elements
- Nope, it´s way more than this - it´s an new mode.
- Sure in the app and on the screens that are not public.

I would like to use ontime as something similar like Shoflo, but with much less functionality.
It´s not just Speaker times, it´s everything of the event:
Intro time
Host time
VT time
Speaker time
Break time
etc.

This in sum is what I would like to use ontime for.
[...]

PR #410, opened 2023-05-22
Context Menu Feature

### LINKED ISSUE #389: Feat: Create context menu for <EventBlock> component
We want to allow quick actions on events without opening the event editor.

The planned strategy is to have a custom context menu which appears when right-clicking the event.

Development tasks:
- Create a reusable event context menu component
- create logic to show menu when right-clicking on <EventBlock>
- show a single entry in the context menu which lists the clicked event ID

We will use this as a springboard to allow quick actions on selection and, eventually multiple selections.

asharonbaltazar:
closes #389 

#### Changelog
- create `ContextMenuProvider` component responsible for determining look, state, position, and options of context menu. 
- wrap `App.tsx` with `ContextMenuProvider` 
- create `useContextMenu` hook to access said context. It receives an `Option[]` arg and returns an event handler that should be passed to a `onContextMenu`. (However, the returned event handler can be used for any of the element's event callbacks, but perhaps this flexibility is welcome.)

The above implementation does have an issue with positioning. I spent a few hours trying to pinpoint this bug, but I believe it comes down to how the elements on the page are positioned. I'd love some help with this!

https://github.com/cpvalente/ontime/assets/58940073/fce9cecf-5cad-4dcb-9dab-963df8285673

Happy to discuss further feedback!

### DIFF
--- a/apps/client/src/App.tsx
+++ b/apps/client/src/App.tsx
@@ -6,7 +6,6 @@ import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
 
 import ErrorBoundary from './common/components/error-boundary/ErrorBoundary';
 import { AppContextProvider } from './common/context/AppContext';
-import { ContextMenuProvider } from './common/context/ContextMenuContext';
 import useElectronEvent from './common/hooks/useElectronEvent';
 import { ontimeQueryClient } from './common/queryClient';
 import { connectSocket } from './common/utils/socket';
@@ -50,18 +49,16 @@ function App() {
     <ChakraProvider resetCSS theme={theme}>
       <QueryClientProvider client={ontimeQueryClient}>
         <AppContextProvider>
-          <ContextMenuProvider>
-            <BrowserRouter>
-              <div className='App'>
-                <ErrorBoundary>
-                  <TranslationProvider>
-                    <AppRouter />
-                  </TranslationProvider>
-                </ErrorBoundary>
-                <ReactQueryDevtools initialIsOpen={false} />
-              </div>
-            </BrowserRouter>
-          </ContextMenuProvider>
+          <BrowserRouter>
+            <div className='App'>
+              <ErrorBoundary>
+                <TranslationProvider>
+                  <AppRouter />
+                </TranslationProvider>
+              </ErrorBoundary>
+              <ReactQueryDevtools initialIsOpen={false} />
+            </div>
+          </BrowserRouter>
         </AppContextProvider>
       </QueryClientProvider>
     </ChakraProvider>
--- a/apps/client/src/App.tsx
+++ b/apps/client/src/App.tsx
@@ -6,6 +6,7 @@ import { ReactQueryDevtools } from '@tanstack/react-query-devtools';
 
 import ErrorBoundary from './common/components/error-boundary/ErrorBoundary';
 import { AppContextProvider } from './common/context/AppContext';
+import { ContextMenuProvider } from './common/context/ContextMenuContext';
 import useElectronEvent from './common/hooks/useElectronEvent';
 import { ontimeQueryClient } from './common/queryClient';
 import { connectSocket } from './common/utils/socket';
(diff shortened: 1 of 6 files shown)


apps/client/src/App.tsx
```diff
@@ -49,16 +50,18 @@ function App() {
...
-            <div className='App'>
-              <ErrorBoundary>
-                <TranslationProvider>
-                  <AppRouter />
-                </TranslationProvider>
-              </ErrorBoundary>
-              <ReactQueryDevtools initialIsOpen={false} />
-            </div>
-          </BrowserRouter>
+          <ContextMenuProvider>
```
cpvalente, line 53: We don't need to wrap the entire app in the context menu.
Wrapping the element with `data-testid:"event-editor"` in `Editor.tsx` would be enough

Although your solution would have generally been correct, in ontime's case, the views are not interactive and often run in devices like TVs. No need to add code there

cpvalente (changes requested):
Honestly, the implementation here is very clean, and I really like your approach.
I have nothing to comment on other than a minor suggestion on the context placement

I also looked into the positioning issue you mentioned.
As far as I could test, the IDs always resolved correctly. The only thing I saw in your video was that the context menu does not follow the screen. This is common behaviour for a non-browser context menu and not an issue for us

We could potentially attach an event listener to the scroll and close the menu. But I wonder whether it would be worth it 

The render time for the context menu is not great (around 20ms on average) but I guess this is due to Chakra UI. You did great into lazy loading the menu so that there is no initial performance hit

Altogether this is great work and I am keen to integrating it into the app.

We would need to add some actual behaviour into the context menu, but we should do that separately.

I look forwards to seeing more code from you

apps/client/src/App.tsx
asharonbaltazar: Addressed with https://github.com/cpvalente/ontime/pull/410/commits/c9f0b08e5da9e4f0e1c297758c0c10b805f43740

asharonbaltazar:
> We could potentially attach an event listener to the scroll and close the menu. But I wonder whether it would be worth it

We could emulate the browser behavior and prevent any scrolling when the context menu is open

cpvalente:
> > We could potentially attach an event listener to the scroll and close the menu. But I wonder whether it would be worth it
> 
> We could emulate the browser behavior and prevent any scrolling when the context menu is open

That is a better idea, thank you for looking out.
Would you like to implement?

asharonbaltazar:
>  Wrapping the element with data-testid:"event-editor" in Editor.tsx would be enough

BTW, `/rundown` doesn't work as it's not sharing the `ContextMenuProvider` context

cpvalente:
> If you're able to give it a test, that'd be much appreciated!

Great work, it looks perfect to me.
Reverting to App.tsx is probably the best for now.

asharonbaltazar:
>  Reverting to App.tsx is probably the best for now.
Reverted

cpvalente:
Thank you again @asharonbaltazar 
I will merge this as is


Could I interest you on another issue? Perhaps something more challenging?

asharonbaltazar:
> Could I interest you on another issue? Perhaps something more challenging?

Let's give it a go. But depending on how difficult it is, I can't promise anything in terms of a timeline.

cpvalente (approved):

cpvalente:
How about one of these?

They are still good in the way that you don't have to know much about the app to be able to do it.

https://github.com/cpvalente/ontime/issues/388 <---- React only
https://github.com/cpvalente/ontime/issues/403 <---- React + node

asharonbaltazar:
#388 looks tame enough. I'll start with that. Can you assign the issue to me?

cpvalente:
Since you are not a collaborator, I am unable to assign it to you without you being involved in the issue. Would you mind dropping a comment in the ticket?

cpvalente:
Sure, if you are up for a challenge, this is it https://github.com/cpvalente/ontime/issues/358
Otherwise we can have a chat, next big task in my list is https://github.com/cpvalente/ontime/issues/100#issuecomment-1556267435

asharonbaltazar:
Alrighty, I'll keep your comment as a bookmark. I'll work on the form ticket, and once finished, I'll check back here. 
Appreciate the quick follow-ups!

PR #412, opened 2023-05-22
Fields' Stale Data Fix

### LINKED ISSUE #388: Event Editor: fields show stale data

asharonbaltazar:
closes #388 

#### Changelog
- add `event.id` as `key` to `<EventEditorTitles />`

React will now know to re-render this component whenever the selected event changes

(I cannot reproduce this issue with `<EventEditorTimes />`, but I suspect if this bug pops up again, this fix will probably address it).

### DIFF
--- a/apps/client/src/features/event-editor/EventEditor.tsx
+++ b/apps/client/src/features/event-editor/EventEditor.tsx
@@ -58,6 +58,7 @@ export default function EventEditor() {
         timerType={event.timerType}
       />
       <EventEditorTitles
+        key={event.id}
         eventId={event.id}
         title={event.title}
         presenter={event.presenter}


cpvalente:
Perfect, I seem to remember there was something in the React docs with this exact case.
You made it look easy, good work!

PR #417, opened 2023-05-27
Allow Switch from Roll Mode to Play Mode

bunday:


cpvalente:
I think the PR solves the issue. it would just need some small modifications in the UI to allow the user to pause a rolling timer

There is a confusing side effect where the automation that counts on ontime would receive a pause if the user makes the switch.
This is some business knowledge that I cannot expect you to have and I should have clarified

We would need the playback to go from "roll" to "play" without any intermediate steps.
We would also need to make sure that we do not trigger any of the sideEffects, namely `_onStart()`
https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/TimerService.ts#L176

To be able to test, the UI needs to allow for this. that is a pretty straightforward change
https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/control/playback/playback-buttons/PlaybackButtons.tsx#L56


Thank you again for your work, please let me know if I can help unblock anything here

cpvalente:
Hi @bunday, any updates here? Could I offer some help to push this through

PR #418, opened 2023-05-28
Fix websocket connection behind HTTPS reverse proxy

markspolakovs:
apiConstants.ts currently hardcodes the websocket protocol as `ws://` which fails if the ontime page is loaded on HTTPS (as the browser rejects it as mixed content). This PR ensures `wss://` is used if the page is loaded over HTTPS.


cpvalente:
Hi @markspolakovs, this is an excellent PR on a detail that we overlooked 

Thank you for this

PR #420, opened 2023-05-30
URL Params Form

### LINKED ISSUE #399: Create dynamic form for modifying URL parameters
### Why

Currently, users can modify most of the view routes through URL parameters
What can be defined is in the documentation, which tends to become out of date as we go along. The process is also very error prone and generally unfriendly

### The task

We want to create a reusable form component which posts its data to the current hostname URL params.

**Requirements**
- The form should receive an object containing field and type, and generate a form to modify the given fields with an appropriate element (eg: text input for strings, switches for booleans). 
- On submitting the form, the fields and values should be pushed to the URL parameters of the current host
- We can leave validation for later

### Technical details

The URL params are now read inline in each view component. That is not ideal, but we will leave it as it is for now. 
We will use them to extract from which params a view have. We hope that the code and comments will suffice for documentation here. You can also see the documentation under [View Modifiers](https://ontime.gitbook.io/v2/views/minimal-timer)

Eg: for /minimal (the component with most overrides)
```
const key = searchParams.get('key');
const offsetX = searchParams.get('offsetx');
const alignX = searchParams.get('alignx');
const hideOvertime = searchParams.get('hideovertime');
```

Could be described like
```
[
  {
    field: 'key',
    description: 'Key colour (as an hex value, eg: #FF0000)',
    type: 'string',
  },
  {
    field: 'offsetx',
    description: 'Horizontal offset (-1000, 1000)',
    type: 'number',
  },
  {
    field: 'alignx',
    description: 'Horizontal alignment of timer',
    type: 'option',
    values: ['start', 'center', 'end']
  },
  {
    field: 'hideovertime',
    description: 'Whether styles should change in overtime',
    type: 'boolean',
  },
];
```

and generate an URL looking like
`localhost:4001/minimal?key=#ff0000&offsetx=10&alignx=start&hideovertime=false`

We can skip validation of the fields for the first draft, but if you are very keen we can leverage React Hook Form validation stuff here, like
```
  validation: {
    min: -2000,
    max: 2000,
    validate: () => function to validate
  }
```

asharonbaltazar:
closes #399 

#### Changelog (so far):

- Create `EditFormDrawer` to house the inputs. Add to `/clock` view for testing (will add to more)
- Create `EditFormInput` to handle granulated logic for interacting with search params.
- Add edit button to `NavigationMenu`. Had to adjust styles to make it possible

#### Some thoughts
- Since I'll be adding `EditFormDrawer` to every clock/timer route (despite not being so DRY, which apparently isn't a concern), I simplified logic by passing the `options` prop to `EditFormDrawer`. 
- Styling is a bit difficult due to the amount of files required. I'm trying my best to emulate other `.scss` files.

I opened this PR so it's obvious this issue isn't in limbo and to serve as a source of motivation to complete this feature within a reasonable time. 

Please let me know if you think this a good step in the right direction. Open to criticism, obviously.

### DIFF
--- a/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
+++ b/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
@@ -63,11 +63,13 @@ export default function EditFormDrawer({ options }: EditFormDrawerProps) {
           </form>
         </DrawerBody>
 
-        <DrawerFooter className={style.drawerContent}>
-          <Button variant='outline' mr={3} onClick={onEditDrawerClose}>
+        <DrawerFooter className={style.drawerFooter}>
+          <Button variant='ontime-subtle' onClick={onEditDrawerClose}>
             Cancel
           </Button>
-          <Button colorScheme='blue'>Save</Button>
+          <Button variant='ontime-filled' form='edit-params-form' type='submit'>
+            Save
+          </Button>
         </DrawerFooter>
       </DrawerContent>
     </Drawer>
--- a/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
+++ b/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
@@ -9,13 +9,19 @@ import {
   DrawerFooter,
   DrawerHeader,
   DrawerOverlay,
-  Input,
   useDisclosure,
 } from '@chakra-ui/react';
 
+import EditFormInput from './EditFormInput';
+import { Field } from './types';
+
 import style from './EditFormDrawer.module.scss';
 
-export const EditFormDrawer = () => {
+interface EditFormDrawerProps {
+  options: Field[];
+}
+
+export default function EditFormDrawer({ options }: EditFormDrawerProps) {
   const [searchParams, setSearchParams] = useSearchParams();
   const { isOpen, onClose, onOpen } = useDisclosure();
 
@@ -44,7 +50,17 @@ export const EditFormDrawer = () => {
         </DrawerHeader>
 
         <DrawerBody className={style.drawerContent}>
-          <Input placeholder='Type here...' variant='ontime-filled' />
+          <form id='edit-params-form'>
+            {options.map((field) => (
+              <div key={field.title} className={style.columnSection}>
+                <label className={style.label}>
+                  <span className={style.title}>{field.title}</span>
+                  <span className={style.description}>{field.description}</span>
+                </label>
+                <EditFormInput field={field} />
+              </div>
+            ))}
+          </form>
         </DrawerBody>
 
         <DrawerFooter className={style.drawerContent}>
@@ -56,4 +72,4 @@ export const EditFormDrawer = () => {
       </DrawerContent>
     </Drawer>
   );
-};
+}
--- a/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
+++ b/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
@@ -13,15 +13,15 @@ import {
 } from '@chakra-ui/react';
 
 import EditFormInput from './EditFormInput';
-import { Field } from './types';
+import { ParamField } from './types';
 
 import style from './EditFormDrawer.module.scss';
 
 interface EditFormDrawerProps {
-  options: Field[];
+  paramFields: ParamField[];
 }
 
-export default function EditFormDrawer({ options }: EditFormDrawerProps) {
+export default function EditFormDrawer({ paramFields }: EditFormDrawerProps) {
   const [searchParams, setSearchParams] = useSearchParams();
   const { isOpen, onClose, onOpen } = useDisclosure();
 
@@ -79,14 +79,14 @@ export default function EditFormDrawer({ options }: EditFormDrawerProps) {
         </DrawerHeader>
 
         <DrawerBody className={style.drawerContent}>
-          <form id='edit-params-form' onSubmit={onSettingsSubmit}>
-            {options.map((field) => (
+          <form id='edit-params-form' onSubmit={onParamsFormSubmit}>
+            {paramFields.map((field) => (
               <div key={field.title} className={style.columnSection}>
                 <label className={style.label} htmlFor={field.id}>
                   <span className={style.title}>{field.title}</span>
                   <span className={style.description}>{field.description}</span>
                 </label>
-                <EditFormInput key={field.title} field={field} />
+                <EditFormInput key={field.title} paramField={field} />
               </div>
             ))}
           </form>
--- a/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
+++ b/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
@@ -1,0 +1,55 @@
+import { useEffect } from 'react';
+import { useSearchParams } from 'react-router-dom';
+import {
+  Button,
+  Drawer,
+  DrawerBody,
+  DrawerCloseButton,
+  DrawerContent,
+  DrawerFooter,
+  DrawerHeader,
+  DrawerOverlay,
+  Input,
+  useDisclosure,
+} from '@chakra-ui/react';
+
+export const EditFormDrawer = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const { isOpen, onClose, onOpen } = useDisclosure();
+
+  useEffect(() => {
+    const isEditing = searchParams.get('edit');
+
+    if (isEditing === 'true') {
+      return onOpen();
+    }
+  }, [searchParams, onOpen]);
+
+  const onEditDrawerClose = () => {
+    onClose();
+
+    searchParams.delete('edit');
+    setSearchParams(searchParams);
+  };
+
+  return (
+    <Drawer isOpen={isOpen} placement='right' onClose={onEditDrawerClose}>
+      <DrawerOverlay />
+      <DrawerContent>
+        <DrawerCloseButton />
+        <DrawerHeader>Create your account</DrawerHeader>
+
+        <DrawerBody>
+          <Input placeholder='Type here...' />
+        </DrawerBody>
+
+        <DrawerFooter>
+          <Button variant='outline' mr={3} onClick={onEditDrawerClose}>
+            Cancel
+          </Button>
+          <Button colorScheme='blue'>Save</Button>
+        </DrawerFooter>
+      </DrawerContent>
+    </Drawer>
+  );
+};
--- a/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
+++ b/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
@@ -71,11 +71,11 @@ export default function EditFormDrawer({ paramFields }: EditFormDrawerProps) {
           <form id='edit-params-form' onSubmit={onParamsFormSubmit}>
             {paramFields.map((field) => (
               <div key={field.title} className={style.columnSection}>
-                <label className={style.label} htmlFor={field.id}>
+                <label className={style.label}>
                   <span className={style.title}>{field.title}</span>
                   <span className={style.description}>{field.description}</span>
+                  <EditFormInput key={field.title} paramField={field} />
                 </label>
-                <EditFormInput key={field.title} paramField={field} />
               </div>
             ))}
           </form>
--- a/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
+++ b/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
@@ -45,17 +45,6 @@ export default function EditFormDrawer({ paramFields }: EditFormDrawerProps) {
 
     const newParamsObject = Object.fromEntries(new FormData(formEvent.currentTarget));
     const newSearchParams = Object.entries(newParamsObject).reduce((newSearchParams, [id, value]) => {
-      const isIdABoolean = paramFields.some((field) => field.id === id);
-
-      // special handling for <Switch /> (aka checkboxes)
-      // unchecked checkboxes DO NOT have a value
-      // checked checkboxes will be 'true' (see <EditFormInput />)
-      if (isIdABoolean && value === 'true') {
-        newSearchParams.set(id, value);
-
-        return newSearchParams;
-      }
-
       if (typeof value === 'string' && value.length) {
         newSearchParams.set(id, value);
 
--- a/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
+++ b/apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
@@ -13,6 +13,8 @@ import {
   useDisclosure,
 } from '@chakra-ui/react';
 
+import style from './EditFormDrawer.module.scss';
+
 export const EditFormDrawer = () => {
   const [searchParams, setSearchParams] = useSearchParams();
   const { isOpen, onClose, onOpen } = useDisclosure();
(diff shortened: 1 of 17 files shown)


asharonbaltazar:
Do all components using the `NavigationMenu` have an edit button? If not, I'll add an optional `hideEditButton` prop to hide it

cpvalente:
> Do all components using the `NavigationMenu` have an edit button? If not, I'll add an optional `hideEditButton` prop to hide it

Good observation. The countdown is the only case without style params, in this view params are used for selecting an event to follow

cpvalente:
Hi, I will pull the code and take a look. Meanwhile I can give some thoughts as for the strategy
 
> closes #399
> 
> #### Changelog (so far):
> * Create `EditFormDrawer` to house the inputs. Add to `/clock` view for testing (will add to more)
> * Add edit button to `NavigationMenu`. Had to adjust styles to make it possible
> 
> #### Some thoughts
> * Since I'll be adding `EditFormDrawer` to every clock/timer route (despite not being so DRY, which apparently isn't a concern), I simplified logic by passing the `options` prop to `EditFormDrawer`.

DRY, as any principal, is important with a grain of salt.

In the case of Ontime, we consider each viewer its own app, the viewers serve as templates for users to create their own viewers. They will likely not be doing it in react, but the javascript logic should be transferrable.

Since the viewers are all non interactive, we feel that performance is not an issue and would instead prioritise lower time to FCP

We also consider that logic repetition in the views is coincidental, preferring to be communicative over DRY.

In the case of this menu. This is specifically an app feature, not a view feature. And less interesting for users to replicate.
We can abstract away the details but declare it in the viewers, exactly as you have done

Of course all of this is highly subjective and I am open for being challenged and discussing it further


> * `options` will be a discriminated union.
> * Styling is a bit difficult due to the amount of files required. I'm trying my best to emulate other `.scss` files.
Can you expand here? I am not sure I understand the challenges in styling... 

Some questions from my side
- I see you chose to use a Chakra UI component with scss modules. Would it be easier to define styles in the chakra theme?
- Would it make it easier if "Edit view style" was an option inside the navigation menu? Similar to "Toggle Fullscreen" and "Flip Screen"? If so, would the UX be poorer?
- I am unsure about keeping the state of the form in the params, but I will wait to see your complete implementation.

Thank you again for your work, I believe this is a very good start.
Hope this helps and feel free to continue the conversation

asharonbaltazar:
> I see you chose to use a Chakra UI component with scss modules. Would it be easier to define styles in the chakra theme?

Could you please elaborate a bit more? I don't mind using `.module.scss` files – I was just commenting that tracking the source of a CSS variable is a bit confusing.

> Would it make it easier if "Edit view style" was an option inside the navigation menu? Similar to "Toggle Fullscreen" and "Flip Screen"? If so, would the UX be poorer?

Originally, I wanted to do that but decided against since those buttons apply changes to _all_ views. An edit button should only apply changes to a specific view, right? I'm aware there are shared keys (font size, color, etc), but some settings are unique to a single view (e.g. `/clock?format=12`). Therefore, there should be a special button for edit. Feel free to disagree!

> I am unsure about keeping the state of the form in the params, but I will wait to see your complete implementation.

I'm thinking of starting off without validations to maintain simplicity for this (initial PR) implementation. Without having to worry about that, I'll have inputs within a `form` element, and when a user presses "Save" (or "Submit"), use `new formData(formEvent.target)` to grab the form values and submit changes to the URL. Essentially and initially, there wouldn't be any state to keep track of, and changes only would be submitted to URL once upon confirmation.

cpvalente:
> > I see you chose to use a Chakra UI component with scss modules. Would it be easier to define styles in the chakra theme?
> 
> Could you please elaborate a bit more? I don't mind using `.module.scss` files – I was just commenting that tracking the source of a CSS variable is a bit confusing.

There are a few competing strategies here. I can speak as to the decision-making so far and hopefully that can help guide.
We use CSS modules for the views which we do not expect (or want) the users to modify. 
That means the `/editor` and `/cuesheet` routes, and things structural to the app (like the navigation menu)
All other things using scss with BEM to facilitate CSS overrides

For other things we chose one of the 2:
- Is the component design strong enough that we would like to make it ourselves, or is the component complex enough that Chakra performance hit is not desirable? Use CSS modules
- Is our use case simple and chakra UI provides a handy component? Can we group the styles into a few variations? Use Chakra with themes
By default we use CSS modules

What I meant, is that you might be able to use chakra theming to define your components in an easy way. We also don't use drawers any where else
https://github.com/cpvalente/ontime/blob/master/apps/client/src/theme/theme.ts

> 
> > Would it make it easier if "Edit view style" was an option inside the navigation menu? Similar to "Toggle Fullscreen" and "Flip Screen"? If so, would the UX be poorer?
> 
> Originally, I wanted to do that but decided against since those buttons apply changes to _all_ views. An edit button should only apply changes to a specific view, right? I'm aware there are shared keys (font size, color, etc), but some settings are unique to a single view (e.g. `/clock?format=12`). Therefore, there should be a special button for edit. Feel free to disagree!

If Edit was a property of the menu, I would assume that <NavigationMenu> would take in as a property, a nullable array of fields. That way the declaration still happens at the view level
I imagined that adding an option to the existing is more straightforward in terms of design, but that might well be from my lack of comfort when deciding UI / UX.
I don't have strong inclinations either way code wise and you are welcome to suggest what you feel comfortable with

> 
> > I am unsure about keeping the state of the form in the params, but I will wait to see your complete implementation.
> 
> I'm thinking of starting off without validations to maintain simplicity for this (initial PR) implementation. Without having to worry about that, I'll have inputs within a `form` element, and when a user presses "Save" (or "Submit"), use `new formData(formEvent.target)` to grab the form values and submit changes to the URL. Essentially and initially, there wouldn't be any state to keep track of, and changes only would be submitted to URL once upon confirmation.

This sounds good to me. I would have done the same.
As to my comment: I see I wasn't particularly clear and I apologise for that. I was referring to the `edit` parameter that you keep in the URL to flag whether the edit drawer is open.

Then again, this is quite an insignificant implementation detail that I wouldn't sweat, just got curious about the choice since you show a pretty comfortable level of react knowledge

asharonbaltazar:
`EditFormDrawer` can render inputs based on options! The inputs will grab their respective param and try to use it as a default value.

asharonbaltazar:
> I was referring to the edit parameter that you keep in the URL to flag whether the edit drawer is open.

Ah, well, in that case, I thought it might be useful to share the boolean in the link. Maybe someone will want to share an open menu so it's possible to immediately edit. 
And since it's handled by the params, I also avoid re-declaring a `useState` (or some other) hook in every Clock route.

apps/client/src/common/components/edit-form-drawer/constants.ts
```diff
@@ -0,0 +1,214 @@
...
+    type: 'number',
+  },
+  {
+    id: 'text',
+    title: 'Text',
+    description: 'Text colour in hexadecimal',
+    type: 'string',
+  },
+  {
+    id: 'text',
```
asharonbaltazar, line 188: Seems like there are two keys using `'text'`

apps/client/src/common/components/navigation-menu/NavigationMenu.tsx
```diff
@@ -49,63 +55,70 @@ export default function NavigationMenu() {
...
-              onKeyDown={(event) => {
-                isKeyEnter(event) && handleFullscreen();
-              }}
-            >
-              Toggle Fullscreen
-              {isFullScreen ? <IoContract /> : <IoExpand />}
+      <div className={`${style.buttonContainer} ${!showButton && !showMenu ? style.hidden : ''}`}>
+        <button onClick={toggleMenu} aria-label='toggle menu' className={style.navButton}>
+          <IoApps />
+        </button>
```
asharonbaltazar, line 68: For some reason, this button interferes with the space button when the `EditFormDrawer` is open. Instead of typing a space in an input, it keeps opening the menu. Any thoughts?

cpvalente:
Hey @asharonbaltazar , let me know when this is ready for code review. 
I am planning on making a new release the next couple of days

asharonbaltazar:
It's ready for review!

apps/client/src/common/components/edit-form-drawer/constants.ts
```diff
@@ -0,0 +1,214 @@
+import { ParamField } from './types';
+
```
cpvalente, line 2: There is probably no way for you to have known this, but there is a field missing in some views.
All views apart from `MinimalTimer` call formatTime which will allow the user to declare whether the times should be in 12 or 24 hour format
https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/utils/time.ts#L29

cpvalente (commented):
I was honestly expecting to find a rough feature needing lots of tidying up and style cleanups.

You have done a great work and I am very impressed with the overall quality, both in aesthetic and implementation decisions.

I have left a few comments for your consideration, mostly nitpicks. Thank you again for the great work. I am grateful to have you as a contributor to our project

apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+        return newSearchParams;
+      }
+
+      return newSearchParams;
+    }, new URLSearchParams());
+
+    onEditDrawerClose();
+    setSearchParams(newSearchParams);
+  };
```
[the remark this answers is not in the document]
asharonbaltazar: Thanks. I could add some comments, if you'd like.

apps/client/src/common/components/edit-form-drawer/EditFormInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <option key={value} value={value}>
+            {value}
+          </option>
+        ))}
+      </Select>
+    );
+  }
+
+  if (type === 'boolean') {
+    const defaultCheckedValue = searchParams.get(id) === 'true' ?? false;
```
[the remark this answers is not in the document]
asharonbaltazar: That's an easy fix. I'll push one up soon. 

Keeping users' expectations consistent is important.

apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <DrawerHeader className={style.drawerHeader}>
+          <DrawerCloseButton _hover={{ bg: '#ebedf0', color: '#333' }} size='lg' />
+          Customise
+        </DrawerHeader>
+
+        <DrawerBody className={style.drawerContent}>
+          <form id='edit-params-form' onSubmit={onParamsFormSubmit}>
+            {paramFields.map((field) => (
+              <div key={field.title} className={style.columnSection}>
+                <label className={style.label} htmlFor={field.id}>
```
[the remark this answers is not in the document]
asharonbaltazar: Woops, I meant to do that! My bad, and good catch.

apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
```diff
@@ -0,0 +1,95 @@
...
+          </form>
+        </DrawerBody>
+
+        <DrawerFooter className={style.drawerFooter}>
+          <Button variant='ontime-subtle' onClick={onEditDrawerClose}>
+            Cancel
+          </Button>
+          <Button variant='ontime-filled' form='edit-params-form' type='submit'>
+            Save
+          </Button>
```
cpvalente, line 90: Mega nitpick, could you add a 2em horizontal padding on this button? it fields it should be more proeminent

apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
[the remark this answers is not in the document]
asharonbaltazar: Addressed with https://github.com/cpvalente/ontime/pull/420/commits/79b25884a547a70cc1f78d237062e16c0ae6aeb5

apps/client/src/common/components/edit-form-drawer/EditFormInput.tsx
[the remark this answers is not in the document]
asharonbaltazar: Addressed with https://github.com/cpvalente/ontime/pull/420/commits/f118aaa10c68d0dbed7217b732a184af0ba81bfd

apps/client/src/common/components/edit-form-drawer/constants.ts
asharonbaltazar: Even `Countdown`?

cpvalente: yes, even countdown. The `formatTime` function is called for `startTime` and `endTime `fields

asharonbaltazar: That means all views should have the edit button then, right?

apps/client/src/common/components/edit-form-drawer/EditFormDrawer.tsx
asharonbaltazar: Addressed with https://github.com/cpvalente/ontime/pull/420/commits/4b84988843b526998b3bce2bf32059fb69b86253

asharonbaltazar:
>  Do you think that is something you could do as part of this PR or should I extract that to another task?

Let's extract that to another PR. I feel like this feature is big enough already. Feel free to disagree 👍

apps/client/src/common/components/edit-form-drawer/constants.ts
cpvalente: that is correct

apps/client/src/common/components/edit-form-drawer/EditFormInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    // checked value should be 'true' (see onParamsFormSubmit in <EditFormDrawer />)
+    return <Switch variant='ontime' name={id} defaultChecked={defaultCheckedValue} value='true' />;
+  }
+
+  if (type === 'number') {
+    const defaultNumberValue = searchParams.get(id) ?? '';
```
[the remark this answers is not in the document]
asharonbaltazar: Let's extract that to another PR. I feel like this feature is big enough already. Feel free to disagree 👍

asharonbaltazar:
I'll stop resolving conversations, in case the issue isn't actually resolved from your end. Would you please be kind enough to let me know if the comment/nitpick has been resolved or not? Appreciate it!

cpvalente:
> Let's extract that to another PR. I feel like this feature is big enough already. Feel free to disagree 👍



> I'll stop resolving conversations, in case the issue isn't actually resolved from your end. Would you please be kind enough to let me know if the comment/nitpick has been resolved or not? Appreciate it!

No problem, I know this tend to get out of hand
I am happy with the implementation and happy to merge.

What do you think about it?

Also, can you doublecheck the description for the task that follows up with the default fields?

asharonbaltazar:
Added a comment, but otherwise, looks good.

apps/client/src/common/components/edit-form-drawer/constants.ts
asharonbaltazar: Addressed with https://github.com/cpvalente/ontime/pull/420/commits/ab8b2005ebf9bc02dbc879e769c18c8269f662f6 & https://github.com/cpvalente/ontime/pull/420/commits/48b5d20b955c334247923b081e2db50db0754132

asharonbaltazar:
It's ready for another look

asharonbaltazar:
> What do you think about it?

I think it's good to go! I don't know when you plan to address the spacebar bug, but I can imagine we can pass a flag to that hook: 

```ts
export const useKeyDown = (callback: () => void, targetKey: string, { isDisabled?: boolean })
```

And then in `NavigationMenu`:
```ts
useKeyDown(toggleMenu, ' ', { isDisabled: searchParams.get('edit') !== null });
```

Or we can change the `targetKey` completely. I have no preference.

cpvalente:
> > What do you think about it?
> 
> I think it's good to go! I don't know when you plan to address the spacebar bug, but I can imagine we can pass a flag to that hook:
> 
> ```ts
> export const useKeyDown = (callback: () => void, targetKey: string, { isDisabled?: boolean })
> ```
> 
> And then in `NavigationMenu`:
> 
> ```ts
> useKeyDown(toggleMenu, ' ', { isDisabled: searchParams.get('edit') !== null });
> ```
> 
> Or we can change the `targetKey` completely. I have no preference.

Lets disable the key for now, I will give it another test tomorrow and potentially merge it 
Would you mind creating this last commit?

And can I help you find something new?
I was wondering if a good follow up would be this simple task that you basically done

PR #430, opened 2023-06-06
refactor: increase update rate

cpvalente:
This PR increases the timer update rate to prevent that automation from accumulating lost time

The UI will still only receive updates once a second, although I have tried pushing this to 32ms intervals (30fps) without any negative consequences.

It seems to me that as far as the data consumers are concerned, nothing has changed. It would be good to sanity check


cpvalente:
Hi @kellhogs and @asharonbaltazar, would you mind taking a look here?
This looked a bit too easy and I wonder if I am overlooking something

asharonbaltazar:
I'll take a look at this later today. Sorry for the delay!

apps/server/src/services/TimerService.ts
```diff
@@ -24,9 +26,10 @@ export class TimerService {
    * @param {object} [timerConfig]
    * @param {number} [timerConfig.refresh]
    */
-  constructor(timerConfig?) {
+  constructor(timerConfig?: { refresh?: number; updateInterval?: number }) {
     this._clear();
-    this._interval = setInterval(() => this.update(), timerConfig?.refresh || 1000);
+    this._interval = setInterval(() => this.update(), timerConfig?.refresh ?? 1000);
+    this._updateInterval = timerConfig?.updateInterval ?? 1000;
```
asharonbaltazar, line 32: Nitpicking, but you can give `timerConfig` a default value of `{}` and not have to deal with the (subjectively) ugly `?` operator:

```ts
constructor(timerConfig?: { refresh?: number; updateInterval?: number } = {}) {
    this._clear();
    this._interval = setInterval(() => this.update(), timerConfig.refresh ?? 1000);
    this._updateInterval = timerConfig.updateInterval ?? 1000;
}
```

apps/server/src/services/TimerService.ts
```diff
@@ -240,77 +245,95 @@ export class TimerService {
...
 
     // force an update
-    this.update();
+    this.update(true);
   }
 
-  update() {
-    this.timer.clock = clock.timeNow();
+  private updateRoll() {
+    const tempCurrentTimer = {
```
asharonbaltazar, line 252: Since you have a type defined for `CurrentTimers`, maybe it's a good idea to type this object as such to prevent any ambiguities.

asharonbaltazar (commented):
Sorry, I don't have much insight as to what this PR accomplishes (being new to the codebase), so don't want to approve this. 
Left a few nits, though.

cpvalente:
> few

Thank you for taking the time either way, I am happy to hear that there are no immediate red flags and, since all my tests were positive, I will merge as is.

PR #435, opened 2023-06-15
refactor: cuesheet v2

cpvalente:
This PR aims to redesign the cuesheet to the new system
It should also add row virtualisation to improve performance on handling several hundreds of items

TODO:
- [x] write cache tests
- [x] unify delay values into columns


cpvalente:
Hi @asharonbaltazar , hoping to ask for some help here:

I am working on rebuilding the current `OntimeTable` at `/cuesheet`
This, admittedly slightly too large PR, includes several changes migrating to react-table v8 as well as style changes to follow the rest of the app

I am slightly stunned on finding a good solution for optimising the table here, in the current PR I have attempted to use react-virtual (also tanstack) since it seemed efficient and lightweight
However the UX is less than ideal with several the user experiences jerks in animation
I would also need to write custom logic to measure the rows individually since all rows have variable height 🤦 

In the [rundown view ](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/rundown/event-block/EventBlock.tsx#L122) we use `IntersectionObserver `in the `EventBlock `to prevent rendering the expensive inner components if not in view, resulting in rendering just the outer div
This works well performance wise and simple to maintain
I wonder if this would be a bad approach as far as tables go though

In terms of specification, the expectation here is that users would have slightly under 1000 elements in the table and will want the fastest experience in re-renders possible. Initial render time is less problematic

Any thoughts?

asharonbaltazar:
Ack. Taking a look later today 👀

apps/client/src/features/table/table-settings/TableSettings.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <div className={style.tableSettings}>
+      <div className={style.leftPanel}>
+        <div className={style.sectionTitle}>Toggle field visibility</div>
+        <div className={style.options}>
+          {columns.map((column) => {
+            const columnHeader = column.columnDef.header;
+            const visible = column.getIsVisible();
+
+            return (
+              <label key={column.id} className={style.option}>
```
asharonbaltazar, line 33: Seems like toggling all the columns doesn't re-enable the checkboxes themselves. Adding `visible` to the `key` seems to do the trick:

```suggestion
              <label key={`${column.id}${visible}`} className={style.option}>
```

asharonbaltazar:
> In the [rundown view ](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/rundown/event-block/EventBlock.tsx#L122)we use IntersectionObserver in the EventBlock to prevent rendering the expensive inner components if not in view, resulting in rendering just the outer div
This works well performance wise and simple to maintain
I wonder if this would be a bad approach as far as tables go though

Have you attempted this approach? I don't see anything wrong with it (haven't tried, though). Your implementation in `EventBlock` is simple and understandable; it'd be worth to try the alternative just for readability. 

Regarding `IntersectionObserver`, some quick research suggests that many people have tried using `IntersectionObserver` with [some varying degrees of success](https://dev.to/hey_yogini/infinite-scrolling-in-react-with-intersection-observer-22fh).

cpvalente:
> > In the [rundown view ](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/rundown/event-block/EventBlock.tsx#L122)we use IntersectionObserver in the EventBlock to prevent rendering the expensive inner components if not in view, resulting in rendering just the outer div
> > This works well performance wise and simple to maintain
> > I wonder if this would be a bad approach as far as tables go though
> 
> Have you attempted this approach? I don't see anything wrong with it (haven't tried, though). Your implementation in `EventBlock` is simple and understandable; it'd be worth to try the alternative just for readability.
> 
> Regarding `IntersectionObserver`, some quick research suggests that many people have tried using `IntersectionObserver` with [some varying degrees of success](https://dev.to/hey_yogini/infinite-scrolling-in-react-with-intersection-observer-22fh).

Thank you for your help here. Is it OK if I drop you a line when I get this ready?

It has taken some time to get this, as I realise I should be offloading some work to the backend and have been working on the architecture around this.

asharonbaltazar:
Sure!

cpvalente:
Hey @asharonbaltazar any opinions on the (assumedly naive) implementation for caching store?
We are adding some layers of data pre-processing in the services before the data is sent to frontend, I thought it would be good to cache this since the requests will generally outweigh the changes. Especially at show time when these are more important

https://github.com/cpvalente/ontime/pull/435/files#diff-17a9e8212be69d703da42a7df72494e44c9ae99c058c04ba09810dc8c40429ac

apps/server/src/stores/cachingStore.ts
```diff
@@ -0,0 +1,52 @@
+interface CacheData {
+  data: unknown;
+}
+
+const runtimeCache: Map<string, CacheData> = new Map();
+
+export function checkCached(key: string): boolean {
+  return Object.hasOwn(runtimeCache, key);
```
asharonbaltazar, line 8: Doesn't `Map` have its own [`.has`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/has) function?

asharonbaltazar, line 8: Doesn't `Map` have its own [`.has`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/has) function?

apps/server/src/stores/cachingStore.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export function getCached<T>(key: string, callback: () => T): T {
+  if (!Object.hasOwn(runtimeCache, key)) {
+    try {
+      const data = callback();
+      runtimeCache.set(key, { data });
+    } catch (error) {
+      console.log(`Failed retrieving data from callback: ${error}`);
+    }
+  }
```
asharonbaltazar, line 19: Getting a missing key should probably just throw an error. Any reason why the default behavior should be setting it?

apps/server/src/stores/cachingStore.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  return runtimeCache.get(key).data as T;
+}
+
+export function setCached<T>(key: string, value: T): T {
+  try {
+    runtimeCache.set(key, { data: value });
+  } catch (error) {
+    console.log(`Failed retrieving data from callback: ${error}`);
```
asharonbaltazar, line 28: Error here doesn't make sense. I don't see a callback like on line 14

asharonbaltazar:
It looks good. Can't wait to see where you're taking this. 

It's been slow on my end, but I'm itching to get back.

apps/server/src/stores/cachingStore.ts
cpvalente: the advantage of throwing an error would be to make sure the services do error handling around caching. not sure if this is more work than worth it?

cpvalente:
> It looks good. Can't wait to see where you're taking this.
> 
> It's been slow on my end, but I'm itching to get back.

No worries, I really enjoyed your feedback here, it was thoughtful and pragmatic.
Look forwards to having you back

apps/server/src/stores/cachingStore.ts
asharonbaltazar: Woops, I meant the error message! Should've been more explicit 😅

apps/server/src/stores/cachingStore.ts
asharonbaltazar: I guess that works. When I think of it as a cache as opposed to a glorified `Map`, it makes more sense. Resolving

apps/server/src/stores/cachingStore.ts
cpvalente: a-ha, great observation! This was definitely an unchecked copy-paste artifact. I guess set will never fail

cpvalente:
Hi @asharonbaltazar, could I please ask you for one last bit of code review?

This PR is larger than I would have liked it to be. I'm sorry about that.
The relevant focus points would be
- the [new cuesheet element](https://github.com/cpvalente/ontime/pull/435/files#diff-90f69a3e74ce915927f0e6056dcf351d67962175e9dc00a30b5581eff685568dR1), using react table v8
- the [store for the table settings](https://github.com/cpvalente/ontime/pull/435/files#diff-f1ee9e02dc87edfa22feb49f8a4188f420aaa178dc1876654200481e8d532711R1)
- I am unsure about the styles in the [table settings](https://github.com/cpvalente/ontime/pull/435/files#diff-fe8f5ffdb872bd9de7a5936540b509086f2293de71ef7658527b774c5825250aR1), maybe you have some ideas?

apps/client/src/features/table/store/CuesheetSettings.tsx
```diff
@@ -0,0 +1,55 @@
...
+  Follow = 'ontime-cuesheet-follow-selected',
+  DelayVisibility = 'ontime-cuesheet-show-delay',
+  PreviousVisibility = 'ontime-cuesheet-show-previous',
+}
+
+export const useCuesheetSettings = create<CuesheetSettings>()((set) => ({
+  showSettings: false,
+  followSelected: booleanFromLocalStorage(CuesheetKeys.Follow, false),
+  showDelayBlock: booleanFromLocalStorage(CuesheetKeys.DelayVisibility, true),
+  showPrevious: booleanFromLocalStorage(CuesheetKeys.PreviousVisibility, true),
```
asharonbaltazar, line 34: Have you tried using the power of closures to use the `useLocalStorage` hook?

```ts
export const useCuesheetSettings = () => {

const [followedSelected, setFollowedSelected] = useLocalStorage(CuesheetKeys.Follow, false);
const [showDelayBlock, setShowDelayBlock] = useLocalStorage(CuesheetKeys.DelayVisibility, false);
const [showPrevious, setShowPrevious] = useLocalStorage(CuesheetKeys.PreviousVisibility, false);

return create<CuesheetSettings>()((set) => ({
  showSettings: false,
  followSelected,
  showDelayBlock,
  showPrevious,
  
  // in the setter functions below, 
  // rather than using `localStorage.setItem`, 
  // use the `setVariable` returned by the `useLocalStorage` hooks
  })
}
```

apps/client/src/features/table/table-settings/TableSettings.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const showDelayBlock = useCuesheetSettings((state) => state.showDelayBlock);
+  const toggleDelayVisibility = useCuesheetSettings((state) => state.toggleDelayVisibility);
+  const showPrevious = useCuesheetSettings((state) => state.showPrevious);
+  const togglePreviousVisibility = useCuesheetSettings((state) => state.togglePreviousVisibility);
+
+  return (
+    <div className={style.tableSettings}>
+      <div className={style.leftPanel}>
+        <div className={style.sectionTitle}>Toggle field visibility</div>
+        <div className={style.options}>
```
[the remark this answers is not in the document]
asharonbaltazar: ##### For the checkboxes
If you use a `grid` (and `flex`, too), you can substitute the `padding` for `gap`. 
Screenshot 2023-07-12 at 1 33 00 PM

##### For the toggles
I don't think this padding looks nice:
Screenshot 2023-07-12 at 1 32 33 PM

asharonbaltazar (commented):
Left some nits and suggestions. Lemme know what you think. 
I didn't leave an approval, but can if you want it :). Not that you need it to merge

apps/client/src/features/table/Cuesheet.tsx
```diff
@@ -0,0 +1,275 @@
...
+  useEffect(() => {
+    function scrollToComponent(
+      componentRef: MutableRefObject<HTMLTableRowElement>,
+      scrollRef: MutableRefObject<HTMLDivElement>,
+    ) {
+      const componentRect = componentRef.current.getBoundingClientRect();
+      const scrollRect = scrollRef.current.getBoundingClientRect();
+      const top = componentRect.top - scrollRect.top + scrollRef.current.scrollTop - 100;
+      scrollRef.current.scrollTo({ top, behavior: 'smooth' });
+    }
```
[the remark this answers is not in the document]
cpvalente: it seems to me that this data is very likely to change and we would not gain much in extracting it apart from potentially avoiding a function declaration

There was a potential bug here with the omission of `followSelected` which was masked by me suppressing the linter due to the suggested omission of the refs. Would you happen to have some insight as to why eslint does not consider those valid?

apps/client/src/features/table/Cuesheet.tsx
```diff
@@ -0,0 +1,275 @@
...
+                      })}
+                    </SortableContext>
+                  </tr>
+                </DndContext>
+              );
+            })}
+          </thead>
+
+          <tbody>
+            {table.getRowModel().rows.map((row) => {
```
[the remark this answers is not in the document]
cpvalente: Good tip. I would like for typescript to see that I resolved all scenarios. but that would probably mean more typescript gymnastics that we would get from it

We are not in a rush to release and there is still some work to be done here while the release candidate is under user review. I will consider extracting to components as it was done in the previous version

apps/client/src/features/table/table-settings/TableSettings.tsx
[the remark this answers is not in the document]
cpvalente: Good observation, thank you!

Do you think there is anything in terms of style that we can do here to make this more expandable?
As it is, I fear this can grow into a complicated UX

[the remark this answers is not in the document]
asharonbaltazar: I think it's fine if they're evened out. 
If you want to get fancy, I think organizing the options as a list in a modal could be an alternative. I'm not sure, but if theoretically, there's a hundred options, that'll be the way to go.

apps/client/src/features/table/store/CuesheetSettings.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  toggleSettings: (newValue?: boolean) => set((state) => ({ showSettings: toggle(state.showSettings, newValue) })),
+  toggleFollow: (newValue?: boolean) =>
+    set((state) => {
+      const value = toggle(state.followSelected, newValue);
```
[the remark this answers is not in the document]
cpvalente: In principle I would agree. However since this code is so small, and will likely never grow beyond a few lines. I wonder if adding meaning to the variable would only add to the load on the reader

So in this case

```
  togglePreviousVisibility: (newValue?: boolean) =>
    set((state) => {
      const value = toggle(state.showPrevious, newValue);
      localStorage.setItem(CuesheetKeys.PreviousVisibility, String(value));
      return { showPrevious: value };
    }),

```

could become
```
  togglePreviousVisibility: (shouldShowPrevious?: boolean) =>
    set((state) => {
      const showPrevious = toggle(state.showPrevious, shouldShowPrevious);
      localStorage.setItem(CuesheetKeys.PreviousVisibility, String(showPrevious));
      return { showPrevious };
    }),

```

I wonder if `newValue` does more to signify to the reader "this is the thing the user wants to set" than a semantic variable name, such as in this example `shouldShowPrevious`) Although renaming the value to its name in the object does seem to improve readability

What do you think?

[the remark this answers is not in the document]
asharonbaltazar: Yeah, I don't mind if you take out the bad parts and use the good parts. Information overload is a thing and can be a problem!

asharonbaltazar:
I still think it looks good without the initial padding.

Screenshot 2023-07-12 at 4 05 20 PM

cpvalente:
> I still think it looks better without the initial padding.
> Screenshot 2023-07-12 at 4 05 20 PM

Fair enough, I will follow your lead here.
One more related question: 

The red indexed cell relates to a user-defined event colour.
You can see in the previous version we added the colour to the whole row
https://github.com/cpvalente/ontime/blob/master/.github/aux-images/overview.png

This brought some issues:
- it became difficult to highlight rows
- there were issues with insufficient contrast between background and text

But I am unsure that the current solution provides a good enough indication
What do you think?

asharonbaltazar:
Sorry, should've been more specific! I was referring to the settings, where the first items under `TOGGLE FIELD VISIBILITY` and `TABLE OPTIONS` both have left padding. I think you should remove the left padding because it looks out of place

cpvalente:
> Sorry, should've been more specific! I was referring to the settings, where the first items under `TOGGLE FIELD VISIBILITY` and `TABLE OPTIONS` both have left padding. I think you should remove the left padding because it looks out of place

hahaha, no need to apologise.
I understood and agree.

The question above was unrelated to the padding changes, I should have been the one to be more specific :D

asharonbaltazar:
> But I am unsure that the current solution provides a good enough indication

Nah, it looks cleaner than the old one.  

I'm wondering if the old one would look better if you fade the subsequent rows' color a bit... Sorta like the contrast in these [buttons](https://tailwindui.com/components/application-ui/elements/buttons#component-cc1fc7c0898d134a5e03a53ede94f622)

PR #436, opened 2023-06-18
Feat/userid

kellhogs:


cpvalente:
I realise you are not finished, just made a few notes.
We also need to remember to add this to the functions that create events, and decide on what we do if there is no alias.
Should it be null or empty string?

cpvalente:
Closing this in favour of #473

PR #440, opened 2023-07-02
feat: operator view

cpvalente:
This sets up routes and directories for the new feature. When running in dev mode you can find the feature at `localhost:3000/op`

I tend to run two terminal instances, one from `apps/clients` and one from `apps/server` and do `pnpm dev` in both, just so that I can have control of the two apps separately. 

You can branch from here for your work. Remember to do one thing at a time and keep the PRs small
Many PRs is ok. as long as we can merge without breaking, they can be as small or as large as you need


cpvalente:
This PR looks biggerthan it is.
The important components are inside operator directory, then there are a few abstractions that we extracted from existing views

@arihanv would you mind taking a look?
The view targets phones and tablets

apps/client/src/features/operator/Operator.tsx
```diff
@@ -0,0 +1,150 @@
...
+    followRef: selectedRef,
+    scrollRef: scrollRef,
+    doFollow: !lockAutoScroll,
+    topOffset: selectedOffset,
+  });
+
+  // Set window title
+  useEffect(() => {
+    document.title = 'ontime - Operator';
+  }, []);
```
asharonbaltazar, line 47: Will this reset when a user navigates away from `/operator`?

apps/client/src/features/operator/Operator.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const handleOffset = () => {
+    if (featureData.selectedEventId) {
+      scrollToComponent();
+    }
+    setLockAutoScroll(false);
+  };
+
+  const handleScroll = () => {
+    if (selectedRef && scrollRef) {
```
asharonbaltazar, line 66: I think you meant to check `scrollRef.current`. You can get rid of the optional operator on the next line

apps/client/src/features/operator/Operator.tsx
```diff
@@ -0,0 +1,150 @@
...
+  };
+
+  if (
+    !data ||
+    status === 'loading' ||
+    !userFields ||
+    userFieldsStatus === 'loading' ||
+    !projectData ||
+    projectDataStatus === 'loading'
+  ) {
```
asharonbaltazar, line 82: Maybe this should be a variable... 

```ts
if (isLoading)
```

apps/client/src/features/operator/Operator.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const lastEvent = getLastEvent(data);
+
+  return (
+    <div className={style.operatorContainer}>
+      <NavigationMenu />
+      <ViewParamsEditor paramFields={operatorOptions} />
+
+      <div className={style.operatorEvents} onScroll={handleScroll} ref={scrollRef}>
+        {data.map((entry) => {
+          if (entry.type === SupportedEvent.Event) {
```
asharonbaltazar, line 105: Don't we have type guard functions now?

apps/client/src/features/operator/operator-event/OperatorEvent.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return <>{formatTime(timer.current, { showSeconds: true, format: 'hh:mm:ss' })}</>;
+}
+
+export default function OperatorEvent(props: OperatorEventProps) {
+  const { data, cue, isSelected, subscribed, subscribedAlias, showSeconds, isPast, selectedRef } = props;
+
+  const start = formatTime(data.timeStart, { showSeconds });
+  const end = formatTime(data.timeEnd, { showSeconds });
+
+  const cueColours = data.colour && getAccessibleColour(data.colour);
```
asharonbaltazar, line 34: I believe this `cueColours` variable type is `boolean | object`. Does that make sense when it's spread into the `styles` prop on line 50?

asharonbaltazar (dismissed):
Left some nits and comments. ~I strongly think we should avoid having `props` variables to maintain consistency~

apps/client/src/features/operator/Operator.tsx
cpvalente: all pages have this, so the subsequent views will re-declare the window title

apps/client/src/features/operator/Operator.tsx
```diff
@@ -0,0 +1,150 @@
...
+  // get fields which the user subscribed to
+  const subscribe = searchParams.get('subscribe') as keyof UserFields | null;
+  const subscribedAlias = subscribe ? userFields[subscribe] : '';
+  const showSeconds = isStringBoolean(searchParams.get('showseconds'));
+
+  const operatorOptions = getOperatorOptions(userFields);
+  let isPast = Boolean(featureData.selectedEventId);
+  const hidePast = isStringBoolean(searchParams.get('hidepast'));
+
+  const firstEvent = getLastEvent(data);
```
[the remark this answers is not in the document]
cpvalente: great catch, this is the kinds of things that we will spend several hours debugging in the future

apps/client/src/common/hooks/useFollowComponent.ts
```diff
@@ -0,0 +1,61 @@
...
+
+interface UseFollowComponentProps {
+  followRef: MutableRefObject<HTMLElement | null>;
+  scrollRef: MutableRefObject<HTMLElement | null>;
+  doFollow: boolean;
+  topOffset?: number;
+}
+
+export default function useFollowComponent(props: UseFollowComponentProps) {
+  const { followRef, scrollRef, doFollow, topOffset = 100 } = props;
```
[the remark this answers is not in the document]
cpvalente: hmm, I never thought about that.
I guess the prop terminology is very react like. The conventional terminology for a function would be arguments or params

This is a good observation, the short answer is that it might not matter other than for semantics. In the current code base props has been used so I will stick to that. For my next project I will keep that in mind and see what happens

Out of curiosity, just snooped around on sentry
https://github.com/getsentry

and it seems they tend to use the term Option, so `useFollowComponentOptions`, which makes sense

apps/client/src/common/hooks/useFollowComponent.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    // eslint-disable-next-line -- the prompt seems incorrect
+  }, [followRef?.current, scrollRef?.current]);
+
+  const scrollToRefComponent = useCallback(
+    (componentRef = followRef, containerRef = scrollRef, offset = topOffset) => {
+      if (followRef.current && containerRef.current) {
+        // @ts-expect-error -- we know this are not null
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        scrollToComponent(componentRef!, scrollRef!, offset);
```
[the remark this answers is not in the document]
cpvalente: hmm, the error is still there, although I don't quite understand why cant typescript correctly infer the type.

apps/client/src/features/operator/operator-event/OperatorEvent.tsx
cpvalente: That is correct. in the case `data.colour `is evaluated to falsy, `cueColours` will be a boolean with value false

I tried having a look on ECMA what the behaviour is for spreading booleans, but without luck on finding the right documentation

However I made a quick sandbox to find that spreading a falsy value always returned an empty object
` {...false}, {...undefined}, {...null}, {...""}, {...0}`

Thank you for the deep dive.
In this case, it seems like we are being graced with a language feature that helps us. The code could potentially be clearer, but I am unsure that the added complexity of the check would make it worth it

apps/client/src/features/operator/operator-event/OperatorEvent.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const subscribedData = (subscribed ? data?.[subscribed] : undefined) || '';
```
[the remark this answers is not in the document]
cpvalente: it absolutely can, but typescript complains and loses the typing for subscribedData. do you have recommendations for a way around?

apps/client/src/features/operator/operator-event/OperatorEvent.tsx
asharonbaltazar: Can't `getAccessibleColour` just return `{}` if `data.colour` is falsey?

apps/client/src/common/hooks/useFollowComponent.ts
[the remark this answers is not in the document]
asharonbaltazar: Sorry for not being clearer. If you're checking `scrollRef`, perhaps remove the null assertions?

```ts
scrollToComponent(componentRef, scrollRef, offset);
```

apps/client/src/features/operator/operator-event/OperatorEvent.tsx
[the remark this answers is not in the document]
asharonbaltazar: What is `subscribed` typed as?

PR #442, opened 2023-07-03
Make Operator Layout 

arihanv:
- Made OpBlock, OpEvent, TimeBlock components with scss files for each
- Added Styling from OnTime colors
- Added overflow scrolling

### DIFF
--- a/apps/client/src/features/operator/Operator.tsx
+++ b/apps/client/src/features/operator/Operator.tsx
@@ -19,32 +19,31 @@ export default function Operator() {
   return (
     <div className={style.operator}>
       <div className={style.allBlocks}>
-      {data.map((entry) => {
-        // there are three types of events, you a filter them by using the type property
-        // for this view, we do not show the delay event
+        {data.map((entry) => {
+          // there are three types of events, you a filter them by using the type property
+          // for this view, we do not show the delay event
 
-        // this is a scheduled event
-        if (entry.type === SupportedEvent.Event) {
-          return (
-            <div key={entry.id} className={style.scheduledEvent}>
-              <OpEvent data={entry} />
-            </div>
-          );
-        }
-        
+          // this is a scheduled event
+          if (entry.type === SupportedEvent.Event) {
+            return (
+              <div key={entry.id} className={style.scheduledEvent}>
+                <OpEvent data={entry} />
+              </div>
+            );
+          }
 
-        // this is a block entry (like a section title)
-        if (entry.type === SupportedEvent.Block) {
-          return (
-            <div key={entry.id} className={style.block}>
-              <OpBlock data={entry} />
-            </div>
-          );
-        }
-        return null;
-      })}
+          // this is a block entry (like a section title)
+          if (entry.type === SupportedEvent.Block) {
+            return (
+              <div key={entry.id} className={style.block}>
+                <OpBlock data={entry} />
+              </div>
+            );
+          }
+          return null;
+        })}
       </div>
-      <TimeBlock/>
+      <TimeBlock />
     </div>
   );
 }
--- a/apps/client/src/features/operator/Operator.tsx
+++ b/apps/client/src/features/operator/Operator.tsx
@@ -2,10 +2,11 @@ import { SupportedEvent } from 'ontime-types';
 
 import useRundown from '../../common/hooks-query/useRundown';
 
-import style from './Operator.module.scss';
-import OpEvent from './op-event/OpEvent';
 import OpBlock from './op-block/OpBlock';
+import OpEvent from './op-event/OpEvent';
 import TimeBlock from './time-block/TimeBlock';
+
+import style from './Operator.module.scss';
 
 export default function Operator() {
   // this is the data that you need, the status flag should give you possibility to create a loading state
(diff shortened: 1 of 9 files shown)


apps/client/src/features/operator/Operator.tsx
```diff
@@ -15,6 +18,7 @@ export default function Operator() {
 
   return (
     <div className={style.operator}>
+      <div className={style.allBlocks}>
```
cpvalente, line 21: could we find a more meaningful name here? Perhaps something like rundown?

apps/client/src/features/operator/op-block/OpBlock.tsx
```diff
@@ -0,0 +1,13 @@
+import style from './OpBlock.module.scss';
+
+type Props = {
+  data: any;
```
cpvalente: the type of this is OntimeBlock
https://github.com/cpvalente/ontime/blob/master/packages/types/src/definitions/core/OntimeEvent.type.ts#L22

Also, we have generally used a few contentions for props.
- props use `interface` while value types use `type`
- interfaces are usually named like `ComponentProps`, in this case it would be `OpBlockProps`

None of this affect the code in any relevant way, it is just a convention that the rest of the code follows and it is good not to diverge

apps/client/src/features/operator/op-event/OpEvent.tsx
```diff
@@ -0,0 +1,26 @@
...
+import style from './OpEvent.module.scss';
+import { formatTime } from '../../../common/utils/time';
+
+type Props = {
+  data: any;
+};
+
+export default function OpEvent({ data }: Props) {
+  const start = formatTime(data.timeStart, { format: 'hh:mm' });
+  const end = formatTime(data.timeEnd, { format: 'hh:mm' });
```
cpvalente: nice, cool to see you are figuring your way around

apps/client/src/features/operator/time-block/TimeBlock.tsx
```diff
@@ -0,0 +1,20 @@
+import styles from './TimeBlock.module.scss';
+import { Play } from 'lucide-react';
+import { useTimer } from '../../../common/hooks/useSocket';
+import { formatTime } from '../../../common/utils/time';
+
+type Props = {};
+
+export default function TimeBlock({}: Props) {
+  const timer = useTimer();
```
cpvalente: cool that you are finding the data, good stuff

apps/client/src/features/operator/op-event/OpEvent.tsx
```diff
@@ -0,0 +1,26 @@
+import { EventData } from 'ontime-types';
+import style from './OpEvent.module.scss';
+import { formatTime } from '../../../common/utils/time';
+
+type Props = {
+  data: any;
```
cpvalente: this is an OntimeEvent
https://github.com/cpvalente/ontime/blob/master/packages/types/src/definitions/core/OntimeEvent.type.ts#L27

Generally uses of any should come with a comment

apps/client/src/features/operator/Operator.tsx
```diff
@@ -3,6 +3,9 @@ import { SupportedEvent } from 'ontime-types';
 import useRundown from '../../common/hooks-query/useRundown';
```
cpvalente: eslint should be complaining about the import order in some of these files

apps/client/src/features/operator/time-block/TimeBlock.tsx
```diff
@@ -0,0 +1,20 @@
+import styles from './TimeBlock.module.scss';
+import { Play } from 'lucide-react';
+import { useTimer } from '../../../common/hooks/useSocket';
+import { formatTime } from '../../../common/utils/time';
+
+type Props = {};
+
+export default function TimeBlock({}: Props) {
```
cpvalente: no need to add this until we need it

cpvalente (approved):
Hey! you did great work here and I am very happy that you start finding your way around the app.

I have left some comments, lots of nitpicking and some help with types that you wouldn't have known. We just need some small changes before moving on.

Also, what happened with pnpm-lock? there are no changes in packages so there shouldn't be a change here. It looks like you re-wrote most of the file (it looks like you might have a different pnpm version, checkout the package json with the appropriate tooling versions)
https://github.com/cpvalente/ontime/blob/master/package.json#L21

Thank you again for your work, after this we can continue implementing the data.

apps/client/src/features/operator/time-block/TimeBlock.tsx
```diff
@@ -1,20 +1,21 @@
...
   const timeNow = formatTime(timer.clock, {
     showSeconds: true,
     format: 'hh:mm:ss a',
   });
+
   return (
-    <div className={styles.block}>
-      <Play size={17} />
-      <div className={styles.clock}>{timeNow} <div>00:10:00</div></div>
+    <div className={styles.TimeBlock}>
```
cpvalente, line 16: super nitpick: capital letter names are usually left for components or classes

cpvalente (approved):
I think this is OK for now, we will probably tweak things as we go further
Good work

PR #446, opened 2023-07-10
Swap Event Data

### LINKED ISSUE #403: Swap event data

asharonbaltazar:
closes #403 

- [x] update `ContextMenu` to use `zustand` rather than `useContext` and `useState`
- [x] create `zustand` store to track events to swap. Suggestions on where to init this store are welcome
- [x] add swapping option to context menu 
      - if `from` event is not set, display `Add to swap`
      - if `from` event is set and user right-clicks on same event, display `Remove from swap`
      - if `from` event is set and user clicks a different event, display `Swap with {fromEventId}`
- [ ] create endpoint to swap events


asharonbaltazar:
@cpvalente Did I miss anything in the steps?

cpvalente:
> @cpvalente Did I miss anything in the steps?

Hi @asharonbaltazar, it seems to me that you got it covered. A few points:

>  create zustand store to track events to swap. Suggestions on where to init this store are welcome
In previous iterations of ontime we used atomic state managers (jotai) to handle these cases. As the app grew it felt like we should be leveraging something that would work on the level of larger stores, so we migrated to Zustand.
If I were to approach this, I would be tempted in creating a new hook useEventContextMenu to gather these, this should make it easy to maintain and build on.
As for location, I feel this belongs at the level of the [Rundown component](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/rundown/Rundown.tsx)

>  add swapping option to context menu
> - if from event is not set, display Add to swap
> - if from event is set and user right-clicks on same event, display Remove from swap
> - if from event is set and user clicks a different event, display Swap with {fromEventId}

This seems to cover all of it, although I generally prefer a simpler, less verbose implementation:
- Add two buttons to context menu: 1) `Add to swap` 2) `Swap with {fromEventId}`, where 2 should exist but be disabled, in case we don't have an event added
- When a swap operation is completed we clear the data from the context, I don't see a need for a "remove from swap" button
- Do you think the event marked for swapping should have some styling to indicate this?

Great work, hope this helps

asharonbaltazar:
> If I were to approach this, I would be tempted in creating a new hook useEventContextMenu to gather these, this should make it easy to maintain and build on.

I'll hold off abstracting right now since we only have one instance of this logic (here) – I don't know how the other will behave to be worth the abstraction effort.

asharonbaltazar:
> Add two buttons to context menu: 1) Add to swap 2) Swap with {fromEventId}, where 2 should exist but be disabled, in case we don't have an event added

Hm, 2nd one could be disabled, but will the label be? Can't be `Swap with {fromEventId}` if nothing's selected. Rather than disabling, I'm thinking of adding this option conditionally. What do you think?

cpvalente:
> Hm, 2nd one could be disabled, but will the label be? Can't be `Swap with {fromEventId}` if nothing's selected. Rather than disabling, I'm thinking of adding this option conditionally. What do you think?

Good point. Could it simply say `"Select event to swap"` ?

cpvalente:
Hi @asharonbaltazar, is there anything I can do to help out here?

cpvalente:
> Hey, 
> 
> Just caught up with work. I've been experimenting with various menus logic to see which one is the most intuitive. Your suggestion is great, however, when the user adds event `1` to the swap and then clicks `1` again, what should the menu display? 
> 
> - `Swap with 1`? 
> 
> - Or should the `Swap with {fromEventId}` be disabled since the user clicked `1` again? 
> 
> 
> 
> I was exploring other options, but can't dedicate as much time as I'd like to.

No problem. There is no deadline and no expectation that you will commit more time than you can

Just let me know if I can help

As for the question, I would say that the button should be disabled. But you are welcome to propose something different. Looking forwards to it

asharonbaltazar:
@cpvalente Is the `clone` action appropriate in `useEventAction`?

asharonbaltazar:
Sorry, brainfart. I meant is a _`swap`_ action appropriate in `useEventAction`?

asharonbaltazar:
It's my fault. Sometimes I forget people need context. I should be more descriptive:

Most of the code relating to events (add, edit, delete) are found in the `useEventAction` file. Since I'm starting to write the API logic for swapping the events, I was wondering whether a new action called `swap` is appropriate in this file. Hope that makes sense

cpvalente:
Hi @asharonbaltazar , just a quick heads up that you might merge master into your branch to avoid late merge conflicts

asharonbaltazar:
Was hoping to hear that 😄

asharonbaltazar:
@cpvalente It's ready for review! I did my best to replicate the logic on the backend. I'm still not sure what everything does, so I'll ask here for future reference:

In the `cachedReorder` function on line 148 in `delayedRundown.utils` file, there's what you call a sanity check. What's the reason behind it?

```ts
if (indexCheck !== from) {
  invalidateFromError();
  throw new Error('ID not found at index');
}
```

A few lines down, you make another check. What's happening here?

```ts
const delayedRundown = getDelayedRundown();
if (eventId !== delayedRundown[from].id) {
 invalidateFromError();
} else {
  // TODO: could we be more granular about updates
  // I fear we need to update both from and to, which could signify more iterations
  runtimeCacheStore.invalidate(delayedRundownCacheKey);
}
```

Thank you!

cpvalente:
> @cpvalente It's ready for review! I did my best to replicate the logic on the backend. I'm still not sure what everything does, so I'll ask here for future reference:
> 
> In the `cachedReorder` function on line 148 in `delayedRundown.utils` file, there's what you call a sanity check. What's the reason behind it?
> 
> ```ts
> if (indexCheck !== from) {
>   invalidateFromError();
>   throw new Error('ID not found at index');
> }
> ```
> 
> A few lines down, you make another check. What's happening here?
> 
> ```ts
> const delayedRundown = getDelayedRundown();
> if (eventId !== delayedRundown[from].id) {
>  invalidateFromError();
> } else {
>   // TODO: could we be more granular about updates
>   // I fear we need to update both from and to, which could signify more iterations
>   runtimeCacheStore.invalidate(delayedRundownCacheKey);
> }
> ```
> 
> Thank you!

Hi @asharonbaltazar I will review your code right now, first impression is very good

As for your questions, thank you for bringing it up, it is great that you question and potentially challenge some of the choices. Please continue to do so, I will try my best to provide answers and context on the decisions

In several places of the app we need to consume aggregated or derived data. This data is calculated in the relation of an element to its position in the rundown.

The easiest example is the delay: delays propagate and add onto each other until they find a block.
An events delay then is the result of all the delays before its position after a block (or from the start if no blocks exist)

Initially, we calculated this in frontend, but two things changed: a) we had to communicate this to integrations b) we needed this in the backend as well.
At this point the decision was
- lets calculate this in backend and pass it to the frontend, save ourselves some duplication and make it easier on integrations
- it feels dirty to save derived data in the persistent json storage. lets save this in a caching layer. Additionally I have toyed with the idea of using this caching layer to normalise the data. One for another time

That means that we had to intercept all modifications to the persisted rundown through something that could also change the caching layer. That is what that code tries to achieve. (Note: I attempted using proxies to do this, but since the app backend is structured around OOP it becomes slightly tricky)

Onto your questions:
``` ts
export async function cachedReorder(eventId: string, from: number, to: number) {
/// STUFF ABOUT PERSISTED DATA
  const indexCheck = DataProvider.getIndexOf(eventId);  
  if (indexCheck !== from) {                                   <--- does the client info match the persisted data
    invalidateFromError();
    throw new Error('ID not found at index');
  }
...
// STUFF ABOUT CACHED DATA (delayedRundown)
  const delayedRundown = getDelayedRundown();
  if (eventId !== delayedRundown[from].id) {           <--- does the client info match the cached data
    invalidateFromError();
```

This function receives an eventId and an index of where to move from and to.
This means that if we were to check the rundown at given index, the event needs to have that ID
I fear we might have some edge cases that are uncovered and can cause the client to be trying to mutate an outdated version of the rundown, and have built some fail safes in some operations.

In this case, if the ID of the thing I am trying to move, is not the ID you think it is, we cancel the operation and invalidate the cache so that is calculated again. We do that twice, once when checking against the persisted storage and once when checking against the cached version

Also note that `invalidateFromData()` throws an error if we are not in Production. I am a big fan of crashing apps since it can help trace tricky edge cases in development

Hope this helps clarify, feel free to follow up if it doesnt

apps/server/src/services/rundown-service/delayedRundown.utils.ts
```diff
@@ -171,6 +171,41 @@ export async function cachedReorder(eventId: string, from: number, to: number) {
...
+    };
+
+    updatedRundown[toEventIndex] = {
+      ...fromEvent,
+      timeStart: toEvent.timeStart,
+      timeEnd: toEvent.timeEnd,
+      duration: toEvent.duration,
+    };
+
+    runtimeCacheStore.invalidate(delayedRundownCacheKey);
```
cpvalente, line 204: I dont think we need to invalidate the entire cache

We know that we would only be swapping event types
We also know that the event types delay is calculated from its position in the rundown

With this in mind swapping the delay values should be safe.
I might be missing something, in which case we can use the `getDelayAt()` which will calculate delay backwards from an index

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -16,6 +17,18 @@ import style from './Rundown.module.scss';
...
+  setEvent: (newEventId: string | null, newEventIndex: number | null) => void;
+  clearEvent: () => void;
+}
+
+export const useEventIdSwapping = create<EventIdSwappingStore>((set) => ({
+  event: { index: null, id: null },
+  setEvent: (newEventId, newEventIndex) => set(() => ({ event: { index: newEventIndex, id: newEventId } })),
+  clearEvent: () => set(() => ({ event: { index: null, id: null } })),
+}));
+
```
cpvalente, line 31: This component and its near tree have admittedly confusing scope. 
Perhaps we should extract this into its own file?

apps/client/src/features/rundown/RundownEntry.tsx
```diff
@@ -95,6 +95,12 @@ export default function RundownEntry(props: RundownEntryProps) {
           addEvent({ type: SupportedEvent.Block }, { after: data.id });
           break;
         }
+        case 'swap': {
+          if (typeof payload === 'number') {
+            swapEvents({ from: payload, to: eventIndex });
```
cpvalente, line 100: I think this might be the source of a bug.

With your current implementation
- Clear the rundown by Clicking the `Events...` button and choosing `Delete All`
- Add an event, we will call it A
- Add a second event, we will call it B

If we test the swap now between A and B everything works correctly

Now navigate again to the `Event...` button and select `Add delay at start` and retry the operation
From what I could see here, the swap fails

In the rundown, the eventIndex keeps track of the order of the entries of type Event.
In our previous example we would have
- delay (array index 0)
- A (array index 1, event index 0)
- B (array index 2, event index 1)

In short, we cannot use eventIndex as a signifier for the items position in the array

cpvalente (changes requested):
Hi @asharonbaltazar , I am extremely impressed on how well you navigate the code base.
Really great work.

I have left some very nitpicky comments for your consideration

There is also a small bug that is potentially caused by a confusing naming convention.
What do you use to make your nice gif videos? Maybe in future I can use that to help out as well

Thank you again

apps/client/src/common/hooks/useEventAction.ts
```diff
@@ -316,5 +324,92 @@ export const useEventAction = () => {
...
+      const fromEvent = previousEvents.at(data.from);
+      const toEvent = previousEvents.at(data.to);
+
+      if (!fromEvent || !toEvent) {
+        // Return a context with the previous and new events
+        return { previousEvents };
+      }
+
+      if (fromEvent.type === SupportedEvent.Event && toEvent.type === SupportedEvent.Event) {
+        const eventsWithSwap = previousEvents.map((event, i) => {
```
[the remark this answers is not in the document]
asharonbaltazar: > I wonder whether it would be necessary to iterate through the array since we know exactly where the events we want to swap are.

I'm a bit confused right now. In this [comment](https://github.com/cpvalente/ontime/pull/446#discussion_r1286357835), you suggest we should stick to `id`s since it'd simplify the work. But unless the `OntimeRundown` is a `Map`, I don't see how it's possible to use `id`s to look up an object in a constant complexity.

Does my question make sense?

apps/client/src/features/rundown/Rundown.tsx
asharonbaltazar: We should. Maybe in a the `rundown` folder?

apps/server/src/services/rundown-service/delayedRundown.utils.ts
```diff
@@ -171,6 +171,41 @@ export async function cachedReorder(eventId: string, from: number, to: number) {
...
 
+/**
+ * Swaps two events
+ * @param {number} fromEventIndex
+ * @param {number} toEventIndex
+ */
+export async function cachedSwap(fromEventIndex: number, toEventIndex: number) {
+  const updatedRundown = DataProvider.getRundown();
+
+  const fromEvent = updatedRundown.at(fromEventIndex);
```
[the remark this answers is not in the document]
asharonbaltazar: It's handy, for sure. 

It also allows us to use `array.at(-1)` to return the last element as opposed to `array[array.length - 1]`

apps/client/src/features/rundown/event-block/EventBlock.tsx
```diff
@@ -75,22 +77,39 @@ export default function EventBlock(props: EventBlockProps) {
...
+      icon: IoAdd,
+      onClick: () => setEvent(eventId, eventIndex),
+      withDivider: true,
+    },
+    {
+      label: `Swap this event with ${event.id ?? ''}`,
+      icon: IoSwapVertical,
+      onClick: () => {
+        if (typeof event.index === 'number') {
+          actionHandler('swap', +event.index);
```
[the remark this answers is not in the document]
asharonbaltazar: I'll leave this one hanging until we decide whether to use `index` or `id`. This code won't be relevant if we use `id`

apps/client/src/features/rundown/RundownEntry.tsx
asharonbaltazar: I didn't test with delays! 
Okay, so we should use `id`s then. Do you think the backend and frontend code will be very similar since we'll be doing the same operation?

apps/server/src/services/rundown-service/delayedRundown.utils.ts
asharonbaltazar: TBH, I'm not so familiar with the implementation here to give a response. I'll keep learning as I go!

apps/client/src/common/api/eventsApi.ts
```diff
@@ -50,6 +50,19 @@ export async function requestApplyDelay(eventId: string) {
   return axios.patch(`${rundownURL}/applydelay/${eventId}`);
 }
 
+export type SwapEntry = {
+  from: number;
+  to: number;
+};
```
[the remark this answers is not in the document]
asharonbaltazar: It would, but using `id`s means I have to iterate through the array at least once to get the events I want to swap. I was hoping indexes would save us the relatively expensive operation

apps/client/src/common/hooks/useEventAction.ts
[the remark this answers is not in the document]
cpvalente: absolutely, sorry for not providing enough context.
I think both statements are correct, with nuance

- The feature / general operation should work with IDs, the context menu will save IDs and it is IDs we send to backend
- In this specific implementation we do a few steps

1. Find events at ids (`fromEvent `and `toEvent`)
2. Check if they are the right type
3. iterate through the array to find the events with matching indexes of data.from and data.to

This seems cumbersome and we can remove a few steps:
1. find event indexes of events at ids (`fromIndex `and `toIndex`) using findIndex
3. replace data between `fromIndex`and `toIndex`

I removed step 2 since I think this should be validated before we get this far

Hope this makes sense

apps/client/src/features/rundown/RundownEntry.tsx
cpvalente: Yes, I think the backend and frontend code will do exactly the same, to the exception that
- the backend needs to mutate both the `DataProvider` and the `delayedRundown`
- the `delayedRundown` also needs to move the delay field

cpvalente (commented):
Hey, let me know if it would be helpful to have a chat on slack or discord. Send me an email and we can arrange something

apps/client/src/features/rundown/RundownEntry.tsx
asharonbaltazar: >  - the backend needs to mutate both the DataProvider and the delayedRundown
>  - the delayedRundown also needs to move the delay field

Can you please explain these two bullet points?

apps/client/src/common/hooks/useEventAction.ts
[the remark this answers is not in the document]
asharonbaltazar: > I removed step 2 since I think this should be validated before we get this far

When should this get validated? Also, that `if` is necessary if we want autocomplete on the events. Since there are multiple types of events, TS requires me to check if it's the supported ones before I spread (`{ ...toEvent }`) and return. Does that make sense?

apps/client/src/features/rundown/RundownEntry.tsx
cpvalente: Absolutely

###  the backend needs to mutate both the DataProvider and the delayedRundown
In the backend we essential keep the rundown duplicated.
Once inside the DataProvider, as is in the persistent store
and again in the delayedRundown, with some derived data

We need to remember to do both, and we use the functions like `cachedEdit` to achieve this
- [updating the data provider ](https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/rundown-service/delayedRundown.utils.ts#L85)
- [updating the cached data](https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/rundown-service/delayedRundown.utils.ts#L100)

### the delayedRundown also needs to move the delay field
the idea of the delayedRundown object is that it contains derived data from the original rundown. That includes a `delay` field which is important for this feature
we just need to make sure that the delay data gets swapped as we discussed [here](https://github.com/cpvalente/ontime/pull/446#discussion_r1286379502)

apps/client/src/features/rundown/Rundown.tsx
cpvalente: That sounds good to me

apps/client/src/common/hooks/useEventAction.ts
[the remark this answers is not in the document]
cpvalente: In the frontend we KNOW that we only allow the feature on events of type Event. So within our UI, it would not be possible for it to have another value

In the backend, validation should move to the controller. The controllers should be in charge of validation before allowing data inside our domain (the services)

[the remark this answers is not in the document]
asharonbaltazar: > In the frontend we KNOW that we only allow the feature on events of type Event. So within our UI, it would not be possible for it to have another value

But the TS compiler doesn't know. I can cast it, but an `if` isn't that huge of a deal, right?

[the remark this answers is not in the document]
cpvalente: no, the `if` is not a problem. The idea is that we should make sure that we don't get that far into it with the wrong data, since by now we would have already sent the mutation request.

If typescript needs an `if` and you think that is best, its cool with me

I would personally have moved the check into the `swapEvents` function (not, not the `_swapEvents` mutation) and cast if needed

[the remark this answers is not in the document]
asharonbaltazar: Okay, do you think it's safe to assume (once I switch to the `id` implementation) that this action will always be triggered on a `OntimeEvent`? If yes, I'll just cast the variables and won't bother with validating. 

Or should validation should still be done before calling `_swapEvents`?

apps/client/src/features/rundown/event-block/EventBlock.tsx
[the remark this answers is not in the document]
asharonbaltazar: Resolving as it's no longer relevant

asharonbaltazar:
Hm, I want to raise a concern with your assumptions about certain values. For instance:

```ts
case 'update': {
  // Handles and filters update requests
  const { field, value } = payload as FieldValue;
  const newData: Partial<OntimeEvent> = { id: data.id };
  // etc
}
```

There's nothing stopping `payload` from being `number`. I'm aware you know your codebase and take steps to avoid these silly mistakes, but others might!

apps/client/src/common/hooks/useEventAction.ts
```diff
@@ -316,5 +324,92 @@ export const useEventAction = () => {
...
+
+      const fromEvent = previousEvents.at(data.from);
+      const toEvent = previousEvents.at(data.to);
+
+      if (!fromEvent || !toEvent) {
+        // Return a context with the previous and new events
+        return { previousEvents };
+      }
+
+      if (fromEvent.type === SupportedEvent.Event && toEvent.type === SupportedEvent.Event) {
```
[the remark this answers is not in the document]
asharonbaltazar: Resolving due to conversation on Discord. I've left a `TODO` [comment](https://github.com/cpvalente/ontime/pull/446/commits/82d91e3be2d85e80a3ed704c6e6644684f01549e) in `useEventAction`

asharonbaltazar:
Resolved all the conversations because they became outdated after my latest changes. 

@cpvalente Please leave new reviews. Thank you!

apps/server/src/services/rundown-service/delayedRundown.utils.ts
```diff
@@ -171,6 +171,43 @@ export async function cachedReorder(eventId: string, from: number, to: number) {
...
+    ...fromEvent,
+    timeStart: toEvent.timeStart,
+    timeEnd: toEvent.timeEnd,
+    duration: toEvent.duration,
+  };
+
+  await DataProvider.setRundown(rundownToUpdate);
+  runtimeCacheStore.setCached(delayedRundownCacheKey, rundownToUpdate);
+}
+
```
cpvalente, line 210: The caching abstraction could do with explaining. I will try my best to do it here:

The app uses a json object as the database, in that JSON object we store the raw rundown as the user enters.

From the persisted rundown (calling here user rundown), we extrapolate some more data and save the result in a new object, I will call this `delayedRundown`.
This serves two purposes:
- avoid keeping redundant / derived data in persistence (potential for bugs)
- normalise the data as we want to use in the app

In practice this means that we have
1. the user rundown, which we get with `DataProvider.getRundown()`
2. the delayedRundown, which we get from cache through the utility function `getDelayedRundown()`

When we mutate data in the rundown, we need to make sure that it happens in both the user rundown in `DataProvider` and the `delayedRundown`

When mutation we make a decision:
1. we are confident on what the change is, we make the mutate the object in the cached rundown and update it manually  `runtimeCacheStore.setCached(delayedRundownCacheKey, delayedRundown);`
2. we are unsure what needs to be recalculated, or we think that the recalculation might be more or as costly as calculating again, so we invalidate the whole thing `runtimeCacheStore.invalidate(delayedRundownCacheKey);`

Usually this would have been handled by a proxy. But since the `DataProvider` leverages a OOP architecture it is not possible to do so. The functions inside `delayedRundown.utils.ts` serve the same process. ie: not as convenient

As for your code, you have updated the user rundown, but not the `delayedRundown`.
You just need to mutate the rundown object as below (not great code, but hopefully explicit)

``` ts
export async function cachedSwap(fromEventId: string, toEventId: string) {
  const fromEventIndex = DataProvider.getIndexOf(fromEventId);
  const toEventIndex = DataProvider.getIndexOf(toEventId);

  if (fromEventIndex === -1 || toEventIndex === -1) {
    invalidateFromError();
    throw new Error('ID not found at index');
  }

  const rundownToUpdate = DataProvider.getRundown();

  const fromEvent = rundownToUpdate.at(fromEventIndex) as OntimeEvent;
  const toEvent = rundownToUpdate.at(toEventIndex) as OntimeEvent;

  rundownToUpdate[fromEventIndex] = {
    ...toEvent,
    timeStart: fromEvent.timeStart,
    timeEnd: fromEvent.timeEnd,
    duration: fromEvent.duration,
  };

  rundownToUpdate[toEventIndex] = {
    ...fromEvent,
    timeStart: toEvent.timeStart,
    timeEnd: toEvent.timeEnd,
    duration: toEvent.duration,
  };


  // ADDED CODE STARTS HERE
  const delayedRundown = getDelayedRundown();

  const fromCachedEvent = delayedRundown.at(fromEventIndex) as OntimeEvent;
  const toCachedEvent = delayedRundown.at(toEventIndex) as OntimeEvent;

  if (fromCachedEvent.id !== fromEvent.id || toCachedEvent.id !== toEvent.id) {
    // something went wrong, we invalidate the cache
    runtimeCacheStore.invalidate(delayedRundownCacheKey);
  } else {
    delayedRundown[fromEventIndex] = {
      ...toCachedEvent,
      timeStart: fromCachedEvent.timeStart,
      timeEnd: fromCachedEvent.timeEnd,
      duration: fromCachedEvent.duration,
    };

    delayedRundown[toEventIndex] = {
      ...fromCachedEvent,
      timeStart: toCachedEvent.timeStart,
      timeEnd: toCachedEvent.timeEnd,
      duration: toCachedEvent.duration,
    };
    runtimeCacheStore.setCached(delayedRundownCacheKey, delayedRundown);
  }

  // ADDED CODE FINISHES HERE

  await DataProvider.setRundown(rundownToUpdate);

```

apps/server/src/services/rundown-service/delayedRundown.utils.ts
```diff
@@ -171,6 +171,43 @@ export async function cachedReorder(eventId: string, from: number, to: number) {
...
+    timeEnd: fromEvent.timeEnd,
+    duration: fromEvent.duration,
+  };
+
+  rundownToUpdate[toEventIndex] = {
+    ...fromEvent,
+    timeStart: toEvent.timeStart,
+    timeEnd: toEvent.timeEnd,
+    duration: toEvent.duration,
+  };
```
cpvalente, line 205: Since we have to do this 3 times 
1. in the optimistic mutation in react-query
2. here, to update the rundown in `DataProvider`
3. here again, to update the rundown in the `runtimeCacheStore`

Would it be best to write a small utility function in ontime-utils that handles this?
like 
``` ts
function swapData(rundown: OntimeRundown, eventA: OntimeEvent, eventB: OntimeEvent): OntimeRundown
```

This has the added benefit of facilitating maintenance of the feature, since the change would only happen in one place

cpvalente (commented):
I think this looks great with the minor issue with the cache which can easily be fixed
I hope I managed to clarify in the comments, otherwise feel free to reach out

apps/client/src/common/hooks/useEventAction.ts
```diff
@@ -316,5 +324,86 @@ export const useEventAction = () => {
...
+        duration: fromEvent.duration,
+      };
+
+      previousEvents[toEventIndex] = {
+        ...fromEvent,
+        timeStart: toEvent.timeStart,
+        timeEnd: toEvent.timeEnd,
+        duration: toEvent.duration,
+      };
+
```
cpvalente, line 364: I got confused when I replied earlier and gave the wrong answer, we need to add the delay here

``` ts
      previousEvents[fromEventIndex] = {
        ...toEvent,
        timeStart: fromEvent.timeStart,
        timeEnd: fromEvent.timeEnd,
        duration: fromEvent.duration, 
        delay: fromEvent.delay
      };

      previousEvents[toEventIndex] = {
        ...fromEvent,
        timeStart: toEvent.timeStart,
        timeEnd: toEvent.timeEnd,
        duration: toEvent.duration,
        delay: toEvent.delay
      };
```

asharonbaltazar: I see. However, `delay` seems to be an optional property. If we add it and it's not present, we risk attaching `delay: undefined` to the property, right?

apps/client/src/features/rundown/event-block/EventBlock.tsx
```diff
@@ -170,7 +187,7 @@ export default function EventBlock(props: EventBlockProps) {
         <span className={style.drag} ref={handleRef} {...dragAttributes} {...dragListeners}>
           <IoReorderTwo />
         </span>
-        {eventIndex}
+        {eventIndex + 1}
```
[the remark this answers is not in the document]
asharonbaltazar: Resolved with https://github.com/cpvalente/ontime/pull/446/commits/26ad0ffe65394cb071927f6fdeb6846dbd7bbc17

apps/server/src/services/rundown-service/delayedRundown.utils.ts
asharonbaltazar: Awesome. Thank you for the lengthy reply!

Resolved with https://github.com/cpvalente/ontime/pull/446/commits/3b4ec45c12a705423fcb5459d267a9bc907b8598

apps/server/src/services/rundown-service/delayedRundown.utils.ts
asharonbaltazar: Resolving with https://github.com/cpvalente/ontime/pull/446/commits/617f107fb19e830ab5d5c051580465350f8fb3f3

The functions is used both in the frontend and backend (once for the `rundown` and `delayedRundown`)

apps/server/src/services/rundown-service/delayedRundown.utils.ts
```diff
@@ -174,7 +176,34 @@ export async function cachedReorder(eventId: string, from: number, to: number) {
...
+  if (fromCachedEvent.id !== fromEventId || toCachedEvent.id !== toEventId) {
+    // something went wrong, we invalidate the cache
+    runtimeCacheStore.invalidate(delayedRundownCacheKey);
+  } else {
+    const delayedRundownToUpdate = swapOntimeEvents(delayedRundown, fromEventIndex, toEventIndex);
+    runtimeCacheStore.setCached(delayedRundownCacheKey, delayedRundownToUpdate);
+  }
+
+  await DataProvider.setRundown(rundownToUpdate);
+  runtimeCacheStore.setCached(delayedRundownCacheKey, rundownToUpdate);
```
cpvalente, line 206: I think this call might be too many? Was it not already done in line 202?

asharonbaltazar: Resolved with https://github.com/cpvalente/ontime/pull/446/commits/a54f2421668ed234632ef171afd6d34c7573f0dc

cpvalente (approved):
Great work

PR #453, opened 2023-07-14
fix: fullscreen permissions

cpvalente:
attempt fixing issues with going fullscreen by configuring the Content Security Policy

### DIFF
--- a/apps/client/index.html
+++ b/apps/client/index.html
@@ -6,6 +6,7 @@
   <meta name='viewport' content='width=device-width, initial-scale=1' />
   <meta name='theme-color' content='#121212' />
   <meta name='ontime' content='ontime - event timer manager' />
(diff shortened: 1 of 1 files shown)


cpvalente:
Hey @asharonbaltazar, I know this is a long stretch, but maybe worth a shot

We get some unhandled errors in sentry relating to the fullscreen option in the views

> TypeError  toggleFullScreen(src/common/hooks/useFullscreen)
> Disallowed by permissions policy

the line in question is the request to go fullscreen
https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/hooks/useFullscreen.ts#L39

From a quick google it seems that we can resolve this by configuring the Content Policy in the meta attributes.
Do you have experience with this? Or at least know how I can recreate the issue myself?

asharonbaltazar:
Sorry for the late response, but I don't have experience with this :/

PR #457, opened 2023-07-20
feat: allow renaming connection

cpvalente:
Adds a simple interface for user to be able to rename the websocket client connection

### DIFF
--- a/apps/client/src/common/components/navigation-menu/rename-client-modal/RenameClientModal.tsx
+++ b/apps/client/src/common/components/navigation-menu/rename-client-modal/RenameClientModal.tsx
@@ -1,0 +1,68 @@
+import { useEffect, useState } from 'react';
+import {
+  Button,
+  Input,
+  Modal,
+  ModalBody,
+  ModalCloseButton,
+  ModalContent,
+  ModalHeader,
+  ModalOverlay,
+} from '@chakra-ui/react';
+
+import { setClientName } from '../../../hooks/useSocket';
+import { useSocketClientName } from '../../../stores/connectionName';
+
+import style from './RenameClientModal.module.scss';
+
+interface RenameClientModalProps {
+  isOpen: boolean;
+  onClose: () => void;
+}
+
+export default function RenameClientModal({ isOpen, onClose }: RenameClientModalProps) {
+  const { name: clientName, persistName } = useSocketClientName();
+  const [newName, setNewName] = useState(clientName);
+
+  useEffect(() => {
+    setNewName(clientName);
+  }, [isOpen, clientName]);
+
+  const handleRename = async () => {
+    if (newName && newName !== clientName) {
+      await setClientName(newName);
+      persistName(newName);
+      onClose();
+    }
+  };
+
+  return (
+    <Modal
+      isOpen={isOpen}
+      onClose={onClose}
+      size='sm'
+      closeOnOverlayClick={false}
+      motionPreset='slideInBottom'
+      scrollBehavior='inside'
+      preserveScrollBarGap
+      variant='ontime-small'
+    >
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>Rename client</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody className={style.modalBody}>
+          <Input
+            placeholder='Connection must have a name'
+            defaultValue={newName}
+            onChange={(e) => setNewName(e.target.value)}
+            variant='ontime-filled-on-light'
+          />
+          <Button onClick={handleRename} width='100%' variant='ontime-filled'>
+            Save
+          </Button>
+        </ModalBody>
+      </ModalContent>
+    </Modal>
+  );
+}
--- a/apps/client/src/common/components/navigation-menu/rename-client-modal/RenameClientModal.tsx
+++ b/apps/client/src/common/components/navigation-menu/rename-client-modal/RenameClientModal.tsx
@@ -29,7 +29,7 @@ export default function RenameClientModal({ isOpen, onClose }: RenameClientModal
   }, [isOpen, clientName]);
 
   const handleRename = async () => {
-    if (newName && newName !== clientName) {
+    if (newName) {
       await setClientName(newName);
       persistName(newName);
       onClose();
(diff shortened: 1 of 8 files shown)


apps/client/src/common/stores/connectionName.ts
```diff
@@ -0,0 +1,26 @@
...
+
+interface SocketClientNameState {
+  name?: string;
+  setName: (newValue: string) => void;
+  persistName: (newValue: string) => void;
+}
+
+const clientNameKey = 'ontime-client-name';
+
+async function persistKeyToStorage(newValue: string) {
```
asharonbaltazar, line 12: Any reason this is `async`? Not seeing `await` anywhere

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -69,20 +76,37 @@ export class SocketServer implements IAdapter {
...
+            );
+            return;
+          }
+
+          if (type === 'set-client-name') {
+            if (payload) {
+              const previousName = clientId;
+              clientId = payload;
+              this.clientIds.delete(previousName);
+              this.clientIds.add(clientId);
```
asharonbaltazar, line 98: So it's basically this:
```ts
this.clientIds.delete(clientId);
this.clientIds.add(payload); 
```

Kinda confusing, IMO. If these reassignments are simply for renaming, I'd suggest putting these `if` statements into functions so the context can be better understood:
```ts
if (payload) {
  return renameClientId(payload, clientId)
}

// renameClientId
const renameClientId = (newClientId, prevClientId) => { /* function stuff */}
```

We also gain the added benefit of self-documenting code (`renameClientId` vs `set-client-name`). But I understand if you wouldn't want to do that.

apps/client/src/common/utils/socket.ts
```diff
@@ -49,6 +54,11 @@ export const connectSocket = () => {
 
       // TODO: implement partial store updates
       switch (type) {
+        case 'client-name': {
+          console.log('got connection name', payload);
+          socketClientName.getState().setName(payload);
+          break
```
asharonbaltazar, line 60: missing semicolon?

apps/client/src/common/components/navigation-menu/rename-client-modal/RenameClientModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <ModalHeader>Rename client</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody className={style.modalBody}>
+          <Input
+            placeholder='Connection must have a name'
+            defaultValue={newName}
+            onChange={(e) => setNewName(e.target.value)}
+            variant='ontime-filled-on-light'
+          />
+          <Button onClick={handleRename} width='100%' variant='ontime-filled'>
```
asharonbaltazar, line 61: Perhaps disable if `newName` and `clientName` are the same. 
Would probably render the `if` statement on line 32 somewhat unnecessary. I'm not sure what's the best practice here.

apps/client/src/common/stores/connectionName.ts
```diff
@@ -0,0 +1,26 @@
...
+  localStorage.setItem(clientNameKey, newValue);
+}
+
+export const socketClientName = createStore<SocketClientNameState>((set) => ({
+  name: localStorage.getItem(clientNameKey) ?? undefined,
```
asharonbaltazar, line 17: Curious to know the added benefit of `?? undefined`. Are you against `localStorage.getItem` returning `null`?

asharonbaltazar (approved):
Left some comments! Logic looks good.

apps/client/src/common/stores/connectionName.ts
cpvalente: yes, the ?? undefined is prevent null or empty string, just to narrow down the possible returns, ie: has a name or is undefined
What do you think?
BTW, really enjoy your code reviews, good stuff

apps/server/src/adapters/WebsocketAdapter.ts
cpvalente: This is really good input-

I was considering something along those lines, especially for the client-name stuff, which is repeated throughout the code

I wonder whether this version of the code is easier to leverage and maintain since it is more linear. The API calls are completely unrelated and any repeated code is purely coincidental

apps/client/src/common/stores/connectionName.ts
asharonbaltazar: So is there something wrong with `null` that you prefer `undefined`? More like a question about preferences than logic.

Good reviews keep everyone on their feet, lol. IMO, a defining difference between a good team vs an excellent team.

apps/server/src/adapters/WebsocketAdapter.ts
asharonbaltazar: I find smaller contexts easier to understand and repeated code more confusing since you have to be sure the repeated code is exactly the same as everywhere else

apps/client/src/common/stores/connectionName.ts
asharonbaltazar: > ?? undefined is prevent null or empty string

FYI, nullish coalescing [doesn't prevent empty strings](https://stackoverflow.com/a/63924076)

cpvalente: Couldn't agree more.

I generally find undefined to be more semantic as to "the value does not exist" while null would be more inclined to convey "this value is set as nothing".
Maybe that is more to do with my C++ background

asharonbaltazar: I see. So then follow your heart!

apps/server/src/adapters/WebsocketAdapter.ts
cpvalente: Interesting. while I would generally agree I wonder on whether this specific example is the wrong context.
I also notice that I tend to do this and get yelled at by sonarqube....

I am inclined to prefer repetition as long as the contexts are isolated.
In this case every if statement runs without care for the remaining context and repetitions are coincidental IMO, this means that the changes would not necessarily replicate across all the usages. I see that there could be a danger for forced abstraction, which I see in many code bases where developers start adding optional flags. Reminds me of a talk I saw some years back from react advance, will try to find it for a future 

Extracting stuff into functions is nice, but tends to make the reader to have to jump a bit around the code, I see it as unnecessary where the code in context is under 5 lines.

I imagine that there is a more elegant pattern to do this, perhaps writing a small middleware that can attach keys to functions as you would do in a route - controller. This could give the best of both worlds, with isolation, linear readability and good semantics. One for a rainy day...

Good discussion

asharonbaltazar: > which I see in many code bases where developers start adding optional flags

Guilty as charged. 

While this all makes sense, I'll agree to disagree, lol.

Issue #431, opened 2023-06-07
Create User ID for events


cpvalente:
### Why

Currently each event has a unique identifier generated on creation / import
We can also distinguish an event by its position in the rundown.

This means that our APIs allow us to interact with an event by using either of these fields like
```
/ontime/startid f6cac
/ontime/startindex 3
```

However, in the user workflows, we would need a static reference to an event that is more meaningful than the random id `f6cac`, but that does not depend on the runtime,
This is common practice in a cued system where a cue remains 101 even if 100.5 is added.

### TODO

The ticket task is to add an extra field to the event definition, which allows the user to provide an identifier to the field. 
This will be called cue in our datasets

The tasks is not particularly complex, but involves several parts of the code. We can aim to have smaller multiple pull requests that achieve the tasks defined rather than the entire feature description

The task would in short achieve:
- [ ] Add new cue field to OntimeEvent type and all its creation / parsing steps
- [ ] create API interfaces for this eg: /ontime/startcue (this would live in [dispatchFromAdapter](https://github.com/cpvalente/ontime/blob/master/apps/server/src/controllers/integrationController.ts#L5)
- [ ] create an editable field in the [EventEditor ](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/event-editor/EventEditor.tsx) for this, potentially where the ID is currently shown 
- [ ] investigate on how to present this in the editor view, could it go in the binder?
- [ ] add a new column in the cuesheet to show this

Some extra requirements:
- all events need to contain a cue, so we need to create one when an event is created 
- the cue value should be an increment on the previous value So after A1 we create A2, after Preshow 1 Preshow 2, and after NotNamed NotNamed 2
- once an event is running if an event is reordered and if a) we are in run mode: nothing changes b) we are in edit mode: we rename the event following the convention above

The [current event definition is here](https://github.com/cpvalente/ontime/blob/master/packages/types/src/definitions/core/OntimeEvent.type.ts#L27)

cpvalente:
Implemented in #473

PR #460, opened 2023-07-21
Some quickly implemented ideas / hotfixes / PoCs

hrueger:
Hi @cpvalente,

I know that you ask for opening an issue before making a PRs, but this is not a PR. It is basically just for reference. Let me explain:

I was working in a production for a concert and we wanted to use ontime for our rundown and organizational scheduling. We did find some limitations and I tried to overcome them. Due to severe time contstraints and unforeseen events, I was not able to implement everything completely. This is have done - and I'm very sorry that I can't finish that. I'm usually not someone who's like "take it from here" and this is not the idea behind this PR. Maybe you and some other people like the concepts and want to use this as a starting point.

I'd really like to finalize all that work but I have too many other things to do. 

Here's the list of changes that might be interesting:
- b455fd742acf36f94fc76fabf8fb829254d08a6e and 7b7f97192b1e532a8511513fc0337115fde83b30 - **User Field Improvements**
	- can be enabled or disabled, then they show up in the editor UI
	- removed some code duplication (https://github.com/cpvalente/ontime/commit/b455fd742acf36f94fc76fabf8fb829254d08a6e#diff--741c0e050152bbab3d949a9b679badc8dac1f41c9432089344598135a261dd29L75-R103) -> easier for changes in the future
	- 7b7f97192b1e532a8511513fc0337115fde83b30 **show notes and user fields in the ui**
- ab50700a8c454849127d76c07e8b0f769a2ac893 and others: replace `is Public` with **multiple departments** 😃🥳
	- Configurable in settings (after fix with d15d9ea056b3bc37a9ca64bc64178036de13eb62)
	- Create the Timeline View (below)
- e67e20ee366700c24b6c5d978e78df424edea9e2: **New Timeline View😃🥳**
	grafik
- 3111454219b75dd3b8127f914cd7e195dbb3b5a8 **User Cue IDs** (see #431) 😃🥳
	- Here we allowed that multiple events (with the same cue) can be played at the same time. The Editor Rundown displays that correctly, although there's a bit of a hack: not the backend tells all events that they are running, but only the first one. The UI then highlights all events with that cue. I'd really like to see that `titles` object removed and instead have every `OntimeEvent` a property called something like `_state` or `_runtimeInfo` which contains all that information (running, timer start date, ...). Then, logic like multiple cues running simultaneously like from different departments would be way easier ;-)
- 63c6ac12c59990cc4a73526260781b71fe186c4d and a0201b231982da9db8202d93d15fc9194d750c7f: **Blink if the event changes**
	- This is important since it draws attention of the operators
- 5c551c8ce02cc1d8ece9ea290c50f0be315660ef Filter events by department (currently works for schedule only, not for the timers...)
- ea5e9ca2ea28f4a2fdaffeae7759de960ed7feae **Fixed connection in dev mode from external devices**
	- we ran Ontime in local development and noticed, that the hostname was not correctly resolved for external devices
- 0bc4dcca1cd09090650b07950e209406bb2a6bbc **Google Sheets Sync** 😃🥳 Please tell me if you need a reference sheet.
- f8128da3992cc354dc67cecdcf6093a9d3307138 **hide all the configuration options in the menu bar in run mode** (for safety reasons)

Generally, I found it a bit difficult to add new properties to objects, since most react components
 get each property individually. I've never worked with react before, but for me it also worked with always providing the whole objects instead of individual properties. This way changes in components are way easier as you don't have to change every single file, where that component is used, because you need another property of an object.

Let me know what you think! I'm happy to explain my thoughts behind the changes, there should always be enough time for that!

Best regards
Hannes

### DIFF
--- a/apps/client/package.json
+++ b/apps/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ontime-ui",
-  "version": "2.3.9",
+  "version": "2.7.0",
   "private": true,
   "dependencies": {
     "@chakra-ui/react": "^2.7.0",
--- a/apps/client/src/features/rundown/Rundown.tsx
+++ b/apps/client/src/features/rundown/Rundown.tsx
@@ -234,7 +234,7 @@ export default function Rundown(props: RundownProps) {
 
               return (
                 <Fragment key={entry.id}>
-                  <div className={style.entryWrapper}>
+                  <div className={style.entryWrapper} data-testid={`entry-${eventIndex}`}>
                     {entry.type === SupportedEvent.Event && <div className={style.entryIndex}>{eventIndex}</div>}
                     <div className={style.entry} key={entry.id} ref={hasCursor ? cursorRef : undefined}>
                       <RundownEntry
--- a/apps/client/src/features/rundown/quick-add-block/QuickAddBlock.tsx
+++ b/apps/client/src/features/rundown/quick-add-block/QuickAddBlock.tsx
@@ -80,7 +80,7 @@ const QuickAddBlock = (props: QuickAddBlockProps) => {
             size='xs'
             variant='ontime-subtle-white'
             className={style.quickBtn}
-            data-testId='quick-add-event'
+            data-testid='quick-add-event'
           >
             Event {showKbd && <span className={style.keyboard}>Alt + E</span>}
           </Button>
@@ -92,7 +92,7 @@ const QuickAddBlock = (props: QuickAddBlockProps) => {
             variant='ontime-subtle-white'
             disabled={disableAddDelay}
             className={style.quickBtn}
-            data-testId='quick-add-delay'
+            data-testid='quick-add-delay'
           >
             Delay {showKbd && <span className={style.keyboard}>Alt + D</span>}
           </Button>
@@ -104,7 +104,7 @@ const QuickAddBlock = (props: QuickAddBlockProps) => {
             variant='ontime-subtle-white'
             disabled={disableAddBlock}
             className={style.quickBtn}
-            data-testId='quick-add-block'
+            data-testid='quick-add-block'
           >
             Block {showKbd && <span className={style.keyboard}>Alt + B</span>}
           </Button>
--- a/apps/client/src/index.tsx
+++ b/apps/client/src/index.tsx
@@ -15,10 +15,10 @@ const root = createRoot(container as Element);
 Sentry.init({
   dsn: 'https://5e4d2c4b57ab409cb98d4c08b2014755@o4504288369836032.ingest.sentry.io/4504288371343360',
   integrations: [new BrowserTracing()],
-  tracesSampleRate: 1.0,
+  tracesSampleRate: 0.3,
   release: ONTIME_VERSION,
   enabled: import.meta.env.PROD,
-  ignoreErrors: ['top.GLOBALS', 'Unable to preload CSS'],
+  ignoreErrors: ['top.GLOBALS', 'Unable to preload CSS', 'Failed to fetch dynamically imported module'],
   denyUrls: [/extensions\//i, /^chrome:\/\//i, /^chrome-extension:\/\//i],
 });
 
--- a/apps/electron/package.json
+++ b/apps/electron/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ontime",
-  "version": "2.3.9",
+  "version": "2.7.0",
   "author": "Carlos Valente",
   "description": "Time keeping for live events",
   "repository": "https://github.com/cpvalente/ontime",
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -2,7 +2,7 @@
   "name": "ontime-server",
   "type": "module",
   "main": "src/index.ts",
-  "version": "2.3.9",
+  "version": "2.7.0",
   "exports": "./src/index.js",
   "dependencies": {
     "body-parser": "^1.20.0",
--- a/e2e/tests/features/204-editor-crud.spec.ts
+++ b/e2e/tests/features/204-editor-crud.spec.ts
@@ -17,40 +17,16 @@ test('CRUD operations on the rundown', async ({ page }) => {
   await page.getByTestId('quick-add-event').click();
 
   // test quick add options - start is last end
-  await page
-    .locator('div')
-    .filter({ hasText: /^22SEDEvent title$/ })
-    .getByTestId('time-input-timeEnd')
-    .click();
-  await page
-    .locator('div')
-    .filter({ hasText: /^22SEDEvent title$/ })
-    .getByTestId('time-input-timeEnd')
-    .fill('20m');
-  await page
-    .locator('div')
-    .filter({ hasText: /^22SEDEvent title$/ })
-    .getByTestId('time-input-timeEnd')
-    .press('Enter');
+  await page.getByTestId('entry-2').getByTestId('time-input-timeEnd').fill('20m');
   await page.getByText('Start time is last end').click();
   await page.getByTestId('quick-add-event').click();
-  await expect(
-    await page
-      .locator('div')
-      .filter({ hasText: /^33SEDEvent title$/ })
-      .getByTestId('time-input-timeStart')
-      .inputValue(),
-  ).toContain('00:20:00');
+  await expect(await page.getByTestId('entry-3').getByTestId('time-input-timeStart').inputValue()).toContain(
+    '00:20:00',
+  );
 
   // test quick add options - event is public
   await page.locator('label').filter({ hasText: 'Event is public' }).click();
   await page.getByTestId('quick-add-event').click();
 
-  await expect(
-    await page
-      .locator('div')
-      .filter({ hasText: /^44SEDEvent title$/ })
-      .getByRole('img')
-      .nth(3),
-  ).toHaveAttribute('data-isPublic', 'true');
+  await expect(await page.getByTestId('entry-4').getByRole('img').nth(3)).toHaveAttribute('data-isPublic', 'true');
 });
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ontime",
-  "version": "2.3.9",
+  "version": "2.7.0",
   "description": "Time keeping for live events",
   "keywords": [
     "lighdev",


cpvalente:
Hi @hrueger,

Thank you very much for this. It is unbelievable the amount of work you have here.
I haven't yet gone through all of it and am very excited to do so, it looks like that might take a while 😅 
If it is OK with you I will start by picking some of the smaller fixes and improvements into a branch to be merged in the next release

I am curious as to which barriers did you find in using Ontime, were you attempting to use it as a replacement to another software?

Meanwhile, it would be a great chance to discuss some of your decisions here if your time allows. 

> [User fields] can be enabled or disabled, then they show up in the editor UI

Can you provide some context for this change?

> https://github.com/cpvalente/ontime/commit/e67e20ee366700c24b6c5d978e78df424edea9e2: New Timeline View😃🥳

Nice! It is almost like you are reading my notes, this will go great with our operator view

Here we allowed that multiple events (with the same cue) can be played at the same time. (...)  I'd really like to see that titles object removed and instead have every OntimeEvent a property called something like _state or _runtimeInfo which contains all that information (running, timer start date, ...). Then, logic like multiple cues running simultaneously like from different departments would be way easier ;-)

I am very curious about this. I have pondered about solutions here but always thought that this should come from the server. How did this work in your opinion? 
I am also not sure I understood the the suggestion for the change, would you mind expanding?

> https://github.com/cpvalente/ontime/commit/0bc4dcca1cd09090650b07950e209406bb2a6bbc Google Sheets Sync 😃🥳 Please tell me if you need a reference sheet.

Wow, this is pretty impressive, could you talk me through the process here? Did you end up using this in production or how far do you think it might be?

> Generally, I found it a bit difficult to add new properties to objects, since most react components get each property individually. I've never worked with react before, but for me it also worked with always providing the whole objects instead of individual properties. This way changes in components are way easier as you don't have to change every single file, where that component is used, because you need another property of an object.

I understand, admittedly Ontime does suffer from prop drilling, especially around the `Rundown`. There is also an "interesting" solution on the `ViewWrapper`.
To me it feels like passing the fields individually gives me better control, there is quite a lot of derived data and partial data being split. I did consider creating a small store for every component to combat this, but never seemed to find the time.

Otherwise, how did you feel about working with ontime? I would like for it to be friendly to new comers (or as friendly as a medium - large application can be).


Thank you again,

hrueger:
Hi @cpvalente,
thanks for the nice and very quick reply 😃

> Thank you very much for this. It is unbelievable the amount of work you have here.
I haven't yet gone through all of it and am very excited to do so, it looks like that might take a while 😅

Sure, you're welcome. Please note that most of this has been written while I was also setting up equipment while already at the venue, so there was not a lot of time to test and as I said, not everything was completed in time ;-)

> If it is OK with you I will start by picking some of the smaller fixes and improvements into a branch to be merged in the next release

Sure, please take whatever you like! Happy to see that you like it.

---

> I am curious as to which barriers did you find in using Ontime, were you attempting to use it as a replacement to another software?

So the production we just finished was Europe's biggest school concert. Because of limited budget we couldn't afford 
software like for example CuePilot. So we looked into alternatives and liked Ontime a lot, especially because it was open source.

The main thing we needed was the ability to give different cues to different departments, for example:
- Doors Open:
	- Sound: Bell 1
	- Sound: Bell 2
	- Sound: Bell 3
	- Light: Turn off Main Hall Lights
	- Stage: Open Curtain
- Song 1:
	- Light: Cue 1001
	- ...

The first idea we had, was using the user fields for that. That's why I made them show up in the editor and in some of the views. However, seeing 10 fields was overkill, so I added the enable buttons to only show enabled fields.

However, we then found sometimes that we needed different start times for those department-specific events. Therefore, I updated the `isPublic` to `department` and added the settings tab.

> I am very curious about this. I have pondered about solutions here but always thought that this should come from the server. How did this work in your opinion?
I am also not sure I understood the the suggestion for the change, would you mind expanding?

Sure. Currently, theres the `OntimeRundown` which is basically just an array of `OntimeEvent`s. Then, we have the `titles` and `titlesPublic` object. The backend copies properties into those objects with the suffix `Now` or `Next`. I find this a bit redundant. In addition, it makes it difficult to implement department filtering in the views, because we'd need a `titles` object per department.

I suggest the following: Let's add a property like `_state` or `_runtimeInfo` to the `OntimeEvent` type. This should not be saved, as it is runtime-only. In this new property we could keep an object containing all relevant status info (`running`, `startedAt`, `deltaOff`). The backend could send update messages like `update-event` with `{ eventId: string; data: Record<string, any> }` with all changed properties of that event, whenever their state changes (play, pause, +-1, +-5, editor changes, ...). When a client receives that event, they update the properties of that event in their rundown variable. This way, not the whole rundown or the whole event has to be sent each time something changes.

This way, it is much easier for views to implement custom logic like filtering by department, displaying the last event, too, or similar things (as they have access to all events including their status)

Does that make sense to you?

>  I assume FOH could be a department and our public screens could remain working focusing on events with that filter

Yes, exactly. We had 4 departments: Audio (for the PA), Light, Video (Live-Content for the screens as well as the livestream) and Stage (Logistics, Curtain, Chairs...)

> How about events like "Band Act 2", which does not necessarily have a department, how do you see this would be treated

I added a department None field for general events. We created one for every song. However, it might also make sense to group events into folders? Then, we could have a folder for every song. I'm not sure how that would be seen in the timeline view though.

> What sort of departments do you see here? are there any departments the app should enforce (eg: FOH with the above thought)?

See above for the list of departments we had. I think it would make sense to provide useful defaults, but not enforce any since every event is very different. I personally wouldn't call FOH a department, as there might be Light and Sound people there which are interested in very different cues.

> How do you see that events that concert multiple departments should be handled?

That is a great question. I talked to @lukas-runge who was at that event with me and we were both unsure. We couldn't come up with a cue which is exactly the same for two departments, since they all get different instructions, at least in our case. For example: at chorus of a specific song: light bright bumps, sound chorus effect for some channels, stage: prepare for next act in the background. General events without any department are interesting for all though, like next song, next break, ...

---

> [Google Sheets Sync] Wow, this is pretty impressive, could you talk me through the process here? Did you end up using this in production or how far do you think it might be?

Sure, let's do that in another comment, this one is already too long 😉

---

Regarding

> To me it feels like passing the fields individually gives me better control, there is quite a lot of derived data and partial data being split. I did consider creating a small store for every component to combat this, but never seemed to find the time.

as well as 

> Otherwise, how did you feel about working with ontime? 

Generally, I found it very easy to setup my development environment and get it running quickly. I'm not a React-guy though, I use Angular and SvelteKit mostly. I found it a bit difficult to find the paths where data travels through, but probably because I'm not familiar with react concepts. It was especially difficult to discover the files where data from the settings is read again. I had an issue where my departments were not loaded after restarting ontime. Turns out: each section is parsed individually.

Regarding the prop drilling: To be honest, I'm not a big fan of that. Also I think that creating a store for each component is over-engineeded, though I don't know how "enterprisey" you want it to be 😉 If my suggestion from above with giving events with their runtime-state to the frontend makes sense, it might be possible to have one store and make each view component use that.

What do you think?

Best regards
Hannes

hrueger:
## Google Sheets Sync

We actually used that feature quite a lot and I'd say it is quite done. As it overwrites the complete rundown, we made sure to not make any changes in the Ontime editor and instead only in the Google Sheet. Here's a demo sheet https://docs.google.com/spreadsheets/d/1NXemoq1cqjuwzaRr-hKSdxxhOpukTMG5dna7gRKLT3o/edit?usp=sharing
Some things are still in German since I just copied our schedule and shortened it a bit.

Authentication currently works via the [API Key option](https://theoephraim.github.io/node-google-spreadsheet/#/guides/authentication?id=api-key). Maybe it would make more sense with other options, but this one seemed the easiest (at least to implement, probably not for the users)

Just go to Integrations, set your sheet id (from the URL, the name of the sheet and the API key).

Currently, the importer expects the sheet to be formatted like the demo sheet, this should probably configurable. If the Cue has a department set in column `I`, it's department is set to that value. However, if the department is `-` and it has notes in the section in the very right, the event is duplicated for each department and `note` and `subtitle` from the respective column is used.
Those events get the same cue, so they are triggered simultaneously.

Does that make sense to you? It would be cool to be able to push changes back to the cloud and therefore create a bidirectional sync, but this requires a different authentication method (API Key is readonly) and probably also ids in the sheet.

cpvalente:
> ## Google Sheets Sync
> 
> We actually used that feature quite a lot and I'd say it is quite done. As it overwrites the complete rundown, we made sure to not make any changes in the Ontime editor and instead only in the Google Sheet. Here's a demo sheet https://docs.google.com/spreadsheets/d/1NXemoq1cqjuwzaRr-hKSdxxhOpukTMG5dna7gRKLT3o/edit?usp=sharing Some things are still in German since I just copied our schedule and shortened it a bit.
> 
> Authentication currently works via the [API Key option](https://theoephraim.github.io/node-google-spreadsheet/#/guides/authentication?id=api-key). Maybe it would make more sense with other options, but this one seemed the easiest (at least to implement, probably not for the users)
> 
> Just go to Integrations, set your sheet id (from the URL, the name of the sheet and the API key).
> 
> Currently, the importer expects the sheet to be formatted like the demo sheet, this should probably configurable. If the Cue has a department set in column `I`, it's department is set to that value. However, if the department is `-` and it has notes in the section in the very right, the event is duplicated for each department and `note` and `subtitle` from the respective column is used. Those events get the same cue, so they are triggered simultaneously.
> 
> Does that make sense to you? It would be cool to be able to push changes back to the cloud and therefore create a bidirectional sync, but this requires a different authentication method (API Key is readonly) and probably also ids in the sheet.

Absolutely.
I am not familiar with the procedures here, in my head:
- we don't want to keep any user information in ontime, we would somehow redirect the user to a page where they could authorise the app
- the process should be bi-directional to allow for collaboration but remain safe inside ontime's domain. ie: things like playmodes and settings related to runtime could be tricky to translate into a sheet integration. But most of the event information (titles and other event info) can be safely exposed

I am currently working on a slightly different roadmap but I believe this would be of major benefit to the users.
I wonder whether a first draft implementation similar to yours is enough of a game changer to consider, or perhaps a bi-directional implementation would not be particularly more expensive and is worth looking into?

Do you have any opinions on this? Would you be available to code review a potential implementation (ie. pretty much copy-paste for the first case)


btw: beautiful work with the sheet 😄

hrueger:
> we don't want to keep any user information in ontime, we would somehow redirect the user to a page where they could authorise the app. This seems to be what a [service account](https://theoephraim.github.io/node-google-spreadsheet/#/guides/authentication?id=service-account) is, unfortunately this also seems to be a paid service

Authentication seems to be the difficult point here, you're right. I was thinking about an OAuth 2.0 authentication, but I'm not sure: I've only done things like that with a real backend on a server which is completely private. Therefore, it was no problem saving a private key or a client secret. I wonder if that is still secure if we were to build that into a desktop application, since everybody can extract the .asar file and find the keys...

> the process should be bi-directional to allow for collaboration but remain safe inside ontime's domain. ie: things like playmodes and settings related to runtime could be tricky to translate into a sheet integration. But most of the event information (titles and other event info) can be safely exposed

I completely agree with you on that 👍

> it should include some sort of stream so that the application can be notified of changes. I did not find this in my (quick) documentation read

If we had a central backend on a server, we could do that using [webhooks](https://developers.google.com/drive/api/guides/push?hl=de) but I didn't find a way using websockets or similar.

> I am currently working on a slightly different roadmap but I believe this would be of major benefit to the users.
I wonder whether a first draft implementation similar to yours is enough of a game changer to consider, or perhaps a bi-directional implementation would not be particularly more expensive and is worth looking into?

I think one-way sync would be ok for now. Maybe we could add a warning in the editor that the data came from a sheet and should not be modified here - or (though I don't know which one I'd prefer) we could add a warning in the sync dialog that the rundown has been modified locally and would be overwritten.

> Would you be available to code review a potential implementation (ie. pretty much copy-paste for the first case)

Sure! I can't promise to be able to do that right away, but I should always find some time here and then.

> btw: beautiful work with the sheet 😄

Thanks, but all props for that go to @lukas-runge, he made that 😉

cpvalente:
## On the google sheet integration

Hi @hrueger, had some time to read and I think I would have been OK with implementing a read-only version with the idea that this is temporary.

As it seems that any bidirectional version would not be possible outside a paid plan. 
Google also doesn't seem to be particularly transparent on the licensing costs and relates this to data usage, which we cannot be in control.

If these assumptions are correct, it makes me reluctant to implement the feature

Do you think I am missing something?

In the case I am not, how could we improve the excel import flow so that it can fill in the gap?

cpvalente:
> I just see a big advantage of google sheets over excel: sheets are easily sharable. We used that quite a lot - multiple people had to make changes to the rundown, so we just shared the link. This would not have been possible with an excel file or at least not as easily.

Hi @hrueger.
Sorry for the slow replies. The short answer is that I don't know the best way forwards and would either a) need to take some time to investigate or b) have someone advising the process.

This is a major improvement for the app feature set so we would like to do it.
For option a) it could take a while for me to tie the existing loose ends. Perhaps you are available to advise?

The bottom line for the app:
We don't want to keep any user secrets, if we must hold it, it would be in memory, so users would have to add this every session
Otherwise we would like to have the best UX possible outside of us needing to pay Google

Thank you again for all your contributions

hrueger:
Hi @cpvalente,
> Sorry for the slow replies. The short answer is that I don't know the best way forwards and would either a) need to take some time to investigate or b) have someone advising the process. [...] Perhaps you are available to advise?

No worries. Sure, I'm always happy to do so.

> We don't want to keep any user secrets, if we must hold it, it would be in memory, so users would have to add this every session
Otherwise we would like to have the best UX possible outside of us needing to pay Google

That's right. I just spend a few minutes trying to get OAuth2 without a client secret working in an electron app. This used to be possible (especially with the iOS application type (search for `OAuth2 with Installed Apps (Electron)` OAuth2 with Installed Apps (Electron)), but apparently Google dropped support.

I was able to get it working fully client side, which I think would not be a problem. We can read and write the sheet client side and process data on the server side. What do you think?
Basically, I just followed this tutorial by Google: https://developers.google.com/sheets/api/quickstart/js

The workflow for the user would be very convenient, since they just have to allow access to the application. 

What do you think, could this be a way?

Thank you again for all your contributions

hrueger:
Ups, I didn't want to close this. Changes are still available in [this branch](https://github.com/hrueger/ontime/tree/my-weird-fixes).
I just needed to update my fork, because I'm looking into the google sheets via OAuth feature.

cpvalente:
Hahahaha, no worries.
Just this morning had a chat about google sheets integration, great news that you are helping on that :D

hrueger:
@cpvalente 
Still stuck with the authentication ([fyi](https://github.com/googleapis/google-auth-library-nodejs/issues/1411#issuecomment-1703804817)).
I found [this workaround](https://stackoverflow.com/a/66790710/13485777) which I found pretty neat. What do you think, could this be a solution? We'd host an authorization page on the GitHub Pages site (then, we have a fixed origin) which opens the electron app via the return url. However, this would mean that the authentication would only work in the Electron app and not from a different device / server... Would this be ok for now or do you think we should rather wait for Google to implement better support? I don't have any idea how long this is going to take.

cpvalente:
Hi @hrueger , thank you for this and great work.

My first impression was that, given googles track record, we are better off not waiting. However there seems to be some activity in the ticket you link and I might just be wrong.

Meanwhile, it is also true that most power users, which would benefit the most from the feature, use ontime behind a cloud service. Alienating those is not great.

On the other hand, to implement this feature successfully, we need to do some prep work in ontime. We would need to create a upsync / downsync process that can synchronise local <>remote changes in a running system.
This can be pretty complex and will take some time to get right.

This is my suggestion:

Can we do something straightforward right now to get it working just for ourselves while we work on the downsync?

We then postpone the decision of how to do the auth flow for users. With some luck there has been some development on this by then
What do you think?

Issue #374, opened 2023-04-30
keep several project files in user folder


iqtreiber:
As a very future request, it would be great to enhance the setup and configuration in case a event runs more than 1 day.
Sometimes events get planed for 2-3-4 days and applications and organiser have to plan multiple days.

Maybe in the long future of Ontime, it would be possible to get tabs on the top of editor app.
Each tab could be a day of event and can be configured.

any consultation or feedback is highly welcome.
Thank you for your support!

je71175:
Multiple days and multiple stages (eg a festival) or rooms (eg concert).

cpvalente:
I appreciate this is probably not what you would like to hear:

Ontime has no concept of time in a schedule sense. All time is tracked in relationships between events in milliseconds.
We could modify the app to deal with real date objects, but I am sceptical whether this is the right direction for the app.

>  Each tab could be a day of event and can be configured.
What would this configuration look like? What things would be configurable in a per day basis?
Is day the right split for an event? Often an event is not constrained by a 24-hour period. Would this mean that an event and a day have to be two different concepts that would need to be fully implemented to realise this feature?

> Multiple days and multiple stages (eg a festival) or rooms (eg concert).
Extending this into a multi-room concept from a single ontime instance is a task for another app since it is such a fundamental re-write

However, I do like the idea of Ontime being able to hold onto multiple projects you could easily load. @iqtreiber and @je71175 would that be a correct step forwards for you?

je71175:
Yes, I think that works fine.  thank you

cpvalente:
Writing some feature requirements here

Currently, Ontime keeps a single json file with its definition.
The file is kept in a public folder resolved in [getAppDataPath](https://github.com/cpvalente/ontime/blob/master/apps/server/src/setup.ts#L11)
When a new file is loaded, we delete the existing and upload and parse the new

Nice...

### Feature requirements
We need ontime to keep several files in the public folder. This would mean a few things:
- We need somewhere to see a list of existing files. I would suggest that we use the <[UploadModal](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/modals/upload-modal/UploadModal.tsx)> and repurpose as a generic "load file modal"
- When we [upload a file](https://github.com/cpvalente/ontime/blob/master/apps/server/src/utils/upload.js), we make a temporary name. We would need to keep the user names

### Some pointers
This is mostly a nodejs issue.
[...]

Issue #194, opened 2022-08-20
Feature Request: Sync with Google Sheets


Fred-DTV:
Hi @cpvalente 

Thanks for this amazing project. It is truly a game changer.

I was wondering if it would be possible to have ontime linked to a rundown in Google Sheets, as were are mostly using Sheets because of the easy collaboration with others.

A simple import button would be enough for the beginning, and a real-time Sync would be the dream scenario

cpvalente:
Hi @DachsbauTV , thank you for your suggestion. This really sounds like a good addition to the feature set.

Currently you can import any XLS file and we have a feature in testing to export CSV. This should provide good foundations to start a feature similarly to what you describe.

I am happy to write a feature description and it would be great to get some help. A few questions:
- Would this feature be targeted specifically at Google Sheets instead of a local instance of Excel? 
- Usually ontime would be in local / offline environments, is it a good idea to lock workflows into a cloud service?
- You would need to add account secrets to ontime which could potentially become a security risk, where would this information reside? I imagine it would be a file that can be added to the ontime installation?
- In what part of production would live sync be in place? There would be a risk for misuse and accidentally resetting the timers if used in live production. Could you give an example of your desired workflow?

Meanwhile it would be great if you could test the existing import, this should work fine with Google Sheets. Roundtrip import-export should be ready as soon as cb1da0b834a7c850b368f8fba1634db4082fc7d4 is merged. This is currently being tested and should be merged in the upcoming weeks. 
If you would like to get involved with testing it would also be very useful, drop me an email at mail@getontime.no and we can look through it.

Thank you again

DachsbauTV:
Hi Carlos,

The current import works quite well with a downloaded .xlsx file from google sheets. (just the little hiccup with the user fields)
Why I think a Sheet implementation would be great is because in a bigger production environment where multiple people work on one rundown, it can be almost impossible to get a "final" file for download and import except just a couple of minutes before the actual live starts. Having to import the import the file at that point is a potential stress factor or even worse.

> * Would this feature be targeted specifically at Google Sheets instead of a local instance of Excel?
Yes, however there is an Excel Online which could use the same functionality. But I think, Sheets is still more common.

> * Usually ontime would be in local / offline environments, is it a good idea to lock workflows into a cloud service?
I think it is no problem, because almost all workflows in a production nowadays are online. And it still would only need a tiny bit of ressources (CPU/bandwidth/etc) which is no problem at that level.

[...]

PR #464, opened 2023-07-22
Change server port

### LINKED ISSUE #358: Help needed: Change server port during runtime

pSnehanshu:
Closes #358


pSnehanshu:
@cpvalente feel free to look at the changes to see if I am going the right path.

Also, how do I test it?

apps/server/src/controllers/ontimeController.ts
```diff
@@ -204,6 +204,7 @@ export const postSettings = async (req, res) => {
     const settings = DataProvider.getSettings();
     const editorKey = extractPin(req.body?.editorKey, settings.editorKey);
     const operatorKey = extractPin(req.body?.operatorKey, settings.operatorKey);
+    const serverPort = extractPin(req.body?.serverPort, settings.serverPort.toString());
```
cpvalente, line 207: nice, few notes from our previous talks
- we should refuse the change if we are running in docker
- we should add this to the validation function

Do you think w could check if the port is available before making the change? If the user were to change this it could potentially "brick" the app.

Unsure if that should happen here or when the server attempts start up

apps/electron/main.js
```diff
@@ -30,21 +30,23 @@ let splash;
...
-    await startOSCServer();
-    await startIntegrations();
-  })();
+  const result = await startServer();
+  loaded = result.message;
+
+  await startOSCServer();
+  await startIntegrations();
+
+  return result.serverPort;
```
cpvalente, line 49: interesting, the IIFE was to workaround the usage of esm module in a cjs environment. curious to see if this still works

cpvalente (commented):
nice! I think you are absolutely going the right direction, see some notes and small comments

apps/server/src/controllers/ontimeController.ts
pSnehanshu: > Do you think w could check if the port is available before making the change? If the user were to change this it could potentially "brick" the app.

I don't know, do you expect the server to immediately reboot and listen on the new port when this is changed or is it ok to use this port when it is restarted manually by the user?

apps/electron/main.js
pSnehanshu: Hmm... Interesting, I see no reason for this to not work, but I am not to able to test it even because for some reason the `build-win` command is failing.

cpvalente: A-ha

It probably fails because of the missing sentry key

You can run `pnpm build:local` on the frontend to fix this

pSnehanshu:
No, I ran `build:local` and it worked, but `pnpm run build-win` fails due to failure in downloading Electron's zip file.----Sincerely,Snehanshu  PhukonFullstack web developerhttps://snehanshu.comAssam, India ---- On Sat, 22 Jul 2023 19:11:51 +0530  ***@***.***  wrote ----
@cpvalente commented on this pull request.



In apps/electron/main.js:
> +  // in dev mode, we expect both UI and server to be running
+  if (!isProduction) {
+    return;
+  }
 
-    const ontimeServer = require(nodePath);
-    const { initAssets, startServer, startOSCServer, startIntegrations } = ontimeServer;
+  const ontimeServer = require(nodePath);
+  const { initAssets, startServer, startOSCServer, startIntegrations } = ontimeServer;
 
-    await initAssets();
+  await initAssets();
 
-    loaded = await startServer();
-    await startOSCServer();
-    await startIntegrations();
-  })();
+  const result = await startServer();
+  loaded = result.message;
+
+  await startOSCServer();
+  await startIntegrations();
+
+  return result.serverPort;

A-ha
It probably fails because of the missing sentry key
You can run pnpm build:local on the frontend to fix this

pSnehanshu:
@cpvalente I was able to run `pnpm build:local` and `pnpm dist-linux`, and the AppImage file got generated. But when I try to run it, I get this error, and the app gets stuck in the splash screen:

Screenshot from 2023-07-24 22-13-40 (1)

Do you know how to fix this?

pSnehanshu:
@cpvalente When I run `pnpm install`, the `pnpm-lock.yaml` file changes like this:

Screenshot from 2023-07-24 22-20-58

Do you want to do anything about this?

cpvalente:
> @cpvalente I was able to run `pnpm build:local` and `pnpm dist-linux`, and the AppImage file got generated. But when I try to run it, I get this error, and the app gets stuck in the splash screen:
> 
> 
> 
> Screenshot from 2023-07-24 22-13-40 (1)
> 
> 
> 
> Do you know how to fix this?

I was working in Ubuntu not long ago, so we know this works fine.
Did you maybe forget to build the server?

cpvalente:
> @cpvalente When I run `pnpm install`, the `pnpm-lock.yaml` file changes like this:
> 
> 
> 
> Screenshot from 2023-07-24 22-20-58
> 
> 
> 
> Do you want to do anything about this?
> 
> 

It looks like you might be using a newer version of pnpm.

The builder should complain, for now, I guess your change has no reason to push the package, so we can just avoid committing it. 

It could be an idea in future to upgrade pnpm in the project

cpvalente:
Hi @pSnehanshu, I am sorry you seem to be stuck trying to test this.
I am happy to try and help, get in touch at mail@getontime.no and we can find a way

cpvalente:
Hi @pSnehanshu ,

Had a quick look at your PR.
Ontime has a process which ensures that the project file is healthy and valid every startup, that process is currently stopping the serverPort from being read

You would need to change the definition here
https://github.com/cpvalente/ontime/blob/master/apps/server/src/utils/parserFunctions.ts#L127

to look something like 
```
      const settings = {
        serverPort: s.serverPort || dbModel.settings.serverPort,
        editorKey: s.editorKey || null,
        operatorKey: s.operatorKey || null,
        timeFormat: s.timeFormat || '24',
        language: s.language || 'en',
      };
```

apps/server/src/controllers/ontimeController.ts
cpvalente: I think as first step we would want the user to restart

cpvalente, line 207: I am not sure we want to get the value from extractPin, this can never be null

We would need to add this value to the validation, so that, if it exists, it must be a numeric value
then you can just convert it using `Number(body.serverPort)`

cpvalente:
If you are having trouble with turbo you can recreate its steps manually

- install dependencies `pnpm i`
- navigate to the react app `apps/client`
- build app without sentry `pnpm build:local`
- navigate to the backend `apps/server`
- build app `pnpm build`
- navigate to the electron app `apps/electron`
- create package `pnpm dist-linux`

Hope this helps

apps/server/src/controllers/ontimeController.ts
pSnehanshu: > I am not sure we want to get the value from extractPin, this can never be null

Well I just followed the pattern.

> We would need to add this value to the validation

Yeah, I will do that too.

pSnehanshu:
I was able to make it work. But there are still a lot of hardcoded 4001s in the whole codebase. Not sure what to do about them.

cpvalente:
> I was able to make it work. But there are still a lot of hardcoded 4001s in the whole codebase. Not sure what to do about them.

It should be trivial to dev with those now. Just replacing with the dynamic values. 

I can do that if it helps. I am unsure that there wouldn't be so many though. Are there?

pSnehanshu:
Help is appreciated 🙏. There are a lot of them, just search for 4001.----Sincerely,Snehanshu  PhukonFullstack web developerhttps://snehanshu.comAssam, India ---- On Wed, 02 Aug 2023 22:24:18 +0530  ***@***.***  wrote ----

I was able to make it work. But there are still a lot of hardcoded 4001s in the whole codebase. Not sure what to do about them.

It should be trivial to dev with those now. Just replacing with the dynamic values.
I can do that if it helps. I am unsure that there wouldn't be so many though. Are there?

apps/electron/electron.config.js
```diff
@@ -4,12 +4,12 @@ module.exports = {
   },
   reactAppUrl: {
     development: 'http://localhost:3000/editor',
-    production: 'http://localhost:4001/editor',
+    production: (port = 4001) => `http://localhost:${port}/editor`,
```
cpvalente, line 7: rather than passing the port, what do you think about having some sort of template like,

``` javascript
// in the config file
production: 'http://localhost:%port%/editor'

// and then when we use in main.js
const clientURL = electronConfig.production.replace('%port%', port)
```

cpvalente (commented):
This seems to work great, thank you very much for your effort.
We have some small tasks to cleanup, 

Either way:
1. There are several references to 4001 in the code, but mostly in tests and placeholder data. Nothing that should affect the feature
2. We would still need to enable the UI for changing this
3. what do you think we should do if the new port is taken and we cant start?
4. whad do you think we should do about the docker users?

I will create a new branch to merge this and cleanup the documentation and links so you don't have to do the tedious work 😅

apps/client/src/common/api/apiConstants.ts
```diff
@@ -12,9 +10,11 @@ export const APP_SETTINGS = ['appSettings'];
 export const VIEW_SETTINGS = ['viewSettings'];
 export const RUNTIME = ['runtimeStore'];
 
-export const serverPort = import.meta.env.DEV ? STATIC_PORT : window.location.port;
-export const serverURL = import.meta.env.DEV ? `http://localhost:${serverPort}` : window.location.origin;
-export const websocketUrl = `${window.location.protocol === "https:" ? "wss" : "ws"}://${window.location.hostname}:${serverPort}/ws`;
+export const serverPort = window.location.port;
+export const serverURL = window.location.origin;
+export const websocketUrl = `${window.location.protocol === 'https:' ? 'wss' : 'ws'}://${
+  window.location.hostname
```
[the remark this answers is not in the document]
pSnehanshu: This one was auto-formatted by prettier. Did you make any changes in your `master`?

apps/client/src/features/info/InfoNif.tsx
```diff
@@ -9,7 +9,7 @@ export default function InfoNif() {
   const { data } = useInfo();
 
   const handleClick = (address: string) => {
-    const baseURL = 'http://__IP__:4001';
+    const baseURL = `http://__IP__:${window.location.port}`;
```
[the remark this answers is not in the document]
pSnehanshu: Sure

apps/electron/electron.config.js
pSnehanshu: IMO I think passing the port is better as this templating isn't standardized and can create mess later on. In another place I found something like `__IP__`.

And I think it's faster to generate a string using string laterals than using `replace` as it has to first find and then replace.

I think we should also turn the `development` value to a function too.

pSnehanshu:
Few things are left from my side like adding `serverPort` to validation, preventing port changing in docker environment, and rebooting when port is changed.

apps/client/src/common/api/apiConstants.ts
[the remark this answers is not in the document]
pSnehanshu: Oh i see there are conflicts, I will resolve them.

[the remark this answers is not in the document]
cpvalente: I have moved your target into a feature branch, you can rebase that instead.

apps/electron/electron.config.js
cpvalente: sounds good on the account is spaghetti code 😅  

Speed at startup is not really a concern since the difference would have been in the microsecodns

cpvalente:
> Few things are left from my side like adding `serverPort` to validation, preventing port changing in docker environment, and rebooting when port is changed.

Nice, I think we can leave the auto rebooting for later.
Until the feature is mature I guess it would be better to have the users rebooting themselves. They would also need to change the browsers to target the new port

cpvalente:
Hi @pSnehanshu, would you like to merge this into the feature branch and start a clean branch for the remaining features?

pSnehanshu:
I have added a few final touches. Feel free to merge.

apps/server/src/controllers/ontimeController.validate.ts
```diff
@@ -63,6 +63,7 @@ export const validateSettings = [
   body('operatorKey').isString().isLength({ min: 0, max: 4 }).optional({ nullable: true }),
   body('timeFormat').isString().isIn(['12', '24']),
   body('language').isString(),
+  body('serverPort').isNumeric().optional(),
```
cpvalente, line 66: it seems that there is a builtin validator for this
https://express-validator.github.io/docs/api/validation-chain/#isport

This should also make sure in is a valid range

apps/server/src/controllers/ontimeController.ts
```diff
@@ -207,6 +207,11 @@ export const postSettings = async (req, res) => {
     const editorKey = extractPin(req.body?.editorKey, settings.editorKey);
     const operatorKey = extractPin(req.body?.operatorKey, settings.operatorKey);
 
+    if (isDocker && req.body?.serverPort) {
+      return res.status(403).json({ message: `Can't change port when running inside docker` });
+    }
+    const serverPort = parseInt(req.body?.serverPort ?? settings.serverPort, 10);
```
cpvalente, line 213: I usually prefer using the casting `Number(value)`.
I guess in this case it would have been validated before getting here, but it cant hurt

apps/client/src/common/api/apiConstants.ts
```diff
@@ -15,8 +13,8 @@ export const RUNTIME = ['runtimeStore'];
 const location = window.location;
 const socketProtocol = location.protocol === 'https:' ? 'wss' : 'ws';
 
-export const serverPort = import.meta.env.DEV ? STATIC_PORT : location.port;
-export const serverURL = import.meta.env.DEV ? `http://${location.hostname}:${serverPort}` : location.origin;
+export const serverPort = location.port;
```
cpvalente, line 16: When developing the app we spin react in dev mode (usually port 4001). The previous logic achieved that we can still connect to the server at whatever port but have react coming from another port

HMR is a must for application development, can you think of a way we can keep this?
worst case scenario the previous logic should still work
`export const serverPort = import.meta.env.DEV ? STATIC_PORT : location.port`

the developer would need to manually change the STATIC_PORT in case they want to test on a different port
What do you think?

cpvalente (commented):
Great work, I left some small comments. Mostly nitpicks.
Once those are resolved I am happy to merge and cleanup the remaining 4001s 😄

apps/server/src/controllers/ontimeController.ts
pSnehanshu: But logically port value can't be fractional. Hence `parseInt` makes more sense i think.

apps/client/src/common/api/apiConstants.ts
pSnehanshu: Makes sense, wasn't aware of this. Reverted the change. It now works in DEV mode too.

cpvalente:
Hi @pSnehanshu , thank you for your work here, I will merge this and cleanup the remaining points. This will likely not be included in the next release but the one after.

Could I interest you in helping out with a new task?

I was thinking that this one is mostly node related and should be straightforward for someone with your knowledge https://github.com/cpvalente/ontime/issues/374

and there is this one which we need some advice on how to proceed, a bit more challenging, if you have some experience with google sheets you might just know what to do 
https://github.com/cpvalente/ontime/issues/194
Read also some comments here

pSnehanshu:
Hi @cpvalente,

Thanks for merging this. Tbh, I still don't know what Ontime is, and how it is used in real life. I am not familiar with the concept of rundowns and timers on shows. I approached this issue by treating this project as just another node+react project.

For the issues you mentioned, I will have to understand the whole picture, and get some domain knowledge. I will see what I can do, but I won't be able to promise anything as I have other things as well. Moreover, I am not able to relate to the problem this software is solving.

Thanks again.

cpvalente:
No problem, it was worth a shot from our side.
There are some issues where you would need or benefit from some domain logic. 
The ones highlighted in "Good first issues" generally are pretty agnostic and are technology problems''

I completely understand and am thankful for your help. If you feel like sparing some time to help further do get in touch so I can make sure we don't have multiple people working on the same task

Best

PR #470, opened 2023-07-26
dist-mac: create separate dists for aarch64, x64

### LINKED ISSUE #469: Add support for Apple Silicon / Universal build target for macOS
Hi there! Ontime is a rather strikingly powerful project! I was looking to build something similar for a project I'm working on, and I was quite surprised to see such a well built tool that hits almost every use case I needed! 

I use a M1 MacBook Pro as my laptop. While Ontime appears to run through Rosetta2 reasonably well, having a native version would improve the performance on that platform. I don't personally plan on using Ontime for live usage from my laptop on my next project, but there's a chance I feasibly would, and having maximum runtime and performance on the platform would be a boon to trusting using Ontime on the platform. 

I'm not familiar enough with the stack to know what the level of effort of adding an Apple Silicon build target is, but given that it appears Ontime uses Github Actions for builds, I do at least know that Actions can be used to cross-compile for aarch64 on the x86 Mac runners. Thanks!

belthesar:
Fixes #469.

Current macOS builds for Ontime only build an x64 target, which is less than ideal for Apple Silicon users. Modifies electron dist-mac script to build separate targets for aarch64 and x64 targets, and updates DMG artifactName to include target architecture.

### DIFF
--- a/.github/workflows/build_v2.yml
+++ b/.github/workflows/build_v2.yml
@@ -35,7 +35,9 @@ jobs:
       - name: Release
         uses: softprops/action-gh-release@v1
         with:
-          files: './apps/electron/dist/ontime-macOS.dmg'
+          files: |
+            './apps/electron/dist/ontime-macOS-x64.dmg'
+            './apps/electron/dist/ontime-macOS-arm64.dmg'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/apps/electron/package.json
+++ b/apps/electron/package.json
@@ -22,7 +22,7 @@
     "postinstall": "",
     "dev:electron": "NODE_ENV=development electron .",
     "dist-win": "electron-builder --publish=never  --x64 --win",
-    "dist-mac": "electron-builder --publish=never  --arm64 --mac && electron-builder --publish=never  --x64 --mac",
+    "dist-mac": "electron-builder --publish=never  --mac",
     "dist-linux": "electron-builder --publish=never  --x64 --linux",
     "cleanup": "rm -rf .turbo && rm -rf node_modules && rm -rf dist"
   },
@@ -35,7 +35,13 @@
       "icon": "icon.icns"
     },
     "mac": {
-      "target": "dmg",
+      "target": {
+        "target": "dmg",
+        "arch": [
+          "x64",
+          "arm64"
+        ]
+      },
       "category": "public.app-category.productivity",
       "icon": "icon.icns"
     },
--- a/apps/electron/package.json
+++ b/apps/electron/package.json
@@ -22,7 +22,7 @@
     "postinstall": "",
     "dev:electron": "NODE_ENV=development electron .",
     "dist-win": "electron-builder --publish=never  --x64 --win",
-    "dist-mac": "electron-builder --publish=never  --x64 --mac",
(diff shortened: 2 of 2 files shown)


cpvalente:
Hi @belthesar , thank you for this.
Running the build locally I end up with 3 dmg packages
```
ontime-macOS-arm64.dmg
ontime-macOS-x64.dmg
ontime-macOS.dmg
```

I am not entirely sure on what this last one is.
Modifying the config as below seems to produce two dmgs, although I was unable to test if these are the correct packages

```
  "dist-mac": "electron-builder --publish=never --mac",

  "mac": {
    "target": {
      "target": "default",
      "arch": [
        "x64",
        "arm64"
      ]
    },
    "category": "public.app-category.productivity",
    "icon": "icon.icns"
  },
```

To finish the feature we would also need to upload the artifacts after the build
https://github.com/cpvalente/ontime/blob/master/.github/workflows/build_v2.yml#L38

Would you mind taking a look into these changes?

Thank you again

belthesar:
Heya @cpvalente - 

> Hi @belthesar , thank you for this. Running the build locally I end up with 3 dmg packages
> 
> ```
> ontime-macOS-arm64.dmg
> ontime-macOS-x64.dmg
> ontime-macOS.dmg
> ```

`ontime-macOS.dmg` is likely an old artifact. Did you clean up old artifacts on your machine at `apps/electron/dist` prior to running? I can't replicate this locally. 

To verify architecture, you can do a couple things: 

1. Open the Disk Image, click on the `ontime.app` app bundle, and Get Info on the bundle ( File -> Get Info, or Right Click / Two-Finger-Click / Option Click the bundle, choose Get Info). Under the General section, Kind will either be "Application (Apple silicon)", "Application (Intel)", or "Application (Universal)"

2. Via a terminal, navigate to the location of the App Bundle, then run either of these commands:

```
$ lipo -archs ontime.app/Contents/MacOS/ontime
arm64
```

```
$ file -b ontime.app/Contents/MacOS/ontime
Mach-O 64-bit executable arm64
```

Regardless...

> 
> ```
>   "dist-mac": "electron-builder --publish=never --mac",
> 
>   "mac": {
>     "target": {
>       "target": "dmg",
>       "arch": [
>         "x64",
>         "arm64"
>       ]
>     },
>     "category": "public.app-category.productivity",
>     "icon": "icon.icns"
>   },
> ```

This is the better way to do this. Tested on my machine, and it works quite fine. Forgive my lack of ergonomics with the Electron build tooling, I was working with what I know, haha. 

> To finish the feature we would also need to upload the artifacts after the build https://github.com/cpvalente/ontime/blob/master/.github/workflows/build_v2.yml#L38

I considered updating the build workflow outside of the scope of the pull request, but if you're okay with me stuffing this in the same PR, I don't mind making the change. I'll add commits to this PR for the changes suggested to the build target config in the electron `package.json` and the Github Workflow.

cpvalente:
> 1. Open the Disk Image, click on the `ontime.app` app bundle, and Get Info on the bundle ( File -> Get Info, or Right Click / Two-Finger-Click / Option Click the bundle, choose Get Info). Under the General section, Kind will either be "Application (Apple silicon)", "Application (Intel)", or "Application (Universal)"

Nice. Thank you for the help with the project and for taking the time to explain. I have learned something today

PR #471, opened 2023-07-31
Adding Alias as a Parameter

### LINKED ISSUE #449: Keep original destination on redirect
It has been pointed out in #429 that the URL Aliases feature uses a redirect in its implementation
That means we lose our original destination once the page is entered. If the editor were to change the URL we would need a refresh on the view client to re-resolve the alias.

URL Aliases has two target specifications:
1. Simplify long URLs
2. Allow remote control of unattended clients

This is a bug on the second point

### The task
Fixing the bug would take two steps:

1. Maintain the original alias
2. Periodically check if this has changed and redirect

### Implementation details

I would suggest that **maintaining the original** alias could simply be done by injecting a parameter in the URL (eg: `localhost:4001/timer?from=my-alias` once the [URL is resolved](https://github.com/cpvalente/ontime/blob/master/apps/client/src/AppRouter.tsx#L45)
Alternatives using local storage or react state could also be considered if it improves reactivity on step 2

As for **knowing the alias has changed**, we would either likely need to attach a listener to the [parameter in the store](https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/hooks-query/useAliases.ts)
Once we know **if the alias has changed**, we can check the stored value from point 1 to see if the current destination is affected

bunday:
Closes  #449

### DIFF
--- a/apps/client/src/AppRouter.tsx
+++ b/apps/client/src/AppRouter.tsx
@@ -1,9 +1,7 @@
-import { lazy, Suspense, useEffect } from 'react';
-import { Navigate, Route, Routes, useLocation, useNavigate, useSearchParams } from 'react-router-dom';
-
-import useAliases from './common/hooks-query/useAliases';
+import { lazy, Suspense } from 'react';
+import { Navigate, Route, Routes } from 'react-router-dom';
 import withData from './features/viewers/ViewWrapper';
-import { getAliasRoute } from './common/utils/aliases';
+import withAlias from './features/AliasWrapper';
 
 const Editor = lazy(() => import('./features/editors/ProtectedEditor'));
 const Table = lazy(() => import('./features/table/ProtectedTable'));
@@ -18,14 +16,14 @@ const Public = lazy(() => import('./features/viewers/public/Public'));
 const Lower = lazy(() => import('./features/viewers/lower-thirds/LowerWrapper'));
 const StudioClock = lazy(() => import('./features/viewers/studio/StudioClock'));
 
-const STimer = withData(TimerView);
-const SMinimalTimer = withData(MinimalTimerView);
-const SClock = withData(ClockView);
-const SCountdown = withData(Countdown);
-const SBackstage = withData(Backstage);
-const SPublic = withData(Public);
-const SLowerThird = withData(Lower);
-const SStudio = withData(StudioClock);
+const STimer = withAlias(withData(TimerView));
+const SMinimalTimer = withAlias(withData(MinimalTimerView));
+const SClock = withAlias(withData(ClockView));
+const SCountdown = withAlias(withData(Countdown));
+const SBackstage = withAlias(withData(Backstage));
+const SPublic = withAlias(withData(Public));
+const SLowerThird = withAlias(withData(Lower));
+const SStudio = withAlias(withData(StudioClock));
 
 const EditorFeatureWrapper = lazy(() => import('./features/EditorFeatureWrapper'));
 const RundownPanel = lazy(() => import('./features/rundown/RundownExport'));
@@ -34,21 +32,6 @@ const MessageControl = lazy(() => import('./features/control/message/MessageCont
 const Info = lazy(() => import('./features/info/InfoExport'));
 
 export default function AppRouter() {
-  const { data } = useAliases();
-  const [searchParams] = useSearchParams();
-  const navigate = useNavigate();
-  const location = useLocation();
-
-  // navigate if is alias route
-  useEffect(() => {
-    if (!data) return;
-    const url = getAliasRoute(location, data, searchParams);
-    // navigate to this route if its not empty
-    if (url) {
-      navigate(url);
-    }
-  }, [data, searchParams, navigate, location]);
-
   return (
     <Suspense fallback={null}>
       <Routes>
--- a/apps/client/src/AppRouter.tsx
+++ b/apps/client/src/AppRouter.tsx
@@ -42,10 +42,23 @@ export default function AppRouter() {
     if (!data) return;
 
     for (const d of data) {
-      if (`/${d.alias}` === location.pathname && d.enabled) {
-        const connector = d.pathAndParams.includes('?') ? '&' : '?';
-        navigate(`/${d.pathAndParams}${connector}alias=${d.alias}`);
-        break;
+      if (location.search.indexOf('alias=') > -1) {
+        // if the alias fits the alias on this page, if the URL is diferent, we redirect user to the new URL
+        let aliasOnPage = location.search.substring(location.search.indexOf('alias=') + 6);
+        if (aliasOnPage.indexOf('&') > -1) {
+          // take care of scenario where there is an & after the alias
+          aliasOnPage = aliasOnPage.substring(0, aliasOnPage.indexOf('&'));
+        }
+        let fullLocationPath = location.pathname + location.search;
+        // we need to remove the &alias=ALIAS_NAME or ?alias=ALIAS_NAME at the end of the URL
+        fullLocationPath = fullLocationPath.replaceAll(`alias=${d.alias}`, '');
+        fullLocationPath = fullLocationPath.substring(0, fullLocationPath.length - 1);
+        if (d.alias !== '' && d.enabled && d.alias === aliasOnPage && d.pathAndParams !== fullLocationPath) {
+          //console.log('found a match, same alias but different route so I will redirect to: ', d.pathAndParams);
+          const connector = d.pathAndParams.includes('?') ? '&' : '?';
+          navigate(`${d.pathAndParams}${connector}alias=${d.alias}`);
+          break;
+        }
       }
     }
   }, [data, location, navigate]);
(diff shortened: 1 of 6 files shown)


apps/client/src/AppRouter.tsx
```diff
@@ -42,9 +42,27 @@ export default function AppRouter() {
...
+        if (d.alias !== '' && d.enabled && d.alias === aliasOnPage) {
+          let fullLocationPath = location.pathname + location.search;
+          // we need to remove the &alias=ALIAS_NAME or ?alias=ALIAS_NAME at the end of the URL
+          fullLocationPath = fullLocationPath.replaceAll(`alias=${d.alias}`, '');
+          fullLocationPath = fullLocationPath.substring(0, fullLocationPath.length - 1);
+          // if the location path is different
+          if (d.pathAndParams !== fullLocationPath) {
+            //console.log('found a match, same alias but different route so I will redirect to: ', d.pathAndParams);
+            const connector = d.pathAndParams.includes('?') ? '&' : '?';
+            navigate(`${d.pathAndParams}${connector}alias=${d.alias}`);
```
cpvalente, line 62: Hmm, I wonder if there is a better way of doing this.
The react router documentation is not particular specific on this, but this thread seems to hint to the ability of being able to pass the values programmatically 
https://github.com/remix-run/react-router/issues/7743

apps/client/src/AppRouter.tsx
```diff
@@ -42,9 +42,27 @@ export default function AppRouter() {
...
+          aliasOnPage = aliasOnPage.substring(0, aliasOnPage.indexOf('&'));
+        }
+        // if we have the same alias and its enabled and its not empty
+        if (d.alias !== '' && d.enabled && d.alias === aliasOnPage) {
+          let fullLocationPath = location.pathname + location.search;
+          // we need to remove the &alias=ALIAS_NAME or ?alias=ALIAS_NAME at the end of the URL
+          fullLocationPath = fullLocationPath.replaceAll(`alias=${d.alias}`, '');
+          fullLocationPath = fullLocationPath.substring(0, fullLocationPath.length - 1);
+          // if the location path is different
+          if (d.pathAndParams !== fullLocationPath) {
```
cpvalente, line 59: I am unsure if we can guarantee that the alias is at the end of the URL. 

How about extracting the params using the react router hook (useSearchParams) and comparing the object to one made from the urlAndParams? We would need to build an object from the urlAndParams string and deeply compare both objects. We have a library for the comparison in the project

apps/client/src/common/utils/linkUtils.js
```diff
@@ -22,9 +22,15 @@ export function openLink(url) {
...
+  let appendAlias = '';
+  const connector = location.includes('?') ? '&' : '?';
+  if (alias && alias !== '') {
+    appendAlias = `${connector}alias=${alias}`;
+  }
   // we handle the link manually
   event.preventDefault();
-  openLink(`http://${host}/${location}`);
+  openLink(`http://${host}${location}${appendAlias}`);
 }
```
cpvalente, line 36: I dont think this should be done here. 
Since this is an implementation detail of the alias feature, we would prefer adding this to the code responsible for this (the useEffect you had). I believe that would provide better encapsulation and help future code maintenance and development

cpvalente (commented):
I think the code does roughly the right thing.
There are some small edge cases that I see you have been attempting to circumvent.

Overall I wonder if react-router can help us here 

I think that if we can use the useSearchParams hook and unify the logic in one use effect. That will provide us a better result

I also wonder whether it would be beneficial to extract this a HOC `withAliasNavigate` to encapsulate this logic since it only relates to some of the app

cpvalente (commented):
This seems to work well, good work.

However I think it is not optimal that the alias keyword has to exist from the beginning.

When the user creates the alias, it does not know that it needs to add the param to the URL, that also might be too complex and error prone for some users and defeat the process

Could we instead add this ourselves?

So in the loop where we check for aliases, if an alias is found that matches the location, we inject the parameter.

What do you think?

apps/client/src/AppRouter.tsx
```diff
@@ -34,20 +35,19 @@ const Info = lazy(() => import('./features/info/InfoExport'));
...
   // navigate if is alias route
   useEffect(() => {
     if (!data) return;
-
-    for (const d of data) {
-      if (`/${d.alias}` === location.pathname && d.enabled) {
-        navigate(`/${d.pathAndParams}`);
-        break;
-      }
+    const url = getAliasRoute(location, data, searchParams);
```
cpvalente, line 45: I usually prefer returning something semantic like null.
This is personal I guess, and probably related to my C++ background, up to you

apps/client/src/common/utils/aliases.ts
```diff
@@ -26,4 +29,42 @@ export const validateAlias = (alias: string) => {
...
+        // if the alias fits the alias on this page, but the URL is diferent, we redirect user to the new URL
+        // if we have the same alias and its enabled and its not empty
+        if (d.alias !== '' && d.enabled && d.alias === aliasOnPage) {
+          const newAliasPath = resolvePath(d.pathAndParams);
+          const urlParams = new URLSearchParams(newAliasPath.search);
+          urlParams.set('alias', d.alias);
+          // we confirm either the url parameters does not match or the url path doesnt
+          if (!isEqual(urlParams, searchParams) || newAliasPath.pathname !== location.pathname) {
+            // we then redirect to the alias route, since the view listening to this alias has an outdated URL
+            redirectURL = `${newAliasPath.pathname}?${urlParams.toString()}`;
```
cpvalente, line 55: I think you might not have to call the toString in a template string?
I might be wrong...

apps/client/src/common/utils/aliases.ts
```diff
@@ -26,4 +29,42 @@ export const validateAlias = (alias: string) => {
...
+    }
+  }
+  return redirectURL;
+};
+const generateURLFromAlias = (aliasData: Alias) => {
+  const newAliasPath = resolvePath(aliasData.pathAndParams);
+  const urlParams = new URLSearchParams(newAliasPath.search);
+  urlParams.set('alias', aliasData.alias);
+
+  return `${newAliasPath.pathname}?${urlParams.toString()}`;
```
cpvalente, line 69: same as above, I think the toString is redundant

apps/client/src/common/utils/aliases.ts
```diff
@@ -26,4 +29,42 @@ export const validateAlias = (alias: string) => {
...
+  } else {
+    for (const d of data) {
+      const aliasOnPage = searchParams.get('alias');
+      if (aliasOnPage) {
+        // if the alias fits the alias on this page, but the URL is diferent, we redirect user to the new URL
+        // if we have the same alias and its enabled and its not empty
+        if (d.alias !== '' && d.enabled && d.alias === aliasOnPage) {
+          const newAliasPath = resolvePath(d.pathAndParams);
+          const urlParams = new URLSearchParams(newAliasPath.search);
+          urlParams.set('alias', d.alias);
```
cpvalente, line 51: I think we might be able to pass the whole thing to the constructor and let it deduplicate?
```
    const urlParams = new URLSearchParams({
      ...newAliasPath.search,
      alias: aliasData.alias,
    });
```

apps/client/src/common/utils/aliases.ts
```diff
@@ -26,4 +29,42 @@ export const validateAlias = (alias: string) => {
   }
 
   return valid;
-};
\ No newline at end of file
+};
+
+export const getAliasRoute = (location: Location, data: Alias[], searchParams: URLSearchParams) => {
+  const currentURL = location.pathname.substring(1);
+  let redirectURL = '';
```
cpvalente, line 36: I think the code could be slightly easier to read if we don't keep this around since we have early returns
so the code would look like

```
if (foundAlias) {
   return generateURLFromAlias(foundAlias);
}

for loop {
    if aliasMatch {
        return `${newAliasPath.pathname}?${urlParams.toString()}`;
   }
}

return null

```'

This means less indentations and clauses to keep track of

File: apps/client/src/common/utils/linkUtils.js [hunk 9]
```diff
@@ -22,6 +24,7 @@ export function openLink(url) {
  * Handles opening external links
  * @param event
  * @param location
+ * @param alias (this is optional)
```
cpvalente, line 27: thank you for keeping up the JSDocs

the convention for optional params is to put it inside square brackets, so 
`* @param [alias] `
This will also let the tooling understand that this is optional

cpvalente (commented):
I think the code looks good, I left some small comments for your consideration.

As we discussed before, we could take some extra steps to improve this by encapsulating it in something like a HOC
That would let us contain the logic and only apply it to relevant routes

Either way, I would be happy to merge this if you feel the extra step is unnecessary.
Some unit tests for the functions you have extracted (`generateURLFromAlias` and `getAliasRoute`) could also help us follow through

apps/client/src/common/utils/aliases.ts
bunday: `newAliasPath.search` is a string, so it cant be spread.

apps/client/src/common/utils/aliases.ts
bunday: you're right, I've updated it

cpvalente (approved):
Hi @bunday , really great work here.
I don't have any comments other than to thank you for your work

I will line this up to be part of the next release

PR #473, opened 2023-08-01
feat: event cue

cpvalente:
Adding cue field to events

Also in this PR:
- remove deprecated logic towards finding delays
- moving some code to shared utilities
- encapsulating studio-clock logic. We do not plan on letting any other element use this

Note for reviewer:
The PR is pretty large due some of the recomposition and extracting of logic to shared utilities
The important stuff:
- EventEditor
- Cuesheet
- cueUtils
- rundownUtils

For clarity, this is the spec for the cue 

There are two phases for cue assignment: 
1. The first phase is when the show is being built. Everything is in order. If you move an event up or down, everything gets renumbered.
2. The second phase is when we want to lock the cues. we need to make sure that the cue does not change automatically and only the user can do so

About the ordering:

> In the case of A1, A2, B1, B2, what would ontime do if B2 was moved to after A1?

If B2 moves after A1, and you are in the "editing" phase, B2 becomes A2, and everything else below gets numbered. If you are in the "show" phase, B2 just moves after A1 and stays B2.

> What should the default names be when creating elements

If we are creating elements, after A1 we create  A2, after Preshow 1 Preshow 2, and after NotNamed NotNamed 2

> ie: with the rundown A1, A2, B1, B2
> If B1 is renamed to A3, what happens to b2?

If you are in the editing phase, B2 becomes B1. If you are in the "show" phase, B2 remains the same.


apps/client/src/common/components/schedule/Schedule.tsx
```diff
@@ -34,8 +34,8 @@ export default function Schedule({ className }: ScheduleProps) {
           <ScheduleItem
             key={event.id}
             selected={selectedState}
-            timeStart={event.timeStart}
-            timeEnd={event.timeEnd}
+            timeStart={event.timeStart + (event?.delay || 0)}
+            timeEnd={event.timeEnd + (event?.delay || 0)}
```
asharonbaltazar, line 38: Not such a big deal, but `||` will default to the `0` even if `event?.delay` is `0` itself. If you want to avoid that, use `??`

apps/client/src/common/hooks/useEventAction.ts
```diff
@@ -67,6 +67,10 @@ export const useEventAction = () => {
           after: options?.after,
         };
 
+        if (typeof newEvent?.cue === 'undefined') {
```
asharonbaltazar, line 70: Why not use `if (newEvent?.cue === undefined)`?

apps/client/src/features/event-editor/composite/CountedTextInput.tsx
```diff
@@ -1,23 +1,26 @@
...
+  field: EditorUpdateFields;
   label: string;
   initialValue: string;
-  submitHandler: (field: TitleActions, value: string) => void;
+  submitHandler: (field: EditorUpdateFields, value: string) => void;
 }
 
 export default function CountedTextInput(props: CountedTextInputProps) {
-  const { field, label, initialValue, submitHandler } = props;
+  const { field, label, initialValue, submitHandler, maxLength } = props;
```
asharonbaltazar, line 18: Move the default value (from line 43) here:

```ts
const { field, label, initialValue, submitHandler, maxLength = 50 } = props
```

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -104,7 +104,7 @@ export default function Rundown(props: RundownProps) {
               return;
             }
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- we check for this before
```
asharonbaltazar, line 106: Is there a null assertion happening on the next line?

apps/client/src/features/viewers/studio/studioClock.utils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    formattedEvents.push({
+      id: event.id,
+      time: showEnd ? `${start} - ${end}` : start,
+      title: event.title,
+      isNow: event.id === selectedId,
+      isNext: event.id === nextId,
+      colour: event.colour,
+    });
+  }
```
asharonbaltazar, line 78: Why wouldn't you use `.map` here?

```ts
return rundown.map(() => {
  const start = formatTime(event.timeStart + (event.delay || 0));
  const end = formatTime(event.timeEnd + (event.delay || 0));
  
  return {
    id: event.id,
    time: showEnd ? `${start} - ${end}` : start,
    title: event.title,
    isNow: event.id === selectedId,
    isNext: event.id === nextId,
    colour: event.colour,
  }
})
```

`.map` returns a copy, I believe.

apps/server/src/services/rundown-service/RundownService.ts
```diff
@@ -164,29 +165,29 @@ export async function addEvent(eventData: Partial<OntimeEvent> | Partial<OntimeD
   let newEvent: Partial<OntimeBaseEvent> = {};
   const id = generateId();
 
-  // TODO: filter the parameters that exist in the event, use the parserUtils
+  let insertIndex = 0;
+  if (typeof eventData?.after !== 'undefined') {
```
asharonbaltazar, line 169: Why not `if (eventData?.after === undefined)`?

packages/utils/src/rundown-utils/rundownUtils.ts
```diff
@@ -1,7 +1,100 @@
...
+ * Gets first scheduled event in rundown, if it exists
+ * @param {OntimeRundownEntry[]} rundown
+ * @return {OntimeEvent | null}
+ */
+export function getFirstEvent(rundown: OntimeRundownEntry[]) {
+  for (let i = 0; i < rundown.length; i++) {
+    if (rundown[i].type === SupportedEvent.Event) {
+      return rundown[i] as OntimeEvent;
+    }
+  }
```
asharonbaltazar, line 24: Why not use `.find`?

```ts
return rundown.find((event) => event.type === SupportedEvent.Event)
```

packages/utils/src/rundown-utils/rundownUtils.ts
```diff
@@ -1,7 +1,100 @@
...
+  const index = rundown.findIndex((event) => event.id === currentId);
+  if (index < 0) {
+    return null;
+  }
+
+  for (let i = index + 1; i < rundown.length; i++) {
+    if (rundown[i].type === SupportedEvent.Event) {
+      return rundown[i] as OntimeEvent;
+    }
+  }
```
asharonbaltazar, line 59: Same comment here about using `.find`

packages/utils/src/rundown-utils/rundownUtils.ts
```diff
@@ -1,7 +1,100 @@
...
+  const index = rundown.findIndex((event) => event.id === currentId);
+  if (index < 0) {
+    return null;
+  }
+
+  for (let i = index - 1; i >= 0; i--) {
+    if (rundown[i].type === SupportedEvent.Event) {
+      return rundown[i] as OntimeEvent;
+    }
+  }
```
asharonbaltazar, line 94: Same comment here about using `.find`

asharonbaltazar (commented):
Left a some comments. Lemme know if I'm being too nitpicky

packages/utils/src/rundown-utils/rundownUtils.ts
cpvalente: There is the same answer to all of these that you correctly suggest the use of `.find`. Starting from here:
In this case, I need to start in a specific place and work my way forward to find the next

To use the `.find`, I would need to create an intermediate copy of the array that only contains the events forwards
like `rundown.slice(index - 1).find()`, then do the same for the backwards iterations like `rundown.slice(0, index).find()`

I just thought that the iteration was more explicit and less error prone

Also, since there was an iteration here, I thought having an iteration everywhere would help with predictability and future maintenance

What do you think?

apps/server/src/services/rundown-service/RundownService.ts
cpvalente: generally checking with typeof is safer as it also covers the case where the value is undeclared

eg:
```
typeof doesNotExist === "undefined" // true
doesNotExist === undefined // ReferenceError: doesNotExist is not defined
```

I guess the better approach would be to evaluate this case by case, but I tend to use the safer one unless I am very concerned with the performance of a function and very certain that it will always be declared

apps/client/src/common/components/schedule/Schedule.tsx
cpvalente: great suggestion

apps/server/src/services/rundown-service/RundownService.ts
asharonbaltazar: Seems to be having the same effect in the browser for non-existent object properties. Just pointing it out. Feel free to resolve!

Screenshot 2023-08-16 at 3 08 55 PM

packages/utils/src/rundown-utils/rundownUtils.ts
asharonbaltazar: I see. I didn't realize it did that. Is my first suggestion of using `.find` valid, though?

packages/utils/src/rundown-utils/rundownUtils.ts
asharonbaltazar: Resolving due to https://github.com/cpvalente/ontime/pull/473#discussion_r1296310121

apps/client/src/features/rundown/Rundown.tsx
cpvalente: yeah.... good catch, dont know why this was there
🤷‍♂️ 

I actually went down in trying to get eslint to mark an error on unused directive, but this would have meant adding a new plugin, which I fear would add more weight to maintenance than benefit

Either way, good catch

apps/client/src/features/viewers/studio/studioClock.utils.ts
cpvalente: I guess it doesnt make a difference, but the map might be more semantic. Good suggestion!

apps/server/src/services/rundown-service/RundownService.ts
cpvalente: Thats right, I guess in an object property it would never be undeclared, good suggestion

packages/utils/src/rundown-utils/rundownUtils.ts
cpvalente: Absolutely, we can still use .find in one of the cases. But would it not be nicer to have a similar logic in all cases? ie: when requirements change the code that achieves similar things has a similar structure?

As I can see, the change is only semantic

asharonbaltazar: Fair enough. Consistency is good 💪🏻

packages/utils/src/rundown-utils/rundownUtils.ts
asharonbaltazar: Resolving due to https://github.com/cpvalente/ontime/pull/473#discussion_r1296330210

apps/client/src/features/viewers/studio/studioClock.utils.ts
cpvalente: this code has no counter part to be consistent with, will change to a map as suggested

PR #476, opened 2023-08-02
feat: unlock changing port in UI

cpvalente:


cpvalente:
Hey  @asharonbaltazar , since we talked about this previously, I was wondering if you are available to sanity check?

I still have some cleanup to do in the documentation and some in-app links, but the feature should be operational

asharonbaltazar:
I'll take a look later today. Is that okay?

apps/server/src/controllers/ontimeController.ts
```diff
@@ -207,6 +207,11 @@ export const postSettings = async (req, res) => {
     const editorKey = extractPin(req.body?.editorKey, settings.editorKey);
     const operatorKey = extractPin(req.body?.operatorKey, settings.operatorKey);
 
+    if (isDocker && req.body?.serverPort) {
+      return res.status(403).json({ message: `Can't change port when running inside docker` });
+    }
+    const serverPort = parseInt(req.body?.serverPort ?? settings.serverPort, 10);
```
asharonbaltazar, line 213: What's the `parseInt` for? Isn't `req.body?.serverPort` a number?

asharonbaltazar (approved):
LGTM. Left a comment

apps/server/src/controllers/ontimeController.ts
cpvalente: that is a good question.
From what I understand, the port validation would pass as a number. so port "4001" is a valid port

Unfortunately I couldn't find any tests in the library that document this, but this seems to be the code


```
export default function isPort(str) {
  return isInt(str, { min: 0, max: 65535 });
}
```

asharonbaltazar: So is`serverPort` a number or a string? Maybe we can cut the unnecessary `parseInt` out.

cpvalente: I guess the answer is that we don't know and both are valid.
We keep the parseInt since that is what we have typed inside the domain

Issue #347, opened 2023-04-19
OSC: Set duration for specific event?


tcconway:
Possible for OSC to update/change the duration of a specific eventID?

I'm exploring having other apps control Ontime. For example QLab sending commands via OSC that set a specific timer duration...then running the event.

Might be a good approach for #318 and #283.

cpvalente:
Hi @tcconway, this is a good suggestion.
Few questions from my side
- what would the action look like? would it be `/change/:timerid/duration/:durationvalue`?
- why is this preferable to the HTTP API which gives full control of the patch?
- how would ontime decide how to modify the timer, would a duration always be a change on the end state?

And a semi-related question: should we, as part of v2, remove the input field for end time in the rundown editor (keep it in the bottom sheet)

tcconway:
I'd imagine this would be exactly as you mentioned. This is preferable over the HTTP API as apps like QLab and Millumin seem to prefer osc over an http call. Side note: I'm a fan of `/app/action/id/value` format, but that's a bigger discussion LOL.

I would definitely keep the end time in the rundown editor. I find that I type in there all the time...It's a great confirmation for the operator... and you have plenty of room in the gui there anyway!

arymoraes:
Hey! You can assign this to me and I can tackle this

cpvalente:
Nice! Obrigado

cpvalente:
HI @arymoraes , could we help you get started here?

arymoraes:
> HI @arymoraes , could we help you get started here?

Hey, sorry about taking long, last couple of weeks have been very busy with work + some other bureocracy I have to do.
So the thing is I'm not extremely familiar with OSC API. I did go through the code and I do have a good idea on what and how to do this task, but I just don't know how to test it (test it locally, not unit/integration/e2e testing)

I do get this on my console when I run the dev mode
`OSC integration client connected to 127.0.0.1:9999`

but theres nothing at localhost:9999, is there a UI client for the OSC integration or is it all just on server on `8888`?

cpvalente:
No problem, I understand OSC is very industry-specific and tricky for someone to jump in.

I normally use `[OSC Tester](https://github.com/ETCLabs/OSCTester)` to send messages and [Protokol](https://hexler.net/protokol) to inspect the output of ontime.
There are better solutions, but these are free and lightweight

The exact input / output configuration are in the integration options. Puzzle icon in the left menu bar
[...]

PR #478, opened 2023-08-05
feat: set duration for specific event

cpvalente:


apps/server/src/classes/event-loader/EventLoader.ts
```diff
@@ -258,6 +259,21 @@ export class EventLoader {
...
+  static updateEventDuration(eventId: string, newDuration: number) {
+    const event = EventLoader.getEventWithId(eventId);
+    if (event) {
+      const newEventTimeEnd = event.timeStart + newDuration;
+      editEvent({ id: eventId, duration: newDuration, timeEnd: newEventTimeEnd });
+    } else {
+      throw new Error(`Event with ID ${eventId} not found`);
+    }
+  }
+
```
cpvalente, line 276: I think the right place for this is in the `rundownService`

apps/server/src/controllers/integrationController.ts
```diff
@@ -186,6 +186,15 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
       return { topic: 'timer', payload: timer };
     }
 
+    case 'update-duration': {
+      try {
+        const parsedPayload = JSON.parse(payload.toString());
+        PlaybackService.updateEventDuration(parsedPayload.id, parsedPayload.duration);
```
cpvalente, line 192: what would the payload look like here?
`{id: string, duration: number)`?

We should probably attempt parsing and validating before it reaches the application domain
Can you show me how the OSC message is formed?

apps/server/src/services/PlaybackService.ts
```diff
@@ -224,4 +224,14 @@ export class PlaybackService {
...
   }
+
+  /**
+   * Updates the duration of the event with the given id
+   * @param {string} eventId
+   * @param {number} newDuration
+   */
+  static updateEventDuration(eventId: string, newDuration: number) {
+    EventLoader.updateEventDuration(eventId, newDuration);
+    logger.info('PLAYBACK', `Updated duration of event with ID ${eventId} to ${newDuration}`);
```
cpvalente, line 235: The duration value is in milliseconds and would make little sense in most cases. I would suggest either omitting it our running through a format function like `formatDisplay`

Also, we have a LogOrigin enum that would take care of the `'PLAYBACK'` constant string

cpvalente (commented):
Hey, good stuff!
I am sorry I didn't realise that the link was fixed and took so long to take a look, thank you for your work

apps/server/src/classes/event-loader/EventLoader.ts
```diff
@@ -3,6 +3,7 @@ import { Loaded, OntimeEvent, TitleBlock } from 'ontime-types';
 import { DataProvider } from '../data-provider/DataProvider.js';
 import { getRollTimers } from '../../services/rollUtils.js';
 import { eventStore } from '../../stores/EventStore.js';
+import { editEvent } from '../../services/RundownService.js';
```
cpvalente, line 6: looks there is an issue with this import.
You might also want to merge master to help with the changes since this file has moved

apps/server/src/controllers/integrationController.ts
arymoraes: that's a fair point, this is the OSC message
`{ eventId: '5a6e1', newDuration: 600 }`

I have a python script that allows me to send an OSC message if you want to take a look

```
from pythonosc import udp_client
import json

client = udp_client.SimpleUDPClient(
    "127.0.0.1", 8888)  # Adjust with your IP and Port

# Create the JSON object
data = {
    "id": "5a6e1",
    "duration": 600,
}

# Stringify the JSON object
json_string = json.dumps(data)

# You can put any value you want here
client.send_message("/ontime/update-duration", json_string)
# client.send_message("/ontime/get-timer", "5a6e1")
```

What should be the behaviour if its wrongly formatted/validate? just throw an error?

arymoraes:
> Hey, good stuff! I am sorry I didn't realise that the link was fixed and took so long to take a look, thank you for your work

I made the changes you suggested, only two things I need now are to know the behaviour in case we try to validate/parse and it doesn't work, and also need to change it to adapt to the new `editEvent`, it has changed 3 weeks ago right?

apps/server/src/controllers/integrationController.ts
cpvalente: I think throwing an error is sane. If the data is incorrect, somehow it is a no-op for us

The adapters should handle it
https://github.com/cpvalente/ontime/blob/master/apps/server/src/adapters/OscAdapter.ts#L40
https://github.com/cpvalente/ontime/blob/master/apps/server/src/adapters/WebsocketAdapter.ts#L126

making an object can be tricky in most software (I think)
Do you think it would be possible (and better) to do something like `/ontime/update-duration/<eventId> value`?

Similar to what the user described in the issue
https://github.com/cpvalente/ontime/issues/347#issuecomment-1514883828

cpvalente:
> > Hey, good stuff! I am sorry I didn't realise that the link was fixed and took so long to take a look, thank you for your work
> 
> I made the changes you suggested, only two things I need now are to know the behaviour in case we try to validate/parse and it doesn't work, and also need to change it to adapt to the new `editEvent`, it has changed 3 weeks ago right?

Hi, I tried answering above, hopefully this provides enough context

In short:
- throw on invalid data
- editEvent has changed, but the API has remained the same (???)

Thank you again for your work

cpvalente:
Hi @arymoraes , any news here?

arymoraes:
> Hi @arymoraes , any news here?

Hey sorry for the lack of updates, I just moved a few days ago and it's been very chaotic. I did set a reminder and will hopefully finish this tomorrow

arymoraes:
> > Hi @arymoraes , any news here?
> 
> Hey sorry for the lack of updates, I just moved a few days ago and it's been very chaotic. I did set a reminder and will hopefully finish this tomorrow

Ok I got it working. 

So a couple things:

I had the change the logic for `dispatchFromAdapter` to be able to accept params, `type` wouldn't accept anything after the `pathname/` as it was split on `OscAdapter.ts`, and anything after `path` was ignored. I made an implementation to accept one param but there might be better ways to implement it. I also updated `WebsocketAdapter` to adapt to the change, but I havent added the logic to accept params on this handler.

I have it working, it successfully updates my `db.json` with the duration I pass in the OSC message, and also merged the master branch.

Let me know what you think of the changes, I can modify stuff if needed, I'll have more time from this week on

cpvalente:
Hi @arymoraes, thank you for your work here. From a first look, it seems well achieved and sane.

Looking at the changes, I wonder on how much further we would need to go to abstract the new endpoint syntax to allow changing any property in a given event id, like

`ontime/change/<event-id>/duration` `12`
`ontime/change/<event-id>/title` `newtitle`

Any ideas?

Would you mind if I merge this into a feature branch to do some testing on my side? I will likely line this up for the next release

arymoraes:
sure go on ahead and merge on a feature branch

that's a very good point to be fair,

`ontime/change/:event_id?duration={...}&title={...}` is the only different one I could think about it, thats a more web approach

but I'd prefer `ontime/change/<event-id>/duration` and `12` as payload, we could have just `change` on the switch case, then the next index after the next `/` would be the event id, and the next one would be the property name, if we go that way we should probably abstract the logic and have another function that you'll pass `event-id` and `{propertyName}` and it'll deal with the logic

do we plan to make it just a `change` endpoint that would allow us to change any property we wanted?

cpvalente:
Thank you @arymoraes 

I agree with you that `ontime/change/<event-id>/duration` `12` is preferable. The user can always call this multiple times for all the fields to change. Large changes are probably out of scope in this protocols anyway.

It wasnt the initial plan. But looking at your code, it seems to me that the cost of abstraction might not be very high, and this would be an excellent feature for the app

I think the best is to merge this as is as an MVP and immediately start working on an abstraction for the next release.
Would you be available to help with that?

arymoraes:
> Thank you @arymoraes
> 
> I agree with you that `ontime/change/<event-id>/duration` `12` is preferable. The user can always call this multiple times for all the fields to change. Large changes are probably out of scope in this protocols anyway.
> 
> It wasnt the initial plan. But looking at your code, it seems to me that the cost of abstraction might not be very high, and this would be an excellent feature for the app
> 
> I think the best is to merge this as is as an MVP and immediately start working on an abstraction for the next release. Would you be available to help with that?

yeah for sure, I'll have way more time now to implement stuff as things are way less chaotic than a month ago so I'd be glad to help

cpvalente:
That is great to hear, I am working on a biggish release.
Do you have the information you need to continue working on the feature?

arymoraes:
> That is great to hear, I am working on a biggish release. Do you have the information you need to continue working on the feature?
> 
> Maybe we can leave this branch up and you can target the new PR at this branch? could make it easier to merge

think that sounds good, and yeah I believe I have enough info, I can let you know if theres something I didn't understand, I'll keep working on this branch then

arymoraes:
ok so this is what I came up with:

`const [, address, path, ...params] = msg[0].split('/');`

so params will be a string array with one param after each `/`
and we'd have something like

```
    // ontime/change/{eventID}/{propertyName}
    case 'change': {
```

```
        const eventID = params[0];
        const propertyName = params[1] as keyof OntimeBaseEvent;
```

(think theres a better way to handle propertyName instead of casting but I havent given it lot of thought yet)

and then we'd use something like
`return PlaybackService.updateEvent(eventID, propertyName, payload);`

my only question is we'd have to deal with some special cases, like for example on `duration` we'd also have to increase `timeEnd` based on `timeStart + newDuration`, or if we changed `timeStart` or `timeEnd` we'd have to also change duration, I could have a function to handle those cases but then we'd have to remember to add all these cases there, is this a good approach?

should we only allow to change some properties?

cpvalente:
> 

Hi. this absolutely looks like the right direction. Few thoughts from my side

> ontime/change/{eventID}/{propertyName}

perfect

should we only 

> (think theres a better way to handle propertyName instead of casting but I havent given it lot of thought yet)
> allow to change some properties?

yes! the nice thing about the code you are writing is that it is easily extendable. However the edit function should be payload specific.
ie: if the user inputs `ontime/change/{eventID}/duration` we would need to take a different approach from  `ontime/change/{eventID}/title`

I didn't look too much into it, and there are likely better solutions. My instinct would be to create a sanitation function which would return a Partial object we can send to the editEvent function, something like

arymoraes:
think that makes sense, I'll do that then, thanks!

arymoraes:
I pushed what I wrote, it works for any property and properly parses the payload properly, though we probably should make some properties not changeable (like id and revision) - so I could probably exclude them from the `payloadMap`

only thing left unhandled is `duration, startTime and endTime` changes, maybe having a handler in `RundownService` `updateEvent` that will check if its any of these, then have a helper that also updates the other ones?

For example event1 with duration: 1000, start:0, end:1000

if I changed the duration to 5000, endTime should probably be updated to 5000, or no?

cpvalente:
>  though we probably should make some properties not changeable (like id and revision)

I agree. perhaps the best way is to create an explicit white list, so that in future, if we add something. we are forced to come here and write logic about how to edit and validate 


> only thing left unhandled is duration, startTime and endTime changes, maybe having a handler in RundownService updateEvent that will check if its any of these, then have a helper that also updates the other ones?

was duration now how we started here? I would say, lets just do duration for now and deal with the rest once this is more mature.

This means our whitelist should be like

```
  title: anyString
  subtitle: anyString
  presenter: anyString
  note: anyString
  cue: anyString

  duration: anyNumber
  
  isPublic: anyBoolean
  skip: anyBoolean
  
  colour: anyString
  user0: anyString
  user1: anyString
  user2: anyString
  user3: anyString
  user4: anyString
  user5: anyString
  user6: anyString
  user7: anyString
  user8: anyString
  user9: anyString
```

Does this make sense? Thank you for the great work here

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +237,51 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
...
+
+        const payloadMap = Object.fromEntries(
+          Object.entries(event).map(([key, value]) => {
+            return [key, typeof value];
+          }),
+        );
+
+        if (!payloadMap.hasOwnProperty(propertyName)) {
+          throw new Error(`Cannot update unknown event property ${propertyName}`);
+        }
```
cpvalente, line 258: thinking out loud, could we make a type guard here `isKeyOfOntimeEvent`, this would give us the validation and the type at once and replace all this logic

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +237,51 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
...
+          switch (payloadMap[propertyName]) {
+            case 'string': {
+              return String(payload);
+            }
+            case 'number': {
+              return Number(payload);
+            }
+            case 'boolean': {
+              return Boolean(payload);
+            }
```
cpvalente, line 270: good work

arymoraes:
done I added a whitelist/accepted map, and I'm handling the special case for duration

cpvalente:
Hi @arymoraes thank you for your work so far
Give me a heads up when you would like me to review

arymoraes:
> Hi @arymoraes thank you for your work so far Give me a heads up when you would like me to review

go ahead and review it whenever you have time

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +237,74 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
...
+        if (params.length < 2) {
+          throw new Error('Invalid parameters');
+        }
+
+        const eventID = params[0];
+        const propertyName = params[1] as keyof OntimeEvent;
+
+        const isKeyOfOntimeEvent = (key: string): key is keyof OntimeEvent => {
+          return key in event;
+        };
```
cpvalente, line 252: it would be nice to abstract this utility to the guards file in the types package. or at least, remove it from the closure scope

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +237,74 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
...
+          user1: 'string',
+          user2: 'string',
+          user3: 'string',
+          user4: 'string',
+          user5: 'string',
+          user6: 'string',
+          user7: 'string',
+          user8: 'string',
+          user9: 'string',
+        };
```
cpvalente, line 281: we would likely prefer not having this being declared in the closure scope. maybe a config file? like `integrationController.config.ts`? Better ideas welcome

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +237,74 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
...
+              return Number(payload);
+            }
+            case 'boolean': {
+              return Boolean(payload);
+            }
+            default: {
+              throw new Error('Unhandled payload type');
+            }
+          }
+        };
```
cpvalente, line 298: extracting both the config above and this would allows us to easily unit test, which would be nice

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +237,74 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
...
+
+        const parsedPayload = parsePayload(payload, propertyName);
+
+        return PlaybackService.updateEvent(eventID, propertyName, parsedPayload);
```
cpvalente, line 302: it is not within scope of the `playbackService` to change events

We have a function `editEvent` in `RundownService`, this function makes sure the change is made in all necessary places.
We would need to construct a patch object to pass to `editEvent`

This means that, in the case of duration, we need the logic to make make the necessary time changes to happen before we call `editEvent`. `editEvent` should not contain business logic

maybe one more function for the extracted utilities :D

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +237,74 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
...
+          return key in event;
+        };
+
+        if (!isKeyOfOntimeEvent(propertyName)) {
+          throw new Error(`Cannot update unknown event property ${propertyName}`);
+        }
+
+        const acceptedPropertyPayloadMap = {
+          title: 'string',
+          subtitle: 'string',
```
cpvalente, line 260: what if the value here is a callback to validate as the correct type, like 


```ts
function maybeString(value: unknown) {
  if (value == null) {
      throw new Error("Value type invalid")
  }
  return String(value)
}


function maybeBoolean(value: unknown) {
  if (value == null) {
      throw new Error("Value type invalid")
  }
  return Boolean(value);
}


// should probably handling force ints too. all timers in ontime need to be ints
function maybeNumber(value: unknown) {
  if (value == null) {
      throw new Error("Value type invalid")
  }
  return Number(value);
}

const whitelistedPayload = {
  title: maybeString,
  subtitle: maybeString,
  presenter: maybeString,
  note: maybeString,
  cue: maybeString,

  duration: maybeNumber,

  isPublic: maybeBoolean,
  skip: maybeBoolean,

  colour: maybeString,
  user0: maybeString,
  user1: maybeString,
  user2: maybeString,
  user3: maybeString,
  user4: maybeString,
  user5: maybeString,
  user6: maybeString,
  user7: maybeString,
  user8: maybeString,
  user9: maybeString,
};
```

cpvalente (commented):
Hey @arymoraes , great work here.
I left some comments, mostly to do with code organisation. Hope that is ok

Let me know if you need any help

arymoraes:
> Hey @arymoraes , great work here. I left some comments, mostly to do with code organisation. Hope that is ok
> 
> Let me know if you need any help

hey sure I'll get this done today/tomorrow

apps/server/src/controllers/integrationController.ts
arymoraes: problem about sending it to the types package, is how would I import `event` from there? I'm assuming they're different packages so I wouldnt be able to import like "../../../../..etc" or would I?

apps/server/src/controllers/integrationController.ts
arymoraes: maybe have it all this inside something like `integrationController.config.ts` then? then I can move a big chunk of the parsing and validation logic to it, maybe also have `isKeyOfOntimeEvent` there?

apps/server/src/controllers/integrationController.ts
cpvalente: there is a guards file in the packages, you can create a function there with your guard.
after that you will need to export the function from the index in the types package

From here, you just import the types like any other package, see an example here
https://github.com/cpvalente/ontime/blob/e2aabe36468057b9ee4febc7b389a64a0d1b2f44/apps/server/src/modules/loadDb.ts#L4

apps/server/src/controllers/integrationController.ts
cpvalente: I think a config file makes sense and it will be helpful to encapsulate business logic related to the integration (in my example, parse and whitelisted)

maybeNumber, maybeBoolean and maybeString are generic utilities that can go in the utils package. They should probably be called coerceBoolean coerceString and CoerceNumber instead

the isKeyOfOntimeEvent is more generic to the domain logic and makes more sense in the types package

arymoraes: thanks, it really makes sense now, thanks for the explanation on the architecture, it is really helpful!

apps/server/src/controllers/integrationController.ts
arymoraes: Yeah that makes sense, the only thing I am confused is 

```
export const isKeyOfOntimeEvent = (key: string): key is keyof OntimeEvent => {
  return key in event;
};
```

needs event which is imported from `/models/eventsDefinition.js`

I could do this, but it just feels very wrong
`import { event } from '../../../../apps/server/src/models/eventsDefinition.js';`

I need the `event` object to be able to do `isKeyOfOntimeEvent`as it is right now

apps/server/src/services/rundown-service/RundownService.ts
```diff
@@ -301,3 +301,30 @@ export async function swapEvents(from: string, to: string) {
...
+) {
+  const event = EventLoader.getEventWithId(eventId);
+
+  if (event) {
+    const propertiesToUpdate = { [propertyName]: newValue };
+
+    // Handles the special case for duration
+    if (propertyName === 'duration') {
+      propertiesToUpdate.timeEnd = event.timeStart + (newValue as number);
+    }
```
cpvalente, line 324: I am sorry I didnt see this earlier. The duration given by the user is in seconds, while the internal times are all in milliseconds. we just need to make sure that value gets x 1000 somewhere

cpvalente (commented):
Hey @arymoraes! Just a few tweaks to go, do you have some availability?
When you are ready I will merge this into a test branch before releasing

apps/server/src/controllers/integrationController.ts
arymoraes: what I previously had is:

`integrationController` calls `PlaybackService` `updateEvent` method, which called `updateEvent` in `RundownService`, that called `editEvent` inside the own module, which to be fair is kinda confusing

what I can do is

have `updateEvent` to be inside another module (config?) and call things in this order:

`integrationController` -> `updateEvent` -> `editEvent`.

`updateEvent` would deal with the business logic for changing things on the update

arymoraes:
> Hey @arymoraes! Just a few tweaks to go, do you have some availability? When you are ready I will merge this into a test branch before releasing

I made the changes, I am just not 100% sure of where to keep the `updateEvent` that I shoved inside the config file, but other than that the tweaks should have been finished, and after testing it, it works

apps/server/src/controllers/integrationController.ts
cpvalente: I agree, the playback service shouldnt have a updateEvent method since it is not in its scope to do so
The management of events is in scope of the RundownService

cpvalente:
Great work @arymoraes , just let me know once you lined up the last question and I will merge this into a test branch.
Good stuff

apps/server/src/controllers/integrationController.ts
arymoraes: I had moved`updateEvent` to `integrationController.config.ts`, should I move it to `RundownService`? problem is it will have both `editEvent` and `updateEvent` on it

arymoraes:
> Great work @arymoraes , just let me know once you lined up the last question and I will merge this into a test branch. Good stuff

sounds good, imo we could keep it the config? I also adjusted the duration for *1000

cpvalente:
@arymoraes , I think moving things around at this point is unproductive.
I will push it into a test branch and check with some controllers

If I find that we can reduce coupling by moving this, would it be OK if I do so myself?

arymoraes:
> @arymoraes , I think moving things around at this point is unproductive. I will push it into a test branch and check with some controllers
> 
> If I find that we can reduce coupling by moving this, would it be OK if I do so myself?

yeah sounds good to me

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +239,32 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
...
     }
 
+    // ontime/change/{eventID}/{propertyName}
+    case 'change': {
+      try {
+        if (params.length < 2) {
+          throw new Error('Invalid parameters');
+        }
+
+        if (!payload || typeof payload !== 'string') {
```
alex-Arc, line 249: this blocks every thing that is not a string, before getting to type coercion
allowed types in osc should at least be: int, float, string and boolean
and should properly test `payload === undefined` so that empty strings and FALSE is allowed

apps/server/src/adapters/OscAdapter.ts
```diff
@@ -15,13 +15,13 @@ export class OscServer implements IAdapter {
...
-      // message should look like /ontime/{path} {args} where
+      // message should look like /ontime/{path}/{params?} {args} where
       // ontime: fixed message for app
       // path: command to be called
       // args: extra data, only used on some API entries (delay, goto)
 
       // split message
-      const [, address, path] = msg[0].split('/');
+      const [, address, path, ...params] = msg[0].split('/');
       const args = msg[1];
```
alex-Arc, line 25: if there is no args this throws `Error: uncaught exception TypeError: Cannot read properties of undefined (reading 'elements')`

apps/server/src/adapters/OscAdapter.ts
```diff
@@ -37,7 +37,14 @@ export class OscServer implements IAdapter {
...
+          path,
+          {
+            payload: args,
+            params,
+          },
+          'osc',
+        );
         if (reply) {
           const { topic, payload } = reply;
           this.osc.emit(topic, payload);
```
alex-Arc, line 50: have you tested this to work?

apps/server/src/controllers/integrationController.config.ts
```diff
@@ -0,0 +1,74 @@
...
+  propertyName: keyof OntimeEvent,
+  newValue: OntimeEvent[typeof propertyName],
+) {
+  const event = EventLoader.getEventWithId(eventId);
+
+  if (event) {
+    let valueToUpdate = newValue;
+
+    // Duration needs to be converted to milliseconds
+    if (propertyName === 'duration' && typeof valueToUpdate === 'number') {
```
alex-Arc, line 57: we should be guaranteed a correct type at this point right?
and if not we should make it so, otherwise we wont get the s to ms conversion..?

apps/server/src/controllers/integrationController.config.ts
```diff
@@ -0,0 +1,74 @@
...
+      valueToUpdate = valueToUpdate * 1000;
+    }
+
+    const propertiesToUpdate = { [propertyName]: valueToUpdate };
+
+    // Handles the special case for duration
+    if (propertyName === 'duration') {
```
alex-Arc, line 64: if we don't need the type check on line 57
these could be combined

alex-Arc (changes requested):
Hey @arymoraes 
I'm testing out this change and have some questions

apps/server/src/controllers/integrationController.ts
arymoraes: hmmmm, that one is my bad, I thought OSC payloads were always strings for some reason, do we need to check for other types here? we already coerce the payload, and check for `null` there

apps/server/src/adapters/OscAdapter.ts
arymoraes: it works in general, I didn't know we needed to reply to `change`.

Would `{ topic: propertyName, payload: newValue };` be a proper reply then?

apps/server/src/controllers/integrationController.config.ts
arymoraes: yes, we parse it on `integrationController`

apps/server/src/controllers/integrationController.config.ts
arymoraes: yep, you're right

arymoraes: actually, we need the if statements to be different, line 57 multiplies the value by 1000

then on 61 we create the object to update,

then on 64 we add the updated `endTime` there.

We could combine them but I'd have to create another variable and it feels it would make the flow a bit harder to follow? I could be wrong though.

You're definitely right we don't need the type check though

apps/server/src/adapters/OscAdapter.ts
arymoraes: I thought `args` should ever exist? I was not even able to send a OSC message without a payload

apps/server/src/adapters/OscAdapter.ts
alex-Arc: sorry this was my mistake, we don't expect a reply from change

alex-Arc:
Hey @arymoraes 
If you are not currently working on this, can pull it to feat. branch and do these last small fixes?

arymoraes:
> Hey @arymoraes If you are not currently working on this, can pull it to feat. branch and do these last small fixes?

sounds good

Issue #477, opened 2023-08-03
Style: scale gracefully in large screens


cpvalente:
## The issue
Testing out ontime on screens larger than HD, we see that there is a large gap left on the right side.

![localhost_3000_editor (1)](https://github.com/cpvalente/ontime/assets/34649812/53cd8983-4864-42c2-8ac6-3687a4634a80)

## The task
The task steps are a bit up to experimentation, ideally by someone with a wide screen and some CSS patience.
The task only concerns with the main app route `/editor` (in the screenshot)

While we might not be able to scale indefinitely there might be some things we can do:

1. Increate root font size to `17px`: this will allow the text to take more space, and potentially improve experience in high-density screens
2. Allow rundown column to expand: Currently the rundown is capped at `44.5rem`, a simple change to allow expansion could go a long way. I have tested with `minmax(44.5rem, 1000px)` and it seemed ok, although a real value might take some experimentation
3. Allow central column to grow: the central column is capped at `26rem`, I have not experimented but assume this could grow slightly without issues. It would take some experimentation
4. Allow rightmost column to grow: similar as the rest, this one is capped at `800px`, It would be interesting to see if we could grow this

bunday:
Will be looking into this next

cpvalente:
This was an issue related to v2.
V3 could likely have similar issues but they should be detailed again

PR #491, opened 2023-08-21
Allow Content render properly on Large Screen 

bunday:


pnpm-lock.yaml
cpvalente, line 1: looks like you might be using another version of pnpm?
in general: this feature should not contain changes to the pnpm-lock

if you want to create a separate PR for upgrading pnpm it is cool with me. I have done it in a few repos and it was painless

apps/client/src/features/editors/Editor.module.scss
```diff
@@ -67,6 +67,14 @@ $playback-width: 26rem;
   }
 }
```
cpvalente, line 69: See the other points from the issue description #477 


- Increate root font size to 17px: this will allow the text to take more space, and potentially improve experience in high-density screens
- Allow rundown column to expand: Currently the rundown is capped at `44.5rem`, a simple change to allow expansion could go a long way. I have tested with `minmax(44.5rem, 1000px)` and it seemed ok, although a real value might take some experimentation
- Allow central column to grow: the central column is capped at `26rem`, I have not experimented but assume this could grow slightly without issues. It would take some experimentation
- Allow rightmost column to grow: similar as the rest, this one is capped at `800px`, It would be interesting to see if we could grow this

PR #493, opened 2023-08-21
refactor: pass events instead of derived data to titles

cpvalente:
This changes the `eventLoader` to no longer pass derived data to the socket, but the whole object
It majorly simplifies the code but will break integrations


cpvalente:
@asharonbaltazar we are planning this for a future release that we are coordinating with some integrations.
It is mostly aiming to improve DX and simplify the logic (less code, less bugs amirite?). Would you mind taking a look?

asharonbaltazar (approved):
Haven't delved too deeply into the logic, but nothing stands out. Approving.

PR #494, opened 2023-08-22
fix: apply delay

cpvalente:
Fixes issues with the delay flow and covers feature with unit and e2e tests
- add event function was unable to resolve eventId due to typo
- delay data was shown in public screen
- applyDelay was not propagated to cachedRundown


apps/server/src/services/delayUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+    return updatedRundown;
+  }
+
+  for (let i = delayIndex + 1; i < rundown.length; i++) {
+    const currentEvent = updatedRundown[i];
+
+    if (currentEvent.type === SupportedEvent.Block) {
+      break;
+    } else if (currentEvent.type === SupportedEvent.Event) {
+      const event = currentEvent as OntimeEvent;
```
asharonbaltazar, line 27: What's the point of this reassignment? If you're trying to create a copy, then you should spread `{ ...oldObj }`

apps/server/src/services/delayUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+      event.timeStart = Math.max(0, event.timeStart + delayValue);
+      event.timeEnd = Math.max(event.duration, event.timeEnd + delayValue);
+      if (event.delay) {
```
asharonbaltazar, line 30: Is `event.delay` a boolean? A number? Something else that can be falsey? Maybe you should add an explicit check

asharonbaltazar: It's a number based on the next line, but would argue the check should still be explicit.

asharonbaltazar (approved):
Looks good. Left a few nits

apps/server/src/services/delayUtils.ts
cpvalente: thats correct. The reason for reassigning was to make typescript happy since it is handy having good typing in the following lines. 
It was a hack that stayed behind, I guess the only way is to have a function that asserts a type with the `is` keyword?

asharonbaltazar: I'm having déjà vu right now — I believe I commented something similar somewhere else. It doesn't work without `as`? I believe the condition should be intelligent enough to narrow down the type.

asharonbaltazar:
I love the new changes! Functions for the win!

PR #495, opened 2023-08-23
css style on large viewport fixed

bunday:


cpvalente:
Hi @bunday , I am not quite sure this is the right direction.

Changing the elements size can have negative consequences on the visual continuity, I believe a good solution should be less opinionated, even if not as effective.
CSS is hard 😓 

If it is OK with you, I will leave this branch around to start from and give it a go when time allows

cpvalente:
hey! I was hoping to pick up some of the improvements we talked about. 
Do you have availability for a follow up? It would be great to extract the fix to stop the rundown panel from shrinking if the rundown array is empty

Issue #448, opened 2023-07-11
New Event-Type: Count to time from now


iqtreiber:
I got a new request showing a countdown from now to defined end-time.
This would be very handy to get a new type of event-timer "count-to-time" which calculate the rest until end-time and show as countdown in the /speaker and /backstage and /public views.

The main use case for this event-type is when just a countdown like "the show starts in mm:ss" upfront the speaker-events. It would be helpful the entry for this event-type is only the end-time (target-time) whereto the counter should be run and calculate the rest.

many thanks!

cpvalente:
Hi @iqtreiber, quick clarification:

is this end-time, the end time of the very last event. or the end time of the running event?

iqtreiber:
Hi,
Its the very first event-row - the starting event which count down to the end-time of first event-row.
The view should show a count down to the time of start the event.

Example:
#1: count down until 15:00 from now (e.g. current time 13:30) shows a countdown 01:30:00
Title - the show is starting soon
#2: countdown 15:00 to 15:15 (15 min duration) shows countdown 00:15:00
Title - welcome

cpvalente:
Hi @iqtreiber, I feel that the new timer type has this covered.
Do you think there is still a case to prefer a separate event type to achieve this?
Otherwise, could we close this ticket?

iqtreiber:
Yes, this issue is resolved and implemented in the last Build.
Thanks for all the effort

PR #500, opened 2023-08-26
feat: count to end

cpvalente:
add a feature that allows for counting to the scheduled end of an event, regardless of when it started

This attempts a more semantic way of approaching the issue described in #448


cpvalente:
Hey @asharonbaltazar , would you mind taking a look?
Not so much here code wise, but good to sanity check

asharonbaltazar:
Will take a look tomorrow

asharonbaltazar (approved):
I feel like `getExpectedFinish` and `getCurrent` can be cleaned up using closures (8 args [some repetitive] feels a bit excessive), but overall looks good 👍🏻

PR #509, opened 2023-09-02
style: localise labels to user device

### LINKED ISSUE #492: Shortcuts should match OS
Shortcuts, like the one at the bottom of `/editor`, should correspond to the OS the user is on. It'd would make the UI less confusing for users on different operating systems.

The `QuickAddBlock` displays `Alt + E` to add a new event. On MacOS, it should be `option + E`.

cpvalente:
Closes #492

### DIFF
--- a/apps/client/src/common/utils/deviceUtils.ts
+++ b/apps/client/src/common/utils/deviceUtils.ts
@@ -1,0 +1,6 @@
+export function isMacOS() {
+  const userAgent = navigator.userAgent.toLowerCase();
+  return userAgent.includes('macintosh') || userAgent.includes('mac os');
+}
+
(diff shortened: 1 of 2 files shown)


cpvalente:
@asharonbaltazar, do you have a mac device you can check this with?
Stack overflow says the code is OK

asharonbaltazar:
Works like a charm. Do we have any other instances of shortcuts?

asharonbaltazar (approved):

cpvalente:
> Works like a charm. Do we have any other instances of shortcuts?

That I could find, only in the documentation which is outside the code base.

I will eventually get around to it

PR #512, opened 2023-09-07
[WIP] - dump curent sate

### LINKED ISSUE #174: Understudy backup workflows
Would it be possible to use this functionality to synchronise 2 separate instances of onTime - i am thinking of a production and backup instance - for example once <eventId> is triggered on the production server, it makes an web or OSC call/message to the backup onTime so start the same <eventID>...

_Originally posted by @je71175 in https://github.com/cpvalente/ontime/issues/173#issuecomment-1173008114_

alex-Arc:
Closes #174

@coderabbitai: ignore


alex-Arc:
@cpvalente Hey can you help me find where the TimerService is initialized?
it doesn't seem to be in the `startServer` function of app.ts

cpvalente:
> @cpvalente Hey can you help me find where the TimerService is initialized? it doesn't seem to be in the `startServer` function of app.ts

Timer service leverages the module to initialise itself as a singleton
https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/TimerService.ts#L446

alex-Arc:
Hmm ok
I was trying to pass the `RuntimeStore` to it on construction
but i will just do it in a separate function

cpvalente:
That might be the wrong place for it.

For ontime to resume a timer, we would need to know:
- which timer was loaded
- when did we start
- what was the playback status

With this we can see what the new state should be
likely the most natural place for this is the playback service
https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/PlaybackService.ts

Hope this helps

alex-Arc:
I am looking at adding the saving of state to EventStore every time `broadcast`  called would that not mean that somwthing hase changed?

It looks right to load it back in through PlaybackService 
but we would also need to initialize TimerService with the correct elapsed time, no?

cpvalente:
> I am looking at adding the saving of state to EventStore every time `broadcast`  called would that not mean that somwthing hase changed?



This might lead to a few too many writes. 

We will need some metrics here to make a good decision. We can work on optimisations once we have a working solution 

> but we would also need to initialize TimerService with the correct elapsed time, no?


You are correct. We might need some tweaks here but just do whatever is simpler for now. Potentially an utility function that loads a potential file and initialises the server with that

alex-Arc:
Yes it writes on every time service update, that seams to be every second when a timer is running
But if I am looking at this correctly we can reconstruct the timer state just with the stated time en any user added time?
So we maybe only need to dump en play state change and load

cpvalente:
> Yes it writes on every time service update, that seams to be every second when a timer is running
> 
> But if I am looking at this correctly we can reconstruct the timer state just with the stated time en any user added time?
> 
> So we maybe only need to dump en play state change and load

That also seems correct to me.

- which timer is playing
- what is the playback mode selected
- what time did it start
Should be enough to resume a timer

I think a good approach would be to work on the resumability and tweak the behaviour later.

The system should understand when the process was closed and when did it crash. 

In retrospect, the backend could have been built around events which would simplify this, but we will figure it out

Great work btw, I am impressed with how you navigated the code base this far

alex-Arc:
thanks happy to hear that I'm not wasting your time to bad.

TODO:
- [ ] dump the state to file
  - [ ] currently saving everything, select what is needed
  - [ ] add some feature so we can recover even if the crash happens while  writing the file
- [ ] clear dump on nice shutdown
- [ ] load state file
  - [ ] guard against corrupted data 
- [ ] load selected event
- [ ] resume timer
  - [ ] mode [armed, pause, play, roll]
  - [ ] user added time
- [ ] ??

cpvalente:
> thanks happy to hear that I'm not wasting your time to bad.
> 
> 
> 
> TODO:
> 
> - [x] dump the state to file
> 
>   - [ ] currently saving everything, select what is needed
> 
>   - [ ] add some feature so we can recover even if the crash happens while  writing the file
> 
> - [ ] clear dump on nice shutdown
> 
> - [x] load state file
> 
>   - [ ] guard against corrupted data 
> 
> - [x] load selected event
> 
> - [ ] resume timer
> 
>   - [ ] mode [armed, pause, play, roll]
> 
>   - [ ] user added time
> 
> - [ ] ??

I think this list pretty much covers it, I am sure there will be details we will need to look into as we go

Meanwhile, it would be important to try and understand the cost of writing. As a timer application, we can't afford long blocking operations. Maybe we can keep an open stream for the duration of the app? Maybe the data should simple by a small CSV like file of the expected values, we can check for a end character to validate a successfully terminated string 

I suspect the cost is in opening the file and creating the stream, not so much on the amount of data to be written.

Could you make some tests and write a small report?

alex-Arc:
yes appending to a file seams like a good idea

> I suspect the cost is in opening the file and creating the stream, not so much on the amount of data to be written.
> 
> Could you make some tests and write a small report?

I have no clue on how to profile nodejs other than pressing F12 in the browser :)
if you have some pointers/links/docs I am happy to give it a look

cpvalente:
> I have no clue on how to profile nodejs other than pressing F12 in the browser :)
> 
> if you have some pointers/links/docs I am happy to give it a look
> 


I don't think you need to worry about accuracy, we just want some baseline numbers to see 
1. What's the cost of opening a file
2. What's the cost of writing to a file
3. What is the added cost of writing the whole object in opposition to writing the minimal data

I would personally create a small sandbox to test these. The easiest way is likely to just add a few console logs with dates so we can see how long the tasks take

Otherwise, for profiling in nodejs you can run in inspect mode

https://github.com/cpvalente/ontime/blob/master/DEVELOPMENT.md#debugging-backend

alex-Arc:
testing file wite
https://github.com/alex-Arc/FileTestJs

alex-Arc:
Results so far:

**1 rep**
- override: 4.674ms
- append: 4.271ms
- overrideSync: 4.571ms
- appendSync: 4.102ms
- createWriteStream: 0.609ms
- appendStream: 2.44ms
- writeStream.end: 0.078ms

**100 rep**
- override: 201.856ms
- append: 200.694ms
- overrideSync: 200.731ms
- appendSync: 200.41ms
- createWriteStream: 0.755ms
- appendStream: 198.016ms
- writeStream.end: 0.104ms

**500 rep**
- override: 527.216ms
- append: 526.748ms
- overrideSync: 526.627ms
- appendSync: 526.184ms
- createWriteStream: 0.516ms
- appendStream: 524.043ms
- writeStream.end: 0.117ms

cpvalente:
> writes

Hi @alex-Arc , thank you for this.
It sounds to me, that keeping the file open during the session and writing straight from memory rather than as a stream is the ideal solution.
We should also clarify the language around append. What we likely want to achieve is that, at every write, we clear the file and write a new object with a particular termination.

I would also be curious about creating a small CSV like string with the values we need rather than stringifying the entire state object. 

I haven't looked into this in depth, but it looks that something like the below would suffice to restore the state
```
// playback status, eventId, startedTime, addedTime
play, idf42, 31500000, 10000
```

Either way, it seems that you have this covered, great work

alex-Arc:
I found this https://www.npmjs.com/package/steno
looks to be well tested and solves what we need?

cpvalente:
This looks great to me, Typicodes stuff is pretty good quality.  
We also already have lowdb so the new dependency comes at no cost for us

Can you add a test to the profiling code so that we can see that it actually helps us?

cpvalente:
Great work, I am looking forwards to testing this

alex-Arc:
Most should work now

the only thing is that i can't trigger a proper `SIGTERM` to test the `shutdown` function

alex-Arc:
also save is still run on `eventStore.broadcast()`
would it be better to attach to the integrationService?

cpvalente:
> Most should work now
> 
> the only thing is that i can't trigger a proper `SIGTERM` to test the `shutdown` function

hmm, processes that [exit with code 99](https://github.com/cpvalente/ontime/blob/master/apps/electron/electron.config.js#L3) have been closed using the UI. We can use this to begin with, by just checking for the exit code on the shutdown function
https://github.com/cpvalente/ontime/blob/master/apps/server/src/app.ts#L202

From top of my head, this is not implemented for docker users closing the service. we can look into it as well

> also save is still run on eventStore.broadcast()
> would it be better to attach to the integrationService?

Unfortunately, the things we need are spread across a few services (rundown service, playback service), I fear there is no one single place we can set this that could act as a catch-all. 
How about a mini event emitter that the resumeservice listens to? Or maybe you have a better idea?
Bottom line is that we definitely don't want to do it on every broadcast (which happens at least once a second)

alex-Arc:
> hmm, processes that [exit with code 99](https://github.com/cpvalente/ontime/blob/master/apps/electron/electron.config.js#L3) have been closed using the UI. We can use this to begin with, by just checking for the exit code on the shutdown function https://github.com/cpvalente/ontime/blob/master/apps/server/src/app.ts#L202

Yes i have added a function there that cleans the file on exit code 0, but cant get it to trigger from my terminal


> Bottom line is that we definitely don't want to do it on every broadcast (which happens at least once a second)

maybe it is fine to just add a `.save()` to all the `_on..` functions in TimerService except `_onUpdate`

cpvalente:
> maybe it is fine to just add a .save() to all the _on.. functions in TimerService except _onUpdate

absolutely, the idea of those functions is to handle side effects. it is a natural place for it and we can always move them if it becomes problematic

.gitignore
```diff
@@ -43,4 +43,4 @@ ontime-external/
 apps/server/src/preloaded-db/db.json
 
 # versioning file
-**/ONTIME_VERSION.js
+**/ONTIME_VERSION.js
```
cpvalente, line 46: if there are no changes in the file, we would prefer reverting to master

apps/server/src/stores/EventStore.ts
```diff
@@ -39,10 +39,10 @@ export const eventStore = {
...
   },
   broadcast() {
-    socket.sendAsJson({
-      type: 'ontime',
-      payload: store,
-    });
+      socket.sendAsJson({
+        type: 'ontime',
+        payload: store,
+      });
```
cpvalente, line 45: if there are no changes in the file, we would prefer reverting to master

apps/server/src/setup.ts
```diff
@@ -84,4 +84,4 @@ export const pathToStartDb = isTest
 export const resolveStylesDirectory = join(externalsStartDirectory, config.styles.directory);
 export const resolveStylesPath = join(resolveStylesDirectory, config.styles.filename);
 
-export const pathToStartStyles = join(currentDirectory, '/external/styles/', config.styles.filename);
+export const pathToStartStyles = join(currentDirectory, '/external/styles/', config.styles.filename);
```
cpvalente, line 87: if there are no changes in the file, we would prefer reverting to master

apps/server/src/app.ts
```diff
@@ -202,13 +204,17 @@ export const startIntegrations = async (config?: { osc: OSCSettings }) => {
 export const shutdown = async (exitCode = 0) => {
   console.log(`Ontime shutting down with code ${exitCode}`);
 
+  //TODO: clear the restore file if it was a normal exit
+  if (exitCode != 0) {
+    restoreService.clear();
+  }
+
```
cpvalente, line 211: Few questions here:
- Can you please clarify the task behind this todo? 
- Would it make more sense to use the exit code we have signified in app (99), since other exit codes might have other meanings?
- is there a reason not to use triple equals here?

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+    * Saves data to csv
+    * @param {Partial<OntimeState>} data
+    * @param {Playback} data.playback
+    * @param {string | null} data.selectedEventId
+    * @param {number | null} data.startedAt
+    * @param {number | null} data.addedTime
+    * @param {number | null} data.pausedAt
+    */
+    async save(data: Partial<OntimeState>) {
+        const newStore = data.playback + ',' + data.selectedEventId + ',' + data.startedAt + ',' + data.addedTime + ',' + data.pausedAt + ',\n';
```
cpvalente, line 66: would a template literal help make this more readable?

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (newStore != this.lastStore) {
+            this.lastStore = newStore;
+            this.file.write(newStore);
+        }
```
cpvalente, line 70: nice work on avoiding double work

should the IO part not be inside a try catch?

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+        try {
+            const data = fs.readFileSync(path.join(filePath, 'RestoreService.csv'), 'utf-8');
+            const elements = data.split(',');
+            if (elements[5] == '\n') {
+                logger.info('RESTORE', 'intact restore file');
+                this.load.playback = elements[0] as Playback;
+                this.load.selectedEventId = elements[1] != 'null' ? elements[1] : null;
+                this.load.startedAt = elements[2] != 'null' ? +elements[2] : null;
+                this.load.addedTime = elements[3] != 'null' ? +elements[3] : null;
+                this.load.pausedAt = elements[4] != 'null' ? +elements[4] : null;
```
cpvalente, line 43: We likely can simplify here.
in short, going through the elements, we can cancel the process if any of them are undefined or do not match 

Try and use shorter syntax to make this easier to read, like (rough code written here, likely will not work)


```ts
const enumValues = Object.values(Playback) <--- we can make an utility 
function to give us the values of the enum

const maybePlayback = elements?.[0]
const isValidPlayback = maybePlayback && enumValues.includes(maybePlayback)

if (!isValidPlayback) { return } <--- short circuit if something is not right

```

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+                this.ok = true;
+            } else {
+                logger.error('RESTORE', 'file not intact');
+                this.ok = false;
+            }
+        } catch (err) {
+            logger.info('RESTORE', 'faild to open RestoreService.csv, ' + err);
```
cpvalente, line 50: we have a logger origin object which we should use.
I would say that the service is not its own origin and we should use `LogOrigin.Server`

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+            if (elements[5] == '\n') {
+                logger.info('RESTORE', 'intact restore file');
+                this.load.playback = elements[0] as Playback;
+                this.load.selectedEventId = elements[1] != 'null' ? elements[1] : null;
+                this.load.startedAt = elements[2] != 'null' ? +elements[2] : null;
+                this.load.addedTime = elements[3] != 'null' ? +elements[3] : null;
+                this.load.pausedAt = elements[4] != 'null' ? +elements[4] : null;
+                this.ok = true;
+            } else {
+                logger.error('RESTORE', 'file not intact');
```
cpvalente, line 46: Not sure this is something worth communicating

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    /**
+    * try to restore timer state from csv
+    */
+    restore() {
+        if (!this.ok) return;
+        switch (restoreService.load?.playback) {
+            case (Playback.Armed):
```
cpvalente, line 79: I dont think the restore function should invoke other services, but return a partial object that the app can use to initialise services with

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+            default:
+                logger.info('RESTORE', 'unknown Playback state');
+                break;
+        }
+    }
+
+    /**
+    * delete file
+    */
+    clear() {
```
cpvalente, line 101: comments of this nature add very little other than verbosity to the file. I think it is best to omit

apps/server/src/services/TimerService.ts
```diff
@@ -77,8 +85,9 @@ export class TimerService {
   /**
    * Reloads information for currently running timer
    * @param timer
+   * @param initialData
    */
-  hotReload(timer) {
+  hotReload(timer, initialData?: initialLoadingData, playback?: Playback, addedTime?: number, pausedAt?: number) {
```
cpvalente, line 90: I dont think we should re-use the hot reload, that is not the scope of this function.

We would likely prefer adding a method `init()` to the TimerService to allow it to be re-initialised after the constructer being called

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+/**
+ * Service manages saveing of timer state
+ * that can then be resored when reopening
+ */
+export class RestoreService {
+    private lastStore: string = '';
+    private readonly file;
+    private load: Partial<OntimeState> = { playback: Playback.Armed };
+    private readonly ok;
+    private readonly filePath;
```
cpvalente, line 30: some questions here:
- can this be named more semantically to what they achieve? specifically ok (hasValidPayload?) and load (payload?)
- why is load initialised with armed?

apps/server/src/services/TimerService.ts
```diff
@@ -43,6 +45,12 @@ export class TimerService {
     this._updateInterval = timerConfig?.updateInterval ?? 1000;
   }
 
+  
+  public get paused() : number | null {
+    return this.pausedAt;
+  }
+  
+
```
cpvalente, line 53: I am unsure I understand what this achieves

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+    * delete file
+    */
+    clear() {
+        fs.unlinkSync(path.join(this.filePath, 'RestoreService.csv'));
+    }
+}
+
+export const restoreService = new RestoreService(resolveDbDirectory);
```
cpvalente, line 106: we should likely make a new constant for this and remove it from db directory
perhaps a temp folder with a restore.csv?

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+    startedAt: number | null;
+    addedTime: number | null;
+    pausedAt: number | null
+}
+
+/**
+ * Service manages saveing of timer state
+ * that can then be resored when reopening
+ */
+export class RestoreService {
```
cpvalente, line 25: this class should be exported since we export it initialised as a singleton here

also: could you add a few tests to the constructor and restore functions?

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+    * Saves data to csv
+    * @param {Partial<OntimeState>} data
+    * @param {Playback} data.playback
+    * @param {string | null} data.selectedEventId
+    * @param {number | null} data.startedAt
+    * @param {number | null} data.addedTime
+    * @param {number | null} data.pausedAt
```
cpvalente, line 63: is there a reason it is not preferable to pass the variables we need instead of an object?

cpvalente (commented):
Hey @alex-Arc , really great work navigating this.

I think we need to look into the architecture slightly, I understand this is challenging for someone new to the code base and I am happy to help. Generally:
- keep in mind the scope of work of a function, we likely do not want to add to it
- be mindful of the service hierarchy
- does the restoreService need to be a service with state? would it not suffice to have a write and a restore function? the write function writes its payload to a file, the restore function looks for a file and checks if it contains valid payload

Also left some small nitpicks for code style:
- Avoid changes in formatting to files otherwise unchanged
- Generally use `===`. I am usually OK with using `==` to check for nullish values, but prefer strict equality otherwise
- Be mindful of typos

apps/server/src/services/RestoreService.ts
alex-Arc: I think the try is handled by steno 
https://github.com/typicode/steno/blob/main/src/index.ts
but I will add a catch to log potential errors

apps/server/src/services/TimerService.ts
alex-Arc: was needed when eventStore was responsible for saving
forgot to delete it.

apps/server/src/services/RestoreService.ts
alex-Arc: Yes, but it would be nice to be able to control this especially when using it in a cluster
maybe a command line input?

apps/server/src/services/RestoreService.ts
alex-Arc: > this class should be exported since we export it initialised as a singleton here

Sorry I am not understanding the grammar here i think 🇩🇰 

I will look in to some testing

apps/server/src/services/RestoreService.ts
cpvalente: I am unsure I understand the question.
The comment was for the constant that resolved the location of the file. This is agnostic of any features.

We can expose the temp folder to docker the same way we do with others

apps/server/src/services/RestoreService.ts
cpvalente: Hahaha. No need to apologise, my spaghetti English.

I meant that we didn't need to export the class. Sorry for the poorly verbalised comment

apps/server/src/services/RestoreService.ts
alex-Arc: ah yes you are right one can mount the docker /tmp where ever in the host machine I did not know

apps/server/src/services/RestoreService.ts
alex-Arc: should I build a new object type or use the RuntimeStore ?

cpvalente: I am unsure I understand the question
If you are relating to the types you need in your object, I am not particularly opinionated, I would personally use primitive types for the things that you need

apps/server/src/services/RestoreService.ts
cpvalente: ah, I tend to prefer the first version

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+            this.file.write(newStore).catch((err) => { logger.error(LogOrigin.Server, err) });
+        }
+    }
+
+    /**
+    * try to restore timer state from csv
+    */
+    restore() {
+        if (!this.hasValidData) return;
+        switch (this.playback) {
```
cpvalente, line 99: I am not a big fan of this use for switch cases.
Would prefer to see this as separate if statements

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+                    const event = EventLoader.getEventWithId(this.selectedEventId);
+                    eventTimer.init(event, this.playback, this.selectedEventId, this.startedAt, this.addedTime, this.pausedAt);
+                }
+                break;
+            case (Playback.Roll):
+                PlaybackService.roll();
+                break;
+            default:
+                logger.info(LogOrigin.Server, 'unknown Playback state');
+                break;
```
cpvalente, line 115: no need for this, we handle in the if statements the actionable cases and ignore the rest

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+        switch (this.playback) {
+            case (Playback.Armed):
+                PlaybackService.loadById(this.selectedEventId);
+                break;
+            case (Playback.Pause):
+            case (Playback.Play):
+                if (PlaybackService.loadById(this.selectedEventId)) {
+                    const event = EventLoader.getEventWithId(this.selectedEventId);
+                    eventTimer.init(event, this.playback, this.selectedEventId, this.startedAt, this.addedTime, this.pausedAt);
+                }
```
cpvalente, line 108: I dont think it is in the scope of this function to do this. This is likely to cause issues down the line (for example with the current code the `eventLoader `is out of sync)

How about creating a `resume()` function in the Playback service that does this?
the function itself should be roughly based in the existing `startById`, with the difference that it passes the remaining data to event timer

In that case, the function in event timer might as well be called resume, since that is more descriptive of what it achieves

cpvalente:
> There are 2 things I am not sure how to handle without having state. I could pull it out of the class, but there would stile need to be some persistent data no?
> 
>     make sure to read the file before it is potentially overridden
>     and the file handler needs to persist to get the benefits
> 
> should I name it a utility instead?


You are correct, I just had a different architecture in mind and didn't adjust my comment to what you done.
Your solution is perfectly valid

I made a few small comments, thank you for being patient with the code review :D

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+                this.hasValidData = false;
+                return;
+            }
+        }
+
+        try {
+            const data = fs.readFileSync(path.join(this.filePath, 'restore.csv'), 'utf-8');
+            const elements = data.split(',');
+            if (elements[5] == '\n') {
+                logger.info(LogOrigin.Server, 'intact restore file');
```
cpvalente, line 45: logger.info(LogOrigin.Server, Found resumable state');

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+
+                const maybePlayback = elements[0] as Playback
+                if (!Object.values(Playback).includes(maybePlayback)) { this.hasValidData = false; return; }
+                this.playback = maybePlayback;
+
+                const maybeId = elements[1] === 'null' ? null : elements[1];
+                //TODO: Cannot access 'EventLoader' before initialization
+                // if (maybeId && EventLoader.getEventWithId(maybeId)) { this.hasValidData = false; return; }
```
cpvalente, line 53: is this something I need to worry about? or is it resolved?

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+                const maybePausedAt = elements[4] === 'null' ? null : parseInt(elements[4]);
+                if (maybePausedAt !== null && Number.isNaN(maybePausedAt)) { this.hasValidData = false; return; }
+                this.pausedAt = maybePausedAt;
+
+                this.hasValidData = true;
+            } else {
+                this.hasValidData = false;
+            }
+        } catch (err) {
+            logger.info(LogOrigin.Server, 'faild to open restore.csv, ' + err);
```
cpvalente, line 73: Mind the typos, either way, the error could be 
logger.info(LogOrigin.Server, `Failed to restore state: ${error}`);

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+                logger.error(LogOrigin.Server, err);
+                this.hasValidData = false;
+                return;
+            }
+        }
+
+        try {
+            const data = fs.readFileSync(path.join(this.filePath, 'restore.csv'), 'utf-8');
+            const elements = data.split(',');
+            if (elements[5] == '\n') {
```
cpvalente, line 44: this code is very clear now, but we will forget, can you please add a comment?
// we expect that a well terminated string, has a newline in the 5th element

also: triple equals please

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+                logger.info(LogOrigin.Server, 'intact restore file');
+
+                const maybePlayback = elements[0] as Playback
+                if (!Object.values(Playback).includes(maybePlayback)) { this.hasValidData = false; return; }
```
cpvalente, line 48: I hope this is not getting too nitpicky, but can we do some nicer formatting here ? 😅

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+                this.playback = maybePlayback;
+
+                const maybeId = elements[1] === 'null' ? null : elements[1];
```
cpvalente, line 51:
```ts
if (elements[i] == null) {
  return
}
this.selectedEventId = elements[i]
```

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,124 @@
...
+                //TODO: Cannot access 'EventLoader' before initialization
+                // if (maybeId && EventLoader.getEventWithId(maybeId)) { this.hasValidData = false; return; }
+                this.selectedEventId = maybeId;
+
+                const maybeStartedAt = elements[2] === 'null' ? null : parseInt(elements[2]);
```
cpvalente, line 56: same for all time values

```ts
if (elements[2] == null) {
  return
} 

const maybeNumber = Number(elements[2])
if (isNaN(maybeNumber)) {
  return
}
this.startedAt = maybeNumber
```

apps/server/src/services/RestoreService.ts
alex-Arc: Yes I'm working on this check
And this first try did not work do to that error
If you have a way to fix it that would be nice 
Otherwise I will try to find  another way

cpvalente: I see, I dont know if this is the issue, but it seems that you are trying to call a static property on the class EventLoader rather than on the instance of eventLoader, 
so it should be 
`eventLoader.getEventWithId `// note the lower case E

alex-Arc: I did try that
Property 'getEventWithId' does not exist on type 'EventLoader'. Did you mean to access the static member 'EventLoader.getEventWithId' instead?ts(2576)

cpvalente: hm, you are correct. ignore me, it is a static method.
In that case, you need to ensure the correct order of initialisation as the error suggests

It sort of looks like that should happen according to your code, looks like this might be a race condition
```ts
  // provide initial payload to event store
  eventLoader.init();
  eventStore.init(getInitialPayload());
  restoreService.restore();
```

Either way, I am unsure if loading the event should be of the scope of the service. as we mentioned before, the restore sequence could return to the app a possibly restorable state and let the app deal with it.

in turn, the app would ask the playback service to load an event, if that doesn't exist it would fail

alex-Arc: > in turn, the app would ask the playback service to load an event, if that doesn't exist it would fail

Ok I can live with this

cpvalente:
Hi @alex-Arc , how are you getting along with this?
let me know if I can help or if you would like this to be reviewed.

If you would like to chat feel free to get in touch at mail@getontime.no

alex-Arc:
Hey I'm full time on events right now
but planing to pick backup next week

cpvalente:
> I'm full time on events right now
> 
> but planing to pick backup next week 👍 

No problem, thank you for the heads up and good luck with the events 
I also didn't realise that you worked as an event technician, it would be great to get in touch, if you feel like it hit me at mail@getontime.no when you are back

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,119 @@
...
+    * @param {number} addedTime
+    * @param {number} pausedAt
+    */
+    async save(playback: Playback, selectedEventId: string | null, startedAt: number | null, addedTime: number | null, pausedAt: number | null) {
+        const newStore = `${playback},${selectedEventId},${startedAt},${addedTime},${pausedAt},\n`;
+        if (newStore != this.lastStore) {
+            this.lastStore = newStore;
+            this.file.write(newStore).catch((err) => { logger.error(LogOrigin.Server, err) });
+        }
+    }
```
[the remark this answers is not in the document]
alex-Arc: @cpvalente what do you think about this?
I think we should keep the check.

apps/server/src/services/PlaybackService.ts
```diff
@@ -241,6 +241,27 @@ export class PlaybackService {
...
+  static resumeById(eventId: string, playback: Playback, selectedEventId: string | null, startedAt: number | null, addedTime: number | null, pausedAt: number | null): boolean {
+    const event = EventLoader.getEventWithId(eventId);
+    const success = PlaybackService.loadEvent(event);
+    if (success) {
+      logger.info(LogOrigin.Playback, `Resume event with ID ${event.id}`);
+      eventTimer.resume(event, playback, selectedEventId, startedAt, addedTime, pausedAt);
+    }
+    return success;
+
+  }
```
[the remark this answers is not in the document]
alex-Arc: loadEvent checks for undefined on line 22

apps/server/src/services/TimerService.ts
```diff
@@ -74,6 +76,52 @@ export class TimerService {
...
+      this.pausedTime,
+      this.timer.addedTime,
+      this.loadedTimerEnd,
+      this.timer.timerType,
+    );
+    if (this.timer.startedAt === null) {
+      this.timer.current = this.timer.duration;
+    }
+    this.update(true);
+  }
```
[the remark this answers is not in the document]
alex-Arc: its OK for these to be null, that means they are not set yet

alex-Arc:
@cpvalente 
I think we are approaching something 
one thing the bot brings up 
should I log and return or throw if we can't create the file?

cpvalente:
> @cpvalente I think we are approaching something one thing the bot brings up should I log and return or throw if we can't create the file?
> 
> and the exitCode, right now the file is only cleared if the exitcode is 0 and this works for me when trying it out in an electron build I don't know if this is correct

I wouldnt think so, if you cannot create the file, we likely have some issues and the app will not start.
If, for some reason, the issue is isolated to this process. Since the process is not essential to the app function we do not throw, but can log a warning "Unable to start resumability service"

We dont want the service to unnecessarily block the thread.
We need to make sure that the app doesnt keep trying to make the file, we could give it a few attemps (3?) and from there we just log the error and give up.

apps/server/src/services/RestoreService.ts
```diff
@@ -0,0 +1,119 @@
...
+        if (this.file === undefined) {
+            logger.error(LogOrigin.Server, `Can not save restore file before it is created`);
+            return;
+        }
+        const newStore = `${playback},${selectedEventId},${startedAt},${addedTime},${pausedAt},\n`;
+        if (newStore != RestoreService.lastStore) {
+            RestoreService.lastStore = newStore;
+            this.file.write(newStore).catch((err) => { logger.error(LogOrigin.Server, err) });
+        }
+    }
```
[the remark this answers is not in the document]
cpvalente: hey @alex-Arc , I am leaving you to deal with this, but please be skeptical of what the bot says. In this case: playback cannot be null

apps/server/src/app.ts
```diff
@@ -7,6 +7,11 @@ import cors from 'cors';
...
 import { join, resolve } from 'path';
 
+import { Playback } from 'ontime-types';
+import path from 'path';
+import { getAppDataPath } from './setup.js'
+
+
 import { currentDirectory, environment, externalsStartDirectory, isProduction, resolvedPath } from './setup.js';
 import { ONTIME_VERSION } from './ONTIME_VERSION.js';
 import { LogOrigin, OSCSettings } from 'ontime-types';
```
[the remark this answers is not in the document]
alex-Arc: at what line is the second import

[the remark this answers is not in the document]
cpvalente: @alex-Arc , I suggest we disable the AI review in this PR for the sake of productivity 

As for the import, it does seem you have two imports of path.
In this diff is lines 7 and 11

PR #529, opened 2023-09-21
feat: added export options

### LINKED ISSUE #373: enhancement of export function in the editor app
At the editor app it would be great to enhance the export function
Today the JSON configuration will be exported at this function.
Maybe it is possible to add a small prompt asking what to export?
It could be only the CSV or only the JSON but selectable for the editor

any consultation or feedback is highly welcome.
Thank you for your support!

Enubia:
- Renamed CSV to Export
- Added modal with the option to either download as a CSV or JSON file

image
image
image

resolves #373


.eslintrc
```diff
@@ -21,6 +21,6 @@
     }
   ],
   "rules": {
-    "no-console": "warn"
+    "no-console": ["warn", { "allow": ["warn", "error"]}]
```
Enubia, line 24: this is just a proposal so that `console.warn` and `console.error` aren't marked by the linter plugin since they are valid logs usually

please disregard this if you think it's not needed

Enubia:
I was not sure where to implement this feature since the initial request was for the editor view but the example provided was in the cue sheet so I've gone ahead and added it to both views

.eslintrc
cpvalente: I think this is a good change.
the linter rule attempts to help the fact that occasionally `console.log` are left behind from debugging, but `console.warn` and `console.error` are rarely used in debugging

Good stuff

apps/client/src/common/api/ontimeApi.ts
```diff
@@ -112,29 +113,51 @@ export async function postOscSubscriptions(data: OscSubscription) {
  * @description HTTP request to download db
  * @return {Promise}
  */
-export const downloadRundown = async () => {
-  await axios({
+export const downloadRundown = async (exportType: 'csv' | 'json') => {
+  const response = await axios({
```
cpvalente, line 117: great work.

I think that the strategies should be separate, the CSV file is technically not related to the project, but a feature.
With that in mind, the code related to it should be detachable from the remaining code

Could we have separate `downloadRundown` and `downloadCSV` functions, potentially with a auxiliary function for shared logic (if any)?

apps/client/src/features/cuesheet/CuesheetWrapper.tsx
```diff
@@ -69,37 +74,90 @@ export default function CuesheetWrapper() {
...
+          <ModalBody className={styles.modalBody}>
+            <Button onClick={() => onModalClose('csv')} variant='ontime-filled' width='48%'>
+              CSV
+            </Button>
+            <Button onClick={() => onModalClose('json')} variant='ontime-filled' width='48%'>
+              JSON
+            </Button>
+          </ModalBody>
+        </ModalContent>
+      </Modal>
```
cpvalente, line 160: Few nitpicks:

We would likely prefer isolating this in its own component.
Since we also implemented this both here and rundown it would also help with maintenance

The premise of the modal "Choose your format" is incorrect in the sense that the datasets are different, something more generic like "Download options" is maybe a better fit.
In the modal the user should be given two download options "Download project file" "Download CSV export" (terrible wording from my side too)

apps/client/src/features/cuesheet/CuesheetWrapper.tsx
```diff
@@ -69,37 +74,90 @@ export default function CuesheetWrapper() {
...
 
   return (
     <div className={styles.tableWrapper} data-testid='cuesheet'>
-      <CuesheetTableHeader handleCSVExport={exportHandler} featureData={featureData} />
+      <CuesheetTableHeader handleExport={handleOpenModal} featureData={featureData} />
       <Cuesheet data={rundown} columns={columns} handleUpdate={handleUpdate} selectedId={featureData.selectedEventId} />
+      <Modal isOpen={isModalOpen} onClose={onModalClose} motionPreset='scale' size='xl' colorScheme='blackAlpha'>
+        <ModalOverlay />
+        <ModalContent>
+          <ModalHeader fontWeight={400}>Choose your format</ModalHeader>
```
cpvalente, line 149: is this not the class in the CSS module?

I believe we can either remove the inline style or the class declaration, where I would prefer keeping local styling in the module

apps/client/src/features/cuesheet/CuesheetWrapper.tsx
```diff
@@ -69,37 +74,90 @@ export default function CuesheetWrapper() {
...
+          <ModalCloseButton />
+          <ModalBody className={styles.modalBody}>
+            <Button onClick={() => onModalClose('csv')} variant='ontime-filled' width='48%'>
+              CSV
+            </Button>
+            <Button onClick={() => onModalClose('json')} variant='ontime-filled' width='48%'>
+              JSON
+            </Button>
```
cpvalente, line 157: from my poor UX experience, we should only have one action button at any time
In this case: either the JSON export is the primary, or none of them are, which would make both buttons be of the `ghosted` type 
I will leave a decision up to you

apps/client/src/features/menu/MenuBar.tsx
```diff
@@ -128,12 +167,14 @@ const MenuBar = (props: MenuBarProps) => {
...
         icon={<IoSaveOutline />}
         isDisabled={appMode === AppMode.Run}
-        clickHandler={downloadRundown}
+        clickHandler={() => setIsModalOpen(true)}
         tooltip='Export project file'
         aria-label='Export project file'
         size='sm'
       />
 
+      {exportOptionsModal()}
```
cpvalente, line 176: Few code style nitpicks:

is there a reason for this to be a function and not a react component?
In this case, I think it might just work fine, but I believe it to be a bad pattern to follow

we generally tend to use the `function Component` style rather than arrow function, it is nice to be consistent. I would have generally not mentioned it, but either way, all components must start with Uppercase, I didn't even know that react would let you name it otherwise

we will likely want to share this component so we can extract it in its own file to the `common/components` directory

cpvalente (commented):
Hi, thank you for the impromptu PR ⭐ 
Good work at navigating the feature and the code base.

I left some small comments, mostly nitpicks to do with code style

The feature was indeed only meant for `/editor` but I guess it doesn't hurt to have them in both places

The download from editor is also covered in playwright tests which we would need to update, it should be pretty straightforward to do so

Once this small things are resolved I will test the feature, thank you again

Enubia:
@cpvalente Added your comments and suggestions to my changes. Please check them again

cpvalente:
Hi @Enubia , I see most of the comments from the AI code review are unrelated to your specific changes.
I am happy with your code as it is, although handling of failed `fileDownload `wouldn't be unwise.

Let me know if you think this is done and I will merge it into a test branch

Thank you very much for your work here and apologies for the slow replies

Enubia:
Hey @cpvalente sorry for the late reply

I think the changes are fine to test. Regarding the `fileDownload`, you expect the server to be up anyways right? So there should never be a case where the server can't respond.

cpvalente:
Hi @Enubia, thank you for your work here, I will merge and add it as part of the next release

You did a good job at navigating the code base, if you are available, it would be great to have your help with another task. Can I suggest https://github.com/cpvalente/ontime/issues/374 ? 

Either way, thank you again

cpvalente:
@Enubia perfect!
drop a comment in the ticket when you are ready and I an assign it to you

PR #553, opened 2023-10-23
Feat/change event api

alex-Arc:
#478 
#347

### DIFF
--- a/apps/server/src/controllers/integrationController.ts
+++ b/apps/server/src/controllers/integrationController.ts
@@ -6,7 +6,6 @@ import { eventStore } from '../stores/EventStore.js';
 import { parse, updateEvent } from './integrationController.config.js';
 import { isKeyOfType } from 'ontime-types/src/utils/guards.js';
 import { event } from '../models/eventsDefinition.js';
-import { logger } from '../classes/Logger.js';
 
 //TODO: simplify the Error path
 export function dispatchFromAdapter(
--- a/apps/server/src/controllers/integrationController.ts
+++ b/apps/server/src/controllers/integrationController.ts
@@ -246,8 +246,8 @@ export function dispatchFromAdapter(
           throw new Error('Invalid parameters');
         }
 
-        if (!payload || typeof payload !== 'string') {
-          throw new Error(`Invalid payload: ${payload}`);
+        if (payload === undefined) {
+          throw new Error(`Undefined payload`);
         }
 
         const eventID = params[0];
@@ -261,7 +261,7 @@ export function dispatchFromAdapter(
 
         return updateEvent(eventID, propertyName, parsedPayload);
       } catch (error) {
-        throw new Error(`Error updating duration: ${error}`);
+        throw new Error(`Error updating event: ${error}`);
       }
     }
 
--- a/apps/server/src/controllers/integrationController.ts
+++ b/apps/server/src/controllers/integrationController.ts
@@ -244,26 +244,18 @@ export function dispatchFromAdapter(
     // ontime/change/{eventID}/{propertyName}
     case 'change': {
       if (params.length < 2) {
-        logger.error(LogOrigin.Rx, `${source}: To few parameters`);
-        return;
+        throw new Error(`To few parameters`);
       }
       if (payload === undefined) {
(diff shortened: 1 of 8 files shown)


alex-Arc:
super small thing.
the error chaining is a bit ugly
`[ERROR]          RX      OSC IN: Error: Error updating event: Error: Undefined payload`
but don't know if we can do any better

cpvalente:
> super small thing. the error chaining is a bit ugly `[ERROR] RX OSC IN: Error: Error updating event: Error: Undefined payload` but don't know if we can do any better

sure, we can take a look, just let me know when you have made the changes you think are necessary

alex-Arc:
every thing is functioning
should I look into writing tests for the /change function?

alex-Arc:
@coderabbitai review

packages/types/src/utils/guards.ts
```diff
@@ -1,5 +1,6 @@
 import { OntimeRundownEntry } from '../definitions/core/Rundown.type.js';
 import { OntimeBlock, OntimeDelay, OntimeEvent, SupportedEvent } from '../definitions/core/OntimeEvent.type.js';
+import { event } from '../../../../apps/server/src/models/eventsDefinition.js';
```
cpvalente, line 3: This import shouldn't be allowed. I will make a rule for it. Can we remove it?

apps/server/src/utils/coerceType.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export function coerceBoolean(value: unknown): boolean {
+  if (value == null) {
+    throw new Error('Invalid value received');
+  }
+  return Boolean(value);
+}
+
+/**
+ * @description Converts an unknown value to a number, if it's not null or undefined. Parses the value and checks if it's a valid number. Throws an error otherwise.
```
cpvalente, line 28: This is a little confusing.
This function converts a value to a number if possible, throws otherwise

apps/server/src/utils/coerceType.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * @description Converts an unknown value to a string, if it's not null or undefined. Throws an error otherwise.
```
cpvalente, line 2: Same comment as for the coerceNumber function

apps/server/src/controllers/integrationController.ts
```diff
@@ -222,6 +239,32 @@ export function dispatchFromAdapter(type: string, payload: unknown, source?: 'os
       return { topic: 'timer', payload: timer };
     }
 
+    // ontime/change/{eventID}/{propertyName}
+    case 'change': {
+      try {
```
cpvalente, line 244: I don't think we need to catch here. There is a good opportunity for improvement of the error throwing error logic

cpvalente (changes requested):
Made some small comments

apps/server/src/controllers/integrationController.ts
alex-Arc: I might need some help getting rid of the last part of the try catch

apps/server/src/controllers/integrationController.ts
```diff
@@ -1,9 +1,27 @@
+import { LogOrigin, OntimeEvent } from 'ontime-types';
+
 import { messageService } from '../services/message-service/MessageService.js';
 import { PlaybackService } from '../services/PlaybackService.js';
 import { eventStore } from '../stores/EventStore.js';
+import { parse, updateEvent } from './integrationController.config.js';
+import { isKeyOfType } from 'ontime-types/src/utils/guards.js';
+import { event } from '../models/eventsDefinition.js';
+
+//TODO: simplify the Error path
```
cpvalente, line 10: can you please remove this todo?
It seems that it is no longer relevant?

cpvalente (approved):

cpvalente:
@alex-Arc if the crash scenario we have identified is not related to the changes, lets merge this!

arymoraes:
> I will merge this for now. Thank you very much @arymoraes for your work here If you have availability it would be great to continue having your help, I would suggest #374 which is easily broken in small tasks that we could do together

Sure I'd love to tackle that one!

PR #555, opened 2023-10-26
Async Data Provider

kyelewis:
Part of a series of PRs planned to make it possible to use redis or another solution instead of just a local JSON file for data storage.

This PR updates the Data Provider to be async for all read functions, and all of the drill-down needed to get that through the various layers.

This passes the e2e but I think some additional testing is due around the express API, WS and OSC integrations to make sure everything is still working.

First PR on this project so happy to take any feedback and change requests.

### DIFF
--- a/apps/server/src/adapters/OscAdapter.ts
+++ b/apps/server/src/adapters/OscAdapter.ts
@@ -36,7 +36,7 @@ export class OscServer implements IAdapter {
         return;
       }
 
-      void (async () => {
+      (async () => {
         try {
           const reply = await dispatchFromAdapter(path, args, 'osc');
           if (reply) {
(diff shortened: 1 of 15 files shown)


cpvalente:
Hi @kyelewis , thank you for this. It seems that it took some considerable amount of work to reach this stage and I appreciate your efforts

I am hoping we can clarify the proposition of the change.
In the case of a distributed data provider, would it not make sense to instead synchronise a local dataset with the remote?

We are currently investigating the possibility of integrating with google-sheets. This is made slightly more complex since we are unable to have a paid service-service connection.
However, the work entails creating a worker which syncs the local dataset with an external one. 
@alex-Arc has been doing the tests here. Maybe this can be a good solution for your use case?

Look forwards to hearing from you

cpvalente:
Hi @kyelewis, we are currently planning v3.
It would be nice to discuss the possibilities for async external sources, maybe you have some time to advice?

I will eventually create a discussion for it, but for now you are welcome to get in touch at mail@getontime.no

PR #556, opened 2023-10-26
Documentation- Clarify build:local step before E2E

kyelewis:


cpvalente:
Hi @kyelewis, thank you for this.
If you have any suggestions for other improvements to developer documentation it would be very welcome

PR #558, opened 2023-10-27
partial update

alex-Arc:
fix the // TODO: implement partial store updates
tested with chrome, firefox, samgsung tv and companion with https://github.com/bitfocus/companion-module-getontime-ontime/pull/36


alex-Arc:
We will look at it at a later time

Issue #560, opened 2023-10-28
Enabling Roll Mode leads to WS In Error


lebaston100:
I'm trying to use roll mode but when clicking the roll button i just get a "WS IN: TypeError: Cannot read properties of null (reading 'id')" Error in the Log and no timer is started. I've tested it with both of the latest stable and the beta on a completely new project with just a single event in it. Don't know of any log files i could provide, so if there is anything you need to investigate this further just let me know.
![image](https://github.com/cpvalente/ontime/assets/26111114/3648657a-9166-477b-8497-77880b0bd10a)

alex-Arc:
Hi @lebaston100  I think if found the bug
can you try and see if it works if you set the event to public?

lebaston100:
Yes, it seems to work if the event is set public.

alex-Arc:
Great it should be fixed when #561 gets merged
after we find all the places were this might also happen

PR #561, opened 2023-10-28
wip: strict nullcheck

alex-Arc:
#560


cpvalente:
Hi @alex-Arc, this looks correct, but it is strange that typescript let this pass.
Are you able to investigate why is are the values not being correctly typed as potentially null?

the return of getRollTimers should be typed as 
```
  nowIndex: number | null;
  nowId: string | null;
  publicIndex: number | null;
  nextIndex: number | null;
  publicNextIndex: number | null;
  timeToNext: number | null;
  nextEvent: OntimeEvent | null;
  nextPublicEvent: OntimeEvent | null;
  currentEvent: OntimeEvent | null;
  currentPublicEvent: OntimeEvent | null;
```
Once this is done typescript will throw errors if we don't handle the null cases

alex-Arc:
@cpvalente do we need this branch or did it get fixed else where?

cpvalente:
Hi @alex-Arc , this work was paused to avoid merge conflicts with the work in over-under
I will leave it here for reference but the fix was already done in another merge

thanks for checking in

PR #565, opened 2023-10-31
feat: expand user options for views with schedule

cpvalente:
small PR to add two new options to views with the scheduler component (backstage and public)

1. option to hide events that are in the past
2. option to stop the auto cycling of the schedule

### DIFF
--- a/apps/client/src/common/components/view-params-editor/constants.ts
+++ b/apps/client/src/common/components/view-params-editor/constants.ts
@@ -205,7 +205,7 @@ export const BACKSTAGE_OPTIONS: ParamField[] = [
   {
     id: 'stopCycle',
     title: 'Stop cycling through event pages',
-    description: 'Schedule will auto not cycle through events',
+    description: 'Schedule will not auto-cycle through events',
     type: 'boolean',
   },
 ];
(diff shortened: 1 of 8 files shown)


apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -194,6 +194,37 @@ export const LOWER_THIRDS_OPTIONS: ParamField[] = [
...
+  {
+    id: 'hidePast',
+    title: 'Hide past events',
+    description: 'Scheduler will only show upcoming events',
+    type: 'boolean',
+  },
+  {
+    id: 'stopCycle',
+    title: 'Stop cycling through event pages',
+    description: 'Schedule will auto not cycle through events',
```
alex-Arc, line 208: Schedule will not auto cycle through events

alex-Arc (changes requested):
'hide events that are in the past' dose not work when the publicEventId is null but that seams to be a problem upstream

cpvalente:
> 'hide events that are in the past' dose not work when the publicEventId is null but that seams to be a problem upstream

If `publicEventId` is null we do not have a running event. So there are no events in the past.
Is this not expected?

alex-Arc:
here all events are public and event 3 is loaded and hide past is enabled in both the backstage and public view
image

backstage hides event 1 -> 3 but public shows them all

if i add a console log to ScheduleContext.tsx the public one cant find the evet
image

apps/client/src/common/components/view-params-editor/constants.ts
cpvalente: good catch, thank you

cpvalente:
good catch @alex-Arc, I hope I have managed to cover them all now

alex-Arc (approved):

PR #568, opened 2023-11-01
Save icon for cuesheet

alex-Arc:
from this
image

to this
image


cpvalente:
hey @alex-Arc , we seem to have some issues with the commits here
As for the change, I think the save icon on its own does not explain what is happening, how about adding icon and keeping the "export" text?

alex-Arc:
I was thinking that since it opens the same modal as this button
image
it should just be the same?

cpvalente:
Hi @alex-Arc , can we close this for now? lets revisit once we had some time to think through UI properly

alex-Arc:
Sounds good

PR #572, opened 2023-11-03
Backstage options alt

alex-Arc:
I combined `stopCycle` and `hidePast` into `followSelected` because I don't see `stopCycle` being used without `hidePast` anyway.
And this way we keep the progression overview in the small dots

also added `eventsPerPage` might be very useful for portrait displays

feel free to drop it if it dosen't solve the intended use case


cpvalente (commented):
Few comments here:
- it seems that we have some git mess up with commits from other branches
- I am unsure on the value of joining the two parameters. 
- the vertical items seems to be useful, but the real useful code is to make the component distribute items depending on the size they take. 

if we extract this last bit into a PR I think it could provide some value for now

PR #573, opened 2023-11-03
update node-osc to v9

alex-Arc:
update node-osc to fix this issue https://github.com/MylesBorins/node-osc/issues/104

### DIFF
--- a/apps/server/src/adapters/OscAdapter.ts
+++ b/apps/server/src/adapters/OscAdapter.ts
@@ -5,7 +5,6 @@ import { Server } from 'node-osc';
 import { IAdapter } from './IAdapter.js';
 import { dispatchFromAdapter } from '../controllers/integrationController.js';
 import { logger } from '../classes/Logger.js';
-import { error } from 'console';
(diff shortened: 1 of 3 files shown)


apps/server/src/adapters/OscAdapter.ts
```diff
@@ -5,14 +5,15 @@ import { Server } from 'node-osc';
 import { IAdapter } from './IAdapter.js';
 import { dispatchFromAdapter } from '../controllers/integrationController.js';
 import { logger } from '../classes/Logger.js';
+import { error } from 'console';
```
cpvalente, line 8: can you clarify this import?

cpvalente (commented):
left a small comment for the changes 
great job chasing this, much appreciated

apps/server/src/adapters/OscAdapter.ts
alex-Arc: ah i didn't catch that 
no that is VS code auto importing stuff

PR #575, opened 2023-11-04
Feat/http integration

alex-Arc:
TODO:
- [x] add globale switch for http out
- [x] revert this [create universal Subscription](https://github.com/cpvalente/ontime/pull/575/commits/89c7f8707b5319f310a591d9cf5157d25683da6c) [unify osc and http validateSubscriptionEntry](https://github.com/cpvalente/ontime/pull/575/commits/2b87f72acb024068e63f93401258ffe509973c9d)
- [x] find correct settings for http agent


alex-Arc:
Hey @asharonbaltazar any comments about the [HttpIntegration.ts](https://github.com/cpvalente/ontime/pull/575/files#diff-2e0c9aa4079a76d022b316738aca8e1960b401c93e606a3127bc89207227cc73) would be much appreciated

alex-Arc:
@cpvalente 
I could use some help 

solving this lint error https://github.com/cpvalente/ontime/actions/runs/6922773895/job/18829887020?pr=575#step:6:12

and adding a global toggle 
Screenshot 2023-11-20 112929

apps/server/src/utils/parserFunctions.ts
```diff
@@ -216,20 +221,67 @@ export const parseOsc = (data: { osc?: Partial<OSCSettings> }): OSCSettings => {
...
 };
 
+/**
+ * Parses and validates HTTP subscription cycle options
+ * @param data
+ */
+export const validateHttpSubscriptionCycle = (data: HttpSubscriptionOptions[]): boolean => {
+  for (const subscriptionOption of data) {
+    const isHttp =
+      subscriptionOption.message?.startsWith('http://') || subscriptionOption.message?.startsWith('https://');
```
cpvalente, line 231: would we not need a certificate for https?

have you added code to generate this? maybe we can leave https for next time

cpvalente (commented):
Some small comments, lets cleanup and get this merged
We can extract some code into another branch and make a PR immediately so that you can continue work

apps/server/src/utils/parserFunctions.ts
alex-Arc: we can 
but I don't think a client needs a certificate
https://security.stackexchange.com/questions/110621/ssl-newbie-does-https-client-also-need-a-certificate
https://stackoverflow.com/questions/19383538/what-is-the-difference-between-requiring-an-ssl-cert-and-accepting-an-ssl-cert

in my tests it also works fine.

apps/server/src/services/integration-service/HttpIntegration.ts
```diff
@@ -0,0 +1,137 @@
...
+          };
+        }
+      }
+    });
+  }
+
+  async emit(path: globalThis.URL) {
+    try {
+      await got.get(path, {
+        retry: { limit: this.retryCount },
```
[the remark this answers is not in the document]
alex-Arc: Got has a quite advanced retry API 
https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#retry
the default is an exponential back-off

all 400 request up to 408 are not retired by default. we can also remove `ENOTFOUND` from the allowed try cases

I think our retry count should be 0 and then if a user wants to integrate with a service that might be unreliable (the internet) they can increase it

PR #576, opened 2023-11-05
Addtime in sec

alex-Arc:


apps/server/src/controllers/integrationController.ts
```diff
@@ -174,6 +174,19 @@ export function dispatchFromAdapter(
...
+    case 'addtime': {
+      const time = Number(payload);
+      if (isNaN(time)) {
+        throw new Error(`Time not recognised ${payload}`);
+      }
+      try {
+        PlaybackService.addTime(time);
+      } catch (error) {
+        throw new Error(`Could not add time: ${error}`);
+      }
```
cpvalente, line 186: I see this is a pattern we use here, we should make a todo to improve this. In this case, re-throwing the error does not add any extra information or value

apps/server/src/controllers/integrationController.ts
```diff
@@ -174,6 +174,19 @@ export function dispatchFromAdapter(
       PlaybackService.roll();
       break;
     }
+    case 'addtime': {
```
cpvalente, line 177: I wonder if `addtime` is the best way to call it. 
I realise it matches the internal datapoint, but that doesnt have to be explicit to consumers.

However I dont have a better idea, what do you think?

apps/server/src/services/PlaybackService.ts
```diff
@@ -269,17 +269,26 @@ export class PlaybackService {
     }
   }
 
+  /**
+   * Adds time to current event
+   * @param {number} time time to add in seconds
```
cpvalente, line 274: you wrote time twice

apps/server/src/services/PlaybackService.ts
```diff
@@ -269,17 +269,26 @@ export class PlaybackService {
...
+      eventTimer.addTime(timeInMs);
+      timeInMs > 0
+        ? logger.info(LogOrigin.Playback, `Added ${time} sec`)
+        : logger.info(LogOrigin.Playback, `Removed ${time} sec`);
+    }
+  }
+
   /**
    * Adds delay to current event
+   * @deprecated Use addTime
```
cpvalente, line 288: great use of jsdocs, good stuff

cpvalente (commented):
Code looks good, just some nitpicks and questions

apps/server/src/services/PlaybackService.ts
alex-Arc: ah, it formats like this for me... should I change it
image
image

cpvalente: I usually add a - but didnt remember this was the reason 😅 
up to you, its all good

apps/server/src/controllers/integrationController.ts
alex-Arc: I am also a bit unsure
as I think this function is mostly for the speaker timer the convention from other apps like stagetimer.io seams to be add/remove time

cpvalente:
Hi @alex-Arc , this looks good to me, lets line up the companion changes and we set this up for next release

PR #520, opened 2023-09-11
feat: google sheets

hrueger:
Hi @cpvalente,
I finally found some time to push what I was able to get working regarding the authentication.
You need to create your own google cloud project, because mine is set to test and I don't know your gmail address to add you as a test user. But setting the project up is easy, if I remember everything ;-)
- Go to https://console.cloud.google.com and create a project
- activate the google sheets api
- go to OAuth Consent Screen and setup your application (for now, test is probably fine). Add yourself as a testuser
- go to Credentials and create a new OAuth 2.0 Client ID
    - use type webapplication
    - add the following authorized js origins:
        - http://localhost:3001
        - http://localhost:3000
        - http://localhost:4001
        - http://localhost
    - copy the client id and change the value in [apps/client/src/features/modals/settings-modal/SettingsModal.tsx](https://github.com/cpvalente/ontime/compare/master...hrueger:google-sheets?expand=1#diff-82e144a5530837c06c249dba43315be8dab88e11a32e92d70fb0700e3bbf966a)

Currently, there's no import logic, we just log the document title to make sure we can talk to the API.

ToDo:
- [x] Move Client ID to a .env variable for developing with own cloud projects and setup GitHub Actions + Secrets to add that to the build
- [ ] Make the Sheet ID configurable in the modal (I just copied a different settings tab page)
- [ ] Maybe we should make the sheet selectable? With the drive api, we could get a list of recently used sheets ;-) Problably something for a second step, though.
- [ ] Add actual import / export logic (maybe like with the excel files? I believe I saw a pre-release there) with a Sync button
- [ ] Automatically refresh the access_token via the refresh_token
- [x] Create an external login page hosted with GitHub pages
    - then, we can get logins from outside the Electron app working
    - Hosting the button on GitHub Pages means we have a fixed js origin.
    - When the button returns the access token and refresh token, we could either do a clientside js redirect back to ontime (url specified via query params) - however, this is probably insecure since we cannot verify that we're redirecting to ontime and not to somewhere else.
    - The more secure alternative would be to let the user copy the json structure (maybe encoded as base64 or something) and paste it in ontime. Then, we're sure that the user is doing that on purpose. However, it is not as convenient. 
    - Or we can ask the user if he is sure that his ontime is running at [ip].

What do you think? Do you have some time to assist?

PR #579, opened 2023-11-08
Google sheets lite

### LINKED ISSUE #194: Feature Request: Sync with Google Sheets

alex-Arc:
- [x] Proper UI
- [x] save sheetid and worksheet to settings
- [x] encrypt token (we don't save it at all)
- [x] create `how to` create cloud project
- [x] push control flow
- [x] pull control flow
- [x] check that the given sheet id and worksheet exists
- [x] tests?

Closes #194


cpvalente:
Hey @hrueger, do you have some input you can give here?

hrueger:
Hi, this looks pretty interesting. If I read the code correctly, it's using a different auth method than I began implementing in #520.
@alex-Arc am I correct that people will need to create a project on the google cloud console, download the credentials file and upload that in ontime?
I think @cpvalente and I have been discussing different authentication options already somewhere, I just cannot find the issue right now. I tried client authentication without a secret, because we did not want people to have to create a google cloud console project.

However, the big advantage of your approach is probably that there are no short-lived auth tokens. In my PR, you always have to reauthenticate every 5 (?) minutes or so. That's annoying ;-)
With your approach, ontime could always sync changes back up and no re-logins would be needed.

I do wonder, though, whether an authentication with an API key would be easier if people have to create a gcc project anyway?

It would probably be the cleanest way to setup a simple server hosted by us (which uses our gcc project and keeps our secret secret) where people login. Then, this server would relay sheet api commands. We could for example host this on vercel free.
I'm not sure how much work that is, though. What to you think?
Don't get me wrong @alex-Arc, the work you've done here is great and I really appreciate it. I just wonder whether people are comfortable enough creating a google cloud project. Or did I read your code wrong?

Best
Hannes

alex-Arc:
@hrueger No I thing you got it pretty much right 😄. The reason for the client secret is  to allow us to push to sheets as well.
And I followed the method used by the companion integration.  https://github.com/bitfocus/companion-module-google-sheets

I don't know why i missed that you had a PR open about this😕 

Would you not need to create a cloud project to get an API key anyway? 
that is the way vMix dose it at least https://www.vmix.com/knowledgebase/article.aspx/224/how-to-use-google-sheets-with-vmix-data-sources

for ontime to have a gcc project, one user could easily hit the rate limiter, and block all other uses right?
or am I misunderstanding the way the quotas work.

hrueger:
Interesting. If all other programs require users to create a gcc project, I think we should do that, too. Do you agree, @cpvalente ?

What do you think is better, copying an api key or using the credentials file?

hrueger:
We could use the os native credentials store (https://github.com/atom/node-keytar or my fork https://github.com/makeproaudio/node-keytar). Credentials would then have to be added to each computer individually because they would not be included in the project file.

We could also encrypt the api key and save it in the project file, I just wonder what the encryption key should be. I don't think it is a big difference to plaintext if we just use the few-digit editor pin... 🤷‍♂️

alex-Arc:
currently I save the client file to the appdir 
this file is not dangerous as it still requires a the user to authenticate, from this a token is generated.
It seams the recommendation is a key store like you recommend
https://developers.google.com/identity/protocols/oauth2/resources/best-practices

I saw that you had this stuff under a tab in settings called sync, maybe that is more appropriate than the import section?
especially when/if we add the upload function

hrueger:
Sounds good 👍

> I saw that you had this stuff under a tab in settings called sync, maybe that is more appropriate than the import section?
especially when/if we add the upload function

I'm open to everything 👍

alex-Arc:
@hrueger Would you be able to help me create the UI necessary?
I'm not very experienced in react

hrueger:
To be honest, before my PR here I've never used react either... I just copied other components and fiddled around with them until it worked, but I did not really understand the concepts behind. Probably @cpvalente knows much better ;-)

cpvalente:
Hi @alex-Arc, I am happy to make the UI, can you make me a napkin drawing of what is needed and I will push some code?

alex-Arc:
Screenshot 2023-11-16 112036

Screenshot 2023-11-16 114515

apps/server/package.json
```diff
@@ -12,6 +12,8 @@
     "express-session": "^1.17.3",
     "express-static-gzip": "^2.1.7",
     "express-validator": "^6.14.2",
+    "google-auth-library": "^9.2.0",
+    "googleapis": "^128.0.0",
```
cpvalente, line 16: S**t, this adds like 9MB to the app
2.6MB > 11.7MB

Is there anything we can do about this?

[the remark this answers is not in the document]
hrueger: Related: https://github.com/googleapis/google-api-nodejs-client/issues/806
That issue has been fixed in https://github.com/googleapis/google-api-nodejs-client/pull/2557
and there's now a separate sheets package: https://www.npmjs.com/package/@googleapis/sheets

cpvalente:
@alex-Arc , i made some placeholder UI to get you started, I will look into proper UI design once we have a flow
Let me know if there is anything else I can do to help

Screenshot 2023-11-17 at 15 22 19

apps/client/src/features/modals/sheets-modal/SheetsModal.tsx
```diff
@@ -0,0 +1,250 @@
...
+      _onChange();
+    }
+  };
+
+  const _onChange = async () => {
+    setSheetState(await getSheetstate());
+  };
+
+  if (isOpen) {
+    //TODO: how to get this on modal open
```
alex-Arc, line 80: @cpvalente could you give a tip on this
I would like it to run this onOpen

cpvalente: To run side effects in react, you would generally use a `useEffect`
Keep in mind that is not always the case, the recipe for useEffect is to synchronise react with things outside of react


In your case,
It seems that `getSheetState` should be called in a `useEffect` and its value pushed to a piece of state instead of a ref. You need the state to trigger a re-render in the components.

We can consider whether we want to push this into react-query which would manage this lifecycle for us. see how the others are done such as `useViewSettings`

Send me a message with what you are trying to achieve and I can take a look

apps/client/src/features/modals/sheets-modal/SheetsModal.tsx
```diff
@@ -0,0 +1,250 @@
...
+        <ModalHeader>Sheets!</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          {rundown && (
+            <>
+              <PreviewExcel
+                rundown={rundown ?? []}
+                project={project ?? projectDataPlaceholder}
+                userFields={userFields ?? userFieldsPlaceholder}
+              />
```
cpvalente, line 178: I wonder whether we should skip this for now

It is complicating your code. This is an experimental feature and the main focus should be in the feature core. Once that works well and we have a good idea of how to integrate, we can expand.

This means we can move faster and be more decisive, for all we know, we decide that this logic does not belong here or that the feature is not usable for ontime

asharonbaltazar:
What can I help with? 

Warning: I'm not an expert, by any means 😄

alex-Arc:
If you could take a look at this file apps/client/src/features/modals/sheets-modal/SheetsModal.tsx
It is a bit of a mess from me copy pasting stuff in from all over, any pointers in a more "correct" React way of doing stuff.
image

I would also like stuff moved about something like this. and if the small x icons could be proper icons
image

apps/client/src/features/modals/sheets-modal/SheetsModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+      if (worksheet.current) {
+        worksheet.current.value = data.worksheet;
+      }
+    });
+  };
+
+  const handleAuthenticate = () => {
+    getSheetsAuthUrl().then((data) => {
+      if (data != 'bad') {
+        window.open(data, '_blank', 'noreferrer');
```
alex-Arc, line 110: Also if there is a way to detect a child tab closing?

cpvalente (commented):
some changes and suggestions, good work

packages/types/src/definitions/core/GoogleSheet.type.ts
```diff
@@ -0,0 +1,12 @@
+export type GoogleSheet = {
```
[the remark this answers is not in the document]
alex-Arc: I would personally like it to save but agree that for the app it might make sens to not do that

apps/server/src/utils/parser.ts
```diff
@@ -103,140 +107,228 @@ export const parseExcel = (excelData: unknown[][], options?: Partial<ExcelImport
...
-          if (typeof column === 'string') {
-            const col = column.toLowerCase();
-
-            if (handlers[col]) {
-              handlers[col](j);
-            }
-            // else. we don't know how to handle this column
-            // just ignore it
+  excelData.forEach((row, rowIndex) => {
+    if (row.length === 0) {
```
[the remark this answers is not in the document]
alex-Arc: to push the data back up to a sheet we need to know where the headers are located

apps/server/src/utils/googleSheetUtils.ts
```diff
@@ -0,0 +1,165 @@
...
+ * @description - creates updateCells request from ontime event
+ * @param {ProjectData} projectData
+ * @param {number} worksheetId
+ * @param {any} metadata - object with all the cell positions of the title of each attribute
+ * @returns {sheets_v4.Schema} - list of update requests
+ */
+export function cellRequenstFromProjectData(
+  projectData: ProjectData,
+  worksheetId: number,
+  metadata,
```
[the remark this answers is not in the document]
alex-Arc: it is my name for the headerlocation

packages/types/src/definitions/core/GoogleSheet.type.ts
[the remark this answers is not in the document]
hrueger: Can we make this a setting to opt in (off by default)? I think both cases are valid, I'd personally prever it being saved, too.

apps/electron/main.js
```diff
@@ -138,6 +138,11 @@ function createWindow() {
   });
 
   win.setMenu(null);
+
+  win.webContents.setWindowOpenHandler(({ url }) => {
+    shell.openExternal(url);
+    return { action: 'deny' };
```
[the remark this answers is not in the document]
alex-Arc: it is what opens the auth process in the browser and not a new electron window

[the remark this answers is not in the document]
cpvalente: That is the exact same that the other function though? If it doesn't achieve anything new we should reuse it

packages/types/src/definitions/core/GoogleSheet.type.ts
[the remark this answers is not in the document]
cpvalente: We are trying to simplify the feature so that we can take smaller steps 

Making things optional is the opposite direction

apps/electron/main.js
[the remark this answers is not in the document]
hrueger: As far as I can tell, they're not exactly the same . `send-to-link` needs to be called via IPC like [here](https://github.com/cpvalente/ontime/blob/d6aef5ea316d2a2a82a8fac5ebd325f88ddf0ad3/apps/client/src/common/utils/linkUtils.js#L15) whereas setWindowOpenHandler catches all `window.open()` calls as well as links with `target=_blank`
Maybe it would make more sense to use the new one instead of the existing everywhere since we shouldn't need the if statement in the js anymore.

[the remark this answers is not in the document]
cpvalente: That is a good observation.

It can be true that we can make the app-wide decision that we will never want to open a link in Electron.
If so, we can use the new function, but we would then remove the old one.
We should not have both

apps/client/src/features/modals/sheets-modal/SheetsModal.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+  const [rundown, setRundown] = useState<OntimeRundown | null>(null);
+  const [userFields, setUserFields] = useState<UserFields | null>(null);
+  const [project, setProject] = useState<ProjectData | null>(null);
+
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const sheetid = useRef<HTMLInputElement>(null);
+  const worksheet = useRef<HTMLSelectElement>(null);
+
+  const { data: sheetState, refetch } = useQuery({
```
alex-Arc, line 57: i'm trying to push the id and worksheet values through the Query but can't get it to work
is it possible or should it be done differently

cpvalente: I am not sure I understand the question, lets chat!

apps/client/src/features/modals/sheets-modal/SheetsModal.tsx
```diff
@@ -0,0 +1,281 @@
...
+              </Step>
+
+              <Step title='3 - Add Document ID' completed={Boolean(sheetState?.id)} disabled={!sheetState?.auth}>
+                <HStack>
+                  <Input
+                    type='text'
+                    ref={sheetRef}
+                    id='sheetid'
+                    size='sm'
+                    variant='ontime-filled-on-light'
```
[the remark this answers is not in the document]
arymoraes: you could make this a controlled input, means you store the value in a useState, and change it using a handler function

like 

```
const [value, setValue] = useState(undefined);

const handleChange = (event) => {
  setValue(event.target.value)
}
```

then on the input

```
<Input
  onChange={handleChange}
  value={value}
/>
```

my guess for the current behaviour is maybe when it collapses, it un-renders the DOM node, meaning the value is also gone

apps/client/src/features/modals/sheets-modal/Step.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <div className={style.wrapper}>
+      <div className={style.header} onClick={handleCollapse}>
+        {completed ? (
+          <IoCheckmarkCircle className={style.step} style={{ color: 'green' }} />
+        ) : error ? (
+          <IoCloseCircle className={style.step} style={{ color: 'red' }} />
+        ) : (
+          <IoRadioButtonOffOutline className={style.step} />
+        )}
```
arymoraes, line 36: can swap this for

```
  const icon = useMemo(() => {
    if (completed) return <IoCheckmarkCircle className={style.step} style={{ color: 'green' }} />;

    if (error) return <IoCloseCircle className={style.step} style={{ color: 'red' }} />;

    return <IoRadioButtonOffOutline className={style.step} />;
  }, [completed, error]);
```

apps/server/src/utils/sheetsAuth.ts
```diff
@@ -0,0 +1,340 @@
...
+      !('client_id' in secrets['installed']) ||
+      !('project_id' in secrets['installed']) ||
+      !('auth_uri' in secrets['installed']) ||
+      !('token_uri' in secrets['installed']) ||
+      !('auth_provider_x509_cert_url' in secrets['installed']) ||
+      !('client_secret' in secrets['installed']) ||
+      !('redirect_uris' in secrets['installed'])
+    ) {
+      throw new Error('Sheet: Client secret is missing some keys');
+    }
```
arymoraes, line 229: could do something like

```
const requiredKeys = ['client_id', 'project_id', 'auth_uri', 'token_uri', 'auth_provider_x509_cert_url', 'client_secret', 'redirect_uris'];
const isKeyMissing = requiredKeys.some(key => !(key in secrets['installed']));

if (isKeyMissing) {
  throw new Error('Sheet: Client secret is missing some keys');
}
```

alex-Arc: Nice!

apps/client/src/features/modals/sheets-modal/SheetsModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+      <ModalContent>
+        <ModalHeader>Rundown from sheets</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <Alert status='info' variant='ontime-on-light-info'>
+            <AlertIcon />
+            <div style={{ display: 'flex', flexDirection: 'column' }}>
+              <AlertTitle>Sync with Google Sheets</AlertTitle>
+              <AlertDescription>
+                <ModalLink href='our-docs'>For more information, see the docs</ModalLink>
```
cpvalente, line 199: here is the link
https://ontime.gitbook.io/v2/features/google-sheet-experimental

apps/client/src/features/modals/sheets-modal/Step.tsx
alex-Arc: nice!

alex-Arc:
@hrueger  would you be able to test again https://github.com/cpvalente/ontime/releases/tag/v2.22.0-sheet-beta.3

hrueger:
Sure, will do that tomorrow evening 👍

apps/server/src/utils/sheetsAuth.ts
```diff
@@ -0,0 +1,382 @@
...
+    const appDataPath = getAppDataPath();
+    if (appDataPath === '') {
+      throw new Error('Sheet: Could not resolve sheet folser');
+    }
+    this.sheetsFolder = join(appDataPath, 'sheets');
+    this.clientSecretFile = join(this.sheetsFolder, 'client_secret.json');
+    ensureDirectory(this.sheetsFolder);
+    try {
+      const secrets = JSON.parse(readFileSync(this.clientSecretFile, 'utf-8'));
+      const isKeyMissing = this.requiredClientKeys.some((key) => !(key in secrets['installed']));
```
hrueger, line 51: This does not work for me. When uploading the Client ID (actually, the file is called client_secret_... ; maybe we should call it that, too, at least somewhere to not confuse people who don't read the docs? ;-) ), I get the following error: `Internal Server Error, 'TypeError: Cannot use 'in' operator to search for 'client_id' in undefined"`.
My secret file generated by google looks like this:
```json
{
    "web": {
        "client_id": "redacted.apps.googleusercontent.com",
        "project_id": "redacted",
        "auth_uri": "https://accounts.google.com/o/oauth2/auth",
        "token_uri": "https://oauth2.googleapis.com/token",
        "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
        "client_secret": "redacted",
        "redirect_uris": [
            "http://localhost"
        ]
    }
}
```

apps/server/src/utils/sheetsAuth.ts
```diff
@@ -0,0 +1,382 @@
...
+   * @description SETP 2 - test that the reciveed OAuth2 is still valid
+   * @throws
+   */
+  async testAuthentication() {
+    if (Sheet.client) {
+      const ref = await Sheet.client.refreshAccessToken();
+      if (ref.credentials.expiry_date > 10000) {
+        return true;
+      } else {
+        throw new Error('Unable to use accese token');
```
hrueger, line 200: _access_?

hrueger (commented):
Hi, great work, you two! I really like the new UI.
I added a couple of comments, mostly typos. I was not able to test everything, though, since I could not get past step 1 (see the respective comment).

Just two other things I noticed:
1. in the (really nice) documentation, the link [here](https://ontime.gitbook.io/v2/features/google-sheet-experimental#create-a-token-in-google-cloud-console) is broken, it resolves to `https://console.cloud.google.comhttps/console.cloud.google.com`.
2. also in the docs, in the second bullet point of the same box: it would be nice if `https://console.cloud.google.com/apis/credentials` was a clickable link, too

alex-Arc:
@hrueger are you not able to create a cloud project https://console.cloud.google.com?

hrueger:
I am, but I'm running into [this](https://github.com/cpvalente/ontime/pull/579#discussion_r1444047797)...

alex-Arc:
ahh we need to add this to the docs then, it has to be a desktop app
image

hrueger:
Yep, that works! Thanks.
Then in the docs, the bullet point `In the Authorised redirect URIs section, add a redirect to the localhost URL http://localhost.` can also go.

hrueger:
Now, after uploading the file, the modal looks like this:
grafik
It is not completely obvious what to do next, since the [cursor does not change to the hand when hovering over step ](https://github.com/cpvalente/ontime/pull/579#discussion_r1444047457) 2. Can we change that and maybe open the next step automatically? That would be great.

The other steps work great, though. Really nice work with the worksheet selection and the preview!
A very minor thing I'd like to see (this is probably a personal opinion only) would be to change the button in the last step from "Review" to "Preview". I think review sounds like it was already downloaded.

One last little thing: When pressing import, the modal closes, but it takes (for me) 10 seconds until events appear. Until that, the list is just blank. That might feel a bit concerning to the user ;-) Can we add a loading indicator or - if it is not too complicated - an actual progressbar here?

alex-Arc:
Prefect thanks for the thorough test, and the spellchecking :)
good not with the pointer/link hand
Preview sounds good


@cpvalente do you know if we can do something here? most likely a a separate PR
> One last little thing: When pressing import, the modal closes, but it takes (for me) 10 seconds until events appear. Until that, the list is just blank. That might feel a bit concerning to the user ;-) Can we add a loading indicator or - if it is not too complicated - an actual progressbar here?

hrueger:
Oh, one more thing. I've used a sheet for testing where I have read access only. The upload button still exists, but it does nothing. Can we detect if we don't have write access and disable the button in that case?

hrueger:
And one more 😅
I've just shown it to a friend, he also likes it. However, his first response was `Why is it hidden behind a calendar icon?` and I think that is a point. It seems like the io5 icons don't include a table icon. Maybe we should use that icon from a different icon pack? https://react-icons.github.io/react-icons/search/#q=table
not too important, though.

apps/client/src/features/modals/sheets-modal/SheetsModal.tsx
```diff
@@ -0,0 +1,463 @@
...
+
+  const queryClient = useQueryClient();
+
+  const [rundown, setRundown] = useState<OntimeRundown | null>(null);
+  const [userFields, setUserFields] = useState<UserFields | null>(null);
+  const [project, setProject] = useState<ProjectData | null>(null);
+
+  const [id, setSheetId] = useState('');
+  const [worksheet, setWorksheet] = useState('');
+  const [worksheetOptions, setWorksheetOptions] = useState(new Array<string>());
```
cpvalente, line 61:
```suggestion
  const [worksheetOptions, setWorksheetOptions] = useState<String[]>([]);
```

cpvalente (commented):
small changes, lets merge this!

PR #580, opened 2023-11-09
External field

alex-Arc:
set-external-message-text -> to set the message
set-external-message-visible -> to change the visibility

todo:
- [x] porper css
- [x] implement on all relevant pages
- [x] add section to messagecontrol
- [x] implement rate limiter


apps/server/src/services/message-service/MessageService.ts
```diff
@@ -36,9 +39,42 @@ class MessageService {
...
+
+  /**
+   * @description sets message on stage timer screen
+   */
+  setExternalText(payload: string) {
+    if (this.externalMessage.text !== payload) {
+      this.externalMessage.text = payload;
+      if (!this.reateLimit) {
+        eventStore.set('externalMessage', this.externalMessage);
+        this.reateLimit = setTimeout(this._OnTimeOut.bind(this), 1000);
```
alex-Arc, line 63: @cpvalente what do you think of this as rate limiting?

alex-Arc: hmm the UI text fields don't really work with this method as they send an update on each key press

cpvalente: Few points:
- we want to always have the latest information in the socket, with this logic, we may not send the data if we are in timeout period
- it seems that the rate limiting needs to be a bit less ambitious, maybe around 64ms? we need to test with the UI

Take a look into debouncing code, pretty similar to what you have, but should ensure that we always send the latest call

apps/server/src/stores/EventStore.ts
```diff
@@ -20,14 +20,16 @@ export const eventStore = {
...
     store[key] = value;
     // TODO: Partial updates seems to cause issues on the client
     // socket.send({
     //   type: `ontime-${key}`,
     //   payload: value,
     // });
-    this.broadcast();
+    if (!defere) {
+      this.broadcast();
+    }
```
cpvalente, line 32: I think adding this optional flags are usually a bad idea.
I dont think we need this and rather build debouncing where necessary. Including here

alex-Arc:
there are some problems with the scaling, it changes a lot depending on the amount of text
image
image
image

alex-Arc:
I like the external message being read only in the editor 
but maybe the visibility toggle should be allowed?

apps/client/src/features/control/message/MessageControl.tsx
```diff
@@ -59,6 +61,17 @@ export default function MessageControl() {
...
+        <InputRow
+          label='External Message'
+          placeholder='-'
+          readonly={true}
+          text={data.externalMessage.text || ''}
+          visible={data.externalMessage.visible || false}
+          changeHandler={(newValue) => setMessage.externalText(newValue)}
+          actionHandler={() => setMessage.externalVisible(!data.externalMessage.visible)}
+        />
+      </div>
```
cpvalente, line 74: I am not sure if we should be setting this at all here. I thought we agreed that this was only for visibility and we wouldnt have controls in our interface?

apps/client/src/features/control/message/MessageControl.module.scss
```diff
@@ -31,3 +31,27 @@
...
+  overflow: hidden;
+  text-overflow: ellipsis;
+  color: $label-gray;
+  padding-left: 0.5rem;
+}
+
+.inlineIcon {
+  font-size: 1rem;
+  padding: 0 0.5rem;
+}
```
cpvalente, line 57: if we decide not to use these styles, we should remove the CSS

PR #581, opened 2023-11-09
View Params Persist when Navigating Back

### LINKED ISSUE #503: Remember view settings on navigate
## The issue
We currently use the URL params as a persistence layer for the view settings

However, when the user navigates through the menu, the URL looses the params and, with it, all the hard work config the user has made

## The task
The issue task is to create a second persistence layer for settings.

A possible implementation suggestion:
1. the user clicks save in the `ViewParamsEditor`, we add the same object used to create the params in sessionStorage.
2. the user navigates, we check if there is a current object in session storage and apply that to URL params
3. if the user clicks clear in `ViewParamsEditor` we also clear the session storage object

### Relevant (existing) code
https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx#L48

asharonbaltazar:
closes #503 

- add local storage to persist saved views

### DIFF
--- a/apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
+++ b/apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
@@ -12,10 +12,26 @@ import {
   useDisclosure,
 } from '@chakra-ui/react';
 
+import { useLocalStorage } from '../../../common/hooks/useLocalStorage';
+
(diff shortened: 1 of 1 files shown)


cpvalente (approved):
Great work @asharonbaltazar
🚀

PR #583, opened 2023-11-10
add eventsPerPage to schedule

alex-Arc:
image
image

### DIFF
--- a/apps/client/src/common/components/view-params-editor/constants.ts
+++ b/apps/client/src/common/components/view-params-editor/constants.ts
@@ -210,8 +210,8 @@ export const BACKSTAGE_OPTIONS: ParamField[] = [
   },
   {
     id: 'eventsPerPage',
-    title: 'Number of events on page',
-    description: '',
+    title: 'Events per page',
+    description: 'Sets the number of events on the page, can cause overlow',
     type: 'number',
   },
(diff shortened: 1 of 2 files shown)


apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -224,6 +230,12 @@ export const PUBLIC_OPTIONS: ParamField[] = [
     description: 'Schedule will not auto-cycle through events',
     type: 'boolean',
   },
+  {
+    id: 'eventsPerPage',
+    title: 'Number of events on page',
+    description: '',
```
cpvalente, line 236: Just for sake of consistency we can use the title and description fields
The description field can also warn users that the events will overflow the page

cpvalente (commented):
If the user adds too many events in a column, they will see some overflow. What does this look like? could we add some styling to make sure it doesn't look terrible?

Why only on the Public and not in the Backstage screens?

alex-Arc:
I tied looking in to some overflow effects but it is beyond my ability
https://amorriscode.github.io/use-overflow/

the clipping doesn't look to bad 😉 
image

cpvalente (approved):
Looks good, thank you!

Issue #535, opened 2023-10-07
Roll mode: SMPTE/MTC Timecode


MilkMGN:
Would it be simple to add to use an audio input and take in timecode to base events off in roll mode? 

Could even then use the system clock and figure out what time events are going to happen within the current timecode cycle

cpvalente:
Hi @MilkMGN , thank you for reaching out.

Adding some sort of timecode integration is something we would like do, we also we have a similar request for adding a MTC output from ontime time #227 

However there is some investigation to be done, and several caveats to consider. I would welcome help on this task.

Personally, I would prefer a MTC or some network time integration, it would be interesting to see what companion does and which protocol would have the least friction in integration

MilkMGN:
Hey @cpvalente , 

I have a friend willing to look into making a branch for this, whether through SMPTE or MTC. For my event i'm planning to use this in we need SMPTE, so we may chip away at that. 

Thanks for your reply, i look forward to possibly making this work.

je71175:
Also if you want to look at another external platform Chataigne is a great intergrator for multiple platforms and systems.  http://benjamin.kuperberg.fr/chataigne/en#download
Have been using external platforms with OSC to trigger onTime events.
But I do like the idea of having TimeCode (we would use MTC) native in onTime.  
cheers Jason

alex-Arc:
@pbl4845, @je71175, @MilkMGN 
Hey all sorry to ping you like this
I am working on a MTC prototype and could use some help testing if you are stile interested

alex-Arc:
@viktorfreiman there is also a NTP option

viktorfreiman:
Hi guys, I have just got LTC from Audio device working. Here is a working demo. https://github.com/viktorfreiman/audio-demo
I tried to get it to work with ontime but I had some problems with getting pnpm and turbo working.
If anyone is up to rework it and getting it upstream would be awesome.

@MilkMGN I am also doing a show with LTC.
I working with lights so my fallback to get it working is to use my light console, grandMA3's LTC input and then send OSC cues to bitfocus companion to have it track with ontime.

je71175:
happy to help with testing. 

> @pbl4845, @je71175, @MilkMGN Hey all sorry to ping you like this I am working on a MTC prototype and could use some help testing if you are stile interested

alex-Arc:
> Hi guys, I have just got LTC from Audio device working. Here is a working demo. https://github.com/viktorfreiman/audio-demo 

[...]

PR #584, opened 2023-11-10
Feat clock source

alex-Arc:
- [x] Offset
- [x] Status
- [x] https://github.com/cpvalente/ontime/issues/535#issuecomment-1817700469 can't roll back #604


cpvalente:
Thank you @alex-Arc 

We have to have a more hands off process on our side for external sync. In short, we need to see the deviation from us to an external clock and try and calculate time on our side, syncing every so often with the server.

Before getting into the details of the structure I would prefer seeing some feedback about the feature and what it means for Ontime.

alex-Arc:
I will start over with a different approche

PR #588, opened 2023-11-12
configure lint staged

cpvalente:
Since we are having some more developers contributing here. It has been brought to my attention that we should improve tooling 

@asharonbaltazar and @alex-Arc , let me know what you think


apps/server/src/adapters/websocketAux.ts
```diff
@@ -4,7 +4,7 @@ import { socket } from './WebsocketAdapter.js';
  * Utility function to notify clients that the REST data is stale
  * @param payload -- possible patch payload
  */
-export function sendRefetch(payload: any | null = null) {
+export function sendRefetch(payload: unknown | null = null) {
```
asharonbaltazar, line 7: Super small not: doesn't `unknown` already contain `null`?

asharonbaltazar (approved):
Love the changes!

asharonbaltazar:
For those using VS Code, you can also [recommend extensions](https://code.visualstudio.com/docs/editor/extension-marketplace) to install

cpvalente:
> Love the changes!

Great to hear. Let me know if there is anything else you would like to see when it comes to tooling

alex-Arc (approved):

PR #590, opened 2023-11-14
refactor: better errors

cpvalente:
Small PR to add consistency to error messages from backend

### DIFF
--- a/apps/server/src/modules/loadDb.ts
+++ b/apps/server/src/modules/loadDb.ts
@@ -67,7 +67,7 @@ async function loadDb() {
   return { db, data };
 }
 
-export let db = {};
+export let db = {} as Low<DatabaseModel>;
 export let data = {} as DatabaseModel;
 export const dbLoadingProcess = loadDb();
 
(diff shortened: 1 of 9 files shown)


apps/server/src/modules/loadDb.ts
```diff
@@ -67,7 +67,7 @@ async function loadDb() {
   return { db, data };
 }
 
-export let db = {};
+export let db = {} as Low<DatabaseModel>;
```
asharonbaltazar, line 70: Does

```ts
export let db: Low<DatabaseModel> = {};
```

work?

asharonbaltazar (approved):
Left a question

apps/server/src/modules/loadDb.ts
cpvalente: not really, we are cheating here and the code is less than ideal

we export a mutable variable (uncool) db. and then have an init sequence to create the lowdb object
this is to avoid race conditions on starting, we need to make sure the database gets initialised at a specific test

we could instead do 
`export let db = Low<DatabaseModel> | null = null`
but then we would need to pepper the whole code with nullchecks 

what do you think?

asharonbaltazar: Not worth it. Resolving.

PR #595, opened 2023-11-16
Local build

### LINKED ISSUE #510: Ontime window does not launch in local builds
When the app is packaged inside electron, it is electron's responsibility to start all the processes.
From electron's perspective, it goes something like

1. Open preload window
2. Start Ontime server (returns a port indicating where to find the server)
3. Close the preload window and open a window with the UI build served at the given port

It seems that the app gets stuck somewhere between step 2 and 3.
The server process is started by the preload window is never closed but the server is available

It is puzzling that the pipeline builds work well.
The only apparent difference is the inclusion of the local environment variable

This bug was likely introduced by the changes in #476 

We are likely calling the backend at the wrong place. A good place to start is the handling of the backend start call
https://github.com/cpvalente/ontime/blob/master/apps/electron/main.js#L159

alex-Arc:
fixes #510
the issue comes from build:local in apps/server builds' index and not app that is needed for electron dist

needs testing on mac and linux
- [x] Mac os
- [ ] Linux
- [x] Win

## Summary by CodeRabbit

- **Documentation**
	- Updated the build command instructions for UI and server in the development guide to reflect the new process for generating project distributions.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -76,7 +76,7 @@ You can generate a distribution for your OS by running the following steps.
 From the project root, run the following commands
 
 - __Install the project dependencies__ by running `pnpm i`
-- __Build the UI and server__ by running `turbo build:local`
+- __Build the UI and server__ by running `turbo build:electron`
 - __Create the package__ by running `turbo dist-win`, `turbo dist-mac` or `turbo dist-linux`
 
 The build distribution assets will be at `.apps/electron/dist`
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -49,6 +49,7 @@
     "dev:test": "cross-env IS_TEST=true nodemon --exec \"ts-node-esm\" ./src/index.ts",
     "prebuild": "pnpm setdb",
     "build": "pnpm prebuild && esbuild src/app.ts --log-level=error --platform=node --format=cjs --bundle --minify --outfile=dist/index.cjs",
+    "build:electron": "pnpm prebuild && esbuild src/app.ts --log-level=error --platform=node --format=cjs --bundle --minify --outfile=dist/index.cjs",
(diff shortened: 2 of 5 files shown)


DEVELOPMENT.md
```diff
@@ -76,7 +76,7 @@ You can generate a distribution for your OS by running the following steps.
 From the project root, run the following commands
 
 - __Install the project dependencies__ by running `pnpm i`
-- __Build the UI and server__ by running `turbo build:local`
+- __Build the UI and server__ by running `turbo build:electron`
```
cpvalente, line 79: I am unsure on what I missed, but the `build:electron` scripts seem like copies of `build:local`

I am unsure on how anything is different here?

apps/server/package.json
```diff
@@ -49,6 +49,7 @@
     "dev:test": "cross-env IS_TEST=true nodemon --exec \"ts-node-esm\" ./src/index.ts",
     "prebuild": "pnpm setdb",
     "build": "pnpm prebuild && esbuild src/app.ts --log-level=error --platform=node --format=cjs --bundle --minify --outfile=dist/index.cjs",
+    "build:electron": "pnpm prebuild && esbuild src/app.ts --log-level=error --platform=node --format=cjs --bundle --minify --outfile=dist/index.cjs",
```
alex-Arc, line 52: this is the "big" difference
building `src/app.ts` and not `src/index.ts` 

I made a separate electron script to simplify the build to one turbo command but maybe there is a better way?

cpvalente: oh I see, but then do we still need the `build:local` or is just a broken script? could the new script replace `build:local` instead?

alex-Arc: if one wants to to run the apps/sever an apps/electron separate then  `local`  is the one that is needed

cpvalente (approved):
Good work

PR #597, opened 2023-11-16
refactor: rundown

cpvalente:
This PR is mostly for technical improvements:
- Upgrade react-query to v5
- add `cachedRundown` endpoint which returns a rundown and a revision number, this endpoint is only for app consumption and we can in future extend it to give us a normalised rundown if necessary

I have tried using the revision indicator in the useQuery hook to be able to tell if the data has changed, the logic seems to work but I saw no difference in behaviour from react query (code still in `useRundown.tsx`)

Either way, the addition of the revision seem to have stabilised the rundown re-renders, but it seem to come at cost of the entire rundown being compared

Any of the operations on rundown have the potential to be broken
- Adding, editing, deleting


alex-Arc (approved):

apps/client/src/features/rundown/RundownEntry.tsx
```diff
@@ -100,7 +100,8 @@ export default function RundownEntry(props: RundownEntryProps) {
         }
         case 'clone': {
           const newEvent = cloneEvent(data as OntimeEvent, data.id);
-          newEvent.cue = getCueCandidate(ontimeQueryClient.getQueryData(RUNDOWN_TABLE) || [], data.id);
+          const rundown = ontimeQueryClient.getQueryData<GetRundownCached>(RUNDOWN)?.rundown ?? []
```
asharonbaltazar, line 103: Would it be possible to just do:
```ts
const { rundown = [] } = ontimeQueryClient.getQueryData<GetRundownCached>(RUNDOWN)
```

apps/client/src/common/hooks-query/useRundown.ts
```diff
@@ -1,19 +1,29 @@
...
     retryDelay: (attempt) => attempt * 2500,
     refetchInterval: queryRefetchInterval,
     networkMode: 'always',
+    // structuralSharing: (oldData: GetRundownCached | undefined, newData: GetRundownCached) => {
+    //   if (oldData === undefined) {
+    //     cachedRundownPlaceholder;
+    //   }
+    //   const hasDataChanged = oldData?.revision === newData.revision;
+    //   return hasDataChanged ? oldData : newData;
+    // },
```
asharonbaltazar, line 27: You planning on leaving this in?

PR #604, opened 2023-11-20
trigger doRoll if clock is less than startAt

alex-Arc:
this is needed to have external clock sources like MTC behave correctly if they jump backwards in time

### DIFF
--- a/apps/server/src/services/rollUtils.ts
+++ b/apps/server/src/services/rollUtils.ts
@@ -153,6 +153,7 @@ type CurrentTimers = {
   selectedEventId: string | null;
   current: number | null;
   _finishAt: number | null;
+  _startAt: number | null;
   clock: number | null;
   secondaryTimer: number | null;
   secondaryTarget: number | null;
@@ -164,7 +165,7 @@ type CurrentTimers = {
  * @returns {object} object with selection variables
  */
 export const updateRoll = (currentTimers: CurrentTimers) => {
-  const { selectedEventId, current, _finishAt, clock, secondaryTimer, secondaryTarget } = currentTimers;
+  const { selectedEventId, current, _finishAt, _startAt, clock, secondaryTimer, secondaryTarget } = currentTimers;
 
   // timers
   let updatedTimer = current;
@@ -182,9 +183,13 @@ export const updateRoll = (currentTimers: CurrentTimers) => {
       updatedTimer -= dayInMs;
     }
 
+    console.log(Math.floor(_startAt / 1000), Math.floor(clock / 1000));
+
     if (updatedTimer < 0) {
       isPrimaryFinished = true;
       // we need a new event
+      doRollLoad = true;
+    } else if (clock < _startAt) {
       doRollLoad = true;
     }
   } else if (secondaryTimer >= 0) {
--- a/apps/server/src/services/TimerService.ts
+++ b/apps/server/src/services/TimerService.ts
@@ -339,7 +339,7 @@ export class TimerService {
         this.timer.expectedFinish >= this.timer.startedAt
           ? this.timer.expectedFinish
           : this.timer.expectedFinish + dayInMs,
-
+      _startAt: this.timer.startedAt,
       clock: this.timer.clock,
       secondaryTimer: this.timer.secondaryTimer,
       secondaryTarget: this.secondaryTarget,
--- a/apps/server/src/services/__tests__/rollUtils.test.ts
+++ b/apps/server/src/services/__tests__/rollUtils.test.ts
@@ -703,4 +703,24 @@ describe('typical scenarios', () => {
 
     expect(updateRoll(timers)).toStrictEqual(expected);
   });
+
+  it('rolls backwards', () => {
+    const timers = {
+      selectedEventId: '1',
+      current: 11,
+      _startAt: 10,
+      _finishAt: 15,
+      clock: 9,
+      secondaryTimer: null,
+      secondaryTarget: null,
+    };
+
+    const expected = {
+      updatedTimer: timers._finishAt - timers.clock,
+      updatedSecondaryTimer: null,
+      doRollLoad: true,
+      isFinished: false,
+    };
+    expect(updateRoll(timers)).toStrictEqual(expected);
+  });
 });


cpvalente (commented):
I wonder if this would complicate the integrations a bit.

Thinking out loud:
The next issue might be in the fast forward. 
If the user pushes the timecode forwards, we might need to ask to load another event
Should we then trigger the finished flag?

I would argue not, since the event didnt actually finish


What if, before making the `updateRoll` call, we heck the difference between the time it was previously called and the timeNow.
If that difference is negative (rollback) or large enough that it would finish the event, we call the `loadRoll` instead of the `updateRoll`

From a quick look, it seems that all of this is already available in the `update` function

What do you think?

alex-Arc:
yep, that all makes sense 
what should we set as the skip/jump limit?
it need to be big enough that it doesn't happen by accident and small enough to not hinder an actual attempt to jump

as the timer runs in 32ms time steps maybe 5x32ms would be appropriate?

cpvalente:
> yep, that all makes sense what should we set as the skip/jump limit? it need to be big enough that it doesn't happen by accident and small enough to not hinder an actual attempt to jump
> 
> as the timer runs in 32ms time steps maybe 5x32ms would be appropriate?

that sounds sensible to me. 
I would be tempted to start with a more conservative difference like 3 x 32ms. we should extract this into a config variable so we can tweak without digging too deep into the code

apps/server/src/services/rollUtils.ts
```diff
@@ -183,9 +186,12 @@ export const updateRoll = (currentTimers: CurrentTimers) => {
     }
 
     if (updatedTimer < 0) {
-      isPrimaryFinished = true;
+      if (Math.abs(updatedTimer) < config.rollSkipLimit) isPrimaryFinished = true; //Dont trigger Finished if we are over the skip limit
```
cpvalente, line 189: maybe try to use the code in a way that it is self documenting. like
```ts
      const hasSkipped = Math.abs(updatedTimer) > config.rollSkipLimit;
      if (!hasSkipped) {
        isPrimaryFinished = true;
      }
```

apps/server/src/services/rollUtils.ts
```diff
@@ -164,8 +166,9 @@ type CurrentTimers = {
  * @returns {object} object with selection variables
  */
 export const updateRoll = (currentTimers: CurrentTimers) => {
-  const { selectedEventId, current, _finishAt, clock, secondaryTimer, secondaryTarget } = currentTimers;
+  const { selectedEventId, current, _finishAt, _startAt, clock, secondaryTimer, secondaryTarget } = currentTimers;
 
+  console.log(currentTimers);
```
cpvalente, line 171: remember to check for debug stuff at some point before you are done

apps/server/src/services/rollUtils.ts
```diff
@@ -153,6 +154,7 @@ type CurrentTimers = {
   selectedEventId: string | null;
   current: number | null;
   _finishAt: number | null;
+  _startAt: number | null;
```
cpvalente, line 157: I might be wrong, but I dont think we need to get this far to catch a skip.
in the `update` method of the `TimerService` we have the clock

before setting the new clock, we can check if the timer has skipped. 
we already have the current logic 
```ts
    const previousTime = this.timer.clock;
    this.timer.clock = clock.timeNow();
    if (previousTime > this.timer.clock) {
      force = true;
    }
```

which we can modify to include our logic (this is just an example and will likely not work, we would want something more intelligent)
```ts
    const hasSkipped = Math.abs(previousTime - this.timer.clock) > config.rollSkipLimit;
    if (hasSkipped) {
      // maybe we can even check here if the skip is larger than the duration of the event?
      // we can skip within the event
      PlaybackService.roll();
      force = true;
    }
```

apps/server/src/services/rollUtils.ts
```diff
@@ -183,9 +186,12 @@ export const updateRoll = (currentTimers: CurrentTimers) => {
     }
 
     if (updatedTimer < 0) {
```
cpvalente, line 188: if we do end up with this logic, there is no need for the extra else if
```ts
    if (updatedTimer < 0 || clock < _startAt) {
      const hasSkipped = Math.abs(updatedTimer) > config.rollSkipLimit;
      if (!hasSkipped) {
        isPrimaryFinished = true;
      }
      // we need a new event
      doRollLoad = true;
    }
```

cpvalente (commented):
few comments, feel free to reach out

apps/server/src/services/rollUtils.ts
alex-Arc: we could 
but it looks a bit like we would be duplicating  almost all the functionality of  `updateRoll` to know if we skipped out of an event 
is that preferable? or maybe empty out `updateRoll`

cpvalente: I am unsure on why this would be the case.
The code above likely works. the optimisation I mentioned about the skipping would also only be a few lines of code

like
```ts
    const skipTime = previousTime - this.timer.clock;
    const hasSkipped = Math.abs(skipTime) > config.rollSkipLimit;

    if (hasSkipped) {
      if (skipTime > 0 && skipTime > this.timer.current) {
        // we skipped out of the event to the next one
        PlaybackService.roll();
      } else if (skipTime < 0 && skipTime < this.timer.elapsed) {
        // we might have skipped outside of the event
        PlaybackService.roll();
      }

      // otherwise we just skipped within the event
      force = true;
    }
```

I may be wrong, in which case feel free to ignore. We definitely dont want to duplicate the logic of updateRoll here

alex-Arc: can we write tests for `TimerService.ts`? That is the good thin about having it in `rollUtils.ts`
it would end up somthing like this right?
```typescript
//TimerService.ts
  update(force = false) {
    const previousTime = this.timer.clock;
    this.timer.clock = clock.timeNow();
    if (previousTime > this.timer.clock) {
      force = true;
    }
    const skipTime = previousTime - this.timer.clock;
    const hasSkipped = Math.abs(skipTime) > config.rollSkipLimit;

    // we call integrations if we update timers
    let shouldNotify = false;
    if (this.playback === Playback.Roll) {
      shouldNotify = true;
      if (hasSkipped) {
        if (skipTime > 0 && skipTime > this.timer.current) {
          // we skipped out of the event to the next one
          PlaybackService.roll();
        } else if (skipTime < 0 && skipTime < this.timer.elapsed) { 
          // we might have skipped outside of the event
          PlaybackService.roll();
        } else {
          // otherwise we just skipped within the event
          this.updateRoll();
        }
      } else {
        this.updateRoll();
      }
    } else if (this.timer.startedAt !== null) {
      // we only update timer if a timer has been started
      shouldNotify = true;
      this.updatePlay();
    }
``` 
am I wrong in thinking that `} else if (skipTime < 0 && skipTime < this.timer.elapsed) { ` 
is meant to check if we have skipped backwards out of an event
and should be `} else if (skipTime < 0 && Math.abs(skipTime) > this.timer.elapsed) { `

cpvalente: we can absolutely write tests and you are encouraged to do so.
having said that, the current architecture of this service makes it hard to test. I am hoping to change some of that during the new work.

For now we extract some of the core calculations in `rollUtils` and `timerUtils` for this exact issue.

As for the logic, yes, you are correct.
skip time can be negative, but the elapsed time can only ever be positive
Keep in mind the code I wrote here was just a quick and rough example, it is untested and I did not considered any edge cases

alex-Arc:
there are now test that cover:
- normal roll out of event
- normal roll backwards out of event
- normal roll out of event over midnight
- normal roll backwards out of event over midnight
- skip out of event
- skip backwards out of event
- skip out of event over midnight
- skip backwards out of event over midnight

Am i missing anything and do the tests look sane?

cpvalente:
Hi @alex-Arc , great work

I have made a few changes, let me know if you disagree and we can discuss and roll back. 
In short:

- config is in the domain of timer, we dont want to have it in the application config (also saves us some issues with import order)
- simplify and document logic
- tests are not tied to a skip time, but to the idea that a difference larger than what we allow, is considered a skip

cpvalente (approved):

PR #605, opened 2023-11-20
feat: show progress in op and cuesheet

### LINKED ISSUE #130: Color bar - countdown time & tables view enhancement
Hi
Would it be possible to have a colour bar count down timer - Green Amber Red.   Ideally under the timer displays in UI, back stage and tables views.

For the Table, we would like the green highlight bar to have a different behaviour than current.  Respecting the Green-Amber-Red countdown bar running on the current cue - the highlight bar moves to the next cue event as soon as the current event timer starts.  I think a different highlight color other than green should be used

For example, cue 3 is the current cue and has a Green-Amber-Red bar showing this cue timer is running
and cue 4 is highlighted as the next cue.  Similar to the markup logic applied to the studio clock.

Screenshot 2022-05-22 at 8 48 07 AM

### LINKED ISSUE #376: Countdown bar and highlight cue enhancements
Countdown bar - change the current progress bar to traffic lights status Green / Amber / Red countdown bar as the current running timer approaches zero ( Green > 30s Amber 30-11 Red < 11s).  Include this on the cue sheets and editor views for the current running cue

Highlighted cue.  Sugguest to include a user setting, to highlight the current running cue, or the next cue.  When calling a show, focus is on what is coming up next more than what is currently happening.  In combination with the traffic light status this draws focus at the needed time. 
In the current cue sheet design, i find that if the cue has color, this has priority over the highlight.  Is it possible to change the highlight to the row borders and overwrite the row color?  

Slide1

Slide2

The highlight current in the editor is fine from my perspective, but the follow selected behaviour here and highlight current/next functionality would be great

for the follow selected behaviour, this could be brought from the center of the screen closer to the top of the screen - seldom need to refer to executed cues, and more valueable to use the screen realestate to see upcoming cues. 

thanks

### LINKED ISSUE #517: Add progress bar to operator view
In the operator view, we would like to show a progress bar in the currently running event
The design of this progress bar should be done as to avoid any layout shifts during the runtime

cpvalente:
Shows progress bar in both operator and cuesheet views

Closes #517 
Closes #376 
Closes #130

### DIFF
--- a/apps/client/src/features/cuesheet/cuesheet-progress/CuesheetProgress.tsx
+++ b/apps/client/src/features/cuesheet/cuesheet-progress/CuesheetProgress.tsx
@@ -1,0 +1,24 @@
+import MultiPartProgressBar from '../../../common/components/multi-part-progress-bar/MultiPartProgressBar';
+import { useTimer } from '../../../common/hooks/useSocket';
+import useViewSettings from '../../../common/hooks-query/useViewSettings';
+
+import styles from "./CuesheetProgress.module.scss"
+
+export default function CuesheetProgress() {
+  const { data } = useViewSettings();
+  const timer = useTimer();
+  const totalTime = (timer.duration ?? 0) + (timer.addedTime ?? 0);
+
+  return (
+    <MultiPartProgressBar
+      now={timer.current ?? 0}
+      complete={totalTime}
+      normalColor={data!.normalColor}
+      warning={data!.warningThreshold}
+      warningColor={data!.warningColor}
+      danger={data!.dangerThreshold}
+      dangerColor={data!.dangerColor}
+      className={styles.progressOverride}
+    />
+  );
+}
--- a/apps/client/src/features/cuesheet/cuesheet-progress/CuesheetProgress.tsx
+++ b/apps/client/src/features/cuesheet/cuesheet-progress/CuesheetProgress.tsx
@@ -11,7 +11,7 @@ export default function CuesheetProgress() {
 
   return (
     <MultiPartProgressBar
-      now={timer.current ?? 0}
+      now={timer.current}
       complete={totalTime}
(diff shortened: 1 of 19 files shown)


cpvalente:
@asharonbaltazar , would you mind taking a look here?
After reviewing the changes, I wonder whether <MultiStepProgressBar> should fetch the view changes directly?

The reason why this is passed in is to avoid storming the caching layer with requests. The nature of this component is for it to re-render several times a second

apps/client/src/features/cuesheet/cuesheet-progress/CuesheetProgress.tsx
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <MultiPartProgressBar
+      now={timer.current}
+      complete={totalTime}
+      normalColor={data!.normalColor}
+      warning={data!.warningThreshold}
+      warningColor={data!.warningColor}
+      danger={data!.dangerThreshold}
+      dangerColor={data!.dangerColor}
+      className={styles.progressOverride}
```
asharonbaltazar, line 21: Are these null assertions a good idea?

asharonbaltazar (approved):

alex-Arc (approved):

apps/client/src/features/cuesheet/cuesheet-progress/CuesheetProgress.tsx
cpvalente: I am sure there are better ideas

I had left a question for you related to this

> After reviewing the changes, I wonder whether should fetch the view changes directly?

The reason why this is passed in is to avoid storming the caching layer with requests. The nature of this component is for it to re-render several times a second

Depending on the approach we go here, we could move the request to the component that needs it or just type the hook properly since the data cannot be undefined

PR #607, opened 2023-11-21
feat: customise timer elements

### LINKED ISSUE #589: Improvement: Timer view make all non-essential things optional
Would be great to have everything that's not the timer to be optional. 
This would mean being able to disable the clock, progress bar, upcoming and current section's via URL parameters and the options menu.

cpvalente:
Allows customisation of the timer view
Closes #589

### DIFF
--- a/apps/client/src/common/components/view-params-editor/constants.ts
+++ b/apps/client/src/common/components/view-params-editor/constants.ts
@@ -70,7 +70,33 @@ export const CLOCK_OPTIONS: ParamField[] = [
   },
 ];
 
-export const TIMER_OPTIONS: ParamField[] = [TIME_FORMAT_OPTION];
+export const TIMER_OPTIONS: ParamField[] = [
+  TIME_FORMAT_OPTION,
+  {
+    id: 'hideClock',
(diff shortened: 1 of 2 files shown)


kellhogs (approved):
I like what i see here. I think for the timer we have every option covered we need.
Just one quick thought. IF we extend on this with the options available in the minimal timer view, we could merge these two together and get rid of one.

What do you think?

cpvalente:
> IF we extend on this with the options available in the minimal timer view, we could merge these two together and get rid of one

This is a good thought.

I always saw this as different things, and I think normal users would appreciate having the ready made views
Having said this, I can imagine that now users will want to start configuring the colour of the display in Timer since they can reduce it now to a simple timer.

Not sure, maybe another time?
What do you think @alex-Arc

alex-Arc:
consolidating them seams like a good idea to me. And as you say having it default to the "normal" timer.
It would end up being a lot of settings so if they are consolidated, a divider between the styling and hide/show option would go a long way to make it quick to understand. Consolable sections would be even better, maybe?

adding the toggle for the new external field would also be good

PR #609, opened 2023-11-21
feat: allow editing field from operator

### LINKED ISSUE #516: Edit user fields from operator view
### Context
The `/operator` view is a readonly list of the rundown.
In addition, the user can subscribe to a `userField`, which will be displayed along with the event information

We would like to add the ability for the users to update the data in this user field.

### The task
The suggested implementation is to add a press-and-hold action on the `OperatorEvent` entry that shows a modal with a text input. Submitting a value here should update the relevant user field

We would also like a way to improve visibility of the feature, perhaps we can show a message on scroll "press-and-hold to edit userfield", this should only be shown while the user is scrolling and  if there is a subscribed user field

cpvalente:
Quick PR to implement editing from user fields
Did not do too much testing yet, the values for long press and the show and hide of the overlays can maybe do with tweaking.
It would be great to get some feedback on its usability 

Closes #516

### DIFF
--- a/apps/client/src/common/hooks/useLongPress.tsx
+++ b/apps/client/src/common/hooks/useLongPress.tsx
@@ -1,0 +1,76 @@
+import { useMemo, useRef } from 'react';
+
+type LongPressOptions = {
+  threshold?: number;
+  onStart?: (e: Event) => void;
+  onFinish?: (e: Event) => void;
+  onCancel?: (e: Event) => void;
+};
+
+type LongPressFns = {
+  onMouseDown: (e: React.MouseEvent) => void;
+  onMouseUp: (e: React.MouseEvent) => void;
+  onMouseLeave: (e: React.MouseEvent) => void;
+  onTouchStart: (e: React.TouchEvent) => void;
+  onTouchEnd: (e: React.TouchEvent) => void;
+};
+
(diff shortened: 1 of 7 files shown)


alex-Arc:
very nice 👍🏽 
could there be a tool tip or cursor change of some kind to let the user know that it is possible?

alex-Arc:
ah yes, i do now. Maybe it should show on moves move like the navigation buttons?
(I'm probably not the only one that will miss it)
also I think the modal color should differ a bit from the event color
image

alex-Arc (approved):

cpvalente:
hey @alex-Arc , could you take another look for me please?

alex-Arc:
> hey @alex-Arc , could you take another look for me please?

Looks very nice!!

3 small things. None of these are deal breakers and I think it is fine merge as is:
- I would still love to be able to close the modal with `esc`
- Showing the hint on mouse movement instead of scroll would be more clear
- when there are no prior data in the user filed it is a bit unclear that the whole event area is clickable, maybe a hand pointer type?
image

cpvalente:
Thank you @alex-Arc 
This views are meant for mobile and tablets so touch is the way.

If users start doing strange things with it we can take a look

PR #610, opened 2023-11-23
Default Values in `ParamEditor`

asharonbaltazar:
- add `defaultValue` to `[Union]Field` types
- add `defaultValue`s to constants
- turn several constants into functions to recieve `TimeFormat`
- use `defaultValue`s in `ParamInput`
- update `Clear` to `Reset` to convey the default value resets

### DIFF
--- a/apps/client/src/common/components/view-params-editor/ParamInput.tsx
+++ b/apps/client/src/common/components/view-params-editor/ParamInput.tsx
@@ -11,14 +11,19 @@ interface EditFormInputProps {
 
 export default function ParamInput({ paramField }: EditFormInputProps) {
   const [searchParams] = useSearchParams();
-  const { id, type } = paramField;
+  const { id, type, defaultValue } = paramField;
 
   if (type === 'option') {
     const optionFromParams = searchParams.get(id);
-    const defaultOptionValue = optionFromParams || undefined;
+    const defaultOptionValue = optionFromParams || defaultValue;
 
     return (
-      <Select placeholder='Select an option' variant='ontime' name={id} defaultValue={defaultOptionValue}>
+      <Select
+        placeholder={defaultValue ? undefined : 'Select an option'}
+        variant='ontime'
+        name={id}
+        defaultValue={defaultOptionValue}
+      >
         {Object.entries(paramField.values).map(([key, value]) => (
           <option key={key} value={key}>
             {value}
@@ -36,12 +41,12 @@ export default function ParamInput({ paramField }: EditFormInputProps) {
   }
 
   if (type === 'number') {
-    const defaultNumberValue = searchParams.get(id) ?? '';
+    const defaultNumberValue = searchParams.get(id) ?? defaultValue;
 
     return <Input type='number' step='any' variant='ontime-filled' name={id} defaultValue={defaultNumberValue} />;
   }
 
-  const defaultStringValue = searchParams.get(id) ?? '';
+  const defaultStringValue = searchParams.get(id) ?? defaultValue;
 
   return <Input variant='ontime-filled' name={id} defaultValue={defaultStringValue} />;
 }
--- a/apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
+++ b/apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
@@ -71,7 +71,7 @@ export default function ViewParamsEditor({ paramFields }: EditFormDrawerProps) {
     setSearchParams(searchParams);
   };
 
-  const clearParams = () => {
+  const resetParams = () => {
     setStoredViewParams({ ...storedViewParams, [pathname]: {} });
     setSearchParams();
     onClose();
@@ -111,8 +111,8 @@ export default function ViewParamsEditor({ paramFields }: EditFormDrawerProps) {
         </DrawerBody>
 
         <DrawerFooter className={style.drawerFooter}>
-          <Button variant='ontime-ghosted' onClick={clearParams} type='reset'>
-            Clear
+          <Button variant='ontime-ghosted' onClick={resetParams} type='reset'>
+            Reset
           </Button>
           <Button variant='ontime-subtle' onClick={onEditDrawerClose}>
             Cancel
--- a/apps/client/src/common/components/view-params-editor/constants.ts
+++ b/apps/client/src/common/components/view-params-editor/constants.ts
@@ -1,46 +1,53 @@
 import { UserFields } from 'ontime-types';
+import { TimeFormat } from 'ontime-types/src/definitions/core/TimeFormat.type';
 
 import { ParamField } from './types';
 
-export const TIME_FORMAT_OPTION: ParamField = {
+export const getTimeOption = (timeFormat: TimeFormat): ParamField => ({
   id: 'format',
   title: '12  / 24 hour timer',
   description: 'Whether to show the time in 12 or 24 hour mode. Overrides the global setting from preferences',
   type: 'option',
   values: { '12': '12 hour AM/PM', '24': '24 hour' },
-};
+  defaultValue: timeFormat,
+});
 
-export const CLOCK_OPTIONS: ParamField[] = [
-  TIME_FORMAT_OPTION,
+export const getClockOptions = (timeFormat: TimeFormat): ParamField[] => [
+  getTimeOption(timeFormat),
   {
     id: 'key',
     title: 'Key Colour',
     description: 'Background colour in hexadecimal',
     type: 'string',
+    defaultValue: '101010',
   },
   {
     id: 'text',
     title: 'Text Colour',
     description: 'Text colour in hexadecimal',
     type: 'string',
+    defaultValue: 'fffff',
   },
   {
     id: 'textbg',
     title: 'Text Background',
     description: 'Colour of text background in hexadecimal',
     type: 'string',
+    defaultValue: '101010',
   },
   {
     id: 'font',
     title: 'Font',
     description: 'Font family, will use the fonts available in the system',
     type: 'string',
+    defaultValue: 'Arial Black',
   },
   {
     id: 'size',
     title: 'Text Size',
     description: 'Scales the current style (0.5 = 50% 1 = 100% 2 = 200%)',
     type: 'number',
+    defaultValue: 1,
   },
   {
     id: 'alignx',
@@ -48,12 +55,14 @@ export const CLOCK_OPTIONS: ParamField[] = [
     description: 'Moves the horizontally in page to start = left | center | end = right',
     type: 'option',
     values: { start: 'Start', center: 'Center', end: 'End' },
+    defaultValue: 'center',
   },
   {
     id: 'offsetx',
     title: 'Offset Horizontal',
     description: 'Offsets the timer horizontal position by a given amount in pixels',
     type: 'number',
+    defaultValue: 0,
   },
   {
     id: 'aligny',
@@ -61,16 +70,18 @@ export const CLOCK_OPTIONS: ParamField[] = [
     description: 'Moves the vertically in page to start = left | center | end = right',
     type: 'option',
     values: { start: 'Start', center: 'Center', end: 'End' },
+    defaultValue: 'center',
   },
   {
     id: 'offsety',
     title: 'Offset Vertical',
     description: 'Offsets the timer vertical position by a given amount in pixels',
     type: 'number',
+    defaultValue: 0,
   },
 ];
 
-export const TIMER_OPTIONS: ParamField[] = [TIME_FORMAT_OPTION];
+export const getTimerOptions = (timeFormat: TimeFormat): ParamField[] => [getTimeOption(timeFormat)];
 
 export const MINIMAL_TIMER_OPTIONS: ParamField[] = [
   {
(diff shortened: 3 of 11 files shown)


apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -1,106 +1,122 @@
...
     description: 'Text colour in hexadecimal',
     type: 'string',
+    defaultValue: 'fffff',
   },
   {
     id: 'textbg',
     title: 'Text Background',
     description: 'Colour of text background in hexadecimal',
     type: 'string',
+    defaultValue: '101010',
```
alex-Arc, line 36: maybe this should default to`00000000` <- notice the last to zeros makes it transparent
this way users only have to change the Key Colour

apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -110,189 +126,210 @@ export const MINIMAL_TIMER_OPTIONS: ParamField[] = [
...
```
alex-Arc, line 143: the same here

alex-Arc (approved):

apps/client/src/common/components/view-params-editor/constants.ts
cpvalente: I dont actually understand the comment. The values are used if they exist, otherwise we use the ones defined in CSS. In this case the value of the background colour is `transparent`

since the user cannot use transparent as a key, the 00.00.00.00 suggested is the closest.
It might also be cleaner to have an empty field in the cases where it is unclear

asharonbaltazar: Fair, I'll use `00000000` then

apps/client/src/common/components/view-params-editor/constants.ts
asharonbaltazar: Resolved with https://github.com/cpvalente/ontime/pull/610/commits/837d6de03de7923b490683dd872e213425c186b9

apps/client/src/common/components/view-params-editor/constants.ts
asharonbaltazar: Resolved with https://github.com/cpvalente/ontime/commit/837d6de03de7923b490683dd872e213425c186b9

apps/client/src/common/components/view-params-editor/ParamInput.tsx
```diff
@@ -29,19 +34,19 @@ export default function ParamInput({ paramField }: EditFormInputProps) {
   }
 
   if (type === 'boolean') {
-    const defaultCheckedValue = isStringBoolean(searchParams.get(id)) ?? false;
+    const defaultCheckedValue = isStringBoolean(searchParams.get(id)) ?? defaultValue;
```
cpvalente, line 37: `isStringBoolean` returns a boolean, so the nullish coalescing thing will never do its thing.  This likely not what you meant to write

This happens a few more times down the code

apps/client/src/common/components/view-params-editor/ParamInput.tsx
```diff
@@ -11,14 +11,19 @@ interface EditFormInputProps {
 
 export default function ParamInput({ paramField }: EditFormInputProps) {
   const [searchParams] = useSearchParams();
-  const { id, type } = paramField;
+  const { id, type, defaultValue } = paramField;
```
cpvalente, line 14: took me while to figure out where `paramField` comes from.
According to git blame, it was actually me writing this ... can you please correct my mistake and name this a prop? 😅

apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -1,106 +1,122 @@
...
 
-export const TIME_FORMAT_OPTION: ParamField = {
+export const getTimeOption = (timeFormat: TimeFormat): ParamField => ({
   id: 'format',
   title: '12  / 24 hour timer',
   description: 'Whether to show the time in 12 or 24 hour mode. Overrides the global setting from preferences',
   type: 'option',
   values: { '12': '12 hour AM/PM', '24': '24 hour' },
-};
+  defaultValue: timeFormat,
```
cpvalente, line 12: nice! I was not expecting you to have noticed this

apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -1,106 +1,122 @@
...
+});
 
-export const CLOCK_OPTIONS: ParamField[] = [
-  TIME_FORMAT_OPTION,
+export const getClockOptions = (timeFormat: TimeFormat): ParamField[] => [
```
cpvalente, line 15: is there a way we can avoid the extra step of having to find the times in every view?
we did that for the time string to abstract this into the central formatTime function. maybe here we can do something similar? perhaps exporting a closure that already has the timeFormat and just expects the field definition? (or` -insert-your-better-idea-here-`)

apps/client/src/common/components/view-params-editor/types.ts
```diff
@@ -4,9 +4,13 @@ type BaseField = {
...
 };
 
-type OptionsField = { type: 'option'; values: Record<string, string> };
-type StringField = { type: 'string' };
-type BooleanField = { type: 'boolean' };
-type NumberField = { type: 'number' };
+type OptionsField = {
+  type: 'option';
+  values: Record<string, string>;
+  defaultValue?: string;
```
cpvalente, line 10: Quick question: I am not sure I can think of a case for a select field without a default? am I missing something?

apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
```diff
@@ -111,8 +111,8 @@ export default function ViewParamsEditor({ paramFields }: EditFormDrawerProps) {
         </DrawerBody>
 
         <DrawerFooter className={style.drawerFooter}>
-          <Button variant='ontime-ghosted' onClick={clearParams} type='reset'>
-            Clear
+          <Button variant='ontime-ghosted' onClick={resetParams} type='reset'>
+            Reset
```
cpvalente, line 115: One for the future: I wish reset didnt close the drawer... what do you think?

cpvalente (commented):
Some questions and some nitpicks. Great work here. This feature is coming along great thanks to you

apps/client/src/common/components/view-params-editor/types.ts
asharonbaltazar: It's optional because some selects (like the time one) have a default value. And others, like the operator one, don't have a default value.

apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
asharonbaltazar: I can see two options:
- reset the values and start over
- reset the values and get to the original view

What do you think?

apps/client/src/common/components/view-params-editor/constants.ts
asharonbaltazar: It would probably need to be a hook: `useClockOptions`, but that's a chain reaction to rewriting other constants declarations.

asharonbaltazar: I say let's stick with this verbose solution (unless you really don't like it). I'm sure we'll revisit this again.

apps/client/src/common/components/view-params-editor/ParamInput.tsx
asharonbaltazar: Resolved with https://github.com/cpvalente/ontime/pull/610/commits/41ec894071e1490b26664daa511d7afa97c6fb51

alex-Arc (approved):

apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
cpvalente: I believe we would like to reset the options which would make the view re-render with default attributes 
This already happens, only difference is that we would leave the drawer open

apps/client/src/common/components/view-params-editor/constants.ts
cpvalente: Could the data be exposed from viewWrapper? It might already be

apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
asharonbaltazar: So would you like me to add the drawer to close?

cpvalente: I didnt understand the question, but we can take this another time

cpvalente:
Hi @asharonbaltazar, are you happy to merge the feature?

asharonbaltazar:
I'm okay with this PR because I'll continue working on it, so more changes to come. Are you okay with it?

cpvalente:
You are right @asharonbaltazar , apologies for my lack of focus.

currently when the user opens the drawer and changes a single parameter, they get a URL with every single configuration (as defaulted)

I think the URL should be a list of changes, and this way it becomes harder for users to navigate

I believe we should improve this before merging

asharonbaltazar:
Ah, that's a bug indeed. It has to do with the default values being passed into the `input`. I don't know how I missed that 🫠

cpvalente:
> Ah, that's a bug indeed. It has to do with the default values being passed into the `input`. I don't know how I missed that 🫠

No worries, I didn't notice it either until @alex-Arc mentioned it

asharonbaltazar:
We can prevent those default values from being saved when the new params are built upon submission. What do you think of that?

cpvalente:
> We can prevent those default values from being saved when the new params are built upon submission. What do you think of that?

I am not sure I completely understood. 
What I took away is that we will check the params against the default before redirecting.

Either way, anything that makes it so that the defaults are not in the URL is good with me

asharonbaltazar:
I'll send up a fix soon

apps/client/src/common/components/view-params-editor/ParamInput.tsx
```diff
@@ -9,16 +9,22 @@ interface EditFormInputProps {
...
 
     return (
-      <Select placeholder='Select an option' variant='ontime' name={id} defaultValue={defaultOptionValue}>
+      <Select
+        placeholder={defaultValue ? undefined : 'Select an option'}
+        variant='ontime'
+        name={id}
+        defaultValue={defaultOptionValue}
+      >
         {Object.entries(paramField.values).map(([key, value]) => (
```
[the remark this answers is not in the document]
asharonbaltazar: What do you mean?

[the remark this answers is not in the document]
cpvalente: on line 15 you destructure all the other properties of paramField, why not values as well?

apps/client/src/common/components/view-params-editor/constants.ts
asharonbaltazar: Resolving in favor of newer discussions

apps/client/src/common/components/view-params-editor/ParamInput.tsx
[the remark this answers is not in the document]
asharonbaltazar: Ah, `values` only exists with `type: 'option'`. So TS requires an extra check of 
```ts
if (type === 'option')
``` 
to allow us to access `values`.

I could destructure it after the `if` statement, if you'd like

apps/client/src/features/viewers/studio/StudioClock.tsx
```diff
@@ -40,6 +41,7 @@ export default function StudioClock(props: StudioClockProps) {
   // deferring rendering seems to affect styling (font and useFitText)
   useRuntimeStylesheet(viewSettings?.overrideStyles && overrideStylesURL);
   const { fontSize: titleFontSize, ref: titleRef } = useFitText({ maxFontSize: 500 });
+  const { data: settings } = useSettings();
```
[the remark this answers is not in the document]
asharonbaltazar: Potentially resolved with https://github.com/cpvalente/ontime/pull/610/commits/9d4f19fb6b5593bb1cb96cbda05c62c57db7e28d

apps/client/src/features/operator/Operator.tsx
```diff
@@ -40,6 +41,7 @@ export default function Operator() {
 
   const featureData = useOperator();
   const [searchParams] = useSearchParams();
+  const { data: settings } = useSettings();
```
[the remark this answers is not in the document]
asharonbaltazar: Potentially resolved with https://github.com/cpvalente/ontime/commit/9d4f19fb6b5593bb1cb96cbda05c62c57db7e28d

cpvalente (commented):
left a small comment, good work

apps/client/src/features/viewers/ViewWrapper.tsx
```diff
@@ -82,7 +85,7 @@ const withData = <P extends WithDataProps>(Component: ComponentType<P>) => {
     };
 
     // prevent render until we get all the data we need
-    if (!viewSettings) {
+    if (!viewSettings || !settings) {
```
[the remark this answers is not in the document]
asharonbaltazar: Resolved with https://github.com/cpvalente/ontime/pull/610/commits/7a2f85926d5aede869e7289c69ff517dd61075fd

cpvalente (approved):
Good work! 🚀

PR #611, opened 2023-11-23
Compatibility old chrome

alex-Arc:
This is as far as I can get with adding legacy support, most of the errors are gone
the js and css polyfills should be there but there is sill some strange looking suff when running in chrome 63

Screenshot 2023-11-22 180511
Screenshot 2023-11-22 180439
Screenshot 2023-11-22 180227


alex-Arc:
[CasparCG v2.4.0](https://github.com/CasparCG/server/releases/tag/v2.4.0-stable) has modern chrome
don't see a need to look into this more

Issue #608, opened 2023-11-21
Save view to URL alias


kellhogs:
I really like the new options menu for the views and how this improves workflow with them.
I can imagine users having a favorite way of how they want to display their timer. What if we implement a "Save to URL alias" button which would give the users the ability to save their favorite views and have the easily reproducible when they have to display the on multiple different displays.

This would need us to implement a button in the settings view with maybe a small popup to give it a name. This would then save a new URL alias.

We currently do not check for unique alias names which should probably be added here too.

cpvalente:
Hi @kellhogs, thank you for this suggestion

I think this could be a good improvement for the app. Unsure on the implementation cost
There has been some great work around the configuration of the views done by @asharonbaltazar, maybe he has some opinions?

asharonbaltazar:
The first thing that came to mind was having a third button below the Pencil (Edit) that would display a similar slideout with all the saved URLS for that particular view. The slideout will have a create/edit/delete buttons to manage their saved URLs. The theme will remain consistent. What do you think?

kellhogs:
I really like this approach, although we should think about how we would implement the existing settings menu into this. Or do we ditch it at this point?

Also adding the ability to open saved views from that menu would be good.

asharonbaltazar:
By settings menu, do you mean the customization slideout? I'm thinking either:
- we have two slideouts. One for customizing the view, and one for managing the saved URLs.
- we add a "Saved" tab to the existing customization slideout with all the relevant functionality there

kellhogs:
Sorry for the misunderstanding... I'm talking about the existing URL alias menu in the ontime settings.

asharonbaltazar:
Ah, I think that should be up to @cpvalente to decide

cpvalente:
Maybe it is helpful to understand the scope of the work

### CONTEXT
Currently, the user can customise a view using URL params. The view params helps with this
We also have aliases, which save a view customisation. These are useful for a few scenarios:
- integrations (you dont want to type a long URL)
- sharing a configured view
- saving a view configuration, as some sort of preset manager

In the case where the user has configured a view, they would currently need to copy the URL and paste it in the Alias settings in the `/editor`.

This is the point where we are trying to help the user. I think it is a good feature, but we should keep in mind that the action we are simplifying is already not complex, maybe boring. 
It is important to ensure that our solution doesnt end up being more complex than the problem it solves


### POTENTIAL IMPLEMENTATION
As I see it, we would need something that:
[...]

PR #615, opened 2023-11-24
chore: disable save params feature

cpvalente:
Temporarily disabled the save params feature

Are you ok with this @asharonbaltazar ?

While I was doing this, it occurred to me that the presets work we spoke about might be a simpler way to solve this problem, with less potential edge cases
https://github.com/cpvalente/ontime/issues/608#issuecomment-1823473114


asharonbaltazar (approved):
Approving because admittedly I failed to follow up if there are other functionalities that this might conflict with. We're slowly improving the parame feature, so I don't feel bad about "one step backwards, two steps forward."

PR #617, opened 2023-11-25
Several project files user folder

arymoraes:
[keep several project files in user folder](https://github.com/cpvalente/ontime/issues/374#top)

@cpvalente 

opened a PR (though I still have to polish it a bit), I do have a question on the project load/db.json thing for step 2

let's suppose we havent loaded anything yet, so we only have `db.json` with a blank `lastProject`, great

but then we upload `project.json`, make changes to the project, then close, and open again. Won't  the changes we make be applied to `db.json`, and then we'll try to load `project.json`? We'd have to sync those two. Do we consider the `project.json` in the `uploads` folder as the source of data? I think I'm a bit confused on how we want to approach this


cpvalente:
Hi @arymoraes, this is a good question that I hadnt considered

I guess the user would expect that the old file `project.json` contains all the changes. I wonder if we need a third file that simply contains a reference to the loaded file? if that doesnt exist, we default to loading or creating a `db.json`
This third file could be a simple file written in the same way as we do with the `ResumableService`

What do you think?

arymoraes:
> Hi @arymoraes, this is a good question that I hadnt considered
> 
> I guess the user would expect that the old file `project.json` contains all the changes. I wonder if we need a third file that simply contains a reference to the loaded file? if that doesnt exist, we default to loading or creating a `db.json` This third file could be a simple file written in the same way as we do with the `ResumableService`
> 
> What do you think?

Makes sense imo.

We could have a config file that will contain the reference to the loaded file, and when we load the app, we go to it, ask for it, and try to find that file, if we don't, we could show the user a warning like "we tried to load your latest loaded project but couldn't find it, so we're loading the default.

Where is `ResumableService`? Couldn't find it

cpvalente:
> We could have a config file that will contain the reference to the loaded file, and when we load the app, we go to it, ask for it, and try to find that file, if we don't, we could show the user a warning like "we tried to load your latest loaded project but couldn't find it, so we're loading the default.


Sounds great. We can always tidy up behaviour once we have a proof of concept 


> Where is `ResumableService`? Couldn't find it

Sorry. Meant RestoreService

https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/RestoreService.ts

arymoraes:
> > We could have a config file that will contain the reference to the loaded file, and when we load the app, we go to it, ask for it, and try to find that file, if we don't, we could show the user a warning like "we tried to load your latest loaded project but couldn't find it, so we're loading the default.
> 
> Sounds great. We can always tidy up behaviour once we have a proof of concept
> 
> > Where is `ResumableService`? Couldn't find it
> 
> Sorry. Meant RestoreService
> 
> https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/RestoreService.ts

so I changed the implementation plan a bit since it seemed to make more sense

We had a `config.js` file that we used to fetch this kind of information, so what I did was to change this into a defaultConfig file, and we move the config file to the appData folder.

This config files hold this information:
```

  database: {
    testdb: 'test-db',
    directory: 'db',
    filename: 'db.json',
  },
```

so we can use directory and filename to load and use the latest loaded project json file in the `uploads/` folder

I pushed the changes, let me know what you think. I also broke tests so I probably did something bad, but let me know what you think of the overall approach

apps/server/src/services/ConfigService.ts
```diff
@@ -0,0 +1,61 @@
...
+    filename: string;
+  };
+  styles: {
+    directory: string;
+    filename: string;
+  };
+  restoreFile: string;
+}
+
+export class ConfigService {
```
cpvalente, line 20: why do we export the class if we use as a singleton?

cpvalente (commented):
Hi @arymoraes , good work so far. 

Few notes:

For the read - write from file you can use Steno or LowDB (which is built on top of steno). that simplifies things a bit and makes it more consistent with the similar code

I am not sure this service should take over the rest of the config. Those defaults in config.js are more like app config and do not change at runtime. So there is no reason for them to go into a "volatile" storage
As of now, the only thing we need to save is the name of the project file.

Does this make sense?

As for the tests, it seems that you caused some havoc by adding the config service to the setup file. the setup runs at start to resolve a lot of the dependencies (setup the app) 
We will need this config service to be controlled inside app.ts

arymoraes:
> Hi @arymoraes , good work so far.
> 
> Few notes:
> 
> For the read - write from file you can use Steno or LowDB (which is built on top of steno). that simplifies things a bit and makes it more consistent with the similar code
> 

got it, makes sense

> I am not sure this service should take over the rest of the config. Those defaults in config.js are more like app config and do not change at runtime. So there is no reason for them to go into a "volatile" storage As of now, the only thing we need to save is the name of the project file.
> 
> Does this make sense?

Hmm... it does make a bit of sense. My only concern is `config.js` holds `directory` and `filename`. it only runs on load right? but it still always loads `/db/db.json` and we end up using the generate path to load the db.

Maybe the best solution would be to have the service then only handle a `config.json` thats separate from `config.js` that only holds the projectFile to load, and then when we load the db on:
```

// path to public db
export const resolveDbDirectory = join(
  testDbStartDirectory,
  isTest ? config.database.testdb : config.database.directory,
);
export const resolveDbPath = join(resolveDbDirectory, config.database.filename);

export const pathToStartDb = isTest
  ? join(currentDirectory, '../', config.database.testdb, config.database.filename)
  : join(currentDirectory, '/preloaded-db/', config.database.filename);
```

we'd look for that new config file, if it doesn't have a loadedProject or if we can't load the loadedProject we'd use the default `config.js` stuff like we do above?

>
> As for the tests, it seems that you caused some havoc by adding the config service to the setup file. the setup runs at start to resolve a lot of the dependencies (setup the app) We will need this config service to be controlled inside app.ts

makes sense

cpvalente:
> Hmm... it does make a bit of sense. My only concern is `config.js` holds `directory` and `filename`. it only runs on load right? but it still always loads `/db/db.json` and we end up using the generate path to load the db.
> 
> Maybe the best solution would be to have the service then only handle a `config.json` thats separate from `config.js` that only holds the projectFile to load, and then when we load the db on:
> 

I am not really sure I follow, but as of now, you are much more familiar with the feature than myself.

From my perspective, it looks like the only field from `config.js` we need is the `fileName`.
We could rename that to `defaultFilename` or `fallbackFilename` and use that for the cases where the app starts and doesnt find a definition for another file (from the service you are creating)

Lets flesh out the code a bit more, we can easily tweak these details later

arymoraes:
> > Hmm... it does make a bit of sense. My only concern is `config.js` holds `directory` and `filename`. it only runs on load right? but it still always loads `/db/db.json` and we end up using the generate path to load the db.
> > Maybe the best solution would be to have the service then only handle a `config.json` thats separate from `config.js` that only holds the projectFile to load, and then when we load the db on:
> 
> I am not really sure I follow, but as of now, you are much more familiar with the feature than myself.
> 
> From my perspective, it looks like the only field from `config.js` we need is the `fileName`. We could rename that to `defaultFilename` or `fallbackFilename` and use that for the cases where the app starts and doesnt find a definition for another file (from the service you are creating)
> 
> Lets flesh out the code a bit more, we can easily tweak these details later

Ok, I fixed the feedback and the tests pass, I do have a question though as I had some weird behaviour when testing locally:

what's the difference between localhost:3000 and `4001`? They have similar UIs but are slightly different, and I could only find reference to `4001` on the README.

Is `4001` the client being ran through the docker container and `3000` local?

cpvalente:
> what's the difference between localhost:3000 and `4001`? They have similar UIs but are slightly different, and I could only find reference to `4001` on the README.
> 

When running in dev mode, the react app is available on port 3000. This will be the most up to date version with hot module reloading for when you make changes.

On port 4001 you can see the built app. If you build ontime, both will be the same

I often use the 3000 when working.

arymoraes:
> > what's the difference between localhost:3000 and `4001`? They have similar UIs but are slightly different, and I could only find reference to `4001` on the README.
> 
> When running in dev mode, the react app is available on port 3000. This will be the most up to date version with hot module reloading for when you make changes.
> 
> On port 4001 you can see the built app. If you build ontime, both will be the same
> 
> I often use the 3000 when working.

awesome thanks, I found out I was having some weird behavour cause I had both docker and `npm run dev` running at the same time, but was able to solve it

It's ready for review now for steps 1 and 2

cpvalente:
Hi @arymoraes , this looks good but we can't merge it as is

With this changes the docker build will not be able to know how to load the file. If that is fixed we can make a self contained change and merge the PR

What do you think?

arymoraes:
> Hi @arymoraes , this looks good but we can't merge it as is
> 
> With this changes the docker build will not be able to know how to load the file. If that is fixed we can make a self contained change and merge the PR
> 
> What do you think?

So I was investigating this today and wanted to ask

why would the docker build not be able to load the file?

From what I've seen we use `getAppDataPath()` on `setup.ts` the same way in docker, its just a different root folder (`/external/` from `Dockerfile` env). And we use this to find the folder for the new `config.json` file.

If it's the first time we're running it, it'll create the config file on `const lastLoadedProjectConfigPath = join(getAppDataPath(), 'config.json');`.

When we're uploading the new project it'll go through the controller and upload it to `external/uploads` so we can load it later, won't it?

I might be missing something here though.

cpvalente:
Hi @arymoraes, I apologise for the slow replies and for misleading you here
I think you are correct
Could we continue with the work stacking PRs on this one so that I have some time to organise? Do you have what you need to continue?

@alex-Arc , do you have availability to check docker on this?

alex-Arc:
> do you have availability to check docker on this?

should have a moment to look at in on Saturday

arymoraes:
> Hi @arymoraes, I apologise for the slow replies and for misleading you here I think you are correct Could we continue with the work stacking PRs on this one so that I have some time to organise? Do you have what you need to continue?
> 
> @alex-Arc , do you have availability to check docker on this?

No worries, and sounds good. I might be a bit slower this end of year as I'll have family over, but I'll also be off work so I'm hoping to finish steps 3-5 til early next week

alex-Arc:
is there something specific I should be locking for?

image


contents of `/external/config.json` after uploading a rundown
```json
{
  "lastLoadedProject": "test_rundown.json"
}
```

cpvalente:
> is there something specific I should be locking for?

I think for now we just want to make sure that the docker flow is still working

We also want to make sure that the changes don't break the resume service you created

alex-Arc:
just tested the resume service through a crash and that works 👍🏽

cpvalente:
Perfect, @arymoraes , do you want to merge this into the v3 branch?
Depending on how far you are with the work I could start working on the UI

arymoraes:
> Perfect, @arymoraes , is it ok to merge this into the v3 branch? Depending on how far you are with the work I could start working on the UI

Sure, go ahead and merge. And I've already done planning and started working on steps 3-5 and should be a couple of days til I have it done, so you can start on the UI if you want

arymoraes:
> > Sure, go ahead and merge. And I've already done planning and started working on steps 3-5 and should be a couple of days til I have it done, so you can start on the UI if you want
> 
> Perfect. Just make a PR once you have a draft so we can agree on the API contract

https://github.com/arymoraes/ontime/pull/1/files

Since this isn't merged I honestly don't know how to make a PR into this, just from my fork to my fork, but on that PR steps 3-5 are done

cpvalente:
@arymoraes My bad, merging now

PR #625, opened 2023-11-26
wip: time to next

### LINKED ISSUE #160: Show the time to the next event
At times, we need to know how long until a certian event.  We would like onTime to calculate this automatically.  Inital thoughts are perhaps we could hover the mouse over the cue and this information could appear in a "tips" temporary window.

thanks !

cpvalente:
WIP

Closes #160 

image
image

### DIFF
--- a/apps/client/src/common/hooks/useSocket.ts
+++ b/apps/client/src/common/hooks/useSocket.ts
@@ -84,7 +84,6 @@ export const useInfoPanel = () => {
   const featureSelector = (state: RuntimeStore) => ({
     eventNow: state.eventNow,
     eventNext: state.eventNext,
-    playback: state.playback,
     selectedEventIndex: state.loaded.selectedEventIndex,
     numEvents: state.loaded.numEvents,
   });
--- a/apps/client/src/features/cuesheet/Cuesheet.tsx
+++ b/apps/client/src/features/cuesheet/Cuesheet.tsx
@@ -2,9 +2,11 @@ import { useRef } from 'react';
 import { ColumnDef, flexRender, getCoreRowModel, useReactTable } from '@tanstack/react-table';
 import Color from 'color';
 import { isOntimeBlock, isOntimeDelay, isOntimeEvent, OntimeRundown, OntimeRundownEntry } from 'ontime-types';
+import { millisToString } from 'ontime-utils';
 
 import useFollowComponent from '../../common/hooks/useFollowComponent';
 import { useLocalStorage } from '../../common/hooks/useLocalStorage';
+import { useTimer } from '../../common/hooks/useSocket';
 
 import BlockRow from './cuesheet-table-elements/BlockRow';
 import CuesheetHeader from './cuesheet-table-elements/CuesheetHeader';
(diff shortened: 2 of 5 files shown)


cpvalente:
Hi @alex-Arc , I have started working on this with the idea that we can show in the cuesheet, the time for every event start.

I realised then that we dont have a data source to show the time when a past event has started,
A work around for now is to not calculate this for events that are in the past.

I feel like the feature is confusing without the real start time data, and we are better off not releasing it until we get this
What do you think?

alex-Arc:
I dose feel confusing the way it shows up in the cue sheet.
a popup might be less distracting?

alex-Arc:
just did a quick test
moving the it from the cuesheet into the right-click menu of the rundown
it doesn't live update but maybe that is a good thing?
image

apps\client\src\features\rundown\event-block\EventBlock.tsx
```typescript
  const timer = useTimer();
  const [onContextMenu] = useContextMenu<HTMLDivElement>([
    { label: `Copy ID: ${eventId}`, icon: IoCopyOutline, onClick: () => copyToClipboard(eventId) },
    //....................
    {
      label: millisToDelayString(timeStart - timer.clock) ?? '',
      icon: IoTimeOutline,
      onClick: () => console.log('test'),
      isDisabled: true,
    },
  ]);

```

alex-Arc:
> I am not particularly keen on adding the timer as a dependency to the context menu, we can workshop some other solutions

I can understand that :)

last night I felt that the wall of timers ticking down was very distracting, it is not to bad looking at it now with fresh eyes
but sill I don't know if it is good to have so many fields updating at once?

also a better name might be `Planed start`

cpvalente:
> but sill I don't know if it is good to have so many fields updating at once?
> 

True. The problem to solve here is to provide a overview over how far a particular event is from starting. This wall of timers is likely what you would see in cue lab or grandMA.  

Would this make more sense in the operator where there are less things? 

Or perhaps we only show the time to the next element?

> also a better name might be `Planned start`
> 

Real start might need tweaking but Planned Start is incorrect since it should show the time to either 
a) the time an event will start 
b) the time behind for an event start (previous event running overtime)
c) the time the event has started 

This fields value should only be planned start if the event is in future and we are not having any delays

je71175:
Hi all.  I agree a wall of timers would be distracting and creating pop ups are only useful for a quick look - not a countdown values when other things are happening in the show.  
As a thought, would it be possible to say have the duration field clickable - and have it toggle between the duration value and the countdown to this upcoming event  This way the user can select and choose to see which events display the countdown timer...
thanks Jason

Punkerschaf:
Hey guys, I also would like to leave some various thoughts and this:

- Finding a name: My suggestion is `Log/Exp` Loggend/Expected. On past events it shows the logged time it startet, for future events the expected time to start based on current logging+durations.

- To limit the count of rendered times wouldn't it be the easiest to just have an option to render the next X events?

- In my opinion this is not a necessary information on the operator view. As an operator I rather need the time till I have to do something (to the next highlighted userfield on my view). But I understand this feature is a few steps in the future.

- On the cuesheet it could be helpful to show, that this is a calculated time and not editable in any way. Maybe to not draw the background box could be a small improvement, if we wan't to keep this in a separate column.

cpvalente:
Parking this until we get the reporter going

PR #628, opened 2023-11-29
Improve change api

alex-Arc:
add a more human way to cast strings to boolean or colours


apps/server/src/controllers/integrationController.ts
```diff
@@ -272,6 +272,7 @@ export function dispatchFromAdapter(
     // WS: {type: 'change', payload: { eventId, property, value } }
     case 'change': {
       const { eventId, property, value } = payload as ChangeOptions;
+      //TODO: cue can't pass this check
```
alex-Arc, line 275: I could use a tip on who to get the cue filed past this check

cpvalente: makes sense. it seems that we are using a placeholder event to check this. since we are unable to provide a placeholder cue this fails.

We could
a) add a cue field to the placeholder and make sure it gets overriden on creation of the event

b) modify this check to pass if the field is cue like
` if (property === "cue" ||  !isKeyOfType(property, event)) `

a might be more maintainable, but also more work

apps/server/src/utils/coerceType.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * @description Converts a value to a number if possible, throws otherwise
+ * @description Converts a value to a string if possible, throws otherwise
```
cpvalente, line 2: hahah, copy-paste 😅

apps/server/src/utils/coerceType.ts
```diff
@@ -21,6 +21,22 @@ export function coerceBoolean(value: unknown): boolean {
   if (value == null) {
     throw new Error('Invalid value received');
   }
+  if (typeof value === 'string') {
+    value = value.toLocaleLowerCase();
```
cpvalente, line 25: we probably shouldnt be mutating the value but creating a new one
```ts
  if (typeof value === 'string') {
    const lowerCaseValue = value.toLocaleLowerCase();
    switch (lowerCaseValue) {
```

apps/server/src/utils/coerceType.ts
```diff
@@ -40,3 +56,19 @@ export function coerceNumber(value: unknown): number {
...
+ */
+export function coerceColour(value: unknown): string {
+  if (value == null) {
+    throw new Error('Invalid value received');
+  }
+  if (typeof value === 'string') {
+    value = value.toLocaleLowerCase(); //Red will not work but red will
+  }
+  return String(value);
+}
```
cpvalente, line 74: I dont think this coerces into a colour, it just forces a value to be defined and to be a lower case string, is that not correct?
If we were to pass in a user value 000, we would just get "000" back

Also, 

- remember not to mutate the things we pass in
- in the current code, if the value is already a string it will still be coerced to String on return

cpvalente:
Thank you for the initiative.
I am not particularly onboard with the coerceColour function, I think any attempts here should check if the colour is a valid hexcode or an html named colour (there are only 166, we could make a dictionary in the app)
https://developer.mozilla.org/en-US/docs/Web/CSS/named-color

Apart from it, this is a good change and I just made some small comments, look forwards to merging this

packages/types/src/utils/colour.ts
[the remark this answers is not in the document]
cpvalente: It seems that this is only used in the coerceColour function for now, so maybe it doesnt need to be exported at all?

Also, would you mind adding a 
`sideEffects: false`

to the `package.json` of the utils package?

apps/server/src/utils/coerceType.ts
```diff
@@ -40,3 +58,35 @@ export function coerceNumber(value: unknown): number {
...
+export function coerceColour(value: unknown): string {
+  if (value == null || typeof value != 'string') {
+    throw new Error('Invalid colour value received');
+  }
+  if (value[0] == '#') {
+    const isHexColor = hexColor.exec(value);
+    if (isHexColor) {
+      //TODO: if the hex value is 4 or 8 digits the last 1 or 2 are the trasparent part
+      // we sould remove that to not breake the css formating the app
+      return value;
```
cpvalente, line 78: why would we remove it?
I am not sure we can make these kinds of decisions on behalf of the user

We know sometimes it will break, but not always, should the user not be in control of this?

apps/server/src/utils/coerceType.ts
```diff
@@ -40,3 +58,35 @@ export function coerceNumber(value: unknown): number {
...
   return parsedValue;
 }
+
+/**
+ * @description Converts a value to a colour if possible, throws otherwise
+ * @param {unknown} value - Value to be converted to a colour.
+ * @returns {string} - The converted value as a string.
+ * @throws {Error} Throws an error if the value is null or undefined.
+ */
+const hexColor = new RegExp(/^#((?:[a-f\d]{1,2}){3,4})$/, 'i');
```
cpvalente, line 68: we like tests for regex :D 
Maybe create a function isRegex and test that? we have somewhere in the app a regex file with a few useful regexes

apps/server/src/utils/coerceType.ts
```diff
@@ -40,3 +58,35 @@ export function coerceNumber(value: unknown): number {
...
+export function coerceColour(value: unknown): string {
+  if (value == null || typeof value != 'string') {
```
cpvalente, line 70: I think this checks are exclusive, so something being of type string makes it not null or undefined.
In that case this could simply be
`if (typeof value !== "string")` <---- mind the !== (one more = operator)

apps/server/src/utils/coerceType.ts
```diff
@@ -40,3 +58,35 @@ export function coerceNumber(value: unknown): number {
...
+    const isHexColor = hexColor.exec(value);
+    if (isHexColor) {
+      //TODO: if the hex value is 4 or 8 digits the last 1 or 2 are the trasparent part
+      // we sould remove that to not breake the css formating the app
+      return value;
+    } else {
+      throw new Error('Invalid hex colour received');
+    }
+  } else {
+    const lowerCaseValue = value.toLocaleLowerCase(); //Red will not work but red will
```
cpvalente, line 83: I am not sure the comment is adding much here

apps/server/src/utils/coerceType.ts
```diff
@@ -40,3 +58,35 @@ export function coerceNumber(value: unknown): number {
...
+    const isCssColor = lowerCaseValue in cssColours;
+    if (isCssColor) {
+      return lowerCaseValue;
+    } else {
+      throw new Error('Invalid colour name received');
+    }
+  }
+  //TODO: do we want to change other types to colors?
```
cpvalente, line 91: I am not sure I understand this comment, can you clarify?

apps/server/src/models/eventsDefinition.ts
```diff
@@ -1,10 +1,11 @@
 import { EndAction, OntimeBlock, OntimeDelay, OntimeEvent, SupportedEvent, TimerType } from 'ontime-types';
 
-export const event: Omit<OntimeEvent, 'id' | 'delay' | 'cue'> = {
+export const event: Omit<OntimeEvent, 'id' | 'delay'> = {
   title: '',
   subtitle: '',
   presenter: '',
   note: '',
+  cue: '',
```
cpvalente, line 8: hmm, the rules of our app declare that we cannot have a empty cue field, so we should not provide a placeholder event which breaks our rules


I know I said we could, but I am not sure I am comfortable with the change. 
the reasoning being, what makes `id` different from `cue`? if we add `cue` should we not add `id`? Since for now our use case is very simple, I think we can postpone this change until we have a better use case for it to be necessary

Do you agree?

apps/server/src/controllers/integrationController.ts
cpvalente: The issue you find here is that cue is not part of the object that you use to check against

We do not want to modify that object since it breaks our rules, could we do the second check?
if (property === "cue" ||  !isKeyOfType(property, event))

apps/server/src/controllers/integrationController.ts
```diff
@@ -272,7 +272,8 @@ export function dispatchFromAdapter(
     // WS: {type: 'change', payload: { eventId, property, value } }
     case 'change': {
```
cpvalente, line 273: The checks we made below only apply for events of type OntimeEvent

Is this what we want or do we want to allow the user to change any property that belongs to any event type?
We need to adapt the logic to reflect this decision and maybe document it with a comment

`// we only allow users to modify events of type of OntimeEvent
`
or

`// users can modify any event type, we need to check if the property is valid`

apps/server/src/utils/coerceType.ts
alex-Arc: Yes sound correct.
but we should then maybe look into having the `getAccessibleColour` add the transparent  part to its calculation

apps/server/src/utils/coerceType.ts
alex-Arc: do we want to convert numbers or boolean to a colour?

apps/server/src/utils/coerceType.ts
alex-Arc: the regex file is in the client 
but I added `packages\utils\src\isColourHex.ts` feel free to move or rename it

apps/server/src/models/eventsDefinition.ts
alex-Arc: yes we can wait

apps/server/src/controllers/integrationController.ts
alex-Arc: I think option 2 and it almost already happens in the `parse` function in `integrationController.config.ts`

alex-Arc:
current colour mix, the green is set to 0% transparent by the user
image
new
image

current colour mix, the green is set to 50% transparent by the user
image
new
image

current colour mix, the green is set to 100% transparent by the user
image
new
image

apps/client/src/common/utils/styleUtils.ts
```diff
@@ -13,13 +13,15 @@ type ColourCombination = {
...
+      const orgColour = Color(bgColour);
+      const preMultColour = orgColour.alpha(1).mix(Color('#1a1a1a'), 1 - orgColour.alpha());
+      const textColor = preMultColour.isLight() ? 'black' : '#fffffa';
+      return { backgroundColor: preMultColour.hexa(), color: textColor };
     } catch (_error) {
       /* we do not handle errors here */
     }
   }
-  return { backgroundColor: '#000', color: '#fffffa' };
+  return { backgroundColor: '#1a1a1a', color: '#fffffa' };
```
alex-Arc, line 24: is there a place I could get this default bg value from `#1a1a1a`

PR #629, opened 2023-11-29
refactor: setup global timer state

alexrohleder:
This PR introduces the global server state concept. Here are some benefits:

1. An easy way to derive state across concepts (rundown, timer, playback, etc.)
2. Centralised state mutable only through setters makes it easy to unit test
3. Remove the necessity of service synchronisation (rundown setting timer state, etc.)
4. Reduces orchestration complexity; triggers side-effects on middleware (see state.tsx@mutate fn)


apps/server/src/state.ts
```diff
@@ -0,0 +1,92 @@
...
+      state.timer.current = getCurrent(now, state.timer.duration, 0, 0, now, timer.timeEnd, state.timer.timerType);
+    }
+  });
+}
+
+type RequiredMutable<T> = {
+  -readonly [K in keyof T]-?: NonNullable<T[K]>;
+};
+
+function mutate(fn: (state: RequiredMutable<TState>) => void) {
```
alexrohleder, line 84: We can use this function to set up side-effects for state mutations. For example, we could send the state patches on the socket here.

I enforce at compilation time that you always run the middleware when mutating the state. The state is read-only but mutable when using this function.

---

Examples:

```ts
  const prevState = deepClone(state)
  fn(state)
  const newState = state

  const lifeCycle = {
    "onLoad": prevState.timer !== newState.timer && newState.playback !== prevState.playback,
    "onPlay": prevState.playback !== newState.playback,
  }

  for (const key in lifeCycle) {
    if (lifeCycle[key]) {
      integrationService.dispatch(TimerLifeCycle[key]);
    }
  }

  const patch = generatePatch(prevState, newState);
  websocket.sendAsJson(patch);

  saveToStore()
```

apps/server/src/state.ts
```diff
@@ -0,0 +1,92 @@
...
+    startedAt: null,
+    finishedAt: null,
+    secondaryTimer: null,
+    selectedEventId: null,
+    duration: null,
+    timerType: null,
+    endAction: null,
+    get finishedNow() {
+      return this.current <= 0 && this.finishedAt === null;
+    },
```
alexrohleder, line 31: I opted not to have getters for everything and to allow you to import `state` directly; it feels more straightforward. We can have a derived state using JavaScript getters.

apps/server/src/services/TimerService.ts
```diff
@@ -94,33 +79,23 @@ export class TimerService {
...
     this.pausedAt = restorePoint.pausedAt;
 
-    this.timer.current = this.timer.duration;
-    if (this.timer.timerType === TimerType.TimeToEnd) {
-      const now = clock.timeNow();
-      this.timer.current = getCurrent(now, this.timer.duration, 0, 0, now, timer.timeEnd, this.timer.timerType);
-    }
-
     this._onResume();
   }
```
alexrohleder, line 93: You can see what it will look like when this PR is ready. The service handles its state and executes an action that takes the server state. I'm still deciding if it's best to export `resumeTimer` or use another method.

apps/server/src/state.ts
```diff
@@ -0,0 +1,456 @@
...
+    lastUpdate: null,
+    secondaryTarget: null,
+    get finishedNow() {
+      return this.current <= 0 && this.finishedAt === null;
+    },
+  },
+};
+
+export const stateMutations = {
+  timer: {
```
cpvalente, line 51: I wonder if removing this timer object will be better in the long run since the idea is that all state is treated as a holistic piece of data.

In the cases we have, playback mutations dont make sense to be separed from timer mutation, the same is true for the next piece of state to move (loaded)

apps/server/src/state.ts
```diff
@@ -0,0 +1,456 @@
...
+    patch(timer: Partial<TimerState>) {
+      mutate((state) => {
+        for (const key in timer) {
+          if (key in state.timer) {
+            state.timer[key] = timer[key];
+          }
+        }
+      });
+    },
+    start() {
```
cpvalente, line 61: there is a lot of logic in some of this and I think it can be hard to keep track of where we are.

What do you think about separating the functions from the object? just for readability

apps/server/src/state.ts
```diff
@@ -0,0 +1,456 @@
...
+  }
+
+  restoreService.save({
+    playback: state.playback,
+    selectedEventId: state.timer.selectedEventId,
+    startedAt: state.timer.startedAt,
+    addedTime: state.timer.addedTime,
+    pausedAt: state.timer.pausedAt,
+  });
+
```
cpvalente, line 454: I have separed this into an utility function [here](https://github.com/cpvalente/ontime/pull/692). Could we devise a more elegant way to write utilities that the state mutations use?

apps/server/src/state.ts
```diff
@@ -0,0 +1,456 @@
...
+              if (newState.playback === Playback.Play) {
+                if (newState.timer.endAction === EndAction.Stop) {
+                  PlaybackService.stop();
+                } else if (newState.timer.endAction === EndAction.LoadNext) {
+                  // we need to delay here to put this action in the queue stack. otherwise it won't be executed properly
+                  setTimeout(() => {
+                    PlaybackService.loadNext();
+                  }, 0);
+                } else if (newState.timer.endAction === EndAction.PlayNext) {
+                  PlaybackService.startNext();
```
cpvalente, line 360: it feels strange to loop back to the parent service. 

the order of execution is
Controller > PlaybackService > TimerService > state

How would we return a result from the evaluation that is not in state (as in isFinished?) Maybe this needs to be calculated at the parent?
is there a more elegant way to escalate this information up? 

Perhaps we deprecate the entire `TimerService` in favour of the `PlaybackService`?

apps/server/src/state.ts
```diff
@@ -0,0 +1,456 @@
...
+
+  const newState = state;
+
+  opts?.sideEffect?.(result, newState, prevState);
+
+  try {
+    eventStore.batchSet({
+      playback: newState.playback,
+      timer: newState.timer,
+    });
```
cpvalente, line 437: question: once we move the other services to this process, how can we be granular about only streaming the parts that change?

question2: the previous implementation used the force parameter in update to ensure that we are not streaming changes too often. That meant that we could stream if anything has changed otherwise stream once per second.
Do you have a suggestion on a strategy to achieve this?

PR #633, opened 2023-12-07
feat: show overlap

alex-Arc:
overlap


alex-Arc:
Yes that's fair 
how about:
- space: `#3E75E8` / overlap: `#8064E1`
- next day: `#8a8a8a`
- delay: `#F57C13`

cpvalente:
> Yes that's fair 
> 
> how about:
> 
> - space: `#3E75E8` / overlap: `#8064E1`
> 
> - next day: `#8a8a8a`
> 
> - delay: `#F57C13`

It is difficult having a meaningful opinion without seeing in place.

I would suggest that there is no reason for space and overlap to have different colours

We also could indicate events out of order like

09:00-10:00
11:00-12:00
10:00-11:00

alex-Arc:
image
image
image
image

alex-Arc:
image

alex-Arc:
trying out indicators between cards, would need help moving them to a better position
image

cpvalente:
I think they should be on top of the event as the previous screenshot. What do you think?

alex-Arc:
1:
image

or 
2:
image


I like 2, especially if I can get the spacing to be the same as the icons below

cpvalente:
> 1:
> 
> image
> 
> 
> 
> or 
> 
> 2:
> 
> image
> 
> 
> 
> 
> 
> I like 2, especially if I can get the spacing to be the same as the icons below

Agreed 2nd looks better. I don't think it is extensible enough but I suggest we get this ready to merge and rework in v3

What do you think? Can you clean this up to merge?

I would fuss myself with the spacing

apps/client/src/common/components/input/time-input/TimeInput.module.scss
```diff
@@ -1,7 +1,7 @@
-@use "../../../../theme/v2Styles" as *;
+@use '../../../../theme/v2Styles' as *;
```
cpvalente, line 1: I realise this was probably your linter, but it is generally good to try to stick to the necessary changes.

cpvalente (commented):
Do you mind cleaning up here so we can get this ready for release?

alex-Arc:
should I remove the * indicator from the S in this PR?

alex-Arc:
Looks like a lot of stuff exists only to allow that * warning as there are no other checks in that validation function
image

should take it all out, or leave it in and just change the last part where it is displayed?

cpvalente:
> Looks like a lot of stuff exists only to allow that * warning as there are no other checks in that validation function
> 
> image
> 
> 
> 
> should take it all out, or leave it in and just change the last part where it is displayed?

If you are removing the reason for the code to exist, we should also remove the code. There is no advantage in keeping dead code around.

cpvalente:
Few clarifications:

Screenshot 2023-12-27 at 19 55 38

- When an event is next, the UP NEXT label is over the indicators. is this on purpose?
- I am not sure about the indicators always being in view, how about just showing them when there is reason for it?
- I wonder if all indicators other than delay should be gray. Currently overlap / spacing indicators change colour? blue and violet are do not indicate any relationship (potential schedule issue). It takes significant effort to have consistent colours in the interface and I am not sure this is a good case, it also can add confusion to the app semantic colours (blue for roll). What do you think?

alex-Arc:
> When an event is next, the UP NEXT label is over the indicators. is this on purpose?

that is on purpose, I could not find a better place? do you have a good idea for it?

> I am not sure about the indicators always being in view, how about just showing them when there is reason for it?
that sounds fine to me, it was mostly to follow the same styling as the delay indicators in the playback panel.

cpvalente:
> would loose the quick glace information

good points. I think the colour coding is too much for an user to keep track of.
There are many contexts using similar colours and the circles dont necessarily provide a quick glance information (IMO)

I suggest we run this by some users and merge into v2 to buy us some time to re-think. I believe a better approach would be with the tags on top of the event like this

Screenshot 2023-12-19 at 15 23 38

alex-Arc:
sounds good
so for now I make it gray (except delay) and then we look into it later

alex-Arc:
image
image
image
image

PR #635, opened 2023-12-08
fix: backstage end time

### LINKED ISSUE #634: Expected finish time not showing correctly on Backstage screen
Hi.
Just realized that on the Backstage screen, the Expected finish time is showing the time event stated at. Tested this with mac and windows, Safari and Firefox, Ontime version 2.21.3

cpvalente:
Fixes #634


alex-Arc:
ahh I see that I was looking at an event that was set to `time to end` 
So I agree with the calculation and naming, was just looking at the wrong timer type

image

PR #636, opened 2023-12-08
addTime

alex-Arc:


cpvalente:
This looks ok to me, but I fail to see the reason for the change.
The logic is equivalent and it doesnt save any work either in backend or frontend, does it?

cpvalente (approved):
👍

PR #637, opened 2023-12-09
feat: add user notes to editor

monoxane:
I needed this for a show I'm working on where we have a lot of cues and a lot of operators with their own notes, adding the cues then going to the cue sheet or op view to add their notes was taking a lot of time. 

I've added the 10 user notes fields as a table with EditableCells in the EventEditorDataRight component and moved the colour picker to EventEditorDataLeft so it wasn't too tall.

image

While I do do a fair bit of React I've never really done Typescript and don't fully understand this projects structure yet so I'm absolutely okay with any concerns, complaints, or constructive criticisms related to how I've done it or what exactly it is I've done.

### DIFF
--- a/apps/client/package.json
+++ b/apps/client/package.json
@@ -40,6 +40,7 @@
     "build:local": "cross-env NODE_ENV=local vite build",
     "build:electron": "cross-env NODE_ENV=local vite build",
     "build:docker": "vite build",
+    "build:localdocker": "cross-env NODE_ENV=local vite build",
     "lint": "eslint . --quiet",
     "lint-staged": "eslint",
     "test": "vitest",
--- a/apps/client/src/features/event-editor/EventEditor.module.scss
+++ b/apps/client/src/features/event-editor/EventEditor.module.scss
@@ -11,9 +11,9 @@ $table-header-font-size: calc(1rem - 3px);
 }
 
 .eventEditor {
-  padding: 1rem 2rem 2rem 2rem;
+  padding: 2rem 2rem 2rem 2rem;
   width: 100%;
-  gap: max(1rem, 2vh);
+  gap: 1rem;
 
   display: grid;
   grid-template-areas: 'time left right';
@@ -38,18 +38,21 @@ $table-header-font-size: calc(1rem - 3px);
   display: flex;
   flex-direction: column;
   gap: 0.5rem;
+  height: 100%;
 }
 
 .left {
   grid-area: left;
   padding: 0 1rem;
   border-left: 1px solid $border-color-ondark;
+  max-height: 100%;
 }
 
 .right {
   padding-left: 1rem;
   grid-area: right;
   border-left: 1px solid $border-color-ondark;
+  max-height: 100%;
 }
 
 @mixin input-label() {
@@ -119,6 +122,12 @@ $table-header-font-size: calc(1rem - 3px);
 .notes {
   font-size: $table-font-size;
   font-weight: 400;
+  height: 10rem;
+  overflow-y: auto;
+  transition: height 250ms;
+  table {
+    width: 100%;
+  }
 
   tr {
     display: flex;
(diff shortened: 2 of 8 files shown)


monoxane:
also FYI another set of commands needs to be added to the turbo and package.json files to allow local docker builds without hitting sentry auth issues, #638

cpvalente:
Hi @monoxane , first of all, I greatly appreciate your work and the fact that you share it back into the community. That is imo the most rewarding part of creating a community driven project

I also must congratulate you on your timing, this is something that we recently decided to prioritise and have been discussing in terms of design.

Without looking at the code, I think your implementation is accurate and does the job well.
When we considered this in the past, we found that the event editor (bottom drawer) had issues with expandability, ie: sooner or later we will run out of space to add stuff.
As it is, it already takes a lot of real-estate from the screen

The running idea was to either a) add some tabs to the editor or b) add a scrollable element to contain the editor. Where we thought the latter made more sense

See below a frankenstein suggestion that hopefully is good enough to demonstrate:
- when editing an event, a drawer comes from the right side of the rundown
- this means shifting the info panel out of the interface and the playback area to the right
- this is a vertically scrollable container, ideally with collapsible sections, this would hopefully make it easy to extend and use

Screenshot 2023-12-09 at 15 07 48


I was wondering if I could have your thoughts on the design? Maybe you are willing to help us move with some next steps?


Meanwhile I will take a look at the code and we can try and get it ready to merge
The only thing that immediately jumped to my mind was that the user fields are multi-line, do we have ideas on how we could handle this in the editor?

monoxane:
Yea I definitely agree that putting a large block like that in the bottom editor is unscalable and takes up a lot of realestate, adding the 11 lines for notes increased its height by almost 30% which is not great but I just wanted to get the ball rolling so it is what it is. I think long term something like your mock would be the best solution with an events column, an event editing column, and a "system control" column, all being scrollable. It might also make sense to do something like this (diagrams.net wireframe because I'm bad at real mockups :P) 
image

I think your's might work better on smaller screens though because you could drop the rightmost column <1280 px wide and still have it very usable (especially with the ability to open rundown control in another tab/window and float it on top that I discovered about 10 minutes ago)

Regarding the multiline notes, my implementation uses the same EditableCell as the Cuesheet view so it handles them fine, just makes the whole edit bar taller which, again, isn't great. I think I'll add a fixed height limit to the whole edit bar and overflow:scroll the notes container div to stop it from taking up so much space.

Another point I'd like to bring up is the possibility of making the user notes a more flexible and user definable structure instead of just having user0 through user9 as base level fields on the event object. It might make sense to have it as a an array of key/value pairs where the number of items and the names of them are defined in the modal currently used to just label them. This would make it a lot more flexible for both smaller and larger shows, that might want to have one or two notes sections for "AUDIO" and "LX", or go all the way into having finer notes for specific operators within each department (each ASM on a large theatre show could be doing their own task for each cue point and it can be hard to keep track of these with limited user notes).

Open to working on any ideas or improvements you or anyone else has in regards to this, I'm a pretty heavy user and producer of open source stuff and its nice to have something I use that I comprehend enough to contribute back 😄

monoxane:
Okay so I've made some changes to make it work better on smaller screens, now it will `overflow: scroll` (with 10 rem height) when on a screen smaller than 1080p, and show all note items (with 25 rem height, still scrolls if there's multiline notes) on anything bigger. Screenshots also show a multiline note working the same as in the Cuesheet 

image
image

cpvalente:
Thank you for this @monoxane

From my conversation with @alex-Arc this exposes the need for a director view which we have been hinting at for the past year.

The idea is that we should have a custom purpose view which focuses on the rundown and its playback. I imagine not too dissimilar from the operator view, but with access to more data.

For now I am let wondering how much more data we should move into the editor as it seems we start exposing its limits.
If the `userFields` are to be readonly, could the info panel be re-designed to help in this case?

This would mean that we split the work in two:
- what can we do right now to improve the editor?
- what can the new view do to facilitate this workflow?

It would be great to have your help with the specification. What do you think?

monoxane:
I'm on my phone right now so will check back in later with a response to the majority of the comment but with regard to the use of CSS !important, there seems to be some styles somewhere in the source that resulted in the first column of both the table header and the rows to be white background white text and I could not work out where it was coming from. Happy to take a closer look when I'm not in preprod for this show and fix it up properly.

cpvalente:
Thank you @monoxane , we are moving with a few features at the moment so there is no rush.
Just get in touch when it is convenient to you. Look forwards to it

cpvalente:
Hi @monoxane , this has been implemented in v3.
We can close this PR as it is no longer relevant, but I wonder if you would be available to give us some feedback on the implementation?

PR #638, opened 2023-12-09
add localdocker build commands

monoxane:
as mentioned in #637


cpvalente:
Thank you @monoxane, I see that this migth be what  the `build:local` attempts to achieve, maybe we just need to add the local environment variable there, or is there a reason to maintain both?
What do you think?

PR #644, opened 2023-12-12
minimal demo page

### LINKED ISSUE #530: serve user static files from ontime
## Context

The app leverages a express server which, amongs its responsabilities, its to host the UI and CSS overrides
We would like to let users leverage the server to be able to serve their own files.

This would be useful in situations where the users have made a custom view, using the [view template](https://github.com/cpvalente/ontime-viewer-template-v2/) or otherwise and would facilitate distribution by attaching it to Ontime's infrastructure

## The task
- We would like to create a new folder `static` in the ontime install directory (better naming welcome)
- This folder should be included in [the setup process](https://github.com/cpvalente/ontime/blob/master/apps/server/src/setup.ts#L85), along the CSS files and [be populated at init](https://github.com/cpvalente/ontime/blob/master/apps/server/src/app.ts#L121)
- we would then [allow express to expose this folders to the user](https://github.com/cpvalente/ontime/blob/master/apps/server/src/app.ts#L63), meaning a user could go to `ontime:4001/external/static/my-page` (maybe the paths can be improved, this is quite a mouthful)
- we would ship the app with an example view, potentially just a copy-paste from the template above, this files would not go through a build process so it should be pure js - html - css

It is still to be proved whether this could be detrimental to the app performance, I would expect a developer creating this feature to be critical about this

alex-Arc:
closes #530 

- [x] minimal timer
- [x] reconnecting socket
- [x] 404 in file not found in /external

image
image

### DIFF
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -66,12 +66,7 @@ app.use('/playback', playbackRouter);
 // serve static - css
 app.use('/external', express.static(externalsStartDirectory));
 app.use('/external', (req, res) => {
-  const hasFileEnding = req.url.split('.')[1];
-  if (!hasFileEnding) {
-    res.redirect(`${req.originalUrl}.html`);
-  } else {
-    res.status(404).send(`${req.originalUrl} not found`);
-  }
+  res.status(404).send(`${req.originalUrl} not found`);
 });
 
 // serve static - react, in dev/test mode we fetch the React app from module
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -65,6 +65,14 @@ app.use('/playback', playbackRouter);
 
 // serve static - css
 app.use('/external', express.static(externalsStartDirectory));
+app.use('/external', (req, res) => {
+  const hasFileEnding = req.url.split('.')[1];
+  if (!hasFileEnding) {
+    res.redirect(`${req.originalUrl}.html`);
+  } else {
+    res.status(404).send(`${req.originalUrl} not found`);
+  }
+});
 
 // serve static - react, in dev/test mode we fetch the React app from module
 const reactAppPath = join(currentDirectory, resolvedPath());
--- a/apps/server/src/external/demo/app.js
+++ b/apps/server/src/external/demo/app.js
@@ -2,8 +2,6 @@
 /**
  * This is a very minimal example for a websocket client
  * You could use this as a starting point to creating your own interfaces
- *
- * This example does not handle disconnections
  */
 
 const mts = 1000; // millis to seconds
--- a/apps/server/src/external/demo/app.js
+++ b/apps/server/src/external/demo/app.js
@@ -4,9 +4,7 @@
  *
  * This example does not handle disconnections
  */
-// eslint-disable-file --this is an external demo page
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-nocheck
+/* eslint-disable */
 
 // this would need to resolve to the hostname of where ontime is running
 const socket = new WebSocket(`ws://${window.location.hostname}:4001/ws`);
--- a/apps/server/src/external/demo/app.js
+++ b/apps/server/src/external/demo/app.js
@@ -1,39 +1,71 @@
+/*eslint-env browser*/
 /**
  * This is a very minimal example for a websocket client
  * You could use this as a starting point to creating your own interfaces
  *
  * This example does not handle disconnections
  */
-/* eslint-disable */
 
-// this would need to resolve to the hostname of where ontime is running
-const socket = new WebSocket(`ws://${window.location.hostname}:4001/ws`);
-console.log(`ws://${window.location.hostname}:4001/ws`);
+const mts = 1000; // millis to seconds
+const mtm = 1000 * 60; // millis to minutes
+const mth = 1000 * 60 * 60; // millis to hours
 
-function leftPad(number) {
-  return number.toString().padStart(2, '0');
-}
+const leftPad = (number) => {
+  return Math.floor(number).toString().padStart(2, '0');
+};
 
-socket.addEventListener('message', (event) => {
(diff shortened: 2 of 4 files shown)


apps/server/src/external/demo/app.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * This is a very minimal example for a websocket client
+ * You could use this as a starting point to creating your own interfaces
+ *
+ * This example does not handle disconnections
+ */
+/* eslint-disable */
```
cpvalente, line 7: why do we need to disable eslint?

apps/server/src/external/demo/app.js
```diff
@@ -0,0 +1,39 @@
...
+
+// this would need to resolve to the hostname of where ontime is running
+const socket = new WebSocket(`ws://${window.location.hostname}:4001/ws`);
```
cpvalente, line 10: this will break if we are outside 4001, maybe it is OK but we should log something to the console?

apps/server/src/external/demo/app.js
```diff
@@ -0,0 +1,39 @@
...
+
+  // all objects from ontime are structured with type and payload
+  const { type, payload } = data;
+
+  // we only need to read message type of ontime
+  if (type === 'ontime') {
+    // destructure known data from ontime
+    // see https://cpvalente.gitbook.io/ontime/control-and-feedback/websocket-api
+    const { timer, playback } = payload;
+    console.log('Got message from ontime:', timer);
```
cpvalente, line 28: do you think we should be logging this?

apps/server/src/external/demo/app.js
```diff
@@ -0,0 +1,39 @@
...
+    const timerElement = document.getElementById('timer');
+    if (playback == 'stop') {
+      timerElement.innerText = '--:--:--';
+    } else {
+      const now = new Date(timer.current);
+      timerElement.innerText = `${leftPad(now.getUTCHours())}:${leftPad(now.getMinutes())}:${leftPad(
+        now.getSeconds(),
+      )}`;
```
cpvalente, line 36: is it not easier to calculate it ourselves like we do in the app and module?

apps/server/src/external/demo/app.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * This is a very minimal example for a websocket client
+ * You could use this as a starting point to creating your own interfaces
+ *
+ * This example does not handle disconnections
```
cpvalente, line 5: if you want to handle reconnect, you can check how the code is done in Ontime, we dont use any libraries here.
There also seems to be plenty of code in stack overflow about this, let me know if you want help

cpvalente (commented):
small comments, great initiative

apps/server/src/external/demo/app.js
alex-Arc: there are missing defines
I found this `/*eslint-env browser*/` seams better?

cpvalente: I am not sure I understood. Can you send me a screenshot?

alex-Arc: with out it default browser defines are marked as missing
image

apps/server/src/app.ts
```diff
@@ -65,6 +65,14 @@ app.use('/playback', playbackRouter);
...
 // serve static - css
 app.use('/external', express.static(externalsStartDirectory));
+app.use('/external', (req, res) => {
+  const hasFileEnding = req.url.split('.')[1];
+  if (!hasFileEnding) {
+    res.redirect(`${req.originalUrl}.html`);
+  } else {
+    res.status(404).send(`${req.originalUrl} not found`);
+  }
+});
```
cpvalente, line 75: can you explain what is the context?
does it not load the `index.html` by default

alex-Arc: oh you are right, it dose

how about the 404 shold we show an error messege
otherwise it will just fall through to the react page, that could be a bit confusing

cpvalente: I dont think we should compromise our app behaviour for this feature.

The old behaviour ensured that as long as you were in the right host, you would see ontime. 
There are several reasons why I feel this is important:
- Users are configuring their own URLs and can get it wrong
- Users can have outdated aliases bookmarked
- users on a phone or tablet can easily make mistakes when typing the URL
- a production can be sharing a link to a less tech savy person who could do with some help

in the cases above, it can be frustrating to keep seeing error pages and we do better in assuring them that they are in the right place

apps/server/src/external/demo/app.js
cpvalente: I wouldnt worry about eslint here
One better change could be checking if websocket is available in the environment

apps/server/src/app.ts
cpvalente: btw: just noticed that we copy the contents of the external folder, not the folder itself.
so in production this should be `ontime:4001/demo` in which case we should change the directory to something more meaningful. what do you think?

apps/server/src/external/demo/app.js
```diff
@@ -0,0 +1,69 @@
+/*eslint-env browser*/
+/**
+ * This is a very minimal example for a websocket client
+ * You could use this as a starting point to creating your own interfaces
+ *
+ * This example does not handle disconnections
```
cpvalente, line 6: I think it does now, does it not?

cpvalente (approved):
left a small comment, good work!

apps/server/src/external/demo/app.js
alex-Arc: haha yes it dose

alex-Arc: feel free to add clarifying comments to guide the user if you have some

cpvalente: I think it is good enough as it is, we can create a help page for this outside the code base and add a link here
Good work

PR #645, opened 2023-12-12
Feat: reorder events with alt+ctrl + arrow up/down

alex-Arc:


apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -89,7 +89,7 @@ export default function Rundown(props: RundownProps) {
       // handle held key
       if (event.repeat) return;
       // Check if the alt key is pressed
-      if (event.altKey && (!event.ctrlKey || !event.shiftKey)) {
+      if (event.altKey && !event.ctrlKey && !event.shiftKey) {
```
cpvalente, line 92: maybe we extract these into a few variables like

```ts
const justAlt = event.altKey && !event.ctrlKey && !event.shiftKey
const altAndCtrl = event.altKey && event.ctrlKey && !event.shiftKey
```
what do you think?

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -133,9 +133,32 @@ export default function Rundown(props: RundownProps) {
             break;
           }
         }
+      } else if (event.altKey && event.ctrlKey && !event.shiftKey) {
+        switch (event.code) {
+          case 'ArrowDown': {
+            if (entries.length < 2 || cursor == null) {
```
cpvalente, line 139: how come the check for `ArrowUp` and `ArrowDown` is the same?

cpvalente (changes requested):
small comments, otherwise it looks good to me

apps/client/src/features/rundown/Rundown.tsx
alex-Arc: was just using the same switch case approach as the one above if more shortcuts were added
but I can collapse it for now 👍🏽

alex-Arc:
@cpvalente do you have a idea to sole the scroll?

apps/client/src/features/rundown/Rundown.tsx
cpvalente: ah ok, you dont have to change it. if the logic is the same then thats it

I would have just expected that the logic would have been different, so when merging up we need to check if there is something up, when merging down we need to check if there is something down

alex-Arc:
is it still worth it even if we can't fix the scroll for now?

cpvalente:
> is it still worth it even if we can't fix the scroll for now? 

I would say so. I would suggest merging to v3 and do a review later on

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -133,9 +135,28 @@ export default function Rundown(props: RundownProps) {
...
+      } else if (modKeysCtrlAlt) {
+        if (entries.length < 2 || cursor == null) {
+          return;
+        }
+        if (event.code == 'ArrowDown') {
+          const nextEvent = getNext(entries, cursor);
+          if (nextEvent) {
+            const from = entries.findIndex((event) => event.id === cursor);
+            const to = entries.findIndex((event) => event.id === nextEvent.id);
+            reorderEvent(cursor, from, to);
```
cpvalente, line 147: this means we iterate through the rundown 3 times, could we do something to simplify somehow?

cpvalente (commented):
some small performance concerns, I think we should get this to v3

apps/client/src/features/rundown/Rundown.tsx
alex-Arc: we could get all the results we need from the `getNext` function if we changed it to return `{nextEvent, nextIndex}` as they are all found inside that function and then `currentIndex` would be `nextIndex - 1`

cpvalente: sounds good to me

cpvalente:
I don't think the logic is correct. The next event is not necessarily the next array item. There might be other things in between such as blocks or delays

alex-Arc:
> I don't think the logic is correct. The next event is not necessarily the next array item. There might be other things in between such as blocks or delays

yes I think that is on purpose so that we can also reorder around delays and blocks

Recording 2023-12-23 144458

cpvalente:
I see, but in that case we can use getNextEvent since it always reruns the index of the next OntimeEvent

alex-Arc:
> I see, but in that case we can use getNextEvent since it always reruns the index of the next OntimeEvent

If we have this:
```
-event1
-event2 <- cursor
-block
-event3
```

and the then press move down I would expect this to happen:
```
-event1
-block
-event2 <- cursor
-event3
```

is it therefor not correct to consider the block and delay when looking for the next element?

cpvalente:
> > I see, but in that case we can use getNextEvent since it always reruns the index of the next OntimeEvent
> 
> 
> 
> If we have this:
> 
> ```
> 
> -event1
> 
> -event2 <- cursor
> 
> -block
> 
> -event3
> 
> ```
> 
> 
> 
> and the then press move down I would expect this to happen:
> 
> ```
> 
> -event1
> 
> -block
> 
> -event2 <- cursor
> 
> -event3
> 
> ```
> 
> 
> 
> is it therefor not correct to consider the block and delay when looking for the next element?

What would happen if you had 
```
- event 1
- event 2
- block 1
- block 2
- event 3
```

And affected your change in moving event 2 down?
Am I mistaken in thinking that it would swap with block 2 and not block 1?

alex-Arc:
Recording 2023-12-27 134541

PR #650, opened 2023-12-13
fix: correct link for CSS overrides

### LINKED ISSUE #647: macOS 13.6.3 : Error occurred trying to load the resource
Hello, 

I have made some changes to the css override file and enabled the override in the application. The displays appear to load a new CSS external file (override.css) but safari, chrome, and FF throw errors that an error occurred while trying to load the resource. When attempting to load it directly by link, my css file (stored at /Applications/ontime.app/Contents/Resources/extraResources/external/styles/override.css) is not loaded. Instead a "default" override file, of sorts, appears. 

Any ideas?
Thanks,
Eric

cpvalente:
fixes #647 

@alex-Arc, to test you just need to check in the app settings that the link is to the created css file not to the initial one

### DIFF
--- a/apps/server/src/controllers/ontimeController.ts
+++ b/apps/server/src/controllers/ontimeController.ts
@@ -10,7 +10,7 @@ import { DataProvider } from '../classes/data-provider/DataProvider.js';
 import { failEmptyObjects, failIsNotArray } from '../utils/routerUtils.js';
 import { PlaybackService } from '../services/PlaybackService.js';
 import { eventStore } from '../stores/EventStore.js';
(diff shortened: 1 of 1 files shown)


alex-Arc:
seams to still point to appdata 
image

cpvalente:
I think this is correct?

You can check the code for the getAppData

alex-Arc:
oh OK I thought it should point to the install path, but i guess it makes more sense this way.
but that dose mean that the external path gives access to everything in the appData folder, that seams a bit dangerous, should we maybe copy the whole external folder and point to that?

cpvalente:
You are likely right. We should probably create a folder to expose. Maybe the `assets` folder?

PR #652, opened 2023-12-13
refactor: pass HTTP to integration controller

### LINKED ISSUE #641: Improve HTTP API
We would like the HTTP API to take advantage of the integration controller to be as capable as the other APIs

This means:
- HTTP API should be made of GET methods
- We would extend the `playback` endpoint to call integrationController
- Data should be translated before reaching controller

cpvalente:
a quick PR to test how easy this could be

- makes all get requests go through the integration controller
- changes the playback API to work over GET (much easier for users)
- removes some deprecated (?) logic

you can test with an URL like `localhost:4001/playback?start`

If we think this is good, we should likely look at the return types

Also, do we think the `/playback` route is still the best way? Maybe `/api` would be better?

Closes #641


cpvalente:
@alex-Arc what do you think about this?

a start message could look like `http://localhost:4001/api/start?index=1` 

I also added a fallback for `http://localhost:4001/api` which gives a hello so you can test the connection

Do you think this is simple enough for integrations? What do you usually see in other software?

alex-Arc:
I like the `/api/action?parameter=???` way
the vMix way of doing it is a bit cumbersome https://www.vmix.com/help25/index.htm?DeveloperAPI.html 
they use `/api/?Function=action&parameter=???`

cpvalente:
Nice.

I will work with the return values. Would you mind giving this a test to check if there is anything we should consider? 

Maybe the changes endpoint?

Also testing with vMix would be nice, we can update the docs to work with the GET request which is a lot easier

alex-Arc:
how to we want to unwrap the payload?

right now this
`/api/set-timer-message-text?test`  becomes `set-timer-message-text { payload: { test: '' } }`
and
`/api/set-timer-message-text?payload=test` =>`set-timer-message-text { payload: { payload: 'test' } }`

both of which dont´t work

should users specify payload? or do we take the first label as the payload?

cpvalente:
> should users specify payload? or do

good question, I think the right way to do it would be with

`/api/set-timer-message?text=something&visible=true`

what do you think?
it seems that we should put some effort into documenting the differences in the APIs...

alex-Arc:
`/api/set-timer-message?text=something&visible=true` 
looks good but doesn't quite fit with our current api ?
would we have a translation layer that for this request calls both `set-timer-message-text` and `set-timer-message-visible`

cpvalente:
> `/api/set-timer-message?text=something&visible=true` looks good but doesn't quite fit with our current api ? would we have a translation layer that for this request calls both `set-timer-message-text` and `set-timer-message-visible`

I think we would need to make new api entries for `set-timer-message`

alex-Arc (approved):
works as expected 
`/api/change?eventId=f1e85&property=title&value=HTTP`
changed the title

cpvalente:
@alex-Arc , you should be able to pass a patch timerMessage object now, like `http://localhost:4001/api/set-timer-message?text=testing`

alex-Arc:
> patch

it is a HTTP GET call right?

cpvalente:
> > patch
> 
> it is a HTTP GET call right?

yes, you are correct. I guess it is uncomment to use a get request to make changes in the server, but that is the simplest way for integrations. what do you think?

alex-Arc:
yes, looks like `patch` makes sense in relation to how.
but a lot of our event tech can only do `get` so it is good to keep it like that.

PR #654, opened 2023-12-15
Feat dynamic delaytime buttons

### LINKED ISSUE #366: Request: Adjust seconds
I tend to do a lot of things that we are tracking seconds. Would it be possible to add +/- functionality to adjust running time in seconds?

One approach is to hold CTRL or CMD down and clicking +1, -1, +5, -5 would affect seconds? When holding down CTRL a label "seconds" could pop near the top of the buttons...or something else to identify these are seconds we're clicking on?

alex-Arc:
A quick sketch
the holding down the shift key will reveal the alternate values

image
image
image

closes #366

### DIFF
--- a/apps/client/src/common/stores/editorSettings.ts
+++ b/apps/client/src/common/stores/editorSettings.ts
@@ -9,6 +9,10 @@ type EditorSettings = {
     b: number;
     c: number;
     d: number;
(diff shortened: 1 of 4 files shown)


alex-Arc:
We will solve this in a different way

PR #655, opened 2023-12-15
Add italian client translation

MatteoGheza:
Added italian translation for the client, using the other languages as model.

### DIFF
--- a/apps/client/src/translation/languages/it.ts
+++ b/apps/client/src/translation/languages/it.ts
@@ -1,0 +1,19 @@
+export const langIt = {
+  'common.end_time': 'Ora di Fine',
+  'common.expected_finish': 'Fine Prevista',
+  'common.now': 'Adesso',
+  'common.next': 'Prossimo',
+  'common.public_message': 'Messaggio pubblico',
+  'common.start_time': 'Ora di Inizio',
+  'common.stage_timer': 'Orologio Palco',
+  'common.started_at': 'Iniziato Alle',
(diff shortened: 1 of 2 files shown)


apps/client/src/translation/languages/it.ts
```diff
@@ -0,0 +1,19 @@
...
+  'common.time_now': 'Ora attuale',
+  'countdown.ended': 'Evento finito alle',
+  'countdown.running': 'Evento in corso',
+  'countdown.select_event': 'Seleziona un evento da seguire',
+  'countdown.to_start': 'Tempo alla partenza',
+  'countdown.waiting': 'In attesa dell\'inizio dell\'evento',
+  'countdown.overtime': 'in ritardo',
+};
+
+export type TranslationObject = Record<keyof typeof langIt, string>;
```
cpvalente, line 19: this type will collide with the TranslationObject in the english strings (our baseline).

You dont really need to create a new one since the keys are the same, just import from the `en.ts`
so your object declarion will be

```ts
import { TranslationObject } from './en';

export const langIt: TranslationObject = {...}
```

cpvalente (changes requested):
Thank you for your work @MatteoGheza , good stuff

We just need to make a small change to the type before merging. Once that is done, I will push this into a temporary feature branch to do the necessary changes in the UI

cpvalente (approved):
Thank you! I will take it from here

PR #658, opened 2023-12-17
feat: app settings

cpvalente:
Moves all settings to a central place

Screenshot 2023-12-19 at 21 43 21

## Summary by CodeRabbit

- **Documentation**
  - Updated the development guidelines to reflect a specific version requirement for the `node` and `pnpm` dependencies.


cpvalente:
Very much work in progress, but would you mind taking a look @asharonbaltazar ?

apps/client/src/features/app-settings/AppSettings.tsx
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * TODO: make store to keep settings
+ *       - whether a tab has unsaved changes
+ */
+
+export default function AppSettings(props: AppSettingsProps) {
+  const { settings = settingPanels[0].id, onClose } = props;
+  useKeyDown(onClose, 'Escape');
+
+  const [selectedTab, setSelectedTab] = useState<SettingsOptionId>(settings);
```
asharonbaltazar, line 26: Why aren't you using the Chakra-UI `Tabs` component?

asharonbaltazar (commented):
Left a comment questioning the implementation. What's the advantage to manually writing tabs functionality using a store? Do you want it to sync with the browser params?

apps/client/src/features/app-settings/AppSettings.tsx
cpvalente: We consider abandoning Chakra where possible.

Also the amount of customisation we want would take some effort to do, at which point using the Chakra component no longer provides development speed

PR #659, opened 2023-12-17
refactor: prevent accidental edit in operator view

cpvalente:
Adds user option to prevent the operator view from being editable on long press as suggested by @viktorfreiman in #609


alex-Arc (approved):
we could maybe use something similar in the cuesheet

cpvalente:
> we could maybe use something similar in the cuesheet

This is. 2 minute code change, we can do it if there is something to improve. What would it do?

alex-Arc:
it was in relation to allowing editing of events (titles and everything else), not necessarily something we should do now
but having it default to off might be a good idea, if we ever do add editing to the cuesheet

Issue #97, opened 2022-01-26
Studio clock support for special characters


cpvalente:
rc5_v2(unofficial)
![image](https://user-images.githubusercontent.com/25690871/146659426-b3cfbf59-604c-49a0-a701-e55154ff4d5e.png)
* adding delay doesn't change the countdown timer bellow clock.

* font does not support some special chars. ÅÄÖ for example. it's ok but just noted.
![image](https://user-images.githubusercontent.com/25690871/146659664-428c5393-c4b0-4ac1-857c-49cc5658b0de.png)

List updates times now! great!
Next item text under clock autoscales. Nice!!  (needs a bit of centering covered in graphics couple of posts above)

_Originally posted by @xztraz in https://github.com/cpvalente/ontime/issues/42#issuecomment-997308179_

cpvalente:
We are marking this as Good First Issue to allow collaborators to take this task.
In short, we have a large user base in countries with extended alphabets (nordic and Germanic mostly). We would like to make sure the font here can support some special characters, such as the Norwegian Å Ø Æ

alex-Arc:
maybe the we could switch to a 8-bit like font
https://www.cdnfonts.com/minecraftia.font
![image](https://github.com/cpvalente/ontime/assets/19875125/f2729b35-eea6-4a67-964d-85e4a0339d4c)

cpvalente:
Hi @alex-Arc, thank you for this.
I am not too sure of the readability of that font, but I think it is a good alternative.

I wonder if the better alternative is to use a normal font. for everything but the time.
What do you think?

alex-Arc:
I feel a bit like this clock  is mostly a coolness factor 😉  
but yes a normal font would also be fine

PR #663, opened 2023-12-19
Studio clock

### LINKED ISSUE #662: Studio View - On Air does not change to reflect OnTime status
The 'On Air' in the Studio view does not reflect the status of OnTime.

I think this is due to both states being set to the same colour:

color: var(--studio-idle, #c53030);
color: var(--studio-active, #c53030);

Version 2.24.8 in Docker on Raspberry Pi 
Studio View checked on Safari (17.2) and Firefox (120.0.1) on MacOS (14.2)

alex-Arc:
Fixes #662  and #97


cpvalente:
Very impressive, thank you
I wonder if the test fields on the right side should all be in "normal" font? 

That would increase readability and solve the issues with character support

What do you think?

apps/client/src/features/viewers/studio/StudioClock.scss
```diff
@@ -123,23 +157,23 @@ $cyan-idle: #0aa;
...
       padding-bottom: 2vh;
       font-size: 15vh;
-      line-height: 0.9em;
+      line-height: 1.2em;
       color: var(--studio-active, $red-active);
 
       &--idle {
-        color: var(--studio-idle, $red-active);
+        color: $backlight;
+        // color: var(--studio-idle, $red-idle);
```
cpvalente, line 170: Remember to remove unused code
We also need to maintain the variables where possible

Especially if we are merging to v2

apps/client/src/features/viewers/studio/StudioClock.tsx
```diff
@@ -83,16 +85,28 @@ export default function StudioClock(props: StudioClockProps) {
       <NavigationMenu />
       <ViewParamsEditor paramFields={studioClockOptions} />
       <div className='clock-container'>
-        <div className={`studio-timer ${showSeconds ? 'studio-timer--with-seconds' : ''}`}>{clock}</div>
+        <div className='studio-timer-warp'>
```
cpvalente, line 88: Question: what does timer-warp mean?

apps/client/src/features/viewers/studio/StudioClock.tsx
```diff
@@ -83,16 +85,28 @@ export default function StudioClock(props: StudioClockProps) {
...
+        <div className='next-countdown-warp'>
+          <div
+            className={cx([
+              'next-countdown',
+              isNegative ? ' next-countdown--overtime' : '',
+              isPaused ? ' next-countdown--paused' : '',
+            ])}
+          >
+            {/* ! is used an a blank character //TODO: placeholder moves a bit when becomming active*/}
+            {isNegative ? '-' : '!'}
```
cpvalente, line 107: I wonder if it is best not to have the `-` symbol. This would mean the test moves a bit, but that is maybe OK.
Most of the time the `-` will not be there, and the clock will be off-centre, so I wonder if we are sacrificing the common case for an edge case

apps/client/src/features/viewers/studio/StudioClock.scss
```diff
@@ -1,8 +1,24 @@
...
+  src:
+    local('DSEG14'),
+    url('./../../../assets/fonts/DSEG14ClassicMini-Regular.ttf') format('truetype');
+}
+
+@font-face {
+  font-family: 'mono';
+  src:
+    local('MONO'),
+    url('./../../../assets/fonts/Hack-Regular.ttf') format('truetype');
```
cpvalente, line 21: question: could we use a default monospace font to avoid one more import? I dont know in terms of style if this one is significantly better

alex-Arc: yeah.. i dont like the default monospace that much so i'm trying to finde one that fits better with the lowtech look

apps/client/src/features/viewers/studio/StudioClock.tsx
alex-Arc: the warp is used to create the "backlight" effect

cpvalente (commented):
made a few suggestions. what do you think?

cpvalente:
@alex-Arc , what do you think about the state of this?

apps/client/src/features/viewers/studio/StudioClock.scss
```diff
@@ -1,182 +1,198 @@
...
-    margin: 4vh 0;
-    font-family: digital-clock, monospace;
-    text-transform: uppercase;
+.next-title {
+  font-family: monospace;
+  font-weight: 400;
+  color: var(--studio-active-label, $cyan-active);
+  font-size: calc(var(--clock-size) / 10);
+  text-align: center;
+  overflow: clip visible;
```
alex-Arc, line 116: should we change this to 
`overflow: clip hidden;` just to catch the edge case

image

apps/client/src/features/viewers/studio/StudioClock.scss
```diff
@@ -1,182 +1,198 @@
...
 
-      &--idle {
-        color: var(--studio-idle, $red-active);
-      }
-    }
+  &--overtime {
+    color: var(--studio-active, $red-active);
+  }
+  &--paused {
+    color: var(--studio-active, $orange-active);
```
alex-Arc, line 128: We should add the pause orange to the style override

apps/client/src/features/viewers/studio/StudioClock.scss
```diff
@@ -1,182 +1,198 @@
...
 /* =============== CLOCK STUFF ==================*/
 
-$clock-size: 90vh;
-$size-hours: min(3vh, 20px);
-$half-hours: min(1.5vh, 10px);
-$size-min: min(2.5vh, 18px);
-$half-min: min(1.25vh, 9px);
-$red-active: #c53030;
-$red-idle: #000000;
+$size-hours: clamp(12px, 2.5vw, 22px);
```
alex-Arc, line 13: don't know if we want to fix this but we only scale down the dots proportional to the width and not the height

image

alex-Arc:
am/pm indicator?
image

cpvalente (commented):
good tips, thank you @alex-Arc , let me know if you think this is good enough for now.
We can revisit later

alex-Arc:
Yes this is good

PR #665, opened 2023-12-22
New lower third

alex-Arc:
Recording 2024-01-03 134014


apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -263,6 +263,126 @@ export const LOWER_THIRDS_OPTIONS: ParamField[] = [
...
+  {
+    id: 'upper-src',
+    title: 'Upper Text',
+    description: '',
+    type: 'option',
+    values: {
+      title: 'Title',
+      subtitle: 'Subtitle',
+      presenter: 'Presenter',
+      notes: 'Notes',
```
cpvalente, line 276: I dont think notes should be here, see the options of the operator view

apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -263,6 +263,126 @@ export const LOWER_THIRDS_OPTIONS: ParamField[] = [
...
+    id: 'upper-font',
+    title: 'Upper Text Font',
+    description: 'https://developer.mozilla.org/en-US/docs/Web/CSS/font',
+    type: 'string',
+    placeholder: 'bold 4vh sans-serif',
+  },
+  {
+    id: 'lower-font',
+    title: 'Lower Text Font',
+    description: 'https://developer.mozilla.org/en-US/docs/Web/CSS/font',
```
cpvalente, line 335: Is this maybe too advanced for a setting? at this level of customisation you might as well use the css override

apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -263,6 +263,126 @@ export const LOWER_THIRDS_OPTIONS: ParamField[] = [
...
+    type: 'string',
+    placeholder: 'normal 3.5vh sans-serif',
+  },
+  {
+    id: 'width',
+    title: 'Width',
+    description: 'Width of the element',
```
cpvalente, line 342: is this the width relative to the screen?
Should this not be automated related on the amount of text?

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,216 @@
...
+import { Message, OntimeEvent } from 'ontime-types';
+
+import NavigationMenu from '../../../common/components/navigation-menu/NavigationMenu';
+import { LOWER_THIRD_OPTIONS } from '../../../common/components/view-params-editor/constants';
+import ViewParamsEditor from '../../../common/components/view-params-editor/ViewParamsEditor';
+import { cx } from '../../../common/utils/styleUtils';
+
+import './LowerThrid.scss';
+
+//TODO: add none option
```
cpvalente, line 12: what is a none option?

cpvalente (commented):
small comments about customisation

apps/client/src/common/components/view-params-editor/constants.ts
alex-Arc: For my lowers I like that the element always reaches the same position on the screen even if the name is short
but maybe we cover more users with a dynamic one?
I could change it to be minimum width?

apps/client/src/common/components/view-params-editor/constants.ts
alex-Arc: yes.
What should we expose here in this menu?

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
alex-Arc: If the user only wants one line

alex-Arc: changed this so that no text from the source means that the line won't be shown

apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -215,52 +215,125 @@ export const MINIMAL_TIMER_OPTIONS: ParamField[] = [
...
+    id: 'top-src',
+    title: 'Top Text',
+    description: '',
+    type: 'option',
+    values: {
+      title: 'Title',
+      subtitle: 'Subtitle',
+      presenter: 'Presenter',
+      lowerMsg: 'Lower Thrid Message',
+    },
```
cpvalente, line 239: we should have some defaults here

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,236 @@
...
+import { Message, OntimeEvent, ViewSettings } from 'ontime-types';
+
+import { overrideStylesURL } from '../../../common/api/apiConstants';
+import NavigationMenu from '../../../common/components/navigation-menu/NavigationMenu';
+import { LOWER_THIRD_OPTIONS } from '../../../common/components/view-params-editor/constants';
+import ViewParamsEditor from '../../../common/components/view-params-editor/ViewParamsEditor';
+import { useRuntimeStylesheet } from '../../../common/hooks/useRuntimeStylesheet';
+import { cx } from '../../../common/utils/styleUtils';
+
+import './LowerThrid.scss';
```
cpvalente, line 12: remember to update the name on the import

e2e/tests/features/201-message-control.spec.ts
```diff
@@ -20,7 +20,7 @@ test('message control sends messages to screens', async ({ context }) => {
   await editorPage.getByPlaceholder('Shown in lower third').fill('testing lower');
   await editorPage.getByRole('button', { name: /toggle lower third message/i }).click();
 
-  await featurePage.goto('http://localhost:4001/lower');
+  await featurePage.goto('http://localhost:4001/lower?trigger=manual&bottom-src=lowerMsg');
```
cpvalente, line 23: how does the manual trigger work? I dont understand how the users control it

cpvalente (commented):
I think it looks good, we just need to cleanup code and are good to merge

e2e/tests/features/201-message-control.spec.ts
alex-Arc: by using the show hide button in the message controller

cpvalente: I see, so we remove the other text line and use only the lower?
You have more experience on this than me so I will follow your lead

I wonder whether triggering the in and out of you could be annoying.
How about a button that says trigger which would make it appear and disappear after the timeout period

alex-Arc: > I wonder whether triggering the in and out of you could be annoying.

It could but I think it being a toggle is the most flexible and don't require us to add more settings and buttons
and if uses want it to auto animate out that is super easy to set up in companion
I could make it one of the presets

apps/client/src/common/components/view-params-editor/constants.ts
cpvalente: sorry, didnt see this. It is something we still wonder?

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,236 @@
...
+import { overrideStylesURL } from '../../../common/api/apiConstants';
+import NavigationMenu from '../../../common/components/navigation-menu/NavigationMenu';
+import { LOWER_THIRD_OPTIONS } from '../../../common/components/view-params-editor/constants';
+import ViewParamsEditor from '../../../common/components/view-params-editor/ViewParamsEditor';
+import { useRuntimeStylesheet } from '../../../common/hooks/useRuntimeStylesheet';
+import { cx } from '../../../common/utils/styleUtils';
+
+import './LowerThird.scss';
+
+enum srcKeys {
```
arymoraes, line 14: TS convention is to capitalize enum names and keys, like

```
enum SrcKeys {
 Title = "title"
}
```

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,236 @@
...
+    topBg: '00000000',
+    bottomBg: '00000000',
+    topSize: '65px',
+    bottomSize: '40px',
+    transition: 3,
+    delay: 3,
+    key: 'ffffffff',
+    lineColour: 'ff0000ff',
+    lineHeight: '0.4em',
+  };
```
arymoraes, line 69: for the `options` object, I'm seeing that you are declaring it then making changes to it with the if statements below according to `searchParams` right?

Problem is every time the component re-renders, it'll redo all the logic - from declaring the options to making the changes)

one thing you could do is wrap this on a `useMemo` hook, `useMemo` will allow you to return the options object with all of the modifications you need done, and it'll only run when the properties in the dependency array change, so for example:

you'd have a 
```
const defaultOptions = { 
 width: 45,
 ....
}
```

and then you could do

```
const options = useMemo(() => {
  const newOptions = {...defaultOptions}
  ...
  make all the changes with the if statements
  ...
  return newOptions;
}, [searchParams])
```

Then it'll only recalculate stuff if the `searchParams` change

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const [playState, setplayState] = useState<'pre' | 'in' | 'out'>('pre');
+
+  useEffect(() => {
+    document.title = 'ontime - Lower Third';
+  }, []);
+
+  const _width = searchParams.get('width');
+  if (_width) {
+    const width = Number(_width);
+    if (!Number.isNaN(width) && width != options.width) {
```
arymoraes, line 82: I usually prefer to use `!==` for comparisons as it also ensures both have the same type

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  const topText = options.topSrc == srcKeys.lowerMsg ? lower.text : eventNow ? eventNow[options.topSrc] : '';
+
+  const bottomText = options.bottomSrc == srcKeys.lowerMsg ? lower.text : eventNow ? eventNow[options.bottomSrc] : '';
+
+  const transition = `${options.transition}s`;
+
+  const triggerData =
+    options.trigger == triggerType.event ? eventNow?.id : options.trigger == triggerType.manual ? lower.visible : null;
```
arymoraes, line 167: nesting ternaries usually make it harder to read, you could to something like

```
  const topText = useMemo(() => {
    if (options.topSrc === srcKeys.lowerMsg) {
      return lower.text;
    } else if (eventNow) {
      return eventNow[options.topSrc];
    }
    return '';
  }, [options.topSrc, srcKeys.lowerMsg, eventNow, lower.text]);

  const bottomText = useMemo(() => {
    if (options.bottomSrc === srcKeys.lowerMsg) {
      return lower.text;
    } else if (eventNow) {
      return eventNow[options.bottomSrc];
    }
    return '';
  }, [options.bottomSrc, srcKeys.lowerMsg, eventNow, lower.text]);

  const triggerData = useMemo(() => {
    if (options.trigger === triggerType.event) {
      return eventNow ? eventNow.id : null;
    } else if (options.trigger === triggerType.manual) {
      return lower.visible;
    }
    return null;
  }, [options.trigger, triggerType.event, triggerType.manual, eventNow, lower.visible]);

  const transition = `${options.transition}s`;
```

alex-Arc:
@arymoraes thanks for the review, this is my first time using Memo could you check that it is correct

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,246 @@
...
+import { overrideStylesURL } from '../../../common/api/apiConstants';
+import NavigationMenu from '../../../common/components/navigation-menu/NavigationMenu';
+import { LOWER_THIRD_OPTIONS } from '../../../common/components/view-params-editor/constants';
+import ViewParamsEditor from '../../../common/components/view-params-editor/ViewParamsEditor';
+import { useRuntimeStylesheet } from '../../../common/hooks/useRuntimeStylesheet';
+import { cx } from '../../../common/utils/styleUtils';
+
+import './LowerThird.scss';
+
+enum srcKeys {
```
arymoraes, line 14: enum name is also capitalized

as in `SrcKeys`

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,246 @@
...
+  lineHeight: string;
+};
+
+interface LowerProps {
+  eventNow: OntimeEvent | null;
+  viewSettings: ViewSettings;
+  lower: Message;
+}
+
+const defaultOptions: Readonly<LowerOptions> = {
```
arymoraes, line 50: very good use of `Readonly` here

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+    const topSrc = Object.values(srcKeys).find((s) => s === searchParams.get('top-src'));
+    if (topSrc) {
+      newOptions.topSrc = topSrc;
+    }
+
+    const bottomSrc = Object.values(srcKeys).find((s) => s === searchParams.get('bottom-src'));
+    if (bottomSrc) {
+      newOptions.bottomSrc = bottomSrc;
+    }
```
arymoraes, line 93: as a small optimization, you could run the loop only once (with a forEach) maybe?

Because like this it ends up looping through the array three times, when one would probably be enough.

This is small though since I'm assuming `Object.values(srcKeys)` is a small array, but would be more problematic with larger arrays

arymoraes:
> @arymoraes thanks for the review, this is my first time using Memo could you check that it is correct

That's perfect

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
alex-Arc: ah OK

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
alex-Arc: There are only 4 keys, but I will keep it in mide for the future

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,236 @@
...
+      }
+    } else if (options.trigger == triggerType.manual) {
+      if (triggerData) {
+        setplayState('in');
+      } else {
+        setplayState('out');
+      }
+    }
+    return () => null;
+  }, [options.delay, options.trigger, triggerData]);
```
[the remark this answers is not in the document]
alex-Arc: I tried changing it a bit, let me know if it looks better now

apps/client/src/features/viewers/lower-thirds/LowerThird.tsx
```diff
@@ -0,0 +1,243 @@
...
+  topSize: '65px',
+  bottomSize: '40px',
+  transition: 3,
+  delay: 3,
+  key: 'ffffffff',
+  lineColour: 'ff0000ff',
+  lineHeight: '0.4em',
+};
+
+export default function LowerThird(props: LowerProps) {
```
cpvalente, line 67: the previous solution leveraged a memo to make sure that the component would not re-render unless the relevant props changed (previously presenter and title)

Can we make sure that this is still the case?
I think we would need:
- in automatic mode, check that the lower3 mounts when we go through the events with the << and >> playback
- when a event is running, change the data associated to the event, does the lower3 remount?

alex-Arc: it looks to work correctly on the screen, if the event id changes the timeout is reset and the data is changed in place
should thees be useMemo or useEffect?

````typescript
const trigger = useMemo(() => {
    if (options.trigger === TriggerType.Event) {
      return eventNow?.id;
    } else if (options.trigger === TriggerType.Manual) {
      return lower.visible;
    }
    return false;
  }, [eventNow?.id, lower.visible, options.trigger]);

  useEffect(() => {
    if (options.trigger === TriggerType.Event && trigger) {
      setPlayState('in');
      const animateOutInMs = options.delay * 1000 + options.transition * 1000;
      const timeout = setTimeout(() => {
        setPlayState('out');
      }, animateOutInMs);
      return () => clearTimeout(timeout);
    } else if (options.trigger === TriggerType.Manual) {
      setPlayState(trigger ? 'in' : 'out');
    } else {
      setPlayState('pre');
    }
    return () => null;
  }, [options.delay, options.transition, options.trigger, trigger]);

`````
Recording 2024-01-07 013808

cpvalente: > should thees be useMemo or useEffect?

`useMemo` and `useEffect` achieve different things.
In this case you can read `useMemo` as in "dont recalculate this object unless one of the dependencies changes value"

where the `useEffect` is more like "run this code when one of the dependencies changes value"

We often avoid using `useEffect`, but in this case we need it to coordinate with the animation. The `useMemo` stabilises the trigger so that it only recalculates if the input data has changed.

As you described, if the id changes, the `useMemo` is triggered, and a new object is created. This, in turn, makes the `useEffect` run, and the animation starts.

With these changes we lose the previous behaviour where the animation was triggered if the data of the lower third changed (title, subtitle, presenter...) But maybe that is good? You have more experience in this field and I will leave it up to you.

We just need to make sure that we dont have issues with race conditions. ie: what happens if eventId changes while we are animating the lower? could we end up with multiple animation transitions scheduled?

Hope this helps

PR #670, opened 2023-12-25
Feat: Editable table 

alex-Arc:
Trying out [tanstack table](https://tanstack.com/table/v8)


cpvalente:
You can see how the cuesheet is done, it is already using tanstack table

alex-Arc:
ag-grid seams easy to use so far, but it is a good deal bigger (787 kB vs 36.8 MB)

cpvalente:
if we can do this with the tanstack table it is better.
If you could get it to a point where you can edit in table we can build the customisation from there.
In short, it would end up looking very much like the cuesheet

alex-Arc:
this might be done at a later time

PR #671, opened 2023-12-25
feat: create project list

cpvalente:
@arymoraes , quick experiment to see what data we need from backend.
Will plug it in once your changes are merged


cpvalente:
@arymoraes , it seems that the excel files are not deleted on upload...

Would you mind making sure those get deleted, also the endpoint should only return potential project files, which would end with .json extension

Screenshot 2023-12-28 215510

Also, I need to know which one is currently loaded, any ideas on how to achieve this?

arymoraes:
> @arymoraes , it seems that the excel files are not deleted on upload...
> 
> Would you mind making sure those get deleted, also the endpoint should only return potential project files, which would end with .json extension
> 
> Screenshot 2023-12-28 215510
> 
> Also, I need to know which one is currently loaded, any ideas on how to achieve this?

https://github.com/cpvalente/ontime/pull/679

I made the changes, only thing I want to ask about is what should we do when we upload excel files? I'm not sure what the behaviour should be there.

For the rest I fixed the issues (and also fixed an issue on load), and added the last loaded project (which is also the currently loaded one) to the endpoint response

cpvalente:
> I made the changes, only thing I want to ask about is what should we do when we upload excel files? I'm not sure what the behaviour should be there.


Great work, I will take a look.
I believe the excel files should be deleted after parsing

arymoraes:
> > I made the changes, only thing I want to ask about is what should we do when we upload excel files? I'm not sure what the behaviour should be there.
> 
> Great work, I will take a look. I believe the excel files should be deleted after parsing

makes sense, but if we upload an excel file, should it also create a `.json` that we will store at the `uploads` folder and we can select/load them later?

cpvalente:
> should it also create a .json that we will store at the uploads folder and we can select/load them later?

I dont think so.
The excel import is a way to import data into our project file.
There is no need to create a new file or keep the excel around

arymoraes:
> > should it also create a .json that we will store at the uploads folder and we can select/load them later?
> 
> I dont think so. The excel import is a way to import data into our project file. There is no need to create a new file or keep the excel around

makes sense, so I'll keep the flow as:

`parse -> delete -> returns the response`

and thats it

PR #672, opened 2023-12-25
Seconds for Progress-Bar Input

kellhogs:
Hey, I finally got around to finishing this. Basically, this adds the ability to input seconds for the progress bar color definitions.
Also, we are able to input this as a string like all the other time inputs.

This basically uses a stripped down version of the [TimeInput component](https://github.com/cpvalente/ontime/blob/8afffb987051b738e667dd7e154e6fe399de1e1a/apps/client/src/common/components/input/time-input/TimeInput.tsx)

I chose to copy and modify here to not disturb with the other places this component is used.

I'm still not really confident with the React-Hook-Form stuff, so pls take a look at my implementation there. It's working, but I'm almost sure there is a smarter way to do this.

There is one [place](https://github.com/cpvalente/ontime/blob/16d285b88e11aae60e96273f71e1ace0804d26c2/apps/client/src/features/modals/settings-modal/ViewSettingsForm.tsx#L51) where I wasn't sure how to resolve the typing error. Maybe someone can give me some insight's in how to do this.


cpvalente:
closing for now since we have changed the desired implementation

PR #675, opened 2023-12-27
improvement: hide seconds

### LINKED ISSUE #649: Feature Request: option to hide seconds in views
Feature Request: support ?seconds=false on all views

cpvalente:
I believe this is a good implementation for #649 

It gives options to hide the display of seconds in auxiliary timers (wall clock)
Currently applied to `timer`, `studio`, `countdown`, `backstage` and `public` views

It felt counter-productive to apply on other views 
@alex-Arc , what do you think?


alex-Arc:
> It felt counter-productive to apply on other views

agree

and looks to be working

are we sure it should not applie to the running timer? (or a separate toggle for that)
I would assume that the need to hide the second it to not distract/stress the speaker or who ever is looking at the timer?

cpvalente:
Hi @alex-Arc , I will add the templates in a separate PR since it involves some work in how we format times
What do you think about this for now?

alex-Arc:
look good

alex-Arc (approved):

PR #677, opened 2023-12-28
Warning and danger per event

### LINKED ISSUE #624: Set warning and danger times per event
## Context
We currently have a global setting for warning and danger times.
These allow the setting of the colours in the progress bar

We would like to extend this feature to incorporate alarms as described in #131 
This means that we would need to be able to set the warning and danger times per event

## The task
- for the types of `OntimeEvent`, we would add a new field that describes the two calls for danger and warning times. This could be an object or an array of two elements. The fields value should be a number
- in the `<EventEditor>` there should be a new entry for adding this times. the time added will be the value (in milliseconds) from the end of the event
- connect this to the `<MultiProgressBar>`
- cleanup the `<ViewSettingsForm>` to remove the time input, the colours definition should remain there. We might also need to cleanup data in the endpoints

## Next
I expect that once this is complete, we can reuse the times for the alarms mentioned in #131 
This will be a welcome addition to the feature set of the app

kellhogs:
This PR add's the ability to set "Warning" and "Danger" times for the progress-bar per event.

created the necessary endpoints in the backend for the runtimeStore, parser and types,
added the necessary inputs in the event editor,
progress-bar uses new values from eventNow

cleaned old endpoint and ui elements up

closes #624


apps/client/src/common/components/input/time-input/TimeInput.tsx
```diff
@@ -26,13 +26,17 @@ function ButtonInitial(name: TimeEntryField) {
   if (name === 'timeStart') return 'S';
   if (name === 'timeEnd') return 'E';
   if (name === 'durationOverride') return 'D';
+  if (name === 'timeWarning') return 'Wa';
+  if (name === 'timeDanger') return 'Da';
```
cpvalente, line 30: I think we shouldnt use the element with the button for this.
The button was meant to lock the data input (for example to link events), I dont see a use case for actions here

apps/client/src/features/cuesheet/cuesheet-progress/CuesheetProgress.tsx
```diff
@@ -14,9 +14,9 @@ export default function CuesheetProgress() {
       now={timer.current}
       complete={totalTime}
       normalColor={data!.normalColor}
-      warning={data!.warningThreshold}
+      warning={timer.timeWarning || 0}
```
cpvalente, line 17: would it not be best if the progressbar handled undefined?

cpvalente (commented):
I havent yet tested but code looks solid. good work

Left a small comment for you, meanwhile: should we use one of the other terms instead of alert and danger?

apps/client/src/features/cuesheet/cuesheet-progress/CuesheetProgress.tsx
kellhogs: i agree

apps/client/src/features/viewers/timer/Timer.tsx
```diff
@@ -176,9 +176,9 @@ export default function Timer(props: TimerProps) {
           now={time.current}
           complete={totalTime}
           normalColor={viewSettings.normalColor}
-          warning={eventNow?.timeWarning ?? 0}
+          warning={eventNow!.timeWarning}
```
cpvalente, line 179: do we have to do this?

I wonder if the type could be simply
`warning?: number;`
if we want to save it as nullable is also ok, just one more thing to check

apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.tsx
```diff
@@ -19,8 +19,8 @@ export default function MultiPartProgressBar(props: MultiPartProgressBar) {
 
   const percentComplete = 100 - clamp(100 - (Math.max(now ?? 0, 0) * 100) / complete, 0, 100);
 
-  const dangerWidth = clamp((danger / complete) * 100, 0, 100);
-  const warningWidth = clamp((warning / complete) * 100, 0, 100);
+  const dangerWidth = clamp(((danger || 0) / complete) * 100, 0, 100);
+  const warningWidth = clamp(((warning || 0) / complete) * 100, 0, 100);
```
cpvalente, line 23: would this not be 0 if there is no warning / danger? could we instead write

```ts
const dangerWidth = danger ? clamp(((danger) / complete) * 100, 0, 100) : 0
```

kellhogs: yours is easier to read so sure i'll go with that

apps/client/src/features/viewers/timer/Timer.tsx
kellhogs: We need to check the eventNow here because it's possibly null.

cpvalente: in that case you cant use the null assertion and just pass `warning={eventNow?.timeWarning}` so that it will be undefined if there is no event now

apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
```diff
@@ -135,8 +135,8 @@ export default function MinimalTimer(props: MinimalTimerProps) {
   const showFinished = finished && !userOptions?.hideOvertime && (time.timerType !== TimerType.Clock || showEndMessage);
 
   const showProgress = time.playback !== Playback.Stop;
-  const showWarning = (time.current ?? 1) < viewSettings.warningThreshold;
-  const showDanger = (time.current ?? 1) < viewSettings.dangerThreshold;
+  const showWarning = (time.current ?? 1) < (time.timeWarning ?? 0);
+  const showDanger = (time.current ?? 1) < (time.timeDanger ?? 0);
```
cpvalente, line 139: would this not meant that 
if the timer is not running and the user hasnt set a warning timer, we are in `showWarning`?

I wonder if the logic here can be simplified

apps/server/src/utils/parser.ts
```diff
@@ -149,6 +151,8 @@ export const parseExcel = (excelData: unknown[][], options?: Partial<ExcelImport
         [importMap.user7]: (index: number) => (user7Index = index),
         [importMap.user8]: (index: number) => (user8Index = index),
         [importMap.user9]: (index: number) => (user9Index = index),
+        [importMap.timeWarning]: (index: number) => (timeWarningIndex = index),
+        [importMap.timeDanger]: (index: number) => (timeDangerIndex = index),
```
cpvalente, line 155: nice! does this means we also need to update the components related to the sheet import?

cpvalente (commented):
good work! we need to double check that there are no missing bits in the sheet import

good stuff

cpvalente:
Hi @alex-Arc , can you give me your opinion here?

apps/server/src/utils/parser.ts
kellhogs: If i'm not mistaken no ... I have tested importing and exporting rundowns and everything seems to be working fine.

cpvalente: what I am calling the sheet import is the import of excel xlsx, you need to give the columns to import.
I havent seen this defined anywhere or added to the preview items

apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
kellhogs: Shouldn't this come out as false if there is no timeWarning? Since time.current isn't smaller that timeWarning then. Or in other words, we compare if time.current is smaller than 0 so we won't ever go into showWarning. The only time when this would be true is if the timer goes negative, which is when we are in the red overtime anyway.

cpvalente: timer is loaded (not playing) and there is no `timeWarning`

```ts
const showWarning = (time.current ?? 1) < (time.timeWarning ?? 0);
const showWarning = (null ?? 1) < (null ?? 0);
const showWarning = (1) < (0);
const showWarning = false;
```

timer is loaded (not playing) and there is `timeWarning`

```ts
const showWarning = (time.current ?? 1) < (time.timeWarning ?? 0);
const showWarning = (null ?? 1) < (6000 ?? 0);
const showWarning = (1) < (6000);
const showWarning = true;
```

apps/server/src/utils/parser.ts
kellhogs: will be solved in a separate pr

apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
kellhogs: This logic checks out. But if a timer is loaded we already have the time.current information, which is at that point the starting time of the timer

cpvalente: Sorry, you are correct. I was under the impression that loaded events had null current time payload, but that is not the case

alex-Arc:
> Hi @alex-Arc , can you give me your opinion here?

do you still need me to look?

cpvalente:
> do you still need me to look?

We are only releasing this for v3, so it would be good to have your input

alex-Arc:
should we prevent incorrect timer order? As we can't display this
image
 
can we create a default setting? (maybe the same fields from the settings in v2)

We could call them `Zone 1` and `Zone 2` if we wanted more neutral names 

settings -> views -> timer colours can't be saved

cpvalente:
Hi @alex-Arc , good tips here
I agree that defaults are important and we should find some sensible ones

Name changes are also important, but I wonder if `Alert1` and `Alert2` would be more in line with how we want this dataset to be leveraged

PR #678, opened 2023-12-29
Fix Visual bug on rundown

### LINKED ISSUE #676: Visual Bug: Rundown editor
image

When trying to edit an event that is in the "past" in the current rundown you get this visual bug and are not able to access the options that are behind the other element.

The opacity change for the "past" rundown entry is most likely to cause this issue.

kellhogs:
Im almost certain there is a smarter way to solve this ..... but this works too

closes #676

### DIFF
--- a/apps/client/src/features/rundown/event-block/EventBlock.module.scss
+++ b/apps/client/src/features/rundown/event-block/EventBlock.module.scss
@@ -54,8 +54,21 @@ $skip-opacity: 0.1;
   }
 
   &.past:not(.skip) {
-    opacity: 0.6;
+    .delayNote,
+    .statusElements,
+    .eventTitle,
+    .eventNote,
+    .eventTimers,
+    .eventStatus,
+    .playbackActions,
+    .binder {
+      opacity: 0.2;
+    }
   }
+
+  //   &.past:not(.skip) {
+  //     opacity: 0.6;
+  // }
 
   &.skip {
     border: 1px solid $white-3;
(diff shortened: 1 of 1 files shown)


apps/client/src/features/rundown/event-block/EventBlock.module.scss
```diff
@@ -54,7 +54,16 @@ $skip-opacity: 0.1;
   }
 
   &.past:not(.skip) {
```
cpvalente, line 56: nice!

if what we are trying to do is stop the opacity in an element, could we make it explicit that way?
```
&.past:not(.skip) {
  opacity: 0.4;

  .elementWithoutOpacity {
    opacity: 1;
  }
}
```

kellhogs: for some weird reason this doesn't work. The element i'm sparing with the opacity is the .eventActions

if i write it like you proposed, it doesn't overwrite the opacity on these elements. Even with !Important

someone who is smarter with css than me could probably figure this out.

cpvalente: hmm, probably, but probably not worth losing too much sleep over it
Would you mind adding a comment as to why this is necessary like?
/* we stop the eventActions from having opacity to fix issue with dropdown drawing order */

kellhogs: all done

cpvalente (approved):
🚀

PR #679, opened 2023-12-29
Project files tweaks

arymoraes:
- Fixed issue where the app would crash if we did not have a `Ontime` folder
- Listing files from directory function now accepts optional `allowedExtensions` parameter
- Project list endpoint now only lists `json` files
- Deletes `xlsx` file after upload
- Added last loaded project file to the project list endpoint response


packages/types/src/api/ontime-controller/BackendResponse.type.ts
```diff
@@ -21,6 +21,11 @@ export type ProjectFile = {
 
 export type ProjectFileList = Array<ProjectFile>;
 
+export type ProjectFileListResponse = {
+  files: ProjectFileList;
+  lastLoadedProject: string | null;
```
cpvalente, line 26: There always has to be an open project. in which cases there is not?
How do we handle not having a project? do we create one ourselves? what is it called?

apps/server/src/utils/getFileListFromFolder.ts
```diff
@@ -1,9 +1,23 @@
...
+export const getFileListFromFolder = (folderPath: string, options: Options = {}): Array<ProjectFile> => {
+  let files = readdirSync(folderPath);
+
+  // Filter files if allowedExtensions is provided
+  if (options.allowedExtensions && options.allowedExtensions.length > 0) {
+    files = files.filter((file) => {
+      const ext = extname(file).toLowerCase();
+      return options.allowedExtensions.includes(ext);
+    });
+  }
```
cpvalente, line 19: I might be missing something, but would we not return undefined if the options are not provided?

Would it be best to simplify this to our use case?
I would think about two functions
```
function getFilesFromFolder() // all the node file system stuff
function filterProjectFiles() // the actual logic here
```

we could do unit tests on the second one and potentially generalise it to filter any file

What do you think?

apps/server/src/setup.ts
```diff
@@ -71,14 +72,15 @@ export const currentDirectory = dirname(__dirname);
...
-const lastLoadedProjectConfigPath = join(getAppDataPath(), 'config.json');
+export const lastLoadedProjectConfigPath = join(getAppDataPath(), 'config.json');
 
 let lastLoadedProject;
 
 try {
   lastLoadedProject = JSON.parse(fs.readFileSync(lastLoadedProjectConfigPath, 'utf8')).lastLoadedProject;
 } catch {
   if (!isTest) {
+    ensureDirectory(getAppDataPath());
```
cpvalente, line 83: good catch, there starts to be a lot here, do you think we should have a setup function that makes sure we have everything we need?

cpvalente (commented):
Good work @arymoraes , left some small questions for you

packages/types/src/api/ontime-controller/BackendResponse.type.ts
arymoraes: That's a very good point, for now when we never import a project we just use the regular `db.json`, and `lastLoadedProject` is `null`.

But maybe the best solution would be to create a `default.json` (maybe thats a decent name?) instead

cpvalente: yes, I dont see an issue with it. The user would be able to rename it if they want.
Could we run into trouble with potential duplicate names?

apps/server/src/utils/getFileListFromFolder.ts
arymoraes: my first idea was to have a function that we could pass the extensions argument, and if we didn't we'd just return everything

but I feel your idea is better, it's cleaner and also makes it easier to test. We could keep `getFilesFromFolder` private to the module for now, and only expose `filterProjectFiles`, what do you think?

packages/types/src/api/ontime-controller/BackendResponse.type.ts
arymoraes: perhaps, maybe we could have a warning if the user tries to upload a file with a name that already exists on the uploads folder?

if we want to avoid duplicates on the default one, we could have a timestamp on the default filename maybe?

apps/server/src/setup.ts
arymoraes: yeah imo that would be a good idea

apps/server/src/utils/getFileListFromFolder.ts
cpvalente: sounds good to me.
I guess in that case we might need two private functions `getFilesFromFolder` and `filterProjectFiles`
which we expose as `getProjectFiles`

But I am sure you will find a good way to do it. We just dont want to mess around with mocking too much and is better to extract the business logic from the accidental logic

apps/server/src/setup.ts
cpvalente: I will make a note, maybe this is something you could help with later 😅

packages/types/src/api/ontime-controller/BackendResponse.type.ts
cpvalente: I dont have too many opinions here, I am sure it is ok.
For the backend upload we could just add in increment like your OS does `default (1).json`

arymoraes: will do that, I've been trying for the past hour or so but have ran into some issues, I'll try again tomorrow with a clear mind

cpvalente: sure, I would suggest that you keep in simple for now, we can refine it later
it is ok with me if you just thrown an error if the file exists, we can deal with it later

cpvalente:
Thank you @arymoraes, I am not sure what is the scope you want for this PR. It seems that all the items in the description are completed?

Let me know when you think it is ready to merge so that we can continue with the next steps. 
Meanwhile I keep the UI PR open and you are welcome to consult

apps/server/src/utils/__tests__/getFileListFromFolder.test.ts
```diff
@@ -0,0 +1,34 @@
+import { expect, vi } from 'vitest';
+
+import { getProjectFiles } from '../getFileListFromFolder.js';
+
+describe('getProjectFiles test', async () => {
+  it('should return a list of project .json files', async () => {
```
cpvalente, line 6: I am not sure mocking the fs stuff provides much value. Perhaps we take advantage of the structure and test only the filter part?

apps/server/src/utils/getFileListFromFolder.ts
```diff
@@ -1,11 +1,28 @@
...
+};
+
+const filterProjectFiles = (files: Array<string>): Array<string> => {
+  return files.filter((file) => {
+    const ext = extname(file).toLowerCase();
+    return ext === '.json';
+  });
+};
+
+export const getProjectFiles = (): Array<ProjectFile> => {
```
cpvalente, line 18: I am not sure we should use the synchronous versions of the functions here

We did in reading files, since we are about to change the source of data of the app. Meaning that any other requests would have its source of truth invalidated by this operation

This is not the case here. We are just getting a list of files and will likely not want to block the process with this

apps/server/src/utils/__tests__/getFileListFromFolder.test.ts
arymoraes: fair, honestly with this I'm only testing the filtering anyway since mocking the `fs` mocking doesn't help to test the `fs` part

packages/types/src/api/ontime-controller/BackendResponse.type.ts
arymoraes: made it so it throws an error for now. I had logic working for it but it had a bug that it would also modify the previous file

commit for reference is https://github.com/cpvalente/ontime/pull/679/commits/b896a9b5f543b9e3bc49a9e79decc5ca43c2c780

we can revisist this later

arymoraes:
> Thank you @arymoraes, I am not sure what is the scope you want for this PR. It seems that all the items in the description are completed?
> 
> Let me know when you think it is ready to merge so that we can continue with the next steps. Meanwhile I keep the UI PR open and you are welcome to consult

I addressed the feedback and it should be good to go. I'll work on the new step 5 stuff now

cpvalente:
@arymoraes , did we not agree to simplify the tests in the `getProjectList`? I still see the fs mocks there

arymoraes:
> @arymoraes , did we not agree to simplify the tests in the `getProjectList`? I still see the fs mocks there

my point on the comment is that the tests on `getProjectList` are already testing only the filter, since we're mocking `fs`.

If we wanted to test only `filterProjectFiles` we'd have to export it, but I'm not too keen on the idea of exporting a function just so we can test it.

imho the current tests are basically testing `filterProjectFiles`, just not directly

PR #682, opened 2023-12-30
refactor: switch out steno for lowdb

alex-Arc:


cpvalente (approved):
Thank you @alex-Arc 
Can you make the follow up to update lowdb? It should be a simple name change

PR #684, opened 2023-12-30
Keep several project files in user folder - Steps 5A-E

arymoraes:
> * [x]  **5.A:** create backend endpoint to create a new project file with a given name
> * [x]  **5.B:** create backend endpoint to duplicate a project file, ie: given an existing project file name, it would create a copy with the new name
> * [x]  **5.C:** create backend endpoint to delete project file. we need to be careful about deleting the current loaded file. Either by creating a new one of by refusing the delete
> * [x]  **5.D:** create backend endpoint to change name of a file

https://github.com/cpvalente/ontime/issues/374


arymoraes:
@cpvalente question about `5A` and `5E`:

for `5A`, it would be a file upload, so the request body will be: `file` and `filename`, and we create a file with the `filename` we pass on the req body, correct? So basically like our upload endpoint but we'd be able to choose the filename.

for `5E`, I'm assuming its similar to `5A`, what would be the difference there?

cpvalente:
> for `5A`, it would be a file upload, so the request body will be: `file` and `filename`, and we create a file with the `filename` we pass on the req body, correct? So basically like our upload endpoint but we'd be able to choose the filename.


My idea with `5A` was to have a `make new project file` endpoint, with no upload, just a file name and maybe some initial data.
So the request would be like

> for `5E`, I'm assuming its similar to `5A`, what would be the difference there?

in `5E` we would pass the name of the file to be duplicated and a new name (from and to). The backend would generate a copy of the file
Potentially we can simplify by making duplicate always generate an incremental number, and letting the user leverage the rename endpoint to rename it later

arymoraes:
> > for `5A`, it would be a file upload, so the request body will be: `file` and `filename`, and we create a file with the `filename` we pass on the req body, correct? So basically like our upload endpoint but we'd be able to choose the filename.
> 
> My idea with `5A` was to have a `make new project file` endpoint, with no upload, just a file name and maybe some initial data. So the request would be like
> 
> > for `5E`, I'm assuming its similar to `5A`, what would be the difference there?
> 
> in `5E` we would pass the name of the file to be duplicated and a new name (from and to). The backend would generate a copy of the file Potentially we can simplify by making duplicate always generate an incremental number, and letting the user leverage the rename endpoint to rename it later

oh sorry my bad, about `5E` I meant its similar to `5B`, isn't it?

cpvalente:
yes, sorry, you are correct. 5B and 5E are the same. my bad

arymoraes:
> yes, sorry, you are correct. 5B and 5E are the same. my bad

sounds good, thanks for the clarification, I have all I need now to keep working on them

kellhogs:
Hey, @cpvalente mentioned this could have been caused by your last PR:

I am currently not able to upload the demo excel file from the docs. Here's the error i get 
image

arymoraes:
> Hey, @cpvalente mentioned this could have been caused by your last PR:
> 
> I am currently not able to upload the demo excel file from the docs. Here's the error i get image

Hey, do you have the most up to date `v3`? I had a PR merged today that fixes issues with the excel imports.

It did work for me on the latest `v3` version

image

cpvalente:
> Hey, do you have the most up to date `v3`? I had a PR merged today that fixes issues with the excel imports.
> 
> It did work for me on the latest `v3` version
> 

I wonder if the issue is that he had already imported an excel file in a previous version of the app. So the file already exists in the folder and the backend doesnt have necessary guards to account for this

kellhogs:
hmmm weird i pretty sure i was in the latest v3. but after pulling the last update where my pr got merged its working now

cpvalente:
great to hear. @arymoraes , do you have discord? we sometimes chat there

arymoraes:
> great to hear. @arymoraes , do you have discord? we sometimes chat there

Yep it's fallenefc

apps/server/src/controllers/ontimeController.config.ts
```diff
@@ -0,0 +1,68 @@
+export const emptyProject = {
```
cpvalente, line 1: we should not introduce more variations on the source for our models
https://github.com/cpvalente/ontime/blob/master/apps/server/src/models/dataModel.ts

apps/server/src/controllers/ontimeController.validate.ts
```diff
@@ -166,3 +171,137 @@ export const validateLoadProjectFile = [
...
+/**
+ * @description Validates the existence of project files.
+ * @param {object} projectFiles
+ * @param {string} projectFiles.projectFilename
+ * @param {string} projectFiles.newProjectFilename
+ *
+ * @returns {Promise<Array<string>>} Array of errors
+ *
+ */
+export const validateProjectFiles = async (projectFiles: {
```
cpvalente, line 281: I understand this pattern, but I wonder if the controllers should be accessing files.
Would this not be out of scope for a controller?

apps/server/src/controllers/ontimeController.validate.ts
```diff
@@ -166,3 +171,137 @@ export const validateLoadProjectFile = [
...
+
+      if (newFileError) {
+        errors.push(newFileError);
+      }
+    }
+
+    return errors;
+  } catch (error) {
+    return [error.toString()];
+  }
```
cpvalente, line 306: Question: I am not sure I understand the reason to re-throw if we dont enrich the error. Maybe I am missing something?

cpvalente (commented):
Hey! I know you are not finished but thought of checking in
Hope that is ok, good work

apps/server/src/controllers/ontimeController.config.ts
arymoraes: oops didn't know this existed, I looked for it but apparently not enough

makes much more sense

apps/server/src/routes/ontimeRouter.ts
```diff
@@ -105,3 +113,15 @@ router.get('/projects', listProjects);
 
 // create route between controller and '/ontime/load-project' endpoint
 router.post('/load-project', validateLoadProjectFile, loadProject);
+
+// create route between controller and '/ontime/duplicate-project' endpoint
+router.post('/duplicate-project', validateProjectDuplicate, sanitizeProjectFilename, duplicateProjectFile);
```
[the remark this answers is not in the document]
arymoraes: yeah I agree with it

apps/server/src/controllers/ontimeController.validate.ts
arymoraes: I definitely agree this is out of scope for a controller. I was thinking but didn't reach a conclusion about where to move this kind of logic

apps/server/src/controllers/ontimeController.config.ts
```diff
@@ -0,0 +1,68 @@
...
+      ? newProjectFilename
+      : `${newProjectFilename}.json`;
+  }
+
+  if (projectName) {
+    req.params.projectName = projectName.includes('.json') ? projectName : `${projectName}.json`;
+  }
+
+  next();
+};
```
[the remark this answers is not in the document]
arymoraes: I made

```
export const ensureJsonExtension = (filename: string) => {
    if (!filename) return filename;
    
    return filename.includes('.json') ? filename : `${filename}.json`;
};
```

and we use it like this on the middleware, since we decoupled the json extension logic we can also unit test it

```
import { ensureJsonExtension } from "../utils/ensureJsonExtension.js";

export const sanitizeProjectFilename = (req, res, next) => {
  const { projectFilename, newProjectFilename } = req.body;
  const { projectName } = req.params;

  req.body.projectFilename = ensureJsonExtension(projectFilename);
  req.body.newProjectFilename = ensureJsonExtension(newProjectFilename);
  req.params.projectName = ensureJsonExtension(projectName);

  next();
};
```

apps/server/src/controllers/ontimeController.ts
```diff
@@ -511,3 +523,148 @@ export const loadProject: RequestHandler = async (req, res) => {
...
+ * Receives the original project filename (`projectFilename`) and the filename for the duplicate (`newProjectFilename`) from the request body.
+ *
+ * @param {Request} req - The express request object. Expects `projectFilename` and `newProjectFilename` in the request body.
+ * @param {Response} res - The express response object. Sends a 200 status with a success message upon successful duplication,
+ *                         a 404 status if the original file is not found, a 409 status if the duplicate file name already exists,
+ *                         or a 500 status with an error message in case of an exception.
+ */
+export const duplicateProjectFile: RequestHandler = async (req, res) => {
+  try {
+    const { projectFilename, newProjectFilename } = req.body;
```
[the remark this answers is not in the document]
arymoraes: so `duplicate`, `delete`, `rename`, and `create` (or `createEmpty`) for the fn names?

apps/server/src/controllers/ontimeController.validate.ts
arymoraes: yeah you're right this is silly

apps/server/src/controllers/ontimeController.validate.ts
```diff
@@ -166,3 +171,137 @@ export const validateLoadProjectFile = [
...
+const checkExistingFile = async (projectFilename?: string): Promise<string | null> => {
+  const projectFilePath = join(uploadsFolderPath, projectFilename);
+
+  try {
+    const fh = await open(projectFilePath);
+    close(fh.fd);
+  } catch (error) {
+    return 'Project file does not exist';
+  }
+};
```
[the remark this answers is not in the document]
arymoraes: which ones? I didn't find any that checked for an existing file, only for directory, or `ensureDirectory` also checks for files? I think so right? it checks for a path

apps/server/src/routes/ontimeRouter.ts
```diff
@@ -105,3 +113,15 @@ router.get('/projects', listProjects);
...
+
+// create route between controller and '/ontime/rename-project' endpoint
+router.put('/rename-project', validateProjectRename, sanitizeProjectFilename, renameProjectFile);
+
+// create route between controller and '/ontime/create-project' endpoint
+router.post('/create-project', validateProjectCreate, sanitizeProjectFilename, createProjectFile);
+
+// create route between controller and '/ontime/delete-project' endpoint
+router.delete('/delete-project/:projectName', sanitizeProjectFilename, deleteProjectFile);
```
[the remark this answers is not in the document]
arymoraes: for delete should we use `ontime/project/:id/delete` if we're following the suggestion above?

apps/server/src/controllers/ontimeController.ts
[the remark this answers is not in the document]
arymoraes: Oh you meant for the keys in the object body right? Yeah makes sense

apps/server/src/routes/ontimeRouter.ts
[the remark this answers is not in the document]
cpvalente: delete is a HTTP request, so I suggest `DELETE` `ontime/project/:id`

apps/server/src/controllers/ontimeController.validate.ts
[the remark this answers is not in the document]
cpvalente: you might be correct, maybe I am confusing with checking for a directory.
Could we use `existsSync` from `fs` instead of opening the file?

apps/server/src/controllers/ontimeController.validate.ts
cpvalente: I guess this is not objectively better. but the way we have developed so far is that
- the controllers make sure the data has the right shape
- all domain validation happens at service level

This trades off failing early (the requests make it inside the domain to be rejected) by having all the business logic in one place

In this case we dont have a service, so I guess we could have the logic in the controller function body as you did in `loadProject`
What do you think?

apps/server/src/controllers/ontimeController.validate.ts
```diff
@@ -166,3 +171,137 @@ export const validateLoadProjectFile = [
...
+
+/**
+ * @description Validates the filenames for duplicating a project.
+ */
+export const validateProjectDuplicate = [
+  body('projectFilename')
+    .exists()
+    .withMessage('Project filename is required')
+    .isString()
+    .withMessage('Project filename must be a string'),
```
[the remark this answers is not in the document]
arymoraes: makes sense, do you have a length max in mind?

OS limits apparently are 255

```
BTRFS   255 bytes
exFAT   255 UTF-16 characters
ext2    255 bytes
ext3    255 bytes
ext3cow 255 bytes
ext4    255 bytes
FAT32   8.3 (255 UCS-2 code units with VFAT LFNs)
NTFS    255 characters
XFS     255 bytes
```

apps/server/src/controllers/ontimeController.validate.ts
arymoraes: I think it's probably fine to keep it in the controller body.

Alternatively we could have a fileService that does this kind of thing, since its mostly handling files anyway. I'm fine with keeping it in the controller for now

apps/server/src/controllers/ontimeController.validate.ts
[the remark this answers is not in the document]
cpvalente: oh, I meant the other way, we want to check that the name has at least 1 (?) character.
I wonder if an empty string would pass the isString validation.

I dont think we should worry about the max characters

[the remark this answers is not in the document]
arymoraes: makes sense,

I added a `min: 1`, so it should. I'll double check either way.

Maybe keeping `max: 255` isn't a bad idea, just in case, since the OS might not be able to handle it

apps/server/src/controllers/ontimeController.config.ts
```diff
@@ -0,0 +1,12 @@
...
+export const sanitizeProjectFilename = (req, res, next) => {
+  const { filename, newFilename } = req.body;
+  const { filename: projectName } = req.params;
+
+  req.body.filename = ensureJsonExtension(filename);
+  req.body.newFilename = ensureJsonExtension(newFilename);
+  req.params.filename = ensureJsonExtension(projectName);
+
+  next();
+};
```
cpvalente, line 12: question: i am unsure on why the difference between the config and validate files. should we simplify and move this to the ontimeController.validate.ts?
Or perhaps call it a middleware instead?

What do you think?

apps/server/src/controllers/ontimeController.validate.ts
```diff
@@ -166,3 +170,101 @@ export const validateLoadProjectFile = [
...
+
+    next();
+  },
+];
+
+const checkExistingProjectFile = (projectFilename?: string): boolean => {
+  const projectFilePath = join(uploadsFolderPath, projectFilename);
+
+  return existsSync(projectFilePath);
+};
```
cpvalente, line 240: question: I am not entirely sure this function needs to exist, what do you think?

apps/server/src/routes/ontimeRouter.ts
```diff
@@ -105,3 +113,15 @@ router.get('/projects', listProjects);
...
+// create route between controller and '/ontime/duplicate-project' endpoint
+router.post('/project/:filename/duplicate', validateProjectDuplicate, sanitizeProjectFilename, duplicateProjectFile);
+
+// create route between controller and '/ontime/rename-project' endpoint
+router.put('/project/:filename/rename', validateProjectRename, sanitizeProjectFilename, renameProjectFile);
+
+// create route between controller and '/ontime/create-project' endpoint
+router.post('/project', validateProjectCreate, sanitizeProjectFilename, createProjectFile);
+
+// create route between controller and '/ontime/delete-project' endpoint
```
cpvalente, line 126: todo: remember to update the doc strings...

apps/server/src/utils/__tests__/ensureJsonExtension.test.ts
```diff
@@ -0,0 +1,28 @@
+import { describe, it, expect } from 'vitest';
+import { ensureJsonExtension } from '../ensureJsonExtension.js';
+
+describe('ensureJsonExtension', () => {
```
cpvalente, line 4: nice!

cpvalente (commented):
I think you have done great work here.
There are a few small todos that should be pretty quick, no code changes

Once that is done, I would like to merge it to implement the frontend, from there we might see that some tweaks are needed. is this ok with you?

apps/server/src/controllers/ontimeController.config.ts
arymoraes: my idea is that validate is only responsible for validation, it doesn't change anything, while this one changes data

I think a middleware is probably a better idea then

cpvalente: Sounds good. Just remember we are doing the fastify migration. We don't want to do double work here, it is ok with me to leave this as a function for now and then move it to a middleware in fastify

arymoraes: tbh it's already a middleware, just on the config file.

I moved it to the utils folder (like other middleware we have), and we can place it in a proper middleware folder when we do the migration

apps/server/src/controllers/ontimeController.validate.ts
arymoraes: you're right, it doesn't need to be its own function now that we reduced the logic for that check

PR #689, opened 2024-01-01
build: upgrade lowdb

alex-Arc:


apps/server/src/modules/loadDb.ts
```diff
@@ -53,7 +53,7 @@ async function loadDb() {
   const dbInDisk = populateDb();
 
   const adapter = new JSONFile<DatabaseModel>(dbInDisk);
-  const db = new Low(adapter);
+  const db = new Low(adapter, dbModel);
```
alex-Arc, line 56: i'm unsure if this is correct?

apps/server/src/services/ConfigService.ts
```diff
@@ -15,7 +15,7 @@ class ConfigService {
   constructor() {
     this.configPath = join(getAppDataPath(), 'config.json');
     const adapter = new JSONFile<Config>(this.configPath);
-    this.config = new Low<Config>(adapter);
+    this.config = new Low<Config>(adapter, null);
```
alex-Arc, line 18: same here

cpvalente: maybe a `{}` instead?

cpvalente (commented):
i think this seems correct
https://github.com/typicode/lowdb?tab=readme-ov-file#new-lowadapter-defaultdata

we could try [the preset syntax](https://github.com/typicode/lowdb?tab=readme-ov-file#typescript) if we feel is simpler. But as long as it works I dont see a need for it

alex-Arc:
> i think this seems correct

cool, it works for me
so if you could give it a try I think we are good

PR #692, opened 2024-01-02
global review

cpvalente:
small PR to align usage of new state mutations

### DIFF
--- a/apps/server/src/state.ts
+++ b/apps/server/src/state.ts
@@ -51,7 +51,7 @@ export const state: TState = {
 
 export const stateMutations = {
   timer: {
-    // utility to allow modifying the state from the ouside
+    // utility to allow modifying the state from the outside
     patch(timer: Partial<TimerState>) {
       mutate((state) => {
         for (const key in timer) {
--- a/apps/server/src/state.ts
+++ b/apps/server/src/state.ts
@@ -64,6 +64,8 @@ export const stateMutations = {
     start() {
       mutate(
         (state) => {
+          // TODO: we need an event to start, should we get the event or the whole rundown?
+
           state.timer.clock = clock.timeNow();
           state.timer.secondaryTimer = null;
           state.timer.secondaryTarget = null;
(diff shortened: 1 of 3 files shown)


apps/server/src/state.ts
```diff
@@ -22,6 +24,7 @@ type TState = DeepReadonly<{
 }>;
 
 export const state: TState = {
+  // QUESTION: should merge playback into the timer?
```
arymoraes, line 27: if its only a property then yeah why not?

apps/server/src/state.ts
```diff
@@ -132,49 +143,63 @@ export const stateMutations = {
...
+        state.playback = restorePoint.playback;
+        state.timer.pausedAt = restorePoint.pausedAt;
+        state.timer.addedTime = restorePoint.addedTime;
+
+        // check if event finished meanwhile
         if (state.timer.timerType === TimerType.TimeToEnd) {
-          const now = clock.timeNow();
-          state.timer.current = getCurrent(now, state.timer.duration, 0, 0, now, timer.timeEnd, state.timer.timerType);
+          state.timer.current = getCurrent(
+            state.timer.clock, // TODO: should this be startedAt?
```
arymoraes, line 168: from what I've seen it seems like it

one thing we can think of is to refactor `getCurrent` to use named arguments since we have 7 arguments there, would be easier and less confusing when we need to call it

apps/server/src/state.ts
```diff
@@ -132,49 +143,63 @@ export const stateMutations = {
...
         }
       });
     },
-    load(timer: OntimeEvent) {
+    load(event: OntimeEvent) {
       mutate(
         (state) => {
-          state.timer.selectedEventId = timer.id;
-          state.timer.timeEnd = timer.timeEnd;
+          // TODO: resume and load logic are very similar
```
arymoraes, line 182: I feel they have some similarities, but might be confusing to transform it to a single fn, except for this block which is exactly the same

```
          if (state.timer.timerType === TimerType.TimeToEnd) {
            state.timer.current = getCurrent(
              state.timer.clock,
              state.timer.duration,
              0,
              0,
              state.timer.clock,
              event.timeEnd,
              state.timer.timerType,
            );
          }
        },
```

PR #693, opened 2024-01-03
Refactor: express -> fastify

alex-Arc:
I made a test of migrating from express to fastify
Using the [@fastify/express](https://www.npmjs.com/package/@fastify/express) plugin to wrap all the current express routs so we can move them over one at a time

I switched to using schemas to validate the requests 
and [JsonSchemaToTsProvider](https://github.com/ThomasAribart/json-schema-to-ts) to then derive the types
don't know if this is the way we want to go?

- [x] websocket
- [x] /events
- [x] /project
- [x] /ontime
- [x] /api
- [ ] /external
- [ ] app.use(express.urlencoded({ extended: true }));
- [ ] app.use(express.json({ limit: '1mb' }));
- [ ] app.disable('x-powered-by');
- [ ] compression
- [x] `*`


cpvalente:
> I made a test of migrating from express to fastify Using the [@fastify/express](https://www.npmjs.com/package/@fastify/express) plugin to wrap all the current express routs so we can move them over one at a time
> 
> I switched to using schemas to validate the requests and [JsonSchemaToTsProvider](https://github.com/ThomasAribart/json-schema-to-ts) to then derive the types don't know if this is the way we want to go?
> 
> * [x]  websocket
> * [x]  /events
> * [x]  `*`

How about the validation stuff? how would we migrate that?
could we also remove body parser?

alex-Arc:
> could we also remove body parser?

yes I think so, don't know if it solves all our validation needs, but all where we just require a parameter to be preset of a specific type

```typescript
export const rundownSwapSchema = {
  body: {
    type: 'object',
    required: ['from', 'to'], //<-- this will result in a 400 if it is missing from the http request
    properties: {
      from: { type: 'string' },
      to: { type: 'string' },
    },
  },
} as const;
```

I can also be set to only allow specific strings
```typescript
export const rundownPostSchema = {
  body: {
    type: 'object',
    required: ['type'],
    properties: {
      type: { type: 'string', enum: [SupportedEvent.Block, SupportedEvent.Delay, SupportedEvent.Event] },
      after: { type: 'string' },
      id: { type: 'string' },
    },
  },
} as const;
```

cpvalente:
Hi @arymoraes, do you mind taking a look here and maybe giving some advice?

This will also conflict with your changes so it is a good for you to be aware

arymoraes:
> Hi @arymoraes, do you mind taking a look here and maybe giving some advice?
> 
> This will also conflict with your changes so it is a good for you to be aware

From what Ive seen this is largely good, being able to validate using a json schema would be really helpful. The code so far seems very similar to the docs so seems pretty good.

alex-Arc:
> From what Ive seen this is largely good, being able to validate using a json schema would be really helpful. The code so far seems very similar to the docs so seems pretty good.

That's good to hear.

Do you know of a way to reuse out TS types as schemas?

PR #695, opened 2024-01-05
fix: focus cursor

### LINKED ISSUE #614: Improve focus on text fields in /editor
## CONTEXT
In the `/editor` part of the app, the user is expected to input potentially large amounts of data.

We attempt to simplify it by forcing the selection of the field elements on focus.
This means that if the user clicks on an input field, we should select all its contents so that they can be overridden.
https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/components/input/time-input/TimeInput.tsx#L71

Unfortunately, the process seems not to work consistently.
From my findings, 
- focusing an input field on a selected block selects the entire field contents. We can click again to set the cursor at a specific place in the input. This is the desired behaviour
- focusing an input field on a NOT SELECTED block momentarily selects the field contents, but it is overridden by making the block active. At which point, we lose the content selection

firefox_eNP1UQE7FN

## THE TASK

We want to make the behaviour consistent.

Regardless of whether the block is selected or not:
- On the first click, an input field should have all its contents selected
- On the second click, we should be able to add a cursor at any point

cpvalente:
Fixes an issue where the changing of the cursor overrides the already in focus input
Closes #614 

## Summary by CodeRabbit

- **Documentation**
  - Updated the required versions for `node` and `pnpm` in the development guidelines to ensure compatibility with the latest features and security updates.

### DIFF
--- a/apps/client/src/features/rundown/event-block/EventBlock.tsx
+++ b/apps/client/src/features/rundown/event-block/EventBlock.tsx
@@ -132,9 +132,19 @@ export default function EventBlock(props: EventBlockProps) {
 
   const binderColours = colour && getAccessibleColour(colour);
 
+  // move focus to element if necessary
   useEffect(() => {
-    if (hasCursor) {
-      handleRef?.current?.focus();
+    if (!hasCursor || handleRef?.current == null) {
+      return;
+    }
+
+    const elementInFocus = document.activeElement;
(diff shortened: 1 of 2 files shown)


arymoraes (approved):
looks pretty good, tested and it looks smooth too

PR #696, opened 2024-01-06
Refactor/time formatting

cpvalente:
This PR unifies logic for formatting times.
We will have two functions, 
- one for the timers, which always has the `hh:mm:ss` format 
- one that formats against a user provided format 

The first is used where performance is more important (timers and backend) 
While the second is intended for the views and schedule data, where users like to have input on format

TODO:
- [ ] Check integrations
- [x] user provides format string
- [x] migrate format settings
- [x] what happens if user format string is invalid?
- [ ] documentation! https://moment.github.io/luxon/#/formatting?id=table-of-tokens
- [x] should we migrate from luxon?


cpvalente:
Hey @alex-Arc , I just need to finish the tests and doublecheck integrations.
Do you have time to take a quick sanity check in case something has escaped me?

alex-Arc:
Nice, i think it functionality is good

there are a lot of file listed as changed, is there some in particular I should look at?

PR #697, opened 2024-01-07
Project manager

arymoraes:
## Scope
This initiative allows users to manage the project files inside Ontime.
Note that there isn't much domain-specific behaviour here, so we are free (and should) model around common flows in other applications

## Features

1. Allow users to create a new project
There should be a button in the interface that creates a project with a given name
We can reuse the interface from Quick Start

2. Display a list of available projects
We would like to display a list with all the projects that we have saved in the backend

3. Allow manipulating the project list
The user should be able to edit the project list with common actions `Delete`, `Download`, `Rename` and `Duplicate`
- Delete should contain a confirmation step ("This action is irreversible"). We should also think about what it means to delete the loaded project (maybe we dont allow for now?)
- Rename should prompt the user for a new name (we should be careful on renaming the loaded project)
- Duplicate should also prompt the user for a name
Suggestion: for prompting the user for a name, perhaps we replace the name field in the row with an input and replace the action button field with `Cancel` and `Confirm` buttons?

4. Allow loading a project in the list
In the project list, the user can select any file to become the currently loaded


cpvalente:
hey @arymoraes , can you rebase v3 into this branch? we just synced to master and there is potential for  issues

apps/client/src/features/app-settings/panel/project-panel/ProjectList.tsx
```diff
@@ -1,19 +1,42 @@
...
-import { IoEllipsisHorizontal } from '@react-icons/all-files/io5/IoEllipsisHorizontal';
+import { useMemo, useState } from 'react';
 
 import { useProjectList } from '../../../../common/hooks-query/useProjectList';
 import * as Panel from '../PanelUtils';
 
-import style from './ProjectPanel.module.scss';
+import ProjectListItem from './ProjectListItem';
+
+export type EditMode = 'rename' | 'duplicate' | null;
```
cpvalente, line 8: todo: since this is used in the context of the list items, I think it should be exported from there

I am not sure what is the behaviour of imports that are not explicitly marked as `type`, but moving the export could also avoid circular imports here

apps/client/src/features/app-settings/panel/project-panel/ProjectList.tsx
```diff
@@ -1,19 +1,42 @@
...
-  const currentlyLoadedIndex = files.findIndex((project) => project.filename === lastLoadedProject);
-  const projectFiles = [...files];
-  const current = projectFiles.splice(currentlyLoadedIndex, 1)[0];
+  const [editingMode, setEditingMode] = useState<EditMode | null>(null);
+  const [editingFilename, setEditingFilename] = useState<string | null>(null);
+
+  const handleToggleEditMode = (editMode: EditMode, filename: string | null) => {
+    setEditingMode((prev) => (prev === editMode && filename === editingFilename ? null : editMode));
+    setEditingFilename(filename);
+  };
```
cpvalente, line 20: it seems that in `<ProjectListItem>` we short circuit submits where `filename` is falsy, would that not mean that we would never get a `null` here. Is the same not true for `EditMode`?

apps/client/src/features/app-settings/panel/project-panel/ProjectList.tsx
```diff
@@ -1,19 +1,42 @@
...
+
+  const reorderedProjectFiles = useMemo(() => {
+    if (!data?.files?.length) return [];
+
+    const currentlyLoadedIndex = files.findIndex((project) => project.filename === lastLoadedProject);
+    const projectFiles = [...files];
+    const current = projectFiles.splice(currentlyLoadedIndex, 1)?.[0];
+
+    return [current, ...projectFiles];
+  }, [data?.files?.length, files, lastLoadedProject]);
```
cpvalente, line 39: Question: since we pass the placeholder in the `useProjectList` hook, data will never be nullish, so I dont think we need the optional chaining?

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const handleSubmitDuplicate = async (values: ProjectFormValues) => {
+    try {
+      setSubmitError(null);
+
+      if (!values.filename) {
+        setSubmitError('Filename cannot be blank');
+        return;
+      }
+      await duplicateProject(filename, values.filename);
+      await onRefetch();
```
cpvalente, line 63: question: Don't we `refetch` as well in `<ProjectList>` in the `handleSubmit` function? Who should be responsible for fetching? maybe `<ProjectList>` since it is the one that provides the data?

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.module.scss
```diff
@@ -6,5 +6,42 @@
...
+  color: $red-500;
+}
+
+.createContainer {
+  padding: 0.5rem 0;
+}
+
+.saveButton {
+  text-transform: capitalize;
+}
```
cpvalente, line 41: todo: this style doesnt seem to apply or be necessary, i suggest we remove it

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,162 @@
...
+            onSubmit={editingMode === 'duplicate' ? handleSubmitDuplicate : handleSubmitRename}
+            onCancel={handleCancel}
+            submitError={submitError}
+          />
+        </td>
+      ) : (
+        <>
+          <td>{filename}</td>
+          <td>{createdAt}</td>
+          <td>{updatedAt}</td>
```
cpvalente, line 97: question: I was sure this used to be formatted? We dont need to do much fancy here, we can either use luxon or a simple `toLocaleDateString`

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.tsx
```diff
@@ -1,20 +1,78 @@
...
+
+  const handleSubmitCreate = async (values: ProjectFormValues) => {
+    try {
+      setSubmitError(null);
+
+      if (!values.filename) {
+        setSubmitError('Filename cannot be blank');
+        return;
+      }
+      await createProject(values.filename);
```
cpvalente, line 32: question: what is the expected behaviour for a user that creates a new project? Should the new project be loaded after the creation is successful?

cpvalente (commented):
Hey @arymoraes 

Great work here,
I just left a few small cleanup comments, nothing major and should be quick to tidy up.
Let me know if you need help

apps/client/src/features/app-settings/panel/project-panel/ProjectForm.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <div className={style.actionButtons}>
+          <Button onClick={onCancel} size='sm' variant='ontime-ghosted' disabled={isSubmitting}>
+            Cancel
+          </Button>
+          <Button
+            size='sm'
+            variant='ontime-filled'
+            isDisabled={!isDirty || !isValid || isSubmitting}
+            type='submit'
+            className={style.saveButton}
```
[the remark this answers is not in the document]
arymoraes:
```
.saveButton {
  text-transform: capitalize;
}
```

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.tsx
```diff
@@ -1,20 +1,78 @@
...
+  };
+
+  const handleSubmitCreate = async (values: ProjectFormValues) => {
+    try {
+      setSubmitError(null);
+
+      if (!values.filename) {
+        setSubmitError('Filename cannot be blank');
+        return;
+      }
```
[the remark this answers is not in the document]
arymoraes: actually trim will still try to submit if we have `'  '`

I made it 
```

      const filename = values.filename.trim();
      if (!filename) {
        setSubmitError('Project name cannot be empty');
        return;
      }
      await createProject(filename);
```

apps/client/src/features/app-settings/panel/project-panel/ProjectList.tsx
arymoraes: when we cancel, we set `editingFilename` to `null`, same for `editMode` is `null` if we're not editing anything

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
arymoraes: `ProjectListItem` handles submission for `duplicate` and `rename`, and `ProjectPanel` for `create`.

`ProjectList` is currently responsible for fetching

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.module.scss
arymoraes: Without this the buttons will be lower case, like `duplicate` or `rename`.

I could change the action names to be `Duplicate` | `Rename`, but I usually prefer to use all lowercase for action names, but we can change that

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
arymoraes: should it only display the date then? or datetime?

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.tsx
arymoraes: I thought it only created it, and left if there on the list

isn't creating + loading a part of quick start?

apps/client/src/features/app-settings/panel/project-panel/ProjectList.tsx
arymoraes: I moved the export to `ProjectListItem`

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.tsx
```diff
@@ -1,20 +1,78 @@
...
+                action='create'
+                filename=''
+                onSubmit={handleSubmitCreate}
+                onCancel={handleToggleCreate}
+                submitError=''
+              />
+              {submitError && <span className={style.createSubmitError}>{submitError}</span>}
+            </div>
+          ) : null}
+          {!isCreatingProject && <ProjectList />}
```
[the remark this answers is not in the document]
arymoraes: I agree

I am using this pattern and wanted your opinion if I should just move create logic to inside `ProjectList`:

  ```
const createForm = useMemo(() => {
    if (!isCreatingProject) return null;

    return (
      <tr className={style.createContainer}>
        <td colSpan={99}>
          <ProjectForm
            action='create'
            filename=''
            onSubmit={handleSubmitCreate}
            onCancel={handleToggleCreate}
            submitError=''
          />
          {submitError && <span className={style.createSubmitError}>{submitError}</span>}
        </td>
      </tr>
    );
  }, [handleSubmitCreate, handleToggleCreate, isCreatingProject, submitError]);

  return (
    <>
      <Panel.Header>Project</Panel.Header>
      <Panel.Section>
        <Panel.Card>
          <Panel.SubHeader>
            Manage projects
            <Button variant='ontime-filled' onClick={handleToggleCreate} isDisabled={isCreatingProject}>
              New
            </Button>
          </Panel.SubHeader>
          <ProjectList createForm={createForm} />
        </Panel.Card>
      </Panel.Section>
    </>
  );
}

```

I don't know how bad it is to pass the `JSX.Element` into the child, I remember in vue we had stuff like this called `slots`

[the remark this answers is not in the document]
cpvalente: Passing JSX elements is not inherently bad, and can be a good technique to improve performance.
In this case, performance isnt a problem so I think we should prioritise readability and capacity to change.

In the original code, you have multiple conditional statements which seem to belong to the same state, so it makes sense to me that we separe that state into a view.

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.tsx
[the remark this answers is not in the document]
cpvalente: 👍 
We can also add the trim on the process that makes the button enabled, so that the user cant submit the form to begin with. Maybe for next step

apps/client/src/features/app-settings/panel/project-panel/ProjectForm.tsx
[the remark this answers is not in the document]
cpvalente: oh, my bad... you are completely correct. I thought it was an UPPERCASE
Good work and apologies for the misunderstanding

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
cpvalente: hm, so why are we calling a refetch here?

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.module.scss
cpvalente: you are correct. sorry for the confusion

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
cpvalente: I think just creating the locale date like `23.08.19 16:12:30` is OK

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.tsx
cpvalente: yes, lets talk about this again when we work on quick start

cpvalente (approved):
Some quick feedback on your questions.
You can choose whether to move forwards now or make the tweaks you find necessary

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
arymoraes: not sure I understand, as of now with the changes:

`ProjectList` deals only with `create`, and calls a refetch for it
`ProjectListItem` handles `duplicate` and `rename`, so it calls the `refetch` callback after we made the query so it refreshes the list

arymoraes:
> Some quick feedback on your questions. You can choose whether to move forwards now or make the tweaks you find necessary

Adressing them was fairly quick so I did that, it's ready to merge

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
cpvalente: What I mean is that the fetching of the list is not in the scope of the `ProjectListItem`.

We already have the `onSubmit` callback which allows the parents to handle side effects, why not make the parent (which is responsible for fetching) call the refetch?

PR #699, opened 2024-01-08
fix: use background-color: transparent; for overlap indicators

alex-Arc:
image

should look like this

image

### DIFF
--- a/apps/client/src/features/rundown/event-block/EventBlockInner.tsx
+++ b/apps/client/src/features/rundown/event-block/EventBlockInner.tsx
@@ -141,7 +141,6 @@ const EventBlockInner = (props: EventBlockInnerProps) => {
             <Tooltip
               label={
                 <div>
-                  {' '}
                   {delayTime} <br />
                   New Time: {newTime}
                 </div>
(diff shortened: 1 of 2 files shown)


cpvalente (commented):
Would the user not get the tooltip on hovering? Would that not be a bit weird? Maybe it is best to conditionally render it so that they only show if they need to

What do you think?

apps/client/src/features/rundown/event-block/EventBlockInner.tsx
```diff
@@ -157,7 +157,7 @@ const EventBlockInner = (props: EventBlockInnerProps) => {
               }`}
             />
           </Tooltip>
-          <Tooltip label='Start time is later than end'>
+          <Tooltip label={timeStart > timeEnd ? 'Start time is later than end' : ''}>
             <div className={`${style.indicator} ${timeStart > timeEnd ? style.nextDay : ''}`} />
           </Tooltip>
```
cpvalente, line 162:
```suggestion
{timeStart > timeEnd && (
          <Tooltip label={'Start time is later than end'}>
            <div className={`${style.indicator} ${style.nextDay}`} />
          </Tooltip>
 )}
```

Question: I am not sure I understand the change.

If the element is visible, it should have a tooltip
If the element is not visible, is there a reason to render it? could we just conditionally render the whole thing?

alex-Arc: hmm they would then end up moving about depending on witch one is active
we could maybe place them with a css grid? or is it ok that they move?
image

PR #700, opened 2024-01-08
refactor: errors from integrationcontroller

alex-Arc:
maybe also add to v3?


cpvalente (approved):
Good catch

cpvalente:
> maybe also add to v3?

Yes. After this release we can merge master into v3

Issue #334, opened 2023-04-09
Feature request: Custom OSC Output on Events


Keksstar:
Please add custom OSC command output per Event. OSC Port should be definable in the options menu or in the (new?) OSC dialog per Event.
This could be used to trigger external software.

cpvalente:
Hi @Keksstar 

For now I am skeptic of creating something like this. At least until we have proven that the workflow works well and it is useful.

Ontime is used in different industries and for different purposes. But it is not a show controller.

The fear here is that since ontime is working on a single nodejs thread, we could be touching the limits of what is performant 

However, if we see that the application is performant and can handle overhead, I have no issues in doing this change in future.

Keksstar:
But where is the difference if it spits out custom OSC messages or the ones that is ontime already generating for start/end of event? It's just one more port that needs to be opened. Nodejs will yawn about it...

For show control you use something like Qlab, etc. but not ontime. ;-)
It would just make the life easier if we could get rid of middleware that translates the ontime OSC output to something useful.

cpvalente:
I am definitely not against the idea. 
If we can prove that this offers no significant performance hits, I am all for it. I personally have no experience on what is the cost of opening ports every second and this is something we can try and find. 

The integration feature is new and I would prefer to see it mature before trying to push more nuance into it.

We are also hoping to use this as a proof of concept to implement the same with HTTP which has been in the books for a while

ONCE v2 is out I can unlock the limitations of OSC integration in a separate release. Would you be available to stress test?

Keksstar:
Yup, would be happy to test this :-)

Once a second? What are you running for crazy shows? 😆

herzogmedia:
I'd also volunteer to test this =)

cpvalente:
Hi @herzogmedia, thank you for volunteering.

Could you specify which part of the task you would be able to test? 
We currently have custom messages per event (in v2). Are you using the feature?

This is the first time the feature is in production and I would like to have it ironed out before extending.
Perhaps you have some general feedback as to your experience with the feature?

cpvalente:
> so, the thing I wanted to do is to be able to trigger specific companion buttons from ontime as soon as a timer starts. I think that should be possible utilizing the osc integration and a user field. I'll try and then get back and report

Perfect, thank you.
It sounds like this should be possible with integrations. 
Also look into the companion feedback variables. I don't use companion but it sounds like that should be possible

Let me know if you get into trouble.
[...]

Issue #318, opened 2023-03-30
request: CLIP Remaining time from Millumin


sledzik1984:
I would be great to have clip remaining timer that is actually playing (like in Clock-8001)

Millumin is sending that via OSC. Ontime should allow to choose from which layer to get time from.

That time should be available only on non public screens (like backstage or studio)

cpvalente:
Hi @mmattausch 

Thank you for reaching out.
I am skeptic of creating specific integrations to other software, but have however no objections in creating an open endpoint to set this (which you can configure something else to populate)

The only point standing is that I am unsure on how best to fit this in the current designs
Could you please checkout the most recent release (v2 beta 2 as of now) and provide a suggestion?

https://github.com/cpvalente/ontime/releases

Feel also free to get in touch on mail@getontime.no

Keksstar:
#115 is also part of this request.

Don't open it up for specific software, just provide an entry point for OSC time data.
Everything before (between specific software and ontime) can be solved by middleware.

cpvalente:
Agreed. The server is ready to receive this data and I will implement as soon as I can get some help defining which screens should present this and where should it go

Keksstar:
Should be shown in the event itself... if you have it there, it will be reflected to all other views?

cpvalente:
Hi @Keksstar ,

There are a few challenges here.
As much as ontime can receive a string from another software, it would not understand what to do with it, and its value is purely presentational.

I would be inclined to believe this piece of data should be completely detached from Ontime so as to not have negative consequences to users who do not follow this workflow.

I would suggest Ontime can display this string in the **Info tab** in the Editor or perhaps in the **message control** section (where onAir is)
When it comes to the views (backstage and studio clock as mentioned here), we would need to find a place for it to be shown that follows the same principles. A user that does not follow this workflow should not see an empty section or the UI should not feel incomplete.

If we could gather some suggestions on this presentations. The remaining work is ready

mmattausch:
I use the remaining time display mainly for the preview monitors on stages (for hosts or speakers) and in FOH for all other operators (lighting, sound ...). 
Usually the time for a whole slot runs in the timer - e.g. 20min interview, and this slot is interrupted by a short clip whose timing is announced by the presenter. That's why I find the function of my timer so useful to display both times in parallel, as long as the clip is running, in order to allow the people on stage a clean re-entry into the programme.
[...]

Issue #115, opened 2022-05-15
Get time from CasparCG


Keksstar:
Would be nice, if the remaining time of an videoclip out of CasparCG could be displayed in ontime. 
Time is available via OSC: https://github.com/CasparCG/help/wiki/OSC-Protocol

cpvalente:
Hi @Keksstar, thank you for your suggestion.

It is unlikely that I would create a workflow specific for CasparCG.
However, I like the idea of being able to add dynamic text to screens, specially the stage timer screen.

I could create an API for setting this (OSC, HTTP and websocket), you would then need to add a control layer between Caspar and ontime to achieve what you need. what do you think?

Keksstar:
That would be awesome! The Middle-thingy is then done easily...
It would be great if the time of an timer could be set with this too. This would come in handy for situations like:

Speaker on Stage... has 2min for talking... Intro of Event starts, Speaker walks of stage, can see the remaining time backstage, walks in just before the end of the video clip. That´s how it should be...

Sure, speaker could see it also as dynamic text, but this wouldn´t work at the end with my lovely #100 ;-)

cpvalente:
Hi @Keksstar, I have some availability to work on this feature. 
I would create an endpoint in OSC / HTTP / Websocket for setting a custom message. this would effectively shrink the timer slightly to show inline.

![custom message](https://user-images.githubusercontent.com/34649812/188233118-c4028880-c5c9-4acc-9b5d-487298dc0a4d.PNG)

What do you think of the workflow?
Are there any other screens where it would make sense that this should be present?

cpvalente:
Hi @Keksstar ,
We have some reports of ontime views not rendering in Caspar CG.

We also have a release version which increases support for legacy browsers and hopefully fixes the issue. But none of us here use Caspar CG and can test 

Would you be available to report on this?

Keksstar:
You're referring to the topic where the ontime "website" is not rendered in Caspar. (this is 95% due to outdated browser versions in Caspar and not an ontime topic)

Getting time from Caspar via OSC is the same as getting time from Millumin, Mitti, etc.
Little middleware will solve the translation vom Caspar OSC time to ontime.
But therefore it needs an entry point in ontime as discussed in #318

PR #701, opened 2024-01-08
feat: add `set-time` endpoint

alex-Arc:
osc: `/ontime/set-time/<eventId>/[remaining,elapsed] <time in seconds>`
ws: `{"type": "set-time", "payload": {"eventId" : <eventId>, "property": "[remaining,elapsed]", "value": <time in seconds>}}`
image
image

notes:
https://github.com/cpvalente/ontime/issues/334#issuecomment-1877468901
https://github.com/cpvalente/ontime/issues/334#issuecomment-1877686988
#318 #115

### DIFF
--- a/apps/client/src/features/rundown/event-block/EventBlockInner.tsx
+++ b/apps/client/src/features/rundown/event-block/EventBlockInner.tsx
@@ -10,6 +10,7 @@ import { IoPlayForward } from '@react-icons/all-files/io5/IoPlayForward';
 import { IoPlaySkipForward } from '@react-icons/all-files/io5/IoPlaySkipForward';
 import { IoStop } from '@react-icons/all-files/io5/IoStop';
 import { IoTime } from '@react-icons/all-files/io5/IoTime';
+import { MdInput } from '@react-icons/all-files/md/MdInput';
 import { EndAction, Playback, TimerType } from 'ontime-types';
 import { millisToString } from 'ontime-utils';
 
@@ -242,6 +243,9 @@ function TimerIcon(props: { type: TimerType; className: string }) {
   }
   if (type === TimerType.TimeToEnd) {
     return <BiArrowToBottom className={className} />;
+  }
+  if (type === TimerType.External) {
+    return <MdInput className={className} />;
   }
   return <IoArrowDown className={className} />;
 }
--- a/apps/server/src/controllers/integrationController.ts
+++ b/apps/server/src/controllers/integrationController.ts
@@ -1,5 +1,9 @@
+import { Playback, TimerType } from 'ontime-types';
+import { mts } from '../../../../packages/utils/src/timeConstants.js';
+import { EventLoader } from '../classes/event-loader/EventLoader.js';
 import { messageService } from '../services/message-service/MessageService.js';
 import { PlaybackService } from '../services/PlaybackService.js';
+import { eventTimer } from '../services/TimerService.js';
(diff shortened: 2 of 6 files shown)


packages/types/src/definitions/TimerType.type.ts
```diff
@@ -3,4 +3,5 @@ export enum TimerType {
   CountUp = 'count-up',
   TimeToEnd = 'time-to-end',
   Clock = 'clock',
+  External = 'external',
```
cpvalente, line 6: I am not sure about reusing this. 
The existing behaviour changes the way the time is presented.

I feel the external source is unrelated and I am not sure I would know to look for it here.

On the other hand, we can't provide a countdown countup to an event we don't have control over. 

Not really sure. What do you think?

apps/client/src/features/rundown/event-block/EventBlockInner.tsx
```diff
@@ -10,6 +10,7 @@ import { IoPlayForward } from '@react-icons/all-files/io5/IoPlayForward';
 import { IoPlaySkipForward } from '@react-icons/all-files/io5/IoPlaySkipForward';
 import { IoStop } from '@react-icons/all-files/io5/IoStop';
 import { IoTime } from '@react-icons/all-files/io5/IoTime';
+import { MdInput } from '@react-icons/all-files/md/MdInput';
```
cpvalente, line 13: I would prefer if we stick to Io5.

Either that or find a new icon set which has everything we need

apps/server/src/controllers/integrationController.ts
```diff
@@ -267,6 +271,37 @@ export function dispatchFromAdapter(
       return { topic: 'timer', payload: timer };
     }
 
+    case 'set-time': {
+      const { eventId, property, value } = payload as ChangeOptions;
+      const amout = Number(value);
+      if (isNaN(amout)) {
+        throw new Error(`set-time "${value}" is not a number`);
+      }
+      const event = EventLoader.getEventWithId(eventId);
```
cpvalente, line 280: Would we not like to fail faster?

Maybe add a method in the playback service to get current event and compare that?

apps/client/src/features/rundown/event-block/EventBlockInner.tsx
alex-Arc: agree, it is just a quick place holder

packages/types/src/definitions/TimerType.type.ts
alex-Arc: yes i'm also not sure

TimeToEnd and External can't really work at the same time
Clock and External doesn't make much sense either (I think?)

maybe both External Up and Down could be useful to some one 
although I expect that most people would end up using CountDown with external..?

I think maybe it is OK to say that External will always be a CountDown 
or we could add an ExternalUp and ExternalDown
(^maybe one of thees for v2)
or a separate toggle

and then we can revisit for v3,

cpvalente: How about a single toggle "source: external" which when is set we disable the timer type in the interface?

apps/client/src/features/rundown/event-block/EventBlockInner.tsx
alex-Arc: IoFlash 
image

IoArrowForwardCircleOutline 
image

IoDiscOutline 
image

MdOutlineSettingsInputSvideo 
image

PR #702, opened 2024-01-09
Fix/external path

alex-Arc:


apps/server/src/modules/loadDemo.ts
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    demoInDisk.forEach((to, index) => {
+      const from = pathToStartDemo[index];
+      copyFileSync(from, to);
+    });
+  } catch (_) {
+    /* we do not handle this */
+  }
+
+  return demoInDisk;
```
cpvalente, line 23: question: what does the return do? it seems mostly unused?

apps/server/src/setup.ts
```diff
@@ -80,11 +80,26 @@ export const pathToStartDb = isTest
   ? join(currentDirectory, '../', config.database.testdb, config.database.filename)
   : join(currentDirectory, '/preloaded-db/', config.database.filename);
 
+//TODO: move all to external
```
cpvalente, line 83: suggestion: can you clarify this comment for future us? what is all?
Would the note be TO`DO: move all static files to the external directory?`

apps/server/src/modules/loadDemo.ts
alex-Arc: yes I don't know 
I just kept the same format as the other loader, bu i can remove it from them both?

cpvalente: I guess if they are not used we should remove it

cpvalente (approved):
good stuff 🚀

PR #706, opened 2024-01-10
feat: many timers

cpvalente:
This PR adds the support for extra timers in Ontime.
So far, we dont have support for the extra timers to cause any other actions in the application

Tasks:
- [x] Create simple timer abstraction
- [x] Create API endpoints to control the timer
- [x] Create necessary UI
- [x] Add wanted amount of timers


apps/server/src/classes/simple-timer/SimpleTimer.ts
```diff
@@ -0,0 +1,64 @@
...
+   * Sets the duration of the timer
+   * @param time - time in milliseconds
+   */
+  public setTime(time: number): SimpleTimerState {
+    this.state.duration = time;
+    this.state.current = time;
+    return this.state;
+  }
+
+  public play(timeNow: number): SimpleTimerState {
```
cpvalente, line 32: should probably be called start like the other timer

packages/types/src/definitions/runtime/Extratimer.type.ts
```diff
@@ -1,7 +1,9 @@
 export type SimplePlayback = 'play' | 'pause' | 'stop';
+export type SimpleDirection = 'up' | 'down';
```
cpvalente, line 2: Suggestion: how about something more semantic of timers, like 
`type: 'count-down' | 'count-up'`

It achieves the same but has a similar wording to the other timers, which may make it less confusing for us

apps/client/src/common/components/input/time-input/TimeInput.tsx
```diff
@@ -7,10 +7,13 @@ import { forgivingStringToMillis } from '../../../utils/dateConfig';
 import { TimeEntryField } from '../../../utils/timesManager';
 
 import style from './TimeInput.module.scss';
+
+type OntimeTimeEntry = TimeEntryField | 'extraTimer'
```
alex-Arc, line 11: @cpvalente tcs seams to be complaining about this, I'm unsure how to fix it

alex-Arc:
@cpvalente I think this is a good start for now

PR #709, opened 2024-01-13
update API - part 1

alex-Arc:
https://github.com/cpvalente/ontime-docs/pull/8

## changes
- test-ontime -> version
- ontime-poll -> poll
- start-next -> startnext
- start - add option to send payload with index, id or cue
- load - add option to send payload with index, id or cue
- message - add option to send payload with index, id or cue
- ~~gotocue~~
- ~~gotoid~~
- ~~gotoindex~~
- delay - replaced with addtime and use seconds

nestes object from OSC path
`/ontime/start/index` `1` -> `{type: start, payload: {index: 1}}`

nestes object from HTTP
`/api/start?index=1` -> `{type: start, payload: {index: 1}}`


apps/client/src/common/stores/runtime.ts
```diff
@@ -22,8 +22,8 @@ export const runtimeStorePlaceholder: RuntimeStore = {
   timerMessage: {
     text: '',
     visible: false,
-    timerBlink: false,
-    timerBlackout: false,
+    blink: false,
+    blackout: false,
```
cpvalente, line 26: nice

apps/client/src/features/control/message/MessageControl.tsx
```diff
@@ -42,20 +41,20 @@ export default function MessageControl() {
...
         <Button
           size='sm'
-          className={`${data.timerMessage.timerBlink ? style.blink : ''}`}
-          variant={data.timerMessage.timerBlink ? 'ontime-filled' : 'ontime-subtle'}
-          leftIcon={data.timerMessage.timerBlink ? <IoSunny size='1rem' /> : <IoSunnyOutline size='1rem' />}
-          onClick={() => setMessage.timerBlink(!data.timerMessage.timerBlink)}
+          className={`${data.timerMessage.blink ? style.blink : ''}`}
+          variant={data.timerMessage.blink ? 'ontime-filled' : 'ontime-subtle'}
+          leftIcon={data.timerMessage.blink ? <IoSunny size='1rem' /> : <IoSunnyOutline size='1rem' />}
+          onClick={() => setMessage.timerBlink(!data.timerMessage.blink)}
```
cpvalente, line 47: Would it be helpful to extract this into a variable at the top like `const blink = data.timerMessage.blink`?
This is a little verbose....

Same for blackout

apps/server/src/adapters/OscAdapter.ts
```diff
@@ -59,17 +60,29 @@ export class OscServer implements IAdapter {
...
+        const reply = dispatchFromAdapter(
           path,
           {
             payload: transformedPayload,
           },
           'osc',
         );
+        //We only send simple types back to the OSC client
+        if (typeof reply.payload === 'string' || typeof reply.payload === 'number') {
+          const client = new Client(rinfo.address, rinfo.port);
```
cpvalente, line 82: what is `rinfo`? could we rename it something else?

apps/server/src/adapters/OscAdapter.ts
```diff
@@ -59,17 +60,29 @@ export class OscServer implements IAdapter {
...
-        dispatchFromAdapter(
+        const reply = dispatchFromAdapter(
           path,
           {
             payload: transformedPayload,
           },
           'osc',
         );
+        //We only send simple types back to the OSC client
+        if (typeof reply.payload === 'string' || typeof reply.payload === 'number') {
```
cpvalente, line 81: Is there an issue we have in mind? we could force stringify it and send whatever... although I am unsure on how replies would work in OSC since the client has closed the connection after sending the request

apps/server/src/controllers/integrationController.ts
```diff
@@ -48,62 +48,39 @@ const actionHandlers: Record<string, ActionHandler> = {
...
   /* Message Service */
-  'set-timer-message': (payload) => {
-    if (!isPartialTimerMessage(payload)) {
-      throw new Error('Payload is not a valid timer message');
+  //TODO: maybe coerce the boolean values
+  message: (payload) => {
+    if (payload && typeof payload === 'object') {
+      if ('timer' in payload) {
+        if (!isPartialTimerMessage(payload)) {
+          throw new Error('Payload is not a valid timer message');
```
cpvalente, line 56: instead of maintaining so many error types, should we have a single
`throw new Error('Payload is not valid');` ?

apps/server/src/controllers/integrationController.ts
```diff
@@ -48,62 +48,39 @@ const actionHandlers: Record<string, ActionHandler> = {
     return { payload: updatedEvent };
   },
   /* Message Service */
-  'set-timer-message': (payload) => {
-    if (!isPartialTimerMessage(payload)) {
-      throw new Error('Payload is not a valid timer message');
+  //TODO: maybe coerce the boolean values
```
cpvalente, line 51: maybe we should, can you explain?
We cant get things of the wrong type entering our inner app logic

cpvalente (commented):
Things look generally good to me, just a few suggestions for improvement
Good work.

apps/server/src/adapters/OscAdapter.ts
alex-Arc: we cold send a stringifyed responds to the client but I think very few are ready to handle that
we could also unpack the object and translate it to a bundle OSC messages (this might actually be quite good)
it is just UDP so no connection is open. We are just throwing the reply back to the same port and they can pick it up if they want it

apps/server/src/adapters/OscAdapter.ts
alex-Arc: yes `clientInfo` maybe?

apps/client/src/features/control/message/MessageControl.tsx
alex-Arc: do I need to put it in a memo? or something like that?

apps/server/src/controllers/integrationController.ts
alex-Arc: should we only accept actual boolean values or convert strings and numbers to boolean as well?

apps/client/src/features/control/message/MessageControl.tsx
cpvalente: No. We are just creating a label for a value. No computations here 

Memo should be reserved for expensive computations or computations which trigger invalidation of data or potentially expensive components

apps/server/src/adapters/OscAdapter.ts
cpvalente: I think the best is to skip feedback on UDP.

We have the integration output which should provide adequate feedback to changes in the application state

apps/server/src/controllers/integrationController.ts
cpvalente: I think you can decide that. 

I never remember working with Boolean's in OSC, not sure if for lack of availability or just never had a use case for it

I would suggest we model this against the simplest workflow and wait for users feedback if improvement is needed.

apps/server/src/adapters/OscAdapter.ts
```diff
@@ -31,15 +33,15 @@ export class OscServer implements IAdapter {
       }
 
       // get second part (command)
-      if (!path) {
+      if (!action) {
```
cpvalente, line 36: question: is there a reason for this renaming? to my best knowledge `path` is the name of the element.

If it doesnt do much for us other than being different, I would suggest that we keep it as is, no need for a change

apps/server/src/controllers/integrationController.ts
```diff
@@ -48,62 +47,37 @@ const actionHandlers: Record<string, ActionHandler> = {
...
+        case 'external': {
+          newState = messageService.setExternalMessage(payload[key]);
+          break;
+        }
+        case 'lower': {
+          newState = messageService.setLowerMessage(payload[key]);
+          break;
+        }
+        default: {
+          throw new Error(`Message: ${key} dose not exist`);
```
cpvalente, line 75: If I understand correctly, any bad property would fail the rest of the request, but changes before would have affected. so

```ts
{
  visible: true,
  text: "sadasdas",
  not-value: "going to throw"
}
```
would both throw an error and make changes

I also not sure it is a good idea to affect changes one by one, since this starts a sequence of async actions to notify the clients.
How about composing a patch object. We would then make a single call to validate the object as a whole and another call to make a single change to the message service

Also: do we really need to reuse the code here?
I think separating this into three different `timer` / `public` / `external` / `lower` handlers is fine. The APIs are sometimes different and when they are the same it is mostly coincidental

apps/server/src/controllers/integrationController.ts
```diff
@@ -177,6 +151,31 @@ const actionHandlers: Record<string, ActionHandler> = {
...
+        const reply = actionHandlers.next(payload.next);
+        return reply;
+      }
+      if ('previous' in payload) {
+        const reply = actionHandlers.previous(payload.previous);
+        return reply;
+      }
+    }
+    throw new Error('No load method provided');
+  },
```
cpvalente, line 178:
```suggestion
  load: (payload) => {
    if (payload && typeof payload === 'object') {
      if ('index' in payload) {
        return actionHandlers.loadindex(payload.index);
      }
      if ('id' in payload) {
        return actionHandlers.loadid(payload.id);
      }
      if ('cue' in payload) {
        return actionHandlers.loadcue(payload.cue);
      }
      if ('next' in payload) {
        return actionHandlers.next(payload.next);
      }
      if ('previous' in payload) {
        return actionHandlers.previous(payload.previous);
      }
    }
    throw new Error('No load method provided');
  },
```

apps/server/src/routes/apiRouter.ts
```diff
@@ -14,9 +14,19 @@ router.get('/', (_req, res) => {
 
 // any GET request in /api is sent to the integration controller
 router.get('/*', (req, res) => {
-  const action = req.path.substring(1);
+  let action = req.path.substring(1);
+  const actionArray = action.split('/');
+
```
cpvalente, line 19: todo: remember to run the formatter in your IDE. It is easy to get silly merge conflicts

apps/server/src/routes/apiRouter.ts
```diff
@@ -14,9 +14,19 @@ router.get('/', (_req, res) => {
...
-
+  
+  //TODO: this can be better
+  if (actionArray.length > 1) {
+    action = actionArray.shift();
+    params.payload = actionArray.reduceRight(
+      (parm, key, index) => (index === actionArray.length - 1 ? { [key]: params.payload } : { [key]: parm }),
+      {},
+    );
+  }
```
cpvalente, line 29: todo: if we are doing it this way, I would suggest extracting into a function like `actionsFromParams` and unit testing it, just to make sure we cover the major caveats.

I also think this can be simplified like
```ts
if (actionArray.length > 1) {
  action = actionArray.shift();
  params.payload = actionArray.reverse().reduce((result, key) => ({ [key]: result }), params.payload);
}
```

cpvalente (changes requested):
Good work @alex-Arc , few suggestions for you

apps/server/src/adapters/OscAdapter.ts
```diff
@@ -15,13 +15,15 @@ export class OscServer implements IAdapter {
     this.osc.on('error', (error) => logger.error(LogOrigin.Rx, `OSC IN: ${error}`));
 
     this.osc.on('message', (msg) => {
-      // message should look like /ontime/{path}/{params?} {args} where
+      // TODO: update this comment
```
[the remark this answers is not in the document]
alex-Arc: forgot to remove the todo after adding the comment
// params: used if the to create a nested object to patch with

apps/server/src/controllers/integrationController.ts
alex-Arc: > composing a patch object
I was looking into that, but we have the `throttle` function on the message service
I thing the different services should be throttled separately? maybe it is still possible?

> Also: do we really need to reuse the code here?

I might be thinking to far ahead but maybe the message service would also end up being user definable?

apps/server/src/routes/apiRouter.ts
alex-Arc: nice

apps/server/src/adapters/OscAdapter.ts
cpvalente: I dont really mind, the point is to avoid unnecessary changes. we are making large changes to the app and changing something that works can
- introduce sneaky bugs
- result in merge conflicts

If it doesnt provide any value I would prefer skipping it. But if you think this is good and we can say it is harmless (no bugs and no merge conflicts), go ahead

apps/server/src/controllers/integrationController.ts
cpvalente: > I was looking into that, but we have the throttle function on the message service
> I thing the different services should be throttled separately? maybe it is still possible?


I dont think this stops us from creating the single patch object, I think we should do it. So instead of iterating through the keys, we make a single object (without loops) and a single patch request.
We dont need loops since the keys are not dynamic and we know which ones they are, this is also a good way of filtering out bad keys

> I might be thinking to far ahead but maybe the message service would also end up being user definable?

Maybe, but we should worry about that later. How would that work?

apps/server/src/utils/arrayUtils.ts
```diff
@@ -61,3 +61,15 @@ export const sortArrayByProperty = <T>(arr: T[], property: string): T[] => {
...
+/**
+ * @description Creates a nested object with keys from an array and assigns the `value` to the last key
+ * @param {array} arr - array to be nested
+ * @param {string} value - value to assign
+ * @returns {object | null} nested object or null if no object was created
+ */
+export const nestedObjectFromArray = (arr: string[], value?: unknown): object | null => {
+  const obj = arr.reduceRight((result, key) => ({ [key]: result }), value);
+  if (typeof obj === 'object') return obj;
+  return null;
```
cpvalente, line 74: I dont think we need this generic function. our function should take a path and produce an object. This is only  useful in the context of the APIs and we shouldnt motivate to reuse anywhere else

That would also make our tests more valuable since they would include real paths

apps/server/src/controllers/integrationController.ts
alex-Arc: > how do you see that could work

I could imagine a backstage view in the Green room and in the tech break room
or maybe both the a speaker and the moderators have a screen each with a timer view
if there then were a parameter for the views where you select the message source (kind of like the new lower source)
then you can send separate messages to each

apps/client/src/features/control/message/MessageControl.tsx
```diff
@@ -11,51 +11,52 @@ import InputRow from './InputRow';
 import style from './MessageControl.module.scss';
 
 export default function MessageControl() {
-  const data = useMessageControl();
-
+  const messge = useMessageControl();
```
cpvalente, line 14:
```suggestion
  const message = useMessageControl();
```

apps/client/src/common/stores/runtime.ts
```diff
@@ -19,23 +19,25 @@ export const runtimeStorePlaceholder: RuntimeStore = {
...
-    visible: false,
-  },
-  lowerMessage: {
-    text: '',
-    visible: false,
-  },
-  externalMessage: {
-    text: '',
-    visible: false,
+  message: {
```
cpvalente, line 22: question: what is the motivation to nest this object?

cpvalente (approved):

PR #712, opened 2024-01-15
Project manager: Load, Rename, Duplicate

arymoraes:


apps/server/src/utils/getFileListFromFolder.ts
```diff
@@ -41,7 +41,7 @@ export const getProjectFiles = async (): Promise<ProjectFile[]> => {
       const stats = await stat(filePath);
 
       projectFiles.push({
-        filename: file,
+        filename: file.replace(/\.[^/.]+$/, ''),
```
arymoraes, line 44: TODO: make this into a util fn since we're using it more than once

arymoraes:
@cpvalente this is mostly done, you can see how I did the UI and see what you think about it, I have load, duplicate and rename working. Still missing delete and create, but delete should be easy enough

it's still missing:

input validation - do we have any lib for that?
e2e specs
actually I need error handling for load too which I forgot

cpvalente:
> input validation - do we have any lib for that?
 

We don't have a library and usually just use html validation and manual validation with react hook form. 
I believe the app doesn't have enough complexity to warrant adding an extra library for this. 

> e2e specs

We have an e2e test to load a project file. Maybe we just update that if we have enough elements to?
The UI here is not achieving much and all the functionality is invisible in backend. I am not sure of that value of e2e tests on things like renaming. Would prefer some sensible validation and unit testing in backend 

I would suggest we focus the e2e on the core happy path

What do you think?

arymoraes:
> > input validation - do we have any lib for that?
> 
> We don't have a library and usually just use html validation and manual validation with react hook form. I believe the app doesn't have enough complexity to warrant adding an extra library for this.
> 
> > e2e specs
> 
> We have an e2e test to load a project file. Maybe we just update that if we have enough elements to? The UI here is not achieving much and all the functionality is invisible in backend. I am not sure of that value of e2e tests on things like renaming. Would prefer some sensible validation and unit testing in backend
> 
> I would suggest we focus the e2e on the core happy path
> 
> What do you think?

makes sense, will do it and ask for review when done, hopefully will be done today

apps/client/src/features/app-settings/panel/project-panel/RenameProjectForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  useEffect(() => {
+    setFocus('filename');
+  }, []);
+
+  return (
+    <div>
+      <form onSubmit={handleSubmit(onSubmit)} className={style.form}>
+        <FormControl className={style.formControl}>
```
cpvalente, line 39: question: I dont think the form control does much for us without the associated elements.

We used to use it and decided to move away from it. Cant quite recollect, maybe because it was easier to do it manually with react-hook-form?

I think that, if this doesnt achieve much here, we should aim to remove it

apps/server/src/controllers/ontimeController.ts
```diff
@@ -489,7 +490,9 @@ export const listProjects: RequestHandler = async (_, res: Response<ProjectFileL
   try {
     const fileList = await getProjectFiles();
 
-    const lastLoadedProject = JSON.parse(fs.readFileSync(lastLoadedProjectConfigPath, 'utf8')).lastLoadedProject;
+    const lastLoadedProject = removeFileExtension(
+      JSON.parse(fs.readFileSync(lastLoadedProjectConfigPath, 'utf8')).lastLoadedProject,
+    );
```
cpvalente, line 495: suggestion: could we wrap this into a few steps?

I think this style of function composition tends do impact the ability to read and debug without that many advantages. At least not to this style.
If we were to have some piping system like

```ts

const lastLoadedProject = pipe(
  readFile,
  parse,
  getLastLoadedProject,
  ["lastLoadedProjectConfigPath"]
)
```

There could be some value. but that is not a pattern we use here or have artifices for

apps/server/src/utils/removeFileExtension.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * @description Takes a filename and removes the extension
+ * @param {string} filename - filename with extension
+ */
+export const removeFileExtension = (filename: string): string => {
+  return filename.replace(/\.[^/.]+$/, '');
+};
```
cpvalente, line 7: Suggestion: how about using the `parse` method from `path` to get the filename?

If you feel there is a reason for us to prefer the regex, I would very much like to see some unit tests for it to prove that we handle most edge cases

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,165 @@
...
+    }
+  };
+
+  const handleCancel = () => {
+    onToggleEditMode?.(null, null);
+  };
+
+  useEffect(() => {
+    setSubmitError(null);
+  }, [editingMode]);
```
cpvalente, line 87: can you please double-check this? we likely want to change this in a callback.
In my experience, leveraging `useEffect` this way is more often than not a source for bugs

apps/client/src/features/app-settings/panel/project-panel/RenameProjectForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <Input defaultValue={filename} size='sm' type='text' variant='ontime-filled' {...register('filename')} />
+        </FormControl>
+        <div className={style.actionButtons}>
+          <Button
+            disabled={!isDirty || !isValid || isSubmitting}
+            aria-label='Cancel duplicate project name'
```
cpvalente, line 45: todo: maybe this is a left over from copy paste. 

question: Does it mean that the interface is so similar that it should be the same? I dont think that showing the two inputs in duplicate gives that much value.
How about prepopulating the fields with the name and only activating the action "Rename" and "Duplicate" once the text is differrent?


Also: we dont really have requirements for accessibility this way, the button label provides adequate accessibility for our e2e tests

apps/client/src/features/app-settings/panel/project-panel/RenameProjectForm.tsx
```diff
@@ -0,0 +1,65 @@
...
+            disabled={!isDirty || !isValid || isSubmitting}
+            aria-label='Save duplicate project name'
+            children='Save'
+            onClick={handleSubmit(onSubmit)}
+            size='sm'
+            variant='ontime-filled'
+          />
+        </div>
+      </form>
+      {submitError ? <span className={style.error}>{submitError}</span> : null}
```
cpvalente, line 62:
```suggestion
      {submitError && <span className={style.error}>{submitError}</span>}
```

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.module.scss
```diff
@@ -8,3 +8,48 @@
...
+  align-items: center;
+
+  .cancelRenameButton {
+    margin-right: 0.5rem;
+  }
+  
+  .button {
+    &:first-of-type {
+      margin-right: 0.5rem;
+    }
```
cpvalente, line 49: hmmm... 😅

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,165 @@
...
+}
+
+function ActionMenu({
+  filename,
+  onAction,
+  onChangeEditMode,
+}: {
+  filename: string;
+  onAction?: () => void;
+  onChangeEditMode?: (editMode: EditMode, filename: string) => void;
```
cpvalente, line 133: question: I am not sure why this functions are optional? there is no use case for this menu to exist without them

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,165 @@
...
+  onSubmit,
+  onToggleEditMode,
+  updatedAt,
+}: ProjectListItemProps) {
+  const [submitError, setSubmitError] = useState<string | null>(null);
+
+  const handleRefetch = async () => {
+    await ontimeQueryClient.invalidateQueries({ queryKey: PROJECT_LIST });
+  };
+
```
cpvalente, line 40: nice! 
We also need to invalidate the remaining keys once a project is loaded. I think there is a `invalidateAllCaches` function

cpvalente (changes requested):
Hey @arymoraes , thank you for your work here.
Functionally things seem to work well, havent tested too much in-depth. I just left a few small comments

However I think we should rethink our UI.
- I was unable to understand what happens when we try to delete a loaded file, nothing? This action should be disabled and the backend should contain logic to refuse it

- The rename and duplicate actions should take the whole row, this would prevent overflows and also remove the ability for the user to activate more than one action at the time (right now you can click duplicate and then click delete while the duplicate is active). I also suggest some changes to the UI and action text, see rough explanation below:
- disable all other actions while in edit mode
- edit form takes entire row
- buttons are semantic to the action being completed
- pre-populate input with current name and have placeholder value be the action "Enter new name"

Screenshot 2024-01-18 at 21 47 37

arymoraes:
> * I was unable to understand what happens when we try to delete a loaded file, nothing? This action should be disabled and the backend should contain logic to refuse it

Delete is not yet implemented, I was planning on breaking this down into two PRs (Load, rename and duplicate) and (delete, create) so its easier to review, but I like your idea

> * The rename and duplicate actions should take the whole row, this would prevent overflows and also remove the ability for the user to activate more than one action at the time (right now you can click duplicate and then click delete while the duplicate is active). I also suggest some changes to the UI and action text, see rough explanation below:
> * disable all other actions while in edit mode
> * edit form takes entire row
> * buttons are semantic to the action being completed
> * pre-populate input with current name and have placeholder value be the action "Enter new name"
> 

Makes sense, will make the changes then, probably feels cleaner too

cpvalente:
Hey @arymoraes, I saw that you moved towards divs in the table layout. Is this because of the edit in line?


I suggest that we either move the design to a card list or keep the table layout. We can use `colspan` to solve the issue of the rename form

arymoraes:
> Hey @arymoraes, I saw that you moved towards divs in the table layout. Is this because of the edit in line?
> 
> I suggest that we either move the design to a card list or keep the table layout. We can use `colspan` to solve the issue of the rename form

TIL about `colspan`, let me try with that. I moved to divs due to not being able to have the form fit the width 100%, but that should work and I can go back to table

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
arymoraes: question

if we do `invalidateAllCaches` do we still need to run 

`await ontimeQueryClient.invalidateQueries({ queryKey: PROJECT_LIST });`

on project load? I don't think so right?

apps/client/src/features/app-settings/panel/project-panel/RenameProjectForm.tsx
arymoraes: fair, let's remove it

apps/server/src/utils/removeFileExtension.ts
arymoraes: guess it's another thing learned today, didn't know we could remove it using `path` `parse`, I got rid of the regex

apps/server/src/controllers/ontimeController.ts
arymoraes: agreed

apps/client/src/features/app-settings/panel/project-panel/RenameProjectForm.tsx
arymoraes: I agree the two inputs for duplicate might not have as much value, but I'd still like something to show the user the project they are duplicating

should we have a single input that is prepopulated with the duplicated project name and the `Duplicate` button then? I think that makes sense

arymoraes: I made duplicate single input

image

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
cpvalente: Since `PROJECT_LIST` is new it might not be in `invalidateAllCaches`

I guess if we add it there we are all good.

The gist of it is to tell react with "hey! The whole dataset just changed, if you need stuff fetch it again". There might be a single function to invalidate the entire `queryClient`

arymoraes: makes sense

I've seen this on react-query docs:

```
// Invalidate every query in the cache
queryClient.invalidateQueries()
```

https://tanstack.com/query/latest/docs/react/guides/query-invalidation

should we just invalidate all with `queryClient.invalidateQueries()` or is there a reason we are passing the keys individually to our invalidate function?

cpvalente: I think we should use the function stated in the documentation.

I think the reason might well have been a "did-not-read-the-docs" situation 😅

cpvalente:
> TIL about `colspan`, let me try with that. I moved to divs due to not being able to have the form fit the width 100%, but that should work and I can go back to table

glad that it was helpful. HTML attributes are a minefield 😅 
I typically use `colspan=99` in cases like this so you don't get the UI breaking all of a sudden when someone adds a column

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,189 @@
...
+    try {
+      setSubmitError(null);
+
+      if (!values.newFilename) {
+        setSubmitError('Filename cannot be blank');
+        return;
+      }
+      await duplicateProject(filename, values.newFilename);
+      await handleRefetch();
+      onSubmit?.();
```
cpvalente, line 74: cleanup: there doesnt seem to be a case for `onSubmit` to be undefined

cpvalente (changes requested):
Hey @arymoraes , great work here.
The UX feels really good and things seem to work

There are some issues with linting that I see the tooling should be helping you with, I will look into it

- the import order in files is incorrect, eslint should be prompting you to fix it
- passing children as props to a react component is unsemantic. dont know if there are any other issues apart from that, but I suggest we correct
- we should cleanup the optional stuff, it makes our logic more complex for no added value

Let me know if you need help

apps/client/src/features/app-settings/panel/project-panel/DuplicateProjectForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+          {...register('newFilename')}
+        />
+        <div className={style.actionButtons}>
+          <Button onClick={onCancel} size='sm' variant='ontime-ghosted' children='Cancel' disabled={isSubmitting} />
+          <Button
+            size='sm'
+            variant='ontime-filled'
+            isDisabled={!isDirty || !isValid || isSubmitting}
+            type='submit'
+            children='Duplicate'
```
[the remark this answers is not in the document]
arymoraes: makes sense, I am used to doing it `children={}` at work but I think its mostly because we use a primitive component  and we can pass a `i18n` object into it so its easier to do

`<Text children={translations.something} />` than `<Text>{translations.something}</Text>`

If its a plain string then it doesn't make much sense indeed

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,189 @@
...
+function ActionMenu({
+  current,
+  filename,
+  onAction,
+  onChangeEditMode,
+}: {
+  current?: boolean;
+  filename: string;
+  onAction?: () => Promise<void>;
+  onChangeEditMode?: (editMode: EditMode, filename: string) => void;
```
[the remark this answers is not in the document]
arymoraes: I could swear I changed that

apps/server/src/utils/upload.js
```diff
@@ -6,6 +6,28 @@ import { EXCEL_MIME, JSON_MIME } from './parser.js';
 import { ensureDirectory } from './fileManagement.js';
 import { getAppDataPath } from '../setup.js';
 
+function generateNewFileName(filePath, callback) {
+  let baseName = path.basename(filePath, path.extname(filePath));
+  let extension = path.extname(filePath);
+  let counter = 1;
+
+  const checkExistence = (newPath) => {
```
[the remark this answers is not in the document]
arymoraes: https://nodejs.org/api/fs.html#fsexistspath-callback

supposed to be deprecated

apps/client/src/features/app-settings/panel/project-panel/RenameProjectForm.tsx
```diff
@@ -0,0 +1,69 @@
...
+};
+
+interface RenameProjectFormProps {
+  filename: string;
+  onSubmit: (values: RenameProjectFormValues) => Promise<void>;
+  onCancel: () => void;
+  submitError: string | null;
+}
+
+export default function RenameProjectForm({ filename, onSubmit, onCancel, submitError }: RenameProjectFormProps) {
```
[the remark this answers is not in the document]
arymoraes: done, feels better I agree

apps/server/src/utils/upload.js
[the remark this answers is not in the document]
cpvalente: Nice, thank you for checking this. TIL

apps/client/src/common/api/ontimeApi.ts
```diff
@@ -341,3 +341,27 @@ export const postPushSheet = async (id: string, options: ExcelImportMap) => {
   const response = await axios.post(`${ontimeURL}/sheet-push`, { id, options });
   return response.data.data;
 };
+
+/**
+ * @description HTTP request to rename a project file
+ */
+export async function renameProject(filename: string, newFilename: string): Promise<MessageResponse> {
+  const url = `${ontimeURL}/project/${filename}/rename`;
```
cpvalente, line 349: question: did we check that this works well with names that contain spaces and other characters that we need to escape?

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,176 @@
...
+        await renameProject(filename, values.filename);
+        await handleRefetch();
+        onSubmit();
+      } catch (error) {
+        if (error instanceof AxiosError) {
+          const errorMessage = error?.response?.data?.message;
+
+          setSubmitError(errorMessage);
+        } else {
+          setSubmitError('An unknown error occurred');
```
cpvalente, line 61: I think we have a function for this `maybeAxiosError`

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,176 @@
...
+  filename,
+  onSubmit,
+  onToggleEditMode,
+  updatedAt,
+}: ProjectListItemProps) {
+  const [submitError, setSubmitError] = useState<string | null>(null);
+
+  const handleRefetch = async () => {
+    await ontimeQueryClient.invalidateQueries({ queryKey: PROJECT_LIST });
+  };
```
cpvalente, line 41: question: why not use the actual refetch?

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,176 @@
...
+        <>
+          <td>{createdAt}</td>
+          <td>{updatedAt}</td>
+          <td className={style.actionButton}>
+            <ActionMenu current={current} filename={filename} onChangeEditMode={handleToggleEditMode} />
+          </td>
+        </>
+      )}
+    </tr>
+  );
```
cpvalente, line 132: suggestion: how about this? maybe it is easier to read?

```suggestion
  return (
    <tr key={filename} className={current ? style.current : undefined}>
      {isCurrentlyBeingEdited ? (
        <td colSpan={99}>
          <DuplicateRenameProjectForm
            action={editingMode}
            filename={filename}
            onSubmit={editingMode === 'duplicate' ? handleSubmitDuplicate : handleSubmitRename}
            onCancel={handleCancel}
            submitError={submitError}
          />
        </td>
      ) : (
        <>
          <td>{filename}</td>
          <td>{createdAt}</td>
          <td>{updatedAt}</td>
          <td className={style.actionButton}>
            <ActionMenu current={current} filename={filename} onChangeEditMode={handleToggleEditMode} />
          </td>
        </>
      )}
    </tr>
  );
```

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -0,0 +1,176 @@
...
+          setSubmitError(errorMessage);
+        } else {
+          setSubmitError('An unknown error occurred');
+        }
+      }
+    },
+    [filename, onSubmit],
+  );
+
+  const handleToggleEditMode = useCallback(
```
cpvalente, line 93: question: what is the value of `useCallback` here?

Since the function is consumed locally we might not be doing much more than paying the cost of the memoisation check?

cpvalente (commented):
Great work, I left a few small tips for cleanup. lets get this merged once you had a look

apps/client/src/common/api/ontimeApi.ts
arymoraes: it works for spaces and a few special chars I tested, it doesn't work for `/` as this causes issues with the path

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
arymoraes: `The 'handleToggleEditMode' function makes the dependencies of useCallback Hook (at line 102) change on every render. To fix this, wrap the definition of 'handleToggleEditMode' in its own useCallback() Hook.eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920)`

that line is another `useCallback`, but if I remove that one then we get the warning for the `useMemo`

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
arymoraes: nice

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
arymoraes: ended up forgetting about it

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
arymoraes: it does, nice

apps/client/src/common/api/ontimeApi.ts
cpvalente: Sounds good, we can revisit this later. the validation of the name should make sure that the name is path friendly?

cpvalente (approved):
Good work 🚀

PR #729, opened 2024-01-23
refactor: update timers

cpvalente:
PR for cleaning up duplicate responsibilities when creating new events
It removes dependencies from components and puts the weight of validation to backend

It also changes the rundown indicators to work as in the first draft

the frontend has to deliver an object of the right shape, so it will need to parse the smart time strings


packages/utils/src/validate-events/validateEvent.ts
```diff
@@ -83,3 +87,41 @@ export function validateTimes(_start?: unknown, _end?: unknown, _duration?: unkn
   // we should have covered all cases
   return { timeStart, duration, timeEnd };
 }
+
+/**
+ * Creates a valid patch object for an OntimeEvent
+ * @throws if there is not enough data to do so
+ * @param event
+ */
+export function validatePatch(event: Partial<OntimeEvent>) {
```
alex-Arc, line 96: Is this used anywhere?
Also maybe a good place to add valibot

apps/client/src/features/rundown/event-block/RundownIndicators.module.scss
```diff
@@ -0,0 +1,25 @@
...
+  background-color: $bg-colour;
+}
+
+.delay {
+  @include indicator($ontime-delay);
+}
+
+.gap {
+  @include indicator($blue-500);
+}
```
alex-Arc, line 25: I find the position of the indicators a bit random/off center
what do you think about pinning the gap indicator to the right with the other indicators and the delay inline with the the start timer filed that it is describing
image

```css
.indicators {
  font-size: calc(1rem - 5px);
  position: absolute;
  top: -1.15em;
  z-index: 2;
  margin: 0 8em;
  width: 85%;
  text-align: center;
}

@mixin indicator($bg-colour) {
  padding: 0 0.5rem;
  border-radius: 2px;
  background-color: $bg-colour;
}

.delay {
  @include indicator($ontime-delay);
  width: max-content;
  position: absolute;
  left: 0;
}

.gap {
  @include indicator($blue-500);
  width: max-content;
  position: absolute;
  right: 0;
}

```

cpvalente: I am not sure I agree. since these labels are dynamic, having them in a row improves their visibility.
Otherwise I fear users would try to find hidden meanings int their position, which is untrue.

Maybe you are correct that their current position seem to indicate that they belong to the input field. I just didnt want to have them over the buttons 😓 

Since this PR is so large, I suggest we roll with this and revisit styles in beta

PR #735, opened 2024-01-25
refactor: allow import of blocks and skip import

cpvalente:
Allows importing blocks from spreadsheets, it will also skip the import of rows which contain a timer type we dont know


alex-Arc (approved):
Works good

adding this here will allow export to sheets as well
 https://github.com/cpvalente/ontime/blob/0b5d0711b6c7f92d4186de07317c4b530c0de42a/apps/server/src/utils/sheetUtils.ts#L86
```typescript
    } else if (isOntimeBlock(event)) {
      if (key === 'title') {
        returnRows.push({
          userEnteredValue: { stringValue: event[key] },
        });
      } else if (key === 'timerType') {
        returnRows.push({
          userEnteredValue: { stringValue: 'block' },
        });
      } else {
        returnRows.push({});
      }
```

cpvalente:
> adding this here will allow export to sheets as well


nice, good catch. hopefully i did it right

alex-Arc (approved):
🎉

PR #737, opened 2024-01-26
[DEMO] Filter in views

alex-Arc:
can be cleaned up and generalized, but the general idea is there

https://github.com/cpvalente/ontime/assets/19875125/33c9f29e-0972-4d43-93d9-70ff1aedf85a


cpvalente:
I am unsure if the way we filter the event now is bulletproof. I am afraid we are opening to a bunch of issues.

When we are running a show, there shouldnt be a case for us to show nothing in the event now column. We are always running something.
We would either 
a) look for the last thing of interest to show, like we do not for public
b) show what is actually running but then mark it as "this is not part of your filter"

in the second case, we would not be able to use it in public displays since it would show production data (eg. Mic check) to the audience.

We can revise the logic, and I think the filter is a good idea, but it would need to be fleshed out as a more robust feature

PR #739, opened 2024-01-27
time to end

cpvalente:
this PR makes it so that an event marked as `Time-To-End` will always update as long as it is loaded

### DIFF
--- a/apps/server/src/state.ts
+++ b/apps/server/src/state.ts
@@ -393,6 +393,10 @@ export const stateMutations = {
             const result = play();
             _shouldNotify = true;
             _isFinished = result.isFinished;
+          } else if (state.eventNow?.timerType === TimerType.TimeToEnd) {
+            // or if we are in a time-to-end timer
+            state.timer.current = getCurrent(state);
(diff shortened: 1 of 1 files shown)


alex-Arc:
Sorry I confused the Time-To-End timer type with a question about a expected end
so my opinion is not strong about this.

and is seams like the is sill a difference between a loaded and a playing event
 
image
image

cpvalente:
Hey @alex-Arc, I could not recreate, but I did find a potential misunderstanding.

firefox_Qzn74U4Sza

What is the expected behaviour for a time-to-end timer when we pause or add time?

I am under the impression that we should not allow this operation. the only way to delay a time-to-end event is by changing its scheduled end. Do you think this makes sense?

alex-Arc:
> I am under the impression that we should not allow this operation. the only way to delay a time-to-end event is by changing its scheduled end. Do you think this makes sense?

that makes sense... maybe
i' have only used this type as the first event in the show to have somthing to start without thinking about a duration but it just being a fix point in time. And for that it would make sense

PR #744, opened 2024-01-28
User fields

alex-Arc:
shape `{<label>: {type: 'string'}}`

GET `/project/custom` -> object with all custom fields 

POST `/project/custom`  `{label: <label>, type: 'string'}`  -> object with new custom field

PUT `/project/custom`  `{label: <label>, field: {label?: <new label>, type?: 'string'}}`  -> object with updated custom field 

DELETE`/project/custom/:label` -> removes a custom field


cpvalente:
Thank you @alex-Arc 

I will take a closer look at the code later, meanwhile few comments:

> shape {<id>: {label: <user label>, type: 'string'}}

I understand the intuition to have unique IDs here, but how about a simpler
`{label: {value: string, type: string}}`
Wrapping things in IDs would force us to refer to those IDs in the events. I think the extra abstraction does not provide that much value. 

> GET /project/custom -> object with all custom fields
I dont think this data set should be nested inside project. This refers to the data related to the project, not to the settings.
It is entirely conceivable that the user might want to maintain the `customFields` across projects

What do you think?

alex-Arc:
> I dont think this data set should be nested inside project. This refers to the data related to the project, not to the settings.

while it is possible that a user wants to reuse the customFields across  projects it is almost certain that if you save a project and open it again later the customFields needs to follow along so that you don't loos that information in your rundown
I'm fine with not nesting it inside the project object, where should I put it?

> I understand the intuition to have unique IDs here, but how about a simpler

I thought that you could not use spaces and special charterers in the object key, but now that i look it up it looks like everything is allowed when using the [""] method? 
But could there not a be problems if user enter a reserved word? https://www.w3schools.com/js/js_reserved.asp

cpvalente:
> I'm fine with not nesting it inside the project object, where should I put it?

I think it should be a top level object like the userFields currently are


> I thought that you could not use spaces and special charterers in the object key, but now that i look it up it looks like everything is allowed when using the [""] method? But could there not a be problems if user enter a reserved word? https://www.w3schools.com/js/js_reserved.asp

There are likely a few issues to consider, I am unsure if a reserved word is particularly bad.
Selecting `customFields["delete"]` works just fine

That doesnt mean that there aren't caveats, we should sanitise the input to make sure it is a good string.
From the top of my head I believe that characters like `"` `[]` `.` could be problematic, so maybe we stick to allowing alphanumeric values

apps/server/src/classes/data-provider/DataProvider.ts
```diff
@@ -32,6 +33,22 @@ export class DataProvider {
...
+
+  static async overrideCustomField(newData: CustomFieldDefinitions) {
+    data.customFields = { ...newData };
+    await this.persist();
+    return data.customFields;
+  }
+
+  static getCustomField() {
+    return data.customFields;
+  }
```
cpvalente, line 50: I think the dataprovider should be pretty dumb.
Meaning we should have a single setter function as we do with the other fields

Also: do you mind typing the return values? it will help us when we go intro strict mode

apps/server/src/routes/projectRouter.ts
```diff
@@ -9,3 +16,12 @@ router.get('/', getProject);
 
 // create route between controller and 'POST /project' endpoint
 router.post('/', projectSanitiser, postProject);
+
+router.get('/custom', getCustomFields);
+
+//TODO: validate
+router.post('/custom', postCustomField);
```
cpvalente, line 23: I understand the instinct to keep this brief, but `custom` is a very tricky keyword to understand. `field` would have not been munch better. could we call this endpoint `custom-field`?

apps/server/src/utils/customFields.ts
```diff
@@ -0,0 +1,38 @@
+import { CustomFieldDefinitions, CustomInfo } from 'ontime-types';
+import { isAlphanumeric } from 'ontime-utils';
+import { DataProvider } from '../classes/data-provider/DataProvider.js';
+
+export const createCustomField = async (label: string, field: CustomInfo) => {
```
cpvalente, line 5: we would need some tests for this functions.
the logic can also be simplified here

It is OK with me that we postpone doing it until we decide on the datashape

cpvalente (approved):
This looks ok to me. Just left a few small comments

I see you postponed the validation and testing the logic, it is OK with me to do that once we understand the data better

I realise that my advice on making the custom field object might not have been great and we should experiment.
I have created the UI to receive an array of objects, that should make this operations much simpler

However, I think we should still normalise the fields to make it easier to consume in the client

maybe we just normalise it in the react hook?

alex-Arc:
> I see you postponed the validation and testing the logic, it is OK with me to do that once we understand the data better

yes i would like to not write new validations to move it to valibot later
that might be a reason to add valibot as a dependency now so that we can start using it in all the new stuff we write?

cpvalente:
> that might be a reason to add valibot as a dependency now so that we can start using it in all the new stuff we write?

If we have decided that we definitely want valibot we can add it now and use it in new features

packages/utils/src/regex-utils/isAlphanumeric.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @description Validates a alphanumeric string
+ * @returns {boolean}
+ */
+export const isAlphanumeric = (text: string): boolean => {
+  const regex = /^[a-zA-Z0-9]+$/;
```
cpvalente, line 6:
```suggestion
  const regex = /^[a-z0-9]+$/i
```

PR #747, opened 2024-01-29
fix: timer type clock is not negative in presenter displays

### LINKED ISSUE #746: Time of Day is red
Time of Day is red when the duration is 0. I guess I’d expect the clock should always be white, no matter what, when showing the current time of day.

thanks

cpvalente:
This PR makes it so that presenter displays do not show the current timer as negative when the timer is set to be a clock

Fixes #746


alex-Arc (approved):
Tested 👍🏽

PR #753, opened 2024-02-01
Project Manager: Export

biancahpp:
This includes three changes:

1. Add "Download" button to the ProjectItem overflow menu.
This allows users to download the json project file for any element. 

2. Add "Export CSV Rundown" button to the currently loaded project item overflow menu.
Allowed for the currently selected item.

3. In both cases the file name is the name the user sees eg. Project 1.json or Project 1.csv

### DIFF
--- a/apps/client/src/common/api/ontimeApi.ts
+++ b/apps/client/src/common/api/ontimeApi.ts
@@ -143,15 +143,19 @@ export async function postOscSubscriptions(data: OscSubscription) {
 /**
  * @description HTTP request to download db in CSV format
  */
-export const downloadCSV = () => {
-  return fileDownload(ontimeURL, { name: 'rundown', type: 'csv' }, { type: 'text/csv;charset=utf-8;' });
+export const downloadCSV = (fileName?: string) => {
+  return fileDownload(ontimeURL, { name: fileName ?? 'rundown', type: 'csv' }, { type: 'text/csv;charset=utf-8;' });
 };
 
 /**
  * @description HTTP request to download db in JSON format
  */
-export const downloadRundown = () => {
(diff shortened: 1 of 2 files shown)


apps/client/src/common/api/ontimeApi.ts
```diff
@@ -143,15 +143,19 @@ export async function postOscSubscriptions(data: OscSubscription) {
 /**
  * @description HTTP request to download db in CSV format
  */
-export const downloadCSV = () => {
-  return fileDownload(ontimeURL, { name: 'rundown', type: 'csv' }, { type: 'text/csv;charset=utf-8;' });
+export const downloadCSV = (fileName?: string) => {
+  return fileDownload(ontimeURL, { name: fileName ?? 'rundown', type: 'csv' }, { type: 'text/csv;charset=utf-8;' });
```
biancahpp, line 147: I made this new prop optional and added a fallback value of `rundown`, which was the default value prior to this change.
From a quick search it looks like this was being used in one other place so instead of adding the parameter as "rundown" I just added a fallback value in case the parameter isn't provided.

Should we just remove the "Export" option on the Sidebar?

cpvalente: Thank you for cleaning up.

This makes sense to me. The export function in the `/cuesheet` doesnt have access to the project information. We also likely don't want to add it there so a fallback is good.

> Should we just remove the "Export" option on the Sidebar?

Yes please!

cpvalente (approved):
Hi @biancahpp , thank you very much for your contribution, great to have you on board.

You did great work here, nice and clean. I will approve and you can decide whether you want to remove the existing menu or leave that for later (most of those menus will go away)

If you would like to continuing contributing I am happy to talk through roadmap and make you a collaborator.

Thank you again 🚀

PR #754, opened 2024-02-01
Fix: Project manager list is blank on the first app load

arymoraes:
Project list was blank on the first load


cpvalente (approved):
I am unsure I fully understand the issue and the consequences, but if it works, it works

We have planned to create our setup system which will resolve these

For now, this is fine. When we move forwards I would prefer not having directories as string literals in the code and move this to config instead

PR #756, opened 2024-02-02
refactor: normalise data

cpvalente:
I am aware that this is a difficult PR to follow, but I was hoping to have some feedback as well as sharing knowledge of what is changing.

## Normalising the rundown data

This aims to provide stable references to the rundown elements and hopefully improve rendering and data access performance in the clients and backend.

We have modified the data that comes from backend to be normalised.
This means that we get an object with keys as event IDs and values the event data
`rundown: Record<EventId, OntimeEvent>`

an array with the IDs in order
`order: EventId[]`

a revision number, that increments everytime the rundown changed
`revision: number`

The normalised rundown is calculated in the backend every time anything changes.
This is less efficient than the previous solution, but also less prone to bugs. Especially since we plan to add more logic there.
I have left some simple profiling logs that we can watch for, so far I have experienced around 1ms to calculate around 222 events. We should monitor this and worry about performance if it becomes problematic

**Unfortunately**, we need an array of objects to give to react-table. This is especially unfortunate since this is one of the more performance-starved views. Perhaps there is a workaround to feed it a list of IDs from the `order` array and have it fetch its data from the react-query store?

So far, I have unnormalised the data in another hook `useFlatRundown` which I attempted `memoising` using the revision number

Other changes
- Simplify multi-select logic
- Ensure unique IDs when creating events and bump generated IDs to have 6 digits


apps/client/src/common/hooks/useEventAction.ts
```diff
@@ -111,18 +110,15 @@ export const useEventAction = () => {
...
-        const index = optimisticRundown.findIndex((event) => event.id === newEvent.id);
-        if (index > -1) {
-          // @ts-expect-error -- we expect the event types to match
-          optimisticRundown[index] = { ...optimisticRundown[index], ...newEvent };
-
-          queryClient.setQueryData(RUNDOWN, { rundown: optimisticRundown, revision: -1 });
-        }
+        const newRundown = { ...previousData.rundown };
+        // @ts-expect-error -- we expect the events to be of same type
+        newRundown[eventId] = { ...newRundown[eventId], ...newEvent };
```
arymoraes, line 120: suggestion: I usually prefer to Type cast instead of expecting a ts error, I still don't love type casting but since we know the types we match maybe we could do

```
        const newRundown = { ...previousData.rundown } as NormalisedRundown;
        newRundown[eventId] = { ...newRundown[eventId], ...newEvent } as OntimeRundownEntry;
```

apps/client/src/features/cuesheet/CuesheetWrapper.tsx
```diff
@@ -69,41 +67,21 @@ export default function CuesheetWrapper() {
         console.error(error);
       }
     },
-    [updateEvent, rundown],
+    [flatRundown, rundownStatus, updateEvent],
   );
 
   const exportHandler = useCallback(
```
arymoraes, line 73: nit: I'd prefer we kept the handler names `handleSomething`, so `handleExport` in this case

packages/utils/src/array-utils/arrayUtils.ts
```diff
@@ -58,6 +58,8 @@ export function reorderArray<T>(array: T[], fromIndex: number, toIndex: number)
 
 export const sortArrayByProperty = <T>(arr: T[], property: string): T[] => {
   return [...arr].sort((a, b) => {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore -- its ok
     return a[property] - b[property];
   });
 };
```
arymoraes, line 65: in theory this would make TS happy:

```
export const sortArrayByProperty = <T>(arr: T[], property: keyof T): T[] => {
  return [...arr].sort((a, b) => {
    const valueA = a[property];
    const valueB = b[property];

    if (typeof valueA === 'number' && typeof valueB === 'number') {
      return valueA - valueB;
    } else if (typeof valueA === 'string' && typeof valueB === 'string') {
      return valueA.localeCompare(valueB);
    }

    throw new Error('Property must be a string or number for sorting');
  });
};
```

if we're sure it can only be an integer or a string

arymoraes (approved):
looks good to me, just a couple of nits but you can ignore them

PR #758, opened 2024-02-03
fix: resolve directory to test-db

cpvalente:
trying to fix the issues with resolving the test files

we seem to be creating a folder `test-db` in the `app` directory, this is incorrect
the `test-db` should be in the `server` directory

I am also confused about what all the variables here mean... Lets just get this working safely for now and deal with it on our planned work


arymoraes (approved):
LGTM and it's good enough until we rewrite setup

PR #762, opened 2024-02-05
Project Manager: Quick Start

arymoraes:


cpvalente (commented):
I think this looks good, just a few questions:
- when i first click create, the form submission is not available, should we not let the user create a project with the default `Untitled Project` title?
- What steps do we take to make sure that the filename is unique? Do we need any at all?
- If we are defaulting to Untitled Project, why not let the user submit an empty form, in the backend we can fill in the project title. Having said that, we might not even need one ?
- can we disable autocomplete for the form fields?

arymoraes:
> I think this looks good, just a few questions:
> 
> * when i first click create, the form submission is not available, should we not let the user create a project with the default `Untitled Project` title?
> * What steps do we take to make sure that the filename is unique? Do we need any at all?
> * If we are defaulting to Untitled Project, why not let the user submit an empty form, in the backend we can fill in the project title. Having said that, we might not even need one ?
> * can we disable autocomplete for the form fields?

Good point, I:

- Made is so we can submit without a title, this will generate a `Untitled` project (there is no title actually, just `Untitled` is the filename).
- Incremental for same filename, so `Untitled`, `Untitled (1)` etc
- Removed autocompletes

cpvalente (approved):
Code looks good to me,

I think the `project-panel` directory needs to export a component from `ProjectPanel` from a file `ProjectPanel.tsx`
I dont think this is THE WAY to structure react or even the best way, just the way we have it in the rest of the project.

Also, a `ProjectPanel.module.scss` should apply to a file of the same name. Again, not a hard rule, just what we have been doing

Find which is the name of the entrypoint for the component (`ProjectList`?) and rename it to `ProjectPanel`. Lets merge once that is done

arymoraes:
> Code looks good to me,
> 
> I think the `project-panel` directory needs to export a component from `ProjectPanel` from a file `ProjectPanel.tsx` I dont think this is THE WAY to structure react or even the best way, just the way we have it in the rest of the project.
> 
> Also, a `ProjectPanel.module.scss` should apply to a file of the same name. Again, not a hard rule, just what we have been doing
> 
> Find which is the name of the entrypoint for the component (`ProjectList`?) and rename it to `ProjectPanel`. Lets merge once that is done

`ProjectPanel` is actually the entrypoint, and it renders `ProjectList`

```
import { useState } from 'react';
import { Button } from '@chakra-ui/react';

import * as Panel from '../PanelUtils';

import ProjectList from './ProjectList';

export default function ProjectPanel() {
  const [isCreatingProject, setIsCreatingProject] = useState(false);

  const handleToggleCreate = () => {
    setIsCreatingProject((prev) => !prev);
  };

  return (
    <>
      <Panel.Header>Project</Panel.Header>
      <Panel.Section>
        <Panel.Card>
          <Panel.SubHeader>
            Manage projects
            <Button variant='ontime-filled' onClick={handleToggleCreate}>
              New
            </Button>
          </Panel.SubHeader>
          <ProjectList onToggleCreate={handleToggleCreate} isCreatingProject={isCreatingProject} />
        </Panel.Card>
      </Panel.Section>
    </>
  );
}
```

PR #763, opened 2024-02-05
refactor: UI for linking events

cpvalente:
I still have to test edge cases, but this seems functional.
Before we worry about the code, how do we feel about
- the UI: is it clear what is happening? does it look ok? how about the style and icons?
- the UX: when we reorder events and change data, do things behave as we would expect
- the data shape: are things named well and the structure easy to move forwards and change


alex-Arc:
good start 
nice looking and good icon choice

kellhogs:
Very nice work here on the UI.
I really like the choice of icons you made here. The icon's itself make it very clear what they are doing once you know it. But I would still to see a little tooltip on hover that does a very short explanation of what happens if you click on it.

Do we maybe want to also lock the time input field for the `eventEditor` ? Feels a bit inconsistent to me that we disable it in the rundown but not in the `eventEditor`.

Also, can we get a setting to choose which entry is disabled by default when adding a new event. I often find myself just typing in durations for the speakers without the start and end times. Could be just another checkbox along with `Start time is last end` and `Event is public` 

UX wise, I find it very intuitive what happens if you move events that are linked and events that aren't. 👍🏻

cpvalente:
Thank you @kellhogs, it is valuable to have your feedback here. You were not part of the initial planning, the fact that you understood the concept is reassuring.

> Do we maybe want to also lock the time input field for the eventEditor ? Feels a bit inconsistent to me that we disable it in the rundown but not in the eventEditor.

Yes, I just wanted to have a feeling for how it works first. Done now.

> Also, can we get a setting to choose which entry is disabled by default when adding a new event. I often find myself just typing in durations for the speakers without the start and end times. Could be just another checkbox along with Start time is last end and Event is public

I believe the `QuickAdd` UI is beginning to show its limits. We have planned a feature for `presets`, where you can save any event as a preset. We will also work on defaults for new events.
Do you still think this option should be available from the `QuickAdd`? 

While I assume that there are some edges to iron out I have finished the work for now. Would you mind taking another look?
We have improvements to delays and event presets in the roadmap which should tidy up the feature further

apps/client/src/common/utils/dateConfig.ts
```diff
@@ -155,19 +155,21 @@ export const forgivingStringToMillis = (value: string): number => {
   return millis;
 };
 
-export function millisToDelayString(millis: number | null): undefined | string | null {
+export function millisToDelayString(millis: number | null, small = false): undefined | string | null {
```
arymoraes, line 158: I'm usually not a big fan of having a `boolean` as an optional argument, I feel maybe having "small" | "large" with "large" as default, but its just a nit, maybe its my rails background where the rails linter screams when we do this

apps/client/src/features/cuesheet/Cuesheet.module.scss
```diff
@@ -102,6 +103,9 @@ $table-header-font-size: calc(1rem - 3px);
     position: sticky;
     left: 47.5%; // center of the screen, ish
     padding: 0.5rem 0;
+    &:first-letter {
+      text-transform: uppercase;
+    }
```
arymoraes, line 108: if this does what I'm thinking it does (capitalizes the first letter) `text-transform: capitalize?`

apps/server/src/services/rundown-service/rundownCache.ts
```diff
@@ -6,80 +6,123 @@ import {
...
+/** Returns an ID guaranteed to be unique */
+export function getUniqueId(): string {
+  if (isStale) {
+    generate();
+  }
   let id = '';
   do {
     id = generateId();
-  } while (!isIdUnique(persistedRundown, id));
+  } while (Object.hasOwn(rundown, id));
```
arymoraes, line 114: didn't know this exist, I expected it to be `hasOwnProperty` but apparently its newish

apps/client/src/features/cuesheet/Cuesheet.module.scss
cpvalente: Thank you @arymoraes 
If I am correct, capitalise will make uppercase all the first characters in every word, so like
`My Cuesheet Example`

where we are looking for to capitalising only the very first letter to `My cuesheet example`

apps/client/src/common/utils/dateConfig.ts
cpvalente: I believe you and rails are correct. A union flag here would make the code more descriptive

alex-Arc (approved):
NICE!

there is something about the way End and Duration lock out each other that is a bit confusing but we can look at that later

cpvalente:
> there is something about the way End and Duration lock out each other that is a bit confusing but we can look at that later

Hmm... interesting. My first instinct was also that the locking was optional, but while working with the feature I realised that we would always effectively need to lock one of the values, whether the user has chosen to or not

The decision was to make the evident in the UI

Let's look at improving the UI once we are through with the breaking changes.

apps/client/src/features/cuesheet/Cuesheet.module.scss
arymoraes: good point, I thought capitalise would just do the first one

PR #765, opened 2024-02-07
Move app settings

biancahpp:
This will move the `App`  in Settings 

image

into `General`

image


apps/client/src/features/app-settings/panel/general-panel/GeneralPanelForm.tsx
```diff
@@ -0,0 +1,145 @@
...
+        </GeneralSplitInput>
+        <GeneralFooter
+          formId='app-settings'
+          handleRevert={onReset}
+          isDirty={isDirty}
+          isValid={isValid}
+          isSubmitting={isSubmitting}
+        />
+      </form>
+      {submitError && <span className={style.error}>{submitError}</span>}
```
cpvalente, line 142: hi, thank you for your work,
I think this generally looks fine, we just need a few tweaks to UI that I can help with.

Meanwhile, the old structure of the components is a bit cumbersome as you found with having to pass the references. In the new version of settings I have opted from leaving this at top level.

I hope this makes it easier to develop. I feel that our app settings are not particularly complex and dont warrant this abstraction

have a look here for how we could do it
https://github.com/cpvalente/ontime/blob/v3/apps/client/src/features/app-settings/panel/integrations-panel/OscIntegrations.tsx#L102

Reach out if this is unclear

apps/client/src/features/app-settings/panel/general-panel/GeneralFooter.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+interface GeneralFooterProps {
+  formId: string;
+  handleRevert: () => void;
+  isDirty: boolean;
+  isValid: boolean;
+  isSubmitting: boolean;
+}
+
+export default function GeneralFooter(props: GeneralFooterProps) {
```
cpvalente, line 13: I dont think we need this footer separation anymore
This used to happen as we attempted on reusing the footer elements across tabs of the modal.

We no longer have this structure, this buttons can be in the form, which makes everything much easier (even if more repetitive)

PR #770, opened 2024-02-10
Integrations data

cpvalente:
The objective of this PR is to continue the migration to the new integrations. 
This should completely replace the old solution


apps/client/src/features/app-settings/panel/integrations-panel/HttpIntegrations.tsx
```diff
@@ -32,51 +97,76 @@ export default function HttpIntegrations() {
...
+                            },
+                          })}
+                        />
+                        {maybeError && <Panel.Error>{maybeError}</Panel.Error>}
+                      </td>
+                      <td>
+                        <IconButton
+                          size='sm'
+                          variant='ontime-ghosted'
+                          color='#FA5656' // $red-500
```
biancahpp, line 158: An improvement to keep in mind for the future would be to create an `enum` or something like that to avoid hard coding colors throughout the codebase.

biancahpp (approved):

arymoraes (approved):
looks good to me

apps/client/src/features/app-settings/panel/integrations-panel/HttpIntegrations.tsx
arymoraes: we could even use a script to sync our color scss file into a TS enum so we if we change any color it'll propagate the differences into our enum file, like

```
import fs from 'fs';
import path from 'path';

const inputFilePath = path.join('apps', 'client', 'src', 'theme', '_ontimeColours.scss');
const outputFilePath = path.join('apps', 'client', 'src', 'theme', 'colorsEnum.ts');

fs.readFile(inputFilePath, 'utf8', (err, data) => {
  if (err) {
    console.error(err);
    return;
  }

  const regex = /\$(\w+[-\d]*):\s*(.*?);/g;
  let match;
  let enumEntries = [];

  while ((match = regex.exec(data)) !== null) {
    const [_, name, value] = match;
    const enumName = name.replace(/-\d+/g, (match) => match.toUpperCase().replace('-', ''));
    enumEntries.push(`${enumName} = '${value}'`);
  }

  const enumString = `export enum Colors {\n  ${enumEntries.join(',\n  ')}\n}`;

  fs.writeFile(outputFilePath, enumString, 'utf8', (err) => {
    if (err) {
      console.error(err);
      return;
    }
    console.log('Color enum file has been generated');
  });
});
```

cpvalente: Thank you @biancahpp and @arymoraes  for the input, this is a good idea.

Had a quick google on the subject and it seems that there is a standard for this called [ICSS](https://github.com/css-modules/icss) (Interoperable CSS)
We might be able to configure vite to help us here, would one one of like to give this a go?
https://github.com/vitejs/vite/discussions/13993

alex-Arc (commented):
Nice, will test the server part later

maybe move the +new to the button of  the list
image

right now  if you add one the ui will force scroll down so to add multiple you will be constantly scrolling beck up

PR #774, opened 2024-02-13
Sheets settings

cpvalente:
Converting the import from google sheets to the new app settings

I have attempted simplifying the process by removing steps.
I have also isolated the setup and import steps so that it can be reused with the excel import

We should prioritise merging this PR as soon as it works, the styling needs a lot of tweaking but lets do it separately since the PR is pretty large already


alex-Arc:
the state of authentication seams to not update after closing the auth tab
image

apps/client/src/features/app-settings/panel/sources-panel/useGoogleSheet.ts
```diff
@@ -0,0 +1,145 @@
...
+  };
+
+  /** authenticate with the Google Sheets API */
+  const handleAuthenticate = async () => {
+    try {
+      const authLink = await getSheetsAuthUrl();
+
+      // request windown to open link and check auth when user is back
+      openLink(authLink);
+      window.addEventListener('focus', async () => await getAuthentication(), { once: true });
```
alex-Arc, line 66: When getting the authentication status here you need to handle the result to then enable or disable the next step

PR #775, opened 2024-02-14
Fix danger time not shown

kellhogs:
This fixes a Bug where the danger color isn't shown in the timer view's

This is caused by a wrong order of IF operations.


cpvalente (approved):
Thank you for the initiative and good work with the investigation

PR #780, opened 2024-02-16
refactor: partial state updates

### LINKED ISSUE #346: OSC Integration: On Update
Within the OSC Integration portion of OnTime v2, is "On Update" triggered only when an event is running or is it triggered constantly?

**If it's not supposed to be triggered constantly:**
I assumed (incorrectly?) that this would be triggered only during when an event is running. If that's the case, perhaps this could be renamed to "On Event Running" or something better?

**If it's supposed to be triggered constantly:**
Can On Update be renamed to something like "On Tick", "On Every Second", "On App Update" (or something better than my crappy suggestions). That way users will know this will... Triggers at least once a second, can be more.

cpvalente:
testing the possibility of sending less updates over websocket

Also closes #346


cpvalente:
@coderabbitai review

PR #787, opened 2024-02-22
feat: generate crash report

cpvalente:
Simple PR to add crash reports to ontime. Not that the app ever crashes...


arymoraes (approved):
looks good to me

PR #788, opened 2024-02-22
wip: parse event custom properties

cpvalente:
This is the first part of the work to implement dynamic custom fields.

To avoid creating a large PR I am diving this in parts

In this PR
- [x] add UI and backend to manage custom fields
- [x] add UI to list custom fields in events

To do:
- [ ] list in cuesheet
- [ ] show as option in operator
- [ ] update demo and test DBs
- [ ] refactor rundown cache to repopulate data
- [ ] delete user fields


apps/client/src/common/components/input/colour-input/Swatch.tsx
```diff
@@ -6,21 +6,24 @@ import style from './SwatchSelect.module.scss';
...
+  onClick?: (color: string) => void;
   isSelected?: boolean;
 }
 
 export default function Swatch(props: SwatchProps) {
   const { color, isSelected, onClick } = props;
 
-  const classes = cx([style.swatch, isSelected ? style.selected : null]);
+  const handleClick = () => {
+    onClick?.(color ?? '');
```
arymoraes, line 17: do we need the `?? ''` if color is always a `string`?

cpvalente: good catch, thank you!

PR #792, opened 2024-02-25
[DEMO] add valibot

alex-Arc:


cpvalente:
Hi, it is ok with me to add Valibot to the server, but I am unsure why we would install the library as a standalone PR?

PR #795, opened 2024-02-27
refactor: deprecate presenter and subtitle

cpvalente:
Finally removing the `presenter` and `subtitle` fields

@alex-Arc and @kellhogs , just some sanity-checking here
Removing these fields simplifies data entry and clarifies the use case for custom fields.
`Presenter` and `subtitle` are historically connected to Ontime first being designed for conference centers and dont serve our users well, I think this is a good change

The downside is that there is more configuration for the users. Also some of the cards feel a bit empty (thinking of `Timer` and `Operator` views)

Without worrying about the code, it would be good to have some feedback on the feature.
Once this is complete, we have done all the breaking changes lined up for v3 🥳


cpvalente:
Thank you @alex-Arc 
I will move your feedback to the UI tweaks list

PR #812, opened 2024-03-12
Improve error handling in server

alex-Arc:
implement a toErrorResponse function to streamline the way we handle errors from the controllers


packages/utils/src/types/errors.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function toErrorResponse(maybeError: unknown): ErrorResponse {
+  if (isErrorResponse(maybeError)) return maybeError;
+  try {
+    return new Error(JSON.stringify(maybeError));
+  } catch {
+    return new Error(String(maybeError));
+  }
+}
```
cpvalente, line 19: We start accumulating several of these, from top of my head `maybeAxiosError`, `maybeError`, `unpackError`

I think it is perfectly fine to narrow it down to the context, so, `maybeAxiosError` only makes sense in the client and here we likely just want to handle generic errors of Error.
We dont have a complex use case and I wonder if the `unpackError` function could fit the bill here. Maybe it should be called `getErrorMessage()`

For reference, this is also a good article about it, with a resolution very similar to yours.
In the article we can find the motivation for the different bits of logic so that we can see how much of it applies to our codebase
https://kentcdodds.com/blog/get-a-catch-block-error-message-with-typescript

alex-Arc: > For reference, this is also a good article about it, with a resolution very similar to yours

That is because is is lifted straight from there 😉

cpvalente: hahaha, nice...
In this case I still prefer the original offered by Kent C Dodds
```ts
function getErrorMessage(error: unknown) {
  if (error instanceof Error) return error.message
  return String(error)
}
```

I dont think we have to worry about the edge cases that the more complex solution handles

cpvalente:
Hi @alex-Arc , since we are sticking with express, it would be really nice to make the errors into middleware
maybe something like?

https://dev.to/qbentil/how-to-write-custom-error-handler-middleware-in-expressjs-using-javascript-29j1
https://buttercms.com/blog/express-js-error-handling/

apps/server/src/api-data/index.ts
```diff
@@ -25,3 +25,8 @@ appRouter.use('/sheets', sheetsRouter);
 appRouter.use('/excel', excelRouter);
 appRouter.use('/url-presets', urlPresetsRouter);
 appRouter.use('/view-settings', viewSettingsRouter);
+
+//we don't want to redirect to react index when using api routes
+appRouter.all('/*', (_req, res) => {
+  res.status(404).send();
+});
```
cpvalente, line 32: nice

cpvalente (approved):
Nice, great work here.

Thank you also for the improvement on the fallback

PR #813, opened 2024-03-12
partial ws update for demo

alex-Arc:


apps/server/src/external/demo/app.js
```diff
@@ -44,21 +52,29 @@ const connectSocket = () => {
...
+        break;
+      }
+      case 'ontime-timer': {
+        const { current, playback } = payload;
+        if (playback === 'stop') {
+          timerElement.innerText = '--:--:--';
+        } else {
+          timerElement.innerText = formatTimer(current);
+        }
+        break;
```
cpvalente, line 77: how about something like

```ts
function updateTimerElement(playback, timerValue) {
  const timerElement = document.getElementById('timer');
  if (playback === 'stop') {
    timerElement.innerText = '--:--:--';
  } else {
    timerElement.innerText = formatTimer(timerValue);
  }
}

switch (type) {
  case 'ontime': {
    // destructure known data from ontime
    // see https://docs.getontime.no/api/osc-and-ws/
    const { timer, playback } = payload;
    updateTimerElement(playback, timer.current);
    break;
  }
  case 'ontime-timer': {
    const { current, playback } = payload;
    updateTimerElement(playback, current);
    break;
  }
}
```

cpvalente (approved):
Looks good, thank you

PR #817, opened 2024-03-14
[DEMO] Event presets

alex-Arc:
preset


cpvalente:
Hi @alex-Arc , thank you for your work here

I have doubts about the preset workflow; I believe it raises questions as to what data should be part of a preset or generalised.

Most of an event data doesn't really make too much sense to preset: cue, title, note, times, custom fields...

I could imagine that we would want to make presets for departments ie: Sound preset would give `colour: red` but I cant think of much else that could be premade?
For this case maybe the clone feature is sufficient

I wonder if the real improvement is to make it easier for the user to define defaults such as `isPublic`, `warningTime`, `linkStart`, `timerType` and allow quickly changing things that are more likely to need to be different for different events (both from quick add and context menu)

What do you think?

PR #818, opened 2024-03-14
wip: experiments with adding time

cpvalente:


cpvalente:
lets make the input as text fields

PR #821, opened 2024-03-15
feat: url navigation for settings

cpvalente:
add navigation based on url for settings


alex-Arc (commented):
nice stuff!
although I'm a bit sad to see the `esc` removed, is that necessary?

cpvalente:
> nice stuff!
> 
> although I'm a bit sad to see the `esc` removed, is that necessary?

I removed it because I kept closing things accidentally. If you think we should keep it I will revert that change

PR #823, opened 2024-03-15
refactor: obfuscate pincode in transit

### LINKED ISSUE #755: [Security Issue] Pincode visible (plain) in API endpoint
I recently found a pretty big security issue in Ontime.

The /ontime/settings route returns information about the Ontime instance, but also returns the editor and operator pin codes in plain text.

This endpoint doesn't require any authentication, so is publicly accessible. That means anyone can "hack" into the Editor/Operator pages.

cpvalente:
Adding some simple code to obfuscate the pincodes while in transit

any user who is capable of inspecting the network tab will also easily circumvent this. But it I guess it is a micro 
improvement and harmless (and a little fun to investigate)
Closes #755 

@alex-Arc do you think the pincode workflow needs improvements? I was a bit annoyed with having to type it everytime I refreshed the page

### DIFF
--- a/apps/client/src/common/context/AppContext.tsx
+++ b/apps/client/src/common/context/AppContext.tsx
@@ -26,7 +26,6 @@ export const AppContextProvider = ({ children }: PropsWithChildren) => {
 
   useEffect(() => {
     if (status === 'pending') return;
-    if (!data) return;
     const previousEditor = sessionStorage.getItem(storageKeys.editor);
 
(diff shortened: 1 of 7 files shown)


alex-Arc:
> do you think the pincode workflow needs improvements? I was a bit annoyed with having to type it everytime I refreshed the page

yes it would be nice if that was remembers somehow

alex-Arc (approved):

PR #830, opened 2024-03-18
Fix to `lastLoadedProject` being null

arymoraes:
by throwing an exception there we ensure the code inside the `catch` block runs, which will then create and assign a `lastLoadedProject`

### DIFF
--- a/apps/server/src/setup/index.ts
+++ b/apps/server/src/setup/index.ts
@@ -86,6 +86,9 @@ export const uploadsFolderPath = join(getAppDataPath(), config.uploads);
 const getLastLoadedProject = () => {
   try {
     const appState = JSON.parse(fs.readFileSync(appStatePath, 'utf8'));
+    if (!appState.lastLoadedProject) {
+      throw new Error('No last loaded project');
+    }
     return appState.lastLoadedProject;
   } catch {
     if (!isTest) {
--- a/apps/server/src/setup/index.ts
+++ b/apps/server/src/setup/index.ts
@@ -83,17 +83,21 @@ export const resolveExternalsDirectory = join(isProduction ? getAppDataPath() : 
 export const appStatePath = join(getAppDataPath(), config.appState);
 export const uploadsFolderPath = join(getAppDataPath(), config.uploads);
 
+const ensureAppState = () => {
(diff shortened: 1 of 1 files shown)


apps/server/src/setup/index.ts
```diff
@@ -86,6 +86,9 @@ export const uploadsFolderPath = join(getAppDataPath(), config.uploads);
 const getLastLoadedProject = () => {
   try {
     const appState = JSON.parse(fs.readFileSync(appStatePath, 'utf8'));
+    if (!appState.lastLoadedProject) {
+      throw new Error('No last loaded project');
+    }
```
cpvalente, line 91: if we are throwing inside a try-catch, it seems like we could just be checking for it instead?

arymoraes: that's because we want to execute the code that is inside the `catch` block when it happens, I mean I could extract it to a function and call it here

cpvalente:
```suggestion
async function ensureAppState() {
  try {
    await fs.access(appStatePath);
  } catch {
    if (!isTest) {
      await ensureDirectory(getAppDataPath());
      await fs.writeFile(appStatePath, JSON.stringify({ lastLoadedProject: 'db.json' }));
    }

    if (!appState.lastLoadedProject) {
      throw new Error('No last loaded project');
    }

  }
}

async function getLastLoadedProject() {
  await ensureAppState();

  const appState = JSON.parse(await fs.readFile(appStatePath, 'utf8'));
  return appState.lastLoadedProject || 'db.json';
}
```

arymoraes: should we not keep things synchronous here? otherwise when we use `getLastLoadedProject` here:

`const lastLoadedProject = isTest ? 'db.json' : getLastLoadedProject();`

well I think node now supports top level `await`? If so then its fine

arymoraes: it's on startup, so I extracted the functionality into a function and we call it if it doesnt exist

cpvalente: 👍

cpvalente (approved):

PR #837, opened 2024-03-20
fix: uncaught error in http integration

alex-Arc:


apps/server/src/services/integration-service/HttpIntegration.ts
```diff
@@ -45,18 +45,13 @@ export class HttpIntegration implements IIntegration<HttpSubscription> {
...
     }
   }
 
-  async emit(path: URL) {
-    await got.get(path, {
-      retry: { limit: 0 },
+  emit(path: string) {
+    got.get(path, { retry: { limit: 0 } }).catch((err) => {
+      logger.error(LogOrigin.Tx, `HTTP Integration: ${err.code}`);
     });
```
cpvalente, line 55: this seems ok to me, but since got is not handling anything, why would we not use node fetch instead?
the point of got would have been to simplify some operation (error handling, retrying, parsing...) but we seem to do none of that?

what do you think?

alex-Arc: I would like to add back the option for https at some point
and got handles the reusing of the sockets for us

PR #839, opened 2024-03-20
Improve spreadsheet import flow

alex-Arc:
import


apps/client/src/common/api/db.ts
```diff
@@ -181,6 +181,17 @@ export async function importSpreadsheetPreview(file: File, options: ImportMap):
   return response.data;
 }
 
+export async function getSpreadsheetWorksheetNames(file: File): Promise<string[]> {
```
cpvalente, line 184: I am unsure I follow this endpoint, we already upload the excel, could we not use that instead?

cpvalente (commented):
remember to run fix the linter issues.
I am not sure about the structure here, I will take some time to look into it once we can merge

packages/utils/src/date-utils/isTimeString.ts
```diff
@@ -14,6 +14,6 @@ export const isTimeString = (text: string): boolean => {
   // ([0-5]?\d)          # SS (required)
   // $                   # End of string
 
-  const regex = /^(?:(?:([01]?\d|2[0-3])[:,.])?([0-5]?\d)[:,.])?([0-5]?\d)$/;
+  const regex = /^(?:(?:([01]?\d|2[0-3])[:,.])?([0-5]?\d)[:,.])?([0-5]?\d)?(\s)?([APap][Mm])?$/;
```
cpvalente, line 17: 🥶

e2e/tests/features/207-spreadsheet-import.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  const fileChooserPromise = page.waitForEvent('filechooser');
+  await page.getByRole('button', { name: 'Import from spreadsheet', exact: true }).click();
+  const fileChooser = await fileChooserPromise;
+  await fileChooser.setFiles(fileToUpload);
+
+  const worksheetOption1 = await page.locator('[id="event\\ schedule"]').selectOption('Sheet2');
+  expect(worksheetOption1).toEqual(['Sheet2']);
+
+  const worksheetOption2 = await page.locator('[id="event\\ schedule"]').selectOption('test');
+  expect(worksheetOption2).toEqual(['test']);
```
cpvalente, line 24: really great effort here, I wonder if we could add a `data-testId` to make this selectors simpler?

apps/server/src/utils/time.ts
```diff
@@ -41,16 +41,24 @@ const parse = (valueAsString: string): number => {
  * @param {boolean} fillLeft - autofill left = hours / right = seconds
  * @returns {number} - time string in millis
  */
-export const forgivingStringToMillis = (value: string, fillLeft = true): number => {
+export const forgivingStringToMillis = (value: string, fillLeft: boolean = true): number => {
   let millis = 0;
 
+  //force lower case before AM/PM check
+  value = value.toLowerCase();
```
cpvalente, line 48: we shouldnt be mutating the parameters in a function. That may lead us to hard to find bugs were the data is changing in unpredictable ways, we can just make a new variable 
`const lowerValue = value.toLowerCase();`

apps/server/src/utils/parser.ts
```diff
@@ -194,7 +194,7 @@ export const parseExcel = (excelData: unknown[][], options?: Partial<ImportMap>)
         event.title = makeString(column, '');
         // if this is a block, we have nothing else to import
         if (event.type === SupportedEvent.Block) {
-          continue;
+          /* empty */
```
cpvalente, line 197: i think the reason the continue was here was to stop the execution of that block.
That in turn would make sure we dont continue parsing an element after we found out it was a block

Why is this being removed? maybe I am missing something?

If there is a reason to remove the continuewe should remove the conditional entirely

apps/server/src/api-data/sheets/sheets.controller.ts
```diff
@@ -56,6 +57,16 @@ export async function revokeAuthentication(_req: Request, res: Response) {
   }
 }
 
+export async function getWorksheetNamesFromSheet(req: Request, res: Response) {
+  try {
+    const { sheetId } = req.params;
+    const { worksheetOptions } = await getWorksheetOptions(sheetId);
```
cpvalente, line 63: if this function has a single return, I dont see why we need the overhead of wrapping it in an object? am i missing something?

apps/server/src/api-data/excel/excel.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+import { Request, Response } from 'express';
+import { extname, join } from 'path';
+import { existsSync, renameSync } from 'fs';
+import { generateId } from 'ontime-utils';
+import { uploadsFolderPath } from '../../setup/index.js';
+import { getExcelWorksheets, handleMaybeExcel } from '../../utils/parser.js';
+
+export async function postExcel(req: Request, res: Response) {
```
cpvalente, line 13: I dont really like the idea of having controllers accessing file system. although I know that we have them, we shouldnt make this worse

lets create a excel.service.ts file and push some of the code to do with filesystem there.
Also, I think we have plenty of usages for existSync and renameSync that we could start cleaning up, i reckon we may have reusable functions that do this, or should do if not

cpvalente (commented):
there is a lot of code here
I made a few suggestions for cleanup, lets start with those and I will look into the details more carefully after

packages/utils/src/date-utils/isTimeString.ts
alex-Arc: update the comment

cpvalente (approved):
🚀

alex-Arc:
if you have a chance to skim it that would be great
especially the new service file https://github.com/cpvalente/ontime/blob/improve-spreadsheet-import-flow/apps/server/src/api-data/excel/excel.service.ts

alex-Arc:
image

this doesn't fail localy

apps/server/src/utils/parser.ts
```diff
@@ -250,11 +243,16 @@ export const parseExcel = (excelData: unknown[][], options?: Partial<ImportMap>)
     // if any data was found in row, push to array
     const keysFound = Object.keys(event).length + Object.keys(eventCustomFields).length;
     if (keysFound > 0) {
-      if (timerTypeIndex === null) {
-        event.timerType = TimerType.CountDown;
-        event.type = SupportedEvent.Event;
+      // if it is a Block type drop all other filed
+      if (event.type === SupportedEvent.Block) {
```
alex-Arc, line 247: cleans up the block event

apps/server/src/utils/parser.ts
```diff
@@ -192,10 +189,6 @@ export const parseExcel = (excelData: unknown[][], options?: Partial<ImportMap>)
         }
       } else if (j === titleIndex) {
         event.title = makeString(column, '');
-        // if this is a block, we have nothing else to import
-        if (event.type === SupportedEvent.Block) {
```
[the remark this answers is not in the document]
alex-Arc: yeah

it tried testing out what this skips, and calling continue in an exhaustive else if stack doesn't change anything
and as the type might not be the first thing found all the timer event fields will be added to a block event

so i added the clean up after https://github.com/cpvalente/ontime/pull/839/files/9b459c1fe43fe7359291a83f0f15907d06330d82#r1536170551

PR #850, opened 2024-03-22
Navigation

cpvalente:
consistent navigation across all views
keeps different navigation for production and view 

@alex-Arc 
- in the `cuesheet`, should we make the settings open in a drawer like the OP?
- Should the edit icon be a cogwheel thing instead of a pencil?
- i wonder if the menu disappearing and showing up in the `editor` and `cuesheet` may be annoying since there is a lot of interaction?
(Dont worry about the code, is just presentation. The important thing is whether it feels good)

TODO:
- [x] can the `cuesheet` and `op` use the same overview bar as the editor?
- [x] recover follow mode in cuesheet


alex-Arc:
> in the cuesheet, should we make the settings open in a drawer like the OP?

yes if we can

alex-Arc (commented):
Nice
could we maybe in the "technical views"
make the icons part of the info bar?


Screenshot 2024-03-25 131712

the pop over is a bit annoying

i feel like having the save button close the editor pane like the cancel button dose would also be nice

cpvalente:
> could we maybe in the "technical views"
> 
> make the icons part of the info bar?


Sounds good. I will merge this for now and work on improving UI

> i feel like having the save button close the editor pane like the cancel button dose would also be nice

Which save and cancel button?

PR #852, opened 2024-03-24
Fix to cuesheet table header not resizing

arymoraes:
Problem is we're memoizing the component, and `headerGroups` prop doesn't have the actual sizes, it is hidden behind a getter.

This means React won't think of it as an actual prop change. If we still want to memoize the component, we could pass the header sizes as a prop separately

### DIFF
--- a/apps/client/src/features/cuesheet/cuesheet-table-elements/CuesheetHeader.tsx
+++ b/apps/client/src/features/cuesheet/cuesheet-table-elements/CuesheetHeader.tsx
@@ -1,3 +1,4 @@
+import { memo, useEffect } from 'react';
 import { Tooltip } from '@chakra-ui/react';
 import {
   closestCenter,
@@ -29,6 +30,12 @@ interface CuesheetHeaderProps {
 function CuesheetHeader(props: CuesheetHeaderProps) {
   const { headerGroups } = props;
   const [columnOrder, saveColumnOrder] = useLocalStorage<string[]>('table-order', initialColumnOrder);
+
+  useEffect(() => {
+    if (!localStorage.getItem('table-order')) {
+      saveColumnOrder(initialColumnOrder);
+    }
+  }, [saveColumnOrder]);
 
   const handleOnDragEnd = (event: DragEndEvent) => {
     const { delta, active, over } = event;
@@ -113,4 +120,4 @@ function CuesheetHeader(props: CuesheetHeaderProps) {
   );
 }
 
-export default CuesheetHeader;
+export default memo(CuesheetHeader);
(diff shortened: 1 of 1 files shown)


cpvalente:
`UseLocalStorage` is a stateful hook that should make the component te-render if the value in local storage changes.

Few questions:
- why does this work for all users that have been using the app for a while, just not new users?
- could the hook logic be buggy?

arymoraes:
> `UseLocalStorage` is a stateful hook that should make the component te-render if the value in local storage changes.
> 
> Few questions:
> 
> * why does this work for all users that have been using the app for a while, just not new users?
> * could the hook logic be buggy?

@cpvalente so I'm still investigating

the issue is when we don't have the complete column order on localstorage, adding a default one works, but I'm still not sure what is causing it

cpvalente:
Thank you.

Does your first impression still holds true?

> Problem is we're memoizing the component, and `headerGroups` prop doesn't have the actual sizes, it is hidden behind a getter.
> 
> 
> 
> This means React won't think of it as an actual prop change. If we still want to memoize the component, we could pass the header sizes as a prop separately

We shouldn't need a complete size object since the columns default if no size is given. The issue seems to be that a resized column is not updated

If creating a complete object we could add that logic into a useEffect, however the solution is not in line with the observed problem. So we might be missing something 

You could try to investigate the impact of removing memoisation of the header, maybe it is not worth it

For a useful metric in the cuesheet you should be making yourself something around 800 entries

arymoraes:
> Thank you.
> 
> Does your first impression still holds true?
> 
> > Problem is we're memoizing the component, and `headerGroups` prop doesn't have the actual sizes, it is hidden behind a getter.
> > This means React won't think of it as an actual prop change. If we still want to memoize the component, we could pass the header sizes as a prop separately
> 
> We shouldn't need a complete size object since the columns default if no size is given. The issue seems to be that a resized column is not updated
> 
> If creating a complete object we could add that logic into a useEffect, however the solution is not in line with the observed problem. So we might be missing something
> 
> You could try to investigate the impact of removing memoisation of the header, maybe it is not worth it
> 
> For a useful metric in the cuesheet you should be making yourself something around 800 entries

The problem is if we never change the order (by dragging the header itself) of the columns, `handleOnDragEnd` never gets called, which will never trigger `saveColumnOrder`.

Basically it doesn't trigger on resizing.

The ideal solution was to make `saveColumnOrder` to be called at the end of the resizing, but I tried a few different ways and couldnt, I'm not familiar with dnd kit either.

I think a good enough hacky solution for now would be to check if we have the column order in localStoarge, and if not, we create it. Then it is all good

cpvalente:
> I think a good enough hacky solution for now would be to check if we have the column order in localStoarge, and if not, we create it. Then it is all good

Sounds good

cpvalente:
@arymoraes , good work investigating. I dont have a better idea. lets merge this

PR #853, opened 2024-03-25
Fix: Removed `updatedAt` from project list

arymoraes:


cpvalente:
I believe the problem I observe is the opposite. The created at date is changed everything a file is modified.

Before we remove the dataset, we should understand the issue.

If we find that for some reason, we can't maintain the created at dataset, we should remove it completely, not just the presentation

alex-Arc:
I think I found the reason
We use lowDb and they use steno to do the actual writing and steno uses always writes to a tmp file and then replaces the original, so there will never be any useful "created at" data available

https://github.com/typicode/lowdb/blob/6fc49ccb69666ad9159d1403514549a6af5fa8ef/src/adapters/node/TextFile.ts#L42

alex-Arc:
We should either keep the updatedAt and remove createdAt

or alternatively manually add a createdAt filed to our json project file type

cpvalente:
> We should either keep the updatedAt and remove createdAt
> 
> 
> 
> or alternatively manually add a createdAt filed to our json project file type

Agreed, if we can't maintain a meaningful dataset we should remove it.

If we get to a point where users find a lot of value in a createdAt we can look into persisting it solutions.

arymoraes:
I kept the `updatedAt`, just not sure about `Last Used` as the column header

cpvalente:
> I kept the `updatedAt`, just not sure about `Last Used` as the column header

if you think it should be renamed, it is ok with me

Remember to update the tests

cpvalente:
Hi @arymoraes , is this ready to be reviewed?

arymoraes:
> Hi @arymoraes , is this ready to be reviewed?

hey yes it is

alex-Arc (approved):

PR #854, opened 2024-03-25
Trim api part 2

alex-Arc:


apps/client/src/common/hooks/useSocket.ts
```diff
@@ -105,11 +105,11 @@ export const useExtraTimerControl = () => {
 };
 
 export const setExtraTimer = {
-  start: () => socketSendJson('extratimer', SimplePlayback.Start),
-  pause: () => socketSendJson('extratimer', SimplePlayback.Pause),
-  stop: () => socketSendJson('extratimer', SimplePlayback.Stop),
-  setDirection: (direction: SimpleDirection) => socketSendJson('extratimer', { direction }),
-  setTime: (time: number) => socketSendJson('extratimer', { settime: time }),
+  start: () => socketSendJson('extratimer', { 0: { playback: SimplePlayback.Start } }),
```
cpvalente, line 108: I wonder if this significantly more complex, both for the implementation and consumer, without much gain.

having said that, i understand that the current api is not extensible the way we want it to be.
How about simply extratimer1 extratimer2 … ?

if we don’t find an instantiated timer at that array position we just ignore the message

apps/server/src/utils/__tests__/parserFunctions.test.ts
```diff
@@ -11,19 +11,20 @@ describe('sanitiseOscSubscriptions()', () => {
 
   it('returns an array of valid entries', () => {
     const oscSubscriptions: OscSubscription[] = [
-      { id: '1', cycle: 'onLoad', message: 'test', enabled: true },
-      { id: '2', cycle: 'onStart', message: 'test', enabled: false },
-      { id: '3', cycle: 'onPause', message: 'test', enabled: true },
-      { id: '4', cycle: 'onStop', message: 'test', enabled: false },
-      { id: '5', cycle: 'onUpdate', message: 'test', enabled: true },
-      { id: '6', cycle: 'onFinish', message: 'test', enabled: false },
+      { id: '1', cycle: 'onLoad', path: '/test', message: 'test', enabled: true },
```
cpvalente, line 14: Can you explain what the advantage is in separing the message from the path? 
according to specification, the osc payload contains a single element `address`

alex-Arc: hmmm it is normally an adders and some values
https://github.com/colinbdclark/osc.js/?tab=readme-ov-file#sending-osc-messages

a simple thing like `/go` can often be just the address but if you wanted to do something like setting a fader
you would often do `/fader/6 100` thereby setting fade 6 to 100

apps/client/src/common/hooks/useSocket.ts
alex-Arc: yeah that looks fine to me

how would the structure in the server be set up?
https://github.com/cpvalente/ontime/pull/854/files#diff-f75b0837f76bb68749074b654c112fa3bacadfbde32581a627983ebe0dc077e3

cpvalente: I think the structure is pretty straightforward. 

The external timer service keeps an array of timers 

The adapter prepares the data object and sends it to the external timer service

The service just needs to compare the index against its array of timers and apply the call to the correct instance

apps/server/src/utils/__tests__/parserFunctions.test.ts
cpvalente: It sounds that what we need is an address and a payload, not an address and a message 

Combination of address and payload is already what we do, I guess the payload is not directly exposed to the user since we only allow it to be an ontime value which we have control over

alex-Arc: yeah we can rename it to address and payload

before there was no payload, the ontime value was sent as part of the address

I'm thinking if we should allow the user to choose the payload type or maybe string is enough for now
as the template parser always returns a string anyways

apps/client/src/common/hooks/useSocket.ts
alex-Arc: is this the form you are thinking of?
`start: () => socketSendJson('extratimer', { extratimer1: { playback: SimplePlayback.Start } }),`

apps/server/src/utils/__tests__/parserFunctions.test.ts
alex-Arc: yes it has always been empty

apps/client/src/common/hooks/useSocket.ts
cpvalente: No. I meant that the address would encode the timer index.

I am not following why so we need to maintain all the messages at the same address?

apps/client/src/features/app-settings/panel/integrations-panel/OscIntegrations.tsx
```diff
@@ -234,23 +236,36 @@ export default function OscIntegrations() {
...
                     </td>
+                    <td className={style.halfWidth}>
+                      <Input
+                        key={field.id}
+                        size='sm'
+                        variant='ontime-filled'
+                        autoComplete='off'
+                        placeholder='{{timer.current}}'
+                        {...register(`subscriptions.${index}.payload`, {
+                          //TODO: how to validate this
```
cpvalente, line 264: Can you describe the validation you want to see here?

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -156,35 +150,32 @@ const actionHandlers: Record<string, ActionHandler> = {
...
-        const reply = extraTimerService.pause();
-        return { payload: reply };
-      }
-      if (payload === SimplePlayback.Stop) {
-        const reply = extraTimerService.stop();
-        return { payload: reply };
-      }
-    }
-
     if (payload && typeof payload === 'object') {
```
cpvalente, line 153: I think we have an assertion utility for objects? If not we can make one

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -156,35 +150,32 @@ const actionHandlers: Record<string, ActionHandler> = {
...
-        const reply = extraTimerService.setTime(time);
-        return { payload: reply };
-      }
-      if ('direction' in payload) {
-        if (payload.direction === SimpleDirection.CountUp || payload.direction === SimpleDirection.CountDown) {
-          const reply = extraTimerService.setDirection(payload.direction);
-          return { payload: reply };
-        } else {
-          throw new Error('Invalid direction payload');
+      Object.entries(payload).forEach(([key, value]) => {
```
cpvalente, line 154: I am not really sure I follow the intention here. Why would I be trying to setup multiple timers in the same message? 

I think we should be carefully not to complicate this prematurely

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -156,35 +150,32 @@ const actionHandlers: Record<string, ActionHandler> = {
...
-      }
+        if ('direction' in value) {
+          if (value.direction === SimpleDirection.CountUp || value.direction === SimpleDirection.CountDown) {
+            extraTimerService[key].setDirection(value.direction);
+          }
+        }
+        if ('playback' in value) {
+          switch (value.playback) {
+            case SimplePlayback.Start:
+              extraTimerService[key].start();
```
cpvalente, line 167: I think the service should maintain many timers. And take as an argument which timer (index?) the command is destined to

We could also make a small object that maps the enum keys to actions to help here

apps/server/src/api-integration/integration.router.ts
```diff
@@ -36,14 +36,14 @@ integrationRouter.get('/*', (req: Request, res: Response) => {
 
   try {
     const actionArray = action.split('/');
-    const params = { payload: req.query as object } as { payload: object | null };
-
+    const params = req.query as object;
+    let payload = params;
```
cpvalente, line 40: Would this not simply copy the reference to a new object? If we are trying to dissociate payload so it can be mutated in this context, we likely need to spread params 

We should also likely be more defensive since we don't know for sure what is inside params

apps/server/src/api-integration/integration.utils.ts
```diff
@@ -16,9 +17,35 @@ const whitelistedPayload = {
...
+    case 'string':
+      return {
+        [customValue.customProperty]: {
+          value: coerceString(customValue.value,
+          ),
+        },
+      };
+  }
+  throw new Error('Unable to parese custom value');
+}
```
cpvalente, line 36: There seem to be many assumptions here. I think we need to think about this piece of code 

The assertion is done prematurely, we likely want a type guard instead

The switch statement offers no value here

apps/server/src/api-integration/integration.utils.ts
```diff
@@ -16,9 +17,35 @@ const whitelistedPayload = {
...
+
 export function parse(property: string, value: unknown) {
+  if (property.startsWith('custom:')) {
+    const customProperty = property.split(':')[1];
```
cpvalente, line 40: I believe we have existing logic for checking this? (Or similar)

apps/server/src/services/extra-timer-service/ExtraTimerService.ts
```diff
@@ -76,4 +76,4 @@ function broadcastReturn(_target: any, _propertyKey: string, descriptor: Propert
 const emit = (state: SimpleTimerState) => eventStore.set('timer1', state);
 const timeNow = () => Date.now();
 
-export const extraTimerService = new ExtraTimerService(emit, timeNow);
+export const extraTimerService = new Array(new ExtraTimerService(emit, timeNow));
```
cpvalente, line 79: I don't think it is correct to have an array of services. But a service that manages an array of timers

With the array of services the management logic gets pushed to the consumer, defeating the point of the service

apps/server/src/api-integration/integration.utils.ts
alex-Arc: can't  find this
do you know where i should look?

apps/server/src/utils/parserFunctions.ts
```diff
@@ -142,14 +142,30 @@ export const parseViewSettings = (data: Partial<DatabaseModel>): ViewSettings =>
...
+
+  return sub.filter(
+    ({ id, cycle, address, payload, enabled }) =>
+      typeof id === 'string' &&
+      isOntimeCycle(cycle) &&
+      typeof address === 'string' &&
+      typeof payload === 'string' &&
+      typeof enabled === 'boolean',
   );
 }
```
alex-Arc, line 170: @cpvalente is this the way to do it?

apps/server/src/api-integration/integration.utils.ts
cpvalente: hmmm, i was so sure I wrote this logic but couldnt find it, sorry about this

apps/server/src/utils/parserFunctions.ts
cpvalente: I would personally not write the conversion, v3 is a major release and breaking changes are ok

apps/server/src/adapters/OscAdapter.ts
```diff
@@ -16,14 +16,14 @@ export class OscServer implements IAdapter {
     this.osc.on('error', (error) => logger.error(LogOrigin.Rx, `OSC IN: ${error}`));
 
     this.osc.on('message', (msg) => {
-      // message should look like /ontime/{path}/{params?} {args} where
+      // message should look like /ontime/{type}/{params?} {args} where
```
cpvalente, line 19: I am not sure about renaming this to `type`
`type` is a loaded word with a lot of meaning, none of which relevant in this case

if you dont like path, how about `command`?

apps/server/src/adapters/utils/parse.ts
```diff
@@ -2,9 +2,19 @@
...
+    } else {
+      return { [path[0]]: value };
+    }
+  }
+  if (value === undefined) {
+    value = path.pop();
+  }
   const obj = path.reduceRight((result, key) => ({ [key]: result }), value);
   if (typeof obj === 'object') return obj;
   return null;
```
cpvalente, line 20: I wonder if this can be simplified, few tips:

- remember === is what you want in most cases
- avoid mutating the passed in arguments
- leave us some clues as to what things mean, like `{ [path[0]]: value };` entails a lot of knowledge about the specifics of the object, which someone who comes here to fix a bug might not have. The tests also dont provide context into the abstraction since they show the objects already abstracted. ie, as someone coming here, it is hard to piece the information from the message

how about something like (didnt test)
```ts
export const integrationPayloadFromPath = (path: string[], value?: unknown): object | string | null => {
  if (path.length === 1) {
    const key = path.at(0);
    return value === undefined ? key : { [key]: value };
  }

  const parsedValue = value === undefined ? path.at(-1) : value;

  const obj = path.reduceRight((result, key) => ({ [key]: result }), parsedValue);

  return typeof obj === 'object' ? obj : null;
};
```

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -44,12 +32,21 @@ const actionHandlers: Record<string, ActionHandler> = {
...
+    Object.entries(payload[eventId]).forEach(([property, value], _) => {
+      const prop = parse(property, value);
+      if (patchEvent.custom && prop.custom) {
+        Object.assign(patchEvent.custom, prop.custom);
+      } else {
+        Object.assign(patchEvent, prop);
+      }
+    });
+    //TODO: don't know how to await this
+    updateEvent(eventId, patchEvent);
```
cpvalente, line 48: the function is not async so it cant be awaited. is there an issue here?

apps/server/src/adapters/OscAdapter.ts
alex-Arc: just wanted it to match the `WS` `{type: , payload:}`

but command or action is also ok for me

apps/server/src/api-integration/integration.controller.ts
alex-Arc: before i removed it we were trying to return the update event data, but that was always empty as the result was not awaited

if we want that functionality back we need to fix it somehow

apps/server/src/adapters/utils/parse.ts
alex-Arc: Nice and it almost worked as is

apps/client/src/features/app-settings/panel/integrations-panel/OscIntegrations.tsx
```diff
@@ -234,23 +236,36 @@
...
+                        })}
```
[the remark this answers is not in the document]
alex-Arc: yes

apps/server/src/services/rundown-service/rundownUtils.ts
```diff
@@ -63,9 +63,16 @@ export function getEventWithId(eventId: string): OntimeEvent | undefined {
...
+  // const loaded = curr
+  const timedEvents = getPlayableEvents();
+  const comparCue = cue.toLowerCase();
+  if (currentEventId) {
+    const currentIndex = timedEvents.findIndex((event) => event.id === currentEventId);
+    const futureEvents = timedEvents.slice(currentIndex);
+    return futureEvents.find((event) => event.cue.toLowerCase() === comparCue);
+  }
+  return timedEvents.find((event) => event.cue.toLowerCase() === comparCue);
 }
```
cpvalente, line 76: this only seams to be used inside RundownService, I would suggest that the consumer passes in a cue and a current index. the function would be much simpler just looking for a match after the index. 

```ts
export function getNextEventWithCue(targetCue: string, currentEventIndex = 0): OntimeEvent | undefined {
  const timedEvents = getPlayableEvents();
  const lowerCaseCue = targetCue.toLowerCase();

  for (let i = currentEventIndex; i < timedEvents.length; i++) {
    const event = timedEvents.at(i);
    if (event && event.cue.toLowerCase() === lowerCaseCue) {
      return event;
    }
  }
}
```


I think the function would be even better if it got passed the timedEvents, as the ones in the other rundownUtils file does.
Not sure why we are keeping two definitions of the same functions. the ones where everything is passed are much easier to test

cpvalente (commented):
Thank you very much for all your work here

This PR is pretty large and there is a lot of code here, lets focus on cleaning up and merging. as long as we can get the API right, we can iterate on adding features later

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -204,3 +210,21 @@ function numberOrError(value: unknown) {
...
+    return converted;
+  }
+  throw new Error('Payload is not a valid string or number');
+}
+
+function successPayloadOrError(success: boolean, error: string) {
+  if (!success) {
+    throw new Error(error);
+  }
+  return { payload: 'success' };
```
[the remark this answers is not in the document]
alex-Arc: could we put this on a TODO: list ?

alex-Arc:
@cpvalente I think the API shape is correct now, is it OK for me to merge it?

cpvalente (approved):
Good work!

Can you please open a empty PR with with a tasklist for the cleanups we are leaving behind?

PR #855, opened 2024-03-25
Keyboard shortcuts

alex-Arc:
# Navigation
- Select entry <kbd>Alt</kbd><kbd>↑</kbd>/<kbd>↓</kbd>
- Reorder entry <kbd>Alt</kbd><kbd>Ctrl</kbd><kbd>↑</kbd>/<kbd>↓</kbd>
- Add event below <kbd>Alt</kbd><kbd>E</kbd>
- Add event above <kbd>Alt</kbd><kbd>⇧</kbd><kbd>E</kbd>
- Add block below <kbd>Alt</kbd><kbd>B</kbd>
- Add block above <kbd>Alt</kbd><kbd>⇧</kbd><kbd>B</kbd>
- Add delay below <kbd>Alt</kbd><kbd>D</kbd>
- Add delay above <kbd>Alt</kbd><kbd>⇧</kbd><kbd>D</kbd>
- Deselect entry <kbd>Esc</kbd>
- Delete entry <kbd>Ctrl</kbd><kbd>⌫</kbd>
- Copy selected entry <kbd>Ctrl</kbd><kbd>C</kbd>
- Past above selected entry <kbd>Ctrl</kbd><kbd>⇧</kbd><kbd>V</kbd>
- Past below selected entry <kbd>Ctrl</kbd><kbd>V</kbd>

# Editing
- Submit and leave normal input <kbd>⏎</kbd>
- Submit and leave multi line input <kbd>Ctrl</kbd><kbd>⏎</kbd>
- Cancel and leave input <kbd>Esc</kbd>

# TODO:
- [ ] delete all events in multi select


apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -94,11 +94,27 @@ export default function Rundown({ data }: RundownProps) {
...
 
+      const escapeElement = (e: KeyboardEvent) => {
+        // is something selected
+        if (e.target) {
+          const elm = e.target as HTMLElement;
+          //is it marked as escapable?
+          console.log(elm.getAttributeNames())
+          if (elm.classList.contains('escapable')) {
+            // then escape it
+            elm.blur();
```
cpvalente, line 105: Curious about what is the idea here. Could you please explain the feature?

Blurring the input will cause the input to submit, which is unlikely what you want when you click escape?

Also, is we are just using the class escapable as flag, we likely would prefer using something else, maybe a data-attribute? Not sure

alex-Arc: > Blurring the input will cause the input to submit

ah yeah that was not the intention

> Also, is we are just using the class escapable as flag, we likely would prefer using something else, maybe a data-attribute? Not sure

that sounds good

I want to be able to deselect the event,
but if a text area is active i want be able to leave that using esc also

so 
if in input leave it
else deselect the event

cpvalente: Ah ok. But then it seems that we need a different structure. The input should blur itself on escape capture.
When we blur the input we stop the event bubbling, this will prevent it to getting to the block.

In this case, if we get a escape in the Rundown component, it is safe to clear the selection

alex-Arc: OK
dose that mean adding a keyboard event listener to all inputs?

```
  useEffect(() => {
    document.addEventListener('keydown', handleKeyDown);

    return () => {
      document.removeEventListener('keydown', handleKeyDown);
    };
  }, [handleKeyDown]);
```

alex-Arc: somthing like this?
https://github.com/cpvalente/ontime/pull/855/commits/bb159c48bbc0f2acfa8675a35563e28604efd2c0

cpvalente: Yes.

We should avoid casting types and I am not sure about the settimeout, but this is the gist of it

Good work

cpvalente:
Can you check the shortcuts against the browser and OS? We need to make sure that there are no conflicts, or the conflicts are harmless and don't affect runtime 

From the top of my head ctrl + enter submits the URL, this sounds like it could be pretty frustrating for the user if they do not have the focus where they thing it is

apps/client/src/features/rundown/rundown-header/RundownHeader.tsx
```diff
@@ -11,13 +13,31 @@ import RundownMenu from './RundownMenu';
...
 
-export default function RundownHeader() {
+interface RundownProps {
+  data: RundownCached;
+}
+
+export default function RundownHeader({ data }: RundownProps) {
+  const { rundown } = data;
   const appMode = useAppMode((state) => state.mode);
+  const clipBoard = useAppMode((state) => state.eventClipBoard);
```
cpvalente, line 23: Suggestion: should we name this something else? I spent some time looking for your clipboard implementation... it is good to have clear expectations of what is happening. How about just copiedEvent?

Also, should this not be together with the event swap feature? Should we make a new store for this?

alex-Arc: Yeah sounds good

apps/client/src/common/utils/eventsManager.ts
```diff
@@ -26,6 +26,6 @@ export const cloneEvent = (event: OntimeEvent, after?: string): ClonedEvent => {
     revision: 0,
     timeWarning: event.timeWarning,
     timeDanger: event.timeDanger,
-    custom: {},
+    custom: event.custom,
```
alex-Arc, line 29: @cpvalente 
was there a reason for not copying the custom data in the `cloneEvent()` funciton?

cpvalente: Yes, in the workflows that we use to guide, the custom data belongs to the team, not to the editor or the schedule

In that light, it makes no sense to copy them into a new event

I am sure there are cases where this is not true, I guess we could make a small popup that let's you select the attributes that get copied. Until then I believe we should keep it simple

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -160,9 +209,27 @@ export default function Rundown({ data }: RundownProps) {
...
+            const newEvent = cloneEvent(copyEvent, cursor ?? undefined);
+            addEvent(newEvent);
+            setEventCopyId(null);
+            break;
+          }
+        }
       }
     },
-    [order, cursor, rundown, setCursor, insertAtCursor, reorderEvent],
+    [order, cursor, rundown, eventCopyId],
```
alex-Arc, line 232: I'm also a bit unsure about this dependency array stuff
there is no reason for functions to be in here right?

cpvalente: Just add what eslint says, that is true in 99% of cases

alex-Arc:
> ctrl + enter 

I think this on is save you have to be actively in the address bar for it to do something

apps/client/src/common/utils/eventsManager.ts
alex-Arc: hmmm yes I would very much like to be able to "deeply" clone an event

I have seen some apps call it it clone and duplicate
where duplicate auto increments addresses and so on
and clone is a straight clone

what do you think of that?
or should we rather go with an popup asking to keep or discard custom data?

cpvalente: I am unsure we would have the clarity to modify the user data the way you describe

It wouldn't mean that we have an understanding of how the user is leveraging its data and can make safe assumptions over how to modify it.

I unfortunately don't think this is the case

In the case of cloning, from what i understood, copying the custom fields have the user more cleanup than the value it added. This was especially true before when the editor couldn't even see the custom fields. Again, custom fields are in the domain of operators using the cuesheet and operator view.

Having the popup for asking which fields to clone could be good, but it also has the downside of making the operations slower by having more steps

For now, we should focus on v3 and try not get too distracted and risk introducing bugs in parallel workflows

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -160,9 +209,40 @@ export default function Rundown({ data }: RundownProps) {
...
+            addEvent(newEvent);
+            setEventCopyId(null);
+            break;
+          }
+        }
       }
     },
-    [order, cursor, rundown, setCursor, insertAtCursor, reorderEvent],
+    [
+      setCursor,
```
cpvalente, line 233: Ouch... I think we need to think about this, lets chat

cpvalente:
Suggestion: 
- Add event below <kbd>Alt</kbd><kbd>E</kbd>
- Add event above <kbd>Alt</kbd><kbd>Shift</kbd><kbd>E</kbd>

- Add block below <kbd>Alt</kbd><kbd>B</kbd>
- Add block above <kbd>Alt</kbd><kbd>Shift</kbd><kbd>B</kbd>

- Add delay below <kbd>Alt</kbd><kbd>D</kbd>
- Add delay above <kbd>Alt</kbd><kbd>Shift</kbd><kbd>D</kbd>

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -29,6 +28,44 @@ import style from './Rundown.module.scss';
...
+    checkKeys(event, { altKey: true, ctrlKey: false, shiftKey: true, code: 'KeyB' }),
+  addBlockDown: (event: KeyboardEvent) =>
+    checkKeys(event, { altKey: true, ctrlKey: false, shiftKey: false, code: 'KeyB' }),
+  copy: (event: KeyboardEvent) => checkKeys(event, { altKey: false, ctrlKey: true, shiftKey: false, code: 'KeyC' }),
+  pastUp: (event: KeyboardEvent) => checkKeys(event, { altKey: false, ctrlKey: true, shiftKey: true, code: 'KeyV' }),
+  pastDown: (event: KeyboardEvent) => checkKeys(event, { altKey: false, ctrlKey: true, shiftKey: false, code: 'KeyV' }),
+  deleteEntry: (event: KeyboardEvent) =>
+    checkKeys(event, { altKey: true, ctrlKey: false, shiftKey: false, code: 'Backspace' }),
+} as const;
+
```
alex-Arc, line 68: @cpvalente I tried to make the shortcut definitions more manageable
 do you think this is OK?

apps/client/src/common/utils/eventsManager.ts
```diff
@@ -29,3 +41,33 @@ export const cloneEvent = (event: OntimeEvent, after?: string): ClonedEvent => {
...
+      skip: entry.skip,
+      colour: entry.colour,
+      after,
+      revision: 0,
+      timeWarning: entry.timeWarning,
+      timeDanger: entry.timeDanger,
+      custom: {},
+    };
+  } else if (isOntimeBlock(entry)) {
+    return { type: SupportedEvent.Block, title: entry.title, after };
```
cpvalente, line 68: I am not sure clone should be a operation for delays and blocks. It offers little benefit

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -10,39 +10,81 @@ import {
...
+    keyCombo.altKey === event.altKey &&
+    keyCombo.ctrlKey === event.ctrlKey &&
+    keyCombo.shiftKey === event.shiftKey &&
+    keyCombo.code === event.code
+  );
+};
+
+const keyCombos = {
+  selectUp: (event: KeyboardEvent) =>
+    checkKeys(event, { altKey: true, ctrlKey: false, shiftKey: false, code: 'ArrowUp' }),
```
cpvalente, line 43: I think the sending if modifiers can be improved

How about an array of key codes?
Checkout react-hotkeys for some inspiration

cpvalente (commented):
Nice, this looks better. We should cleanup the api a little and make sure we are not generating multiple listeners for every shortcuts. But a single listener per component 

Also, why not use the existing useKeypress hook?

If there is a reason to separe this logic, CheckKey could probably be renamed to something more related to shortcuts like registerShortcutListener or something

apps/client/src/common/utils/eventsManager.ts
alex-Arc: yeah but I think it is good to have the shortcuts be consistent

if it doesn't bother you too much I would like to keep it

apps/client/src/features/rundown/Rundown.tsx
alex-Arc: yeah I did find that one
but it looks to me like they are manually maintaining an array of pressed keys...
right now I'm managing with only using the KeybordEvent
I would need some help implementing something different

alex-Arc:
@cpvalente 
> Also, why not use the existing useKeypress hook?

https://github.com/cpvalente/ontime/blob/v3/apps/client/src/common/hooks/useKeyDown.ts

it seams to only take in 1 key or am I missing something?

apps/client/src/features/rundown/rundown-header/RundownHeader.tsx
```diff
@@ -35,6 +66,7 @@ export default function RundownHeader() {
           aria-label='Edit mode'
         />
       </ButtonGroup>
+      <span style={{ opacity: '30%' }}>{inClipBoard}</span>
```
alex-Arc, line 69: @cpvalente any comments on the look of this?
image

cpvalente:
> @cpvalente
> 
> > Also, why not use the existing useKeypress hook?
> 
> https://github.com/cpvalente/ontime/blob/v3/apps/client/src/common/hooks/useKeyDown.ts
> 
> it seams to only take in 1 key or am I missing something?

yes, we can choose to modify the code to take an array
alternatively we can use the hook from mantine 
https://mantine.dev/hooks/use-hotkeys/

apps/client/src/features/rundown/rundown-header/RundownHeader.tsx
cpvalente: hm, I am not sure what I am looking at, but if this is some tip to what i being copied, I dont think it belongs here

PR #857, opened 2024-03-27
Alpha 3 cleanup

cpvalente:
small lint cleanups

@alex-Arc, the logic around spreadsheets is a bit heavy and I thought we could slowly try to simplify, can you sanity check the change?

### DIFF
--- a/apps/server/src/services/sheet-service/sheetUtils.ts
+++ b/apps/server/src/services/sheet-service/sheetUtils.ts
@@ -74,7 +74,6 @@ export function cellRequestFromEvent(
   worksheetId: number,
   metadata,
 ): sheets_v4.Schema$Request {
-  const returnRows: sheets_v4.Schema$CellData[] = [];
   const tmp = Object.entries(metadata)
     .filter(([_, value]) => value !== undefined)
     .sort(([_a, a], [_b, b]) => a['col'] - b['col']) as [string, { col: number; row: number }][];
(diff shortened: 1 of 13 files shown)


alex-Arc (approved):

apps/server/src/services/sheet-service/sheetUtils.ts
```diff
@@ -74,7 +74,6 @@ export function cellRequestFromEvent(
   worksheetId: number,
   metadata,
 ): sheets_v4.Schema$Request {
-  const returnRows: sheets_v4.Schema$CellData[] = [];
   const tmp = Object.entries(metadata)
```
cpvalente, line 77: @alex-Arc can you help me find a better name for this? What is in this array?

apps/server/src/services/sheet-service/sheetUtils.ts
```diff
@@ -146,3 +112,36 @@ export function cellRequestFromEvent(
     },
   };
 }
+
+function getCellData(key: string, event: OntimeRundownEntry) {
+  if (isOntimeEvent(event)) {
+    if (key === 'blank') {
+      return {};
+    }
+    if (key === 'blank') {
```
cpvalente, line 121: Note to self: I think this is supposed to say colour

apps/server/src/services/sheet-service/sheetUtils.ts
alex-Arc: it could be rowData as it is just an array of cellData

cpvalente: Makes sense. And what is the filtering doing?

alex-Arc: I think it takes out all the stuff that was not found in the sheet
so if there was no duration in the sheet it is removed here so it dose not become part of the upload
it is then sorted so it lands in the correct order

PR #862, opened 2024-03-29
feat: chaching

alex-Arc:
wip

trying to improve server performance


cpvalente:
@arymoraes , have you got some experience with caching, can you take a look at the suggestion?

arymoraes:
> @arymoraes , have you got some experience with caching, can you take a look at the suggestion?

I think it makes sense, I believe that if a react file content changes it also changes the hash so we don't have to worry about having old cached code

PR #864, opened 2024-04-01
Rundown modes

cpvalente:
Extending behaviour of actions on events

- [x] Settings for rundown modes
- [x] Remove rundown menu
- [x] remove block menu in event, delay and block
- [x] move menu functionality to context menu
- [x] quick add around event

We should cleanup the logic around event actions, that can probably be done in conjunction with the keyboard shortcuts and we can do it once both PRs are merged
Screenshot 2024-04-01 at 20 18 36
Screenshot 2024-04-01 at 20 19 58


alex-Arc (approved):

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -277,9 +287,9 @@ export default function Rundown({ data }: RundownProps) {
                       />
                     </div>
                   </div>
-                  {((showQuickEntry && hasCursor) || isLast) && (
+                  {((isEditMode && hasCursor) || isLast) && (
```
cpvalente, line 290: note to self: 

```suggestion
                  {isEditMode && (hasCursor || isLast) && (
```

PR #866, opened 2024-04-01
V3 feedback

cpvalente:
small tweaks from user feedback on alpha.3

- [x] rename `general` -> `app settings`
- [x] close `param editor` on submit
- [x] rename buttons in `param editor`
- [x] fix blackout
- [x] rename `message blink` > `blink`
- [x] rename `readonly` -> `read only`
- [x] always show all links in navigation menu
- [x] issue with `offset` showing as positive in `time to end` timers
- [x] show pincode for operator

### DIFF
--- a/apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
+++ b/apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
@@ -52,15 +52,16 @@ export default function ViewParamsEditor({ paramFields }: EditFormDrawerProps) {
     }
   }, [searchParams, onOpen]);
 
-  const onCloseWithoutSaving = () => {
-    onClose();
-
+  const handleClose = () => {
     searchParams.delete('edit');
     setSearchParams(searchParams);
+
+    onClose();
   };
 
   const resetParams = () => {
     setSearchParams();
+    onClose();
   };
 
   const onParamsFormSubmit = (formEvent: FormEvent<HTMLFormElement>) => {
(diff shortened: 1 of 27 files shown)


apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
```diff
@@ -96,10 +99,7 @@ export default function ViewParamsEditor({ paramFields }: EditFormDrawerProps) {
...
         <DrawerFooter className={style.drawerFooter}>
           <Button variant='ontime-ghosted' onClick={resetParams} type='reset'>
-            Reset
-          </Button>
-          <Button variant='ontime-subtle' onClick={onCloseWithoutSaving}>
-            Cancel
+            Reset to default
           </Button>
           <Button variant='ontime-filled' form='edit-params-form' type='submit'>
             Save
```
alex-Arc, line 105: did you want to call this submit?

alex-Arc (approved):

apps/client/src/common/components/view-params-editor/ViewParamsEditor.tsx
cpvalente: I didn't understand the question.. is there something with the naming?

cpvalente: Ah yes, sorry.

I tried it and it felt a bit weird. 
The idea of submit was to clarify the expectation of the window disappearing on click, but I think it made the language more complicated.
What do you think?

alex-Arc: both are OK with me
if you think its a bit wired let's not do it

PR #867, opened 2024-04-01
fix: aplly setting to osc server on change

alex-Arc:


apps/server/src/api-data/osc/osc.controller.ts
```diff
@@ -1,10 +1,11 @@
-import type { ErrorResponse, OSCSettings } from 'ontime-types';
+import { type ErrorResponse, type OSCSettings } from 'ontime-types';
```
cpvalente, line 1: is there a reason for this change? both expressions are the same

I am just wondering if your tooling is doing this for some reason

alex-Arc: yeah think it is when intelisens imports stuff for me

apps/server/src/api-data/osc/osc.controller.ts
```diff
@@ -5,6 +5,7 @@ import { Request, Response } from 'express';
 import { DataProvider } from '../../classes/data-provider/DataProvider.js';
 import { failEmptyObjects } from '../../utils/routerUtils.js';
 import { oscIntegration } from '../../services/integration-service/OscIntegration.js';
+import { startOSCServer } from '../../app.js';
```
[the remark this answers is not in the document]
cpvalente: I think this makes sense in the app. maybe there is a better way to do it. but dont think the controller is the place for it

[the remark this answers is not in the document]
alex-Arc: cool let leave it for now

apps/server/src/services/integration-service/OscIntegration.ts
```diff
@@ -100,7 +85,60 @@ export class OscIntegration implements IIntegration<OscSubscription> {
     this.oscClient.send(message);
   }
 
+  private initTX(enabledOut, targetIP, portOut, subscriptions) {
```
cpvalente, line 88: would making the init functions public not help us restart during runtime?
we will likely want to restart the server and client separately 

Currently each private function is handling its own logic to verify the need for change, but this makes that the consumer needs to have access to the entire config instead of just the changes of the part (server / client) that should be restarted

(hope this made sense)

cpvalente (commented):
Changes look good, we can do some cleanup depending on how we want to consume this

Also, remember to type the inputs and run prettier
(turn on strict mode in typescript while you work to get extra points) :D

apps/server/src/services/integration-service/OscIntegration.ts
alex-Arc: hmmm maybe
we always get the whole OSCSettings struct from the client anyway

https://github.com/cpvalente/ontime/blob/22d1181443394d952b94a5268c4a0d6ec66e087d/apps/server/src/api-data/osc/osc.controller.ts#L13-L25


the consumer would then need to check what has changed...?

PR #869, opened 2024-04-04
Typescript null

cpvalente:
Fixing issue raised in #863 by implementing an extra flag in all mutation which indicates whether a mutation actually happenned

We can use this flag to avoid doing side effects.
In future™, we could also use this to do better response codes

Took the chance to do some typescript improvements and simplify initial parsing of json data


alex-Arc (approved):
the issue I had is fixed with this 👍🏽

PR #870, opened 2024-04-05
V3 feedback2

cpvalente:
Small tweaks from user feedback on alpha 3

- [x] remove Add actions from context menu, the feedback was that this made the menu confusing. Since the quick add now shows in edit mode, that should be the way to add events
- [x] restyle quick add buttons to make sure they are always the same size (opinion was that change of sizes was confusing and distracting)
- [x] simplify quick add
- [x] rename action in delay from `Apply` to `Make permanent`, this attempts to clarify that the action is not mandatory
- [x] fix missing negative timer
- [x] small style tweaks
- [x] fix issue with multi progress bar when timer duration is 0
- [x] fix issue with project list being undefined
- [x] cuesheet overview has running timer data  
 
@alex-Arc , what do you think of changing the text in the delay block from Apply to Make permanent?
The idea is that users seem to be confused as to what apply does. Delays do not need to be applied, this only makes them part of the schedule. Another alternative was to call it `Reschedule`, but I thought that could be even more confusing

### DIFF
--- a/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.tsx
+++ b/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.tsx
@@ -19,7 +19,7 @@ interface MultiPartProgressBar {
 export default function MultiPartProgressBar(props: MultiPartProgressBar) {
   const { now, complete, normalColor, warning, warningColor, danger, dangerColor, hidden, className = '' } = props;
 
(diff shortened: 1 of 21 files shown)


alex-Arc:
> what do you think of changing the text in the delay block from Apply to Make permanent?

it's OK, a bit long maybe 
but can't come up with anything better

cpvalente:
> > what do you think of changing the text in the delay block from Apply to Make permanent?
> 
> it's OK, a bit long maybe but can't come up with anything better

same, lets roll with this and see what happens

alex-Arc (approved):

PR #874, opened 2024-04-08
Draft: Feature - Edit multiple events

biancahpp:


apps/client/src/features/rundown/event-editor/EventEditorWrapper.tsx
```diff
@@ -0,0 +1,140 @@
...
+      timeEnd: allHaveSameValue(events, 'timeEnd') ? events[0]['timeEnd'] : undefined,
+      timeStart: allHaveSameValue(events, 'timeStart') ? events[0]['timeStart'] : undefined,
+      timeStrategy: allHaveSameValue(events, 'timeStrategy') ? events[0]['timeStrategy'] : TimeStrategy.Unkown,
+      timeWarning: allHaveSameValue(events, 'timeWarning') ? events[0]['timeWarning'] : undefined,
+      timerType: allHaveSameValue(events, 'timerType') ? events[0]['timerType'] : TimerType.Unkown,
+      title: allHaveSameValue(events, 'title') ? events[0]['title'] : '', // ??
+      type: allHaveSameValue(events, 'type') ? events[0]['type'] : SupportedEvent.Event,
+      cue: allHaveSameValue(events, 'cue') ? events[0]['cue'] : 'unkown',
+    };
+
```
biancahpp, line 124: This is where I left off, trying to figure out what the fallback values should be for when the values are not the same for the selected events.

biancahpp:
Also, I'm not sure why, but it takes a little while for the UI to update after I select more than one event (using ctrl).

apps/client/src/features/rundown/event-editor/EventEditorWrapper.tsx
alex-Arc: Hey Bianca, nice work

here is an example from another event related software

https://github.com/cpvalente/ontime/assets/19875125/98391149-14bf-4eeb-a49d-c37d04ed19f1

values that are identical on both elements stay visible
and element that differ are symbolized with <Multiple Values>

I think that is pretty clear

apps/client/src/common/components/input/colour-input/Swatch.tsx
```diff
@@ -25,5 +25,11 @@ export default function Swatch(props: SwatchProps) {
       </div>
     );
   }
+
+  if (color === 'unkown') {
+    <div className={`${classes} ${style.center}`} onClick={handleClick}>
+      ?
+    </div>;
+  }
```
alex-Arc, line 33: I'm unsure about this
I think that an unknown color should result i nothing being highlighted in the swatch
or dose it serve a function?

alex-Arc (commented):
Good start
I'm no expert in react, but let me know if you need help setting up a batchEdit endpoint with the server

cpvalente:
Hi Bianca, can you send me an example of a rundown with poor performance?
I tried investigating and everything seemed sensible to me

When a new event is selected, we need to re-render all the wrappers (to show the selection outline) and the event editor
This seems expected?
I was also unable to verify any major performance issues. Granted that my rundown has 14 elements, but a full re-render is hitting 40ms
Screenshot 2024-04-27 at 11 20 40

PR #876, opened 2024-04-09
chore: organise settings

cpvalente:
Small PR to improve organisation of settings
also
- adds a code change to prevent cycling when going to the next event
- fixes the stale switches in settings


alex-Arc:
we should gray out the "->" button when at the end
like the other one is when we are at the start

image
image

alex-Arc:
And now i'm nit picking so it is fine to skip thees 

feature settings doesn't really mean anything to me... (but I don't have a better name)
image

and for consistency it would be nice to have a project data subheading also
image

cpvalente:
thank you for the feedback @alex-Arc , can you take another look?

alex-Arc:
the start button should still be available when the last event is loaded
image

alex-Arc:
saw a commit about the progress bar corners
I still see some overflow, don't know if it is related

image

cpvalente:
Thank you @alex-Arc , good catch

The corners dont look great, that is because of how the `MultiProgressbar` is composed
The change you saw was for the simple `ProgressBar` in `/backstage`

alex-Arc (approved):

PR #878, opened 2024-04-10
Granular ws

cpvalente:
I realised that we send the entire websocket payload every second
this PR makes it that we only ever send the things that change

this makes the app transmit around 8x less data. more importantly, it would allow the consumer to do less work on updating
regardless, the amount of data transmitted is very low (around 200b for a full packet) and it is arguable whether it is worth it


alex-Arc (approved):
cool!
we could also add in the event change list to the re-fetch message
or do it as a separate PR

cpvalente:
> cool! we could also add in the event change list to the re-fetch message or do it as a separate PR

Ontime does not benefit from that change, so lets do it later when the companion module is ready

PR #880, opened 2024-04-11
Enforece type imports

alex-Arc:
adds `simple-import-sort/imports` and `@typescript-eslint/consistent-type-imports` to the type and utils packages
I think we will have to wait with adding it to server and client until after v3 or at least after beta

do we wan to add it for export as well?

https://typescript-eslint.io/blog/consistent-type-imports-and-exports-why-and-how/

### DIFF
--- a/packages/types/.eslintrc
+++ b/packages/types/.eslintrc
@@ -3,8 +3,14 @@
     "browser": true,
     "node": true
   },
+  "parser": "@typescript-eslint/parser",
   "parserOptions": {
     "ecmaVersion": 2020,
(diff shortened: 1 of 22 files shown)


cpvalente (approved):
Thank you for this 🚀 

Not sure about the multiple imports from the same package, but this is easy enough to change if it gets annoying

```ts
import type { NormalisedRundown, OntimeEvent, OntimeRundown, OntimeRundownEntry } from 'ontime-types';
import { isOntimeEvent } from 'ontime-types';
```

PR #903, opened 2024-04-17
Link start times from excel import

alex-Arc:
test file
[Ontime rundown demo.xlsx](https://github.com/cpvalente/ontime/files/15022882/Ontime.rundown.demo.xlsx)

## Summary by CodeRabbit

- **New Features**
	- Added a new key 'Link Start' across various components for enhanced link management.
	- Introduced a new style `.linkStartActive` for active link indications.
	- Added functionality to display a link icon next to start time in the `PreviewRundown` component if a link start is present.

- **Enhancements**
	- Improved validation and conversion processes for import maps in settings panel.
	- Enhanced boolean coercion logic to handle additional true option values.

- **Tests**
	- Implemented new test cases to verify the handling of the 'Link Start' field across components and utilities.

- **Style**
	- Reformatted existing CSS styles and added new styles for active link management.

- **Refactor**
	- Updated functions to better handle new 'Link Start' key and improved type assertions and validations.


apps/server/src/utils/parserFunctions.ts
```diff
@@ -48,11 +48,17 @@ export const parseRundown = (data: Partial<DatabaseModel>): OntimeRundown => {
...
+      const tryToLink = event.linkStart;
       newEvent = createEvent(event, eventIndex.toString());
       // skip if event is invalid
       if (newEvent == null) {
         continue;
       }
+      const prevEvent = rundown.at(-1);
+      if (tryToLink && isOntimeEvent(prevEvent)) {
+        newEvent.linkStart = prevEvent.id;
+        newEvent.timeStart = prevEvent.timeEnd;
```
cpvalente, line 60: We should rethink this code.

The previous might not be an event, in which case we don't have a value 

The values here will be calculated by `rundownCache`

I think we either 
a) always provide a value
b) never provide a value 
c) ignore the value

The point being that the output here has to be predicable

apps/server/src/utils/coerceType.ts
```diff
@@ -44,6 +44,16 @@ export function coerceBoolean(value: unknown): boolean {
   return Boolean(value);
 }
 
+export function coerceBooleanLoose(value: unknown, trueOption?: string): boolean {
+  if (value === undefined || value === null) {
```
cpvalente, line 48: We can just use the normal comparison to catch nullish like `value == null`

apps/server/src/utils/coerceType.ts
```diff
@@ -44,6 +44,16 @@ export function coerceBoolean(value: unknown): boolean {
...
+    return false;
+  }
+  if (value === trueOption) {
```
cpvalente, line 51: Not sure about this true option. 

What is the Boolean value that we get from excel and Google sheets??

I think we should check for that Boolean instead. And if the input is a string, any non empty string is a TRUE

alex-Arc: we might get 'TRUE' or 'FALSE' as text withc is handled by the normal `coerceBoolean`
and in the pareser code we had an extra allowens for just an 'x'
I have also noticed a `null` comming through sometimes

cpvalente: but then it sounds like coerceBoolean does everything we need?
How about something like 
```ts
function parseBooleanString(maybeText: string | null) {
  // falsy values would be nullish or empty string
  if (!maybeText) {
    return false;
  }
  return maybeText.toLowerCase() !== 'false';
}

```

apps/client/src/features/app-settings/panel/sources-panel/preview/PreviewRundown.tsx
```diff
@@ -83,7 +85,16 @@ export default function PreviewRundown(props: PreviewRundownProps) {
                 </td>
                 <td className={style.nowrap}>{event.cue}</td>
                 <td>{event.title}</td>
-                <td>{millisToString(event.timeStart)}</td>
+                <td>
+                  {event.linkStart ? (
+                    <HStack>
+                      <Box style={{ opacity: '50%' }}>{millisToString(event.timeStart)}</Box>
```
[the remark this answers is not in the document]
cpvalente: lets chat on discord

cpvalente:
@alex-Arc , did you need help with styling here?

alex-Arc:
[style: simplify layout](https://github.com/cpvalente/ontime/pull/903/commits/1e1127ca58a7a69f0d8d7dee7e3a11393abaf9a5) Nice
I'm happy here

cpvalente (approved):
Just need to solve the merge conflicts, looks good, thank you!

PR #904, opened 2024-04-18
Setup rewrite

arymoraes:
@cpvalente hey, you can look at what I rewrote on the setup file. I'm basically trying to break down most things into functions and then exporting the directories/paths and environment. It's still wip but you can take a look and see how you feel about the approach

## Summary by CodeRabbit

- **Refactor**
	- Streamlined configuration and directory path management across various components to enhance system reliability and maintainability.
	- Improved handling of environmental settings and directory paths, ensuring more robust application behavior across different operating environments.
	- Centralized directory references for file operations, improving code clarity and reducing potential errors in file handling.


cpvalente:
hey @arymoraes , i think this is the gist of it, although I think we can improve the shape of the object for readability.

In the app, there are two locations for things: with the project, or in the user directory
this changes depending on environment

What we are trying to do is to clarify this behaviour so that developers know where their files are and are less prone to make mistakes when writing features

It also seems like we have some race conditions, do you mind taking a look?

arymoraes:
> hey @arymoraes , good work! I think this is the gist of it, although I think we can improve the shape of the object for readability.
> 
> In the app, there are two locations for things: with the project, or in the user directory this changes depending on environment
> 
> What we are trying to do is to clarify this behaviour so that developers know where their files are and are less prone to make mistakes when writing features
> 
> It also seems like we have some race conditions, do you mind taking a look?

Sorry I didn't have as much capacity in the past few days to look at it.

Think it makes sense, I'll polish up the shape of the object so it makes more sense and feels cleaner, and will also fix the tests cause I still have to go and change every place we use the setup exports

alex-Arc:
@arymoraes one of the issues with the test might be that you are not targeting the master branch.
When we went into beta it was merged into master

PR #910, opened 2024-04-22
Event defaults

cpvalente:
Adding some defaults for new events
Also adds some small UI tweaks

@alex-Arc , havent tested around much, but what do you think?

## Summary by CodeRabbit

- **New Features**
    - Introduced new timer settings including default duration, warning time, and danger time.
    - Enhanced user interface for editing timer behaviors and visual settings in the app.

- **Bug Fixes**
    - Updated descriptions in settings to improve clarity on timer functionalities.

- **Style**
    - Updated UI colors and margins in various components for a more cohesive look.

### DIFF
--- a/apps/client/src/common/hooks/useEventAction.ts
+++ b/apps/client/src/common/hooks/useEventAction.ts
@@ -25,7 +25,7 @@ import { forgivingStringToMillis } from '../utils/dateConfig';
  */
 export const useEventAction = () => {
   const queryClient = useQueryClient();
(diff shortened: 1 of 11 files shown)


alex-Arc (approved):
Nice, seams to work good
and thees are disabled on purpose?
image

cpvalente:
> Nice, seams to work good and thees are disabled on purpose? 

Yes, I am just trying to keep the PR simple
Since v3 introduces warn and danger time per component it makes sense to be able to define this globally as before
Eventually users will have opinions on what properties should be part of event defaults and we can extend this

I felt it was counter productive to try to add too much as a first implementation

PR #913, opened 2024-04-25
style: use user given capitalisation

cpvalente:


cpvalente:
@alex-Arc are right, sorry about that. It should be fixed now

alex-Arc (approved):

PR #918, opened 2024-04-27
chore: version bump

cpvalente:
preparing a new beta, just pending the fixes on custom fields


alex-Arc:
Would like to get the shortcut in here as well

PR #925, opened 2024-04-29
Fix: custom fields with undefined colour is not accepted by `parseCustomFields`

alex-Arc:
solves #927

### DIFF
--- a/apps/server/src/utils/parserFunctions.ts
+++ b/apps/server/src/utils/parserFunctions.ts
@@ -248,17 +248,40 @@ export const parseCustomFields = (data: Partial<DatabaseModel>): CustomFields =>
   if (typeof data.customFields !== 'object') {
     return { ...dbModel.customFields };
   }
-
   console.log('Found Custom Fields, importing...');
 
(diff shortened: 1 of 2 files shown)


apps/server/src/utils/parserFunctions.ts
```diff
@@ -255,7 +255,7 @@ export const parseCustomFields = (data: Partial<DatabaseModel>): CustomFields =>
 
   for (const fieldLabel in data.customFields) {
     const field = data.customFields[fieldLabel];
-    if (!field.label || !field.type || !field.colour) {
+    if (!('label' in field) || !('type' in field) || !('colour' in field)) {
```
cpvalente, line 258: I wonder if we should be a bit more cautious in how we merge the fields after. is it worth checking the types?
type should always be "string"
colour should be of type string
label should be of type string and cannot be empty

like

```ts
    newCustomFields[key] = {
      type: field.type, 
      colour: field.colour,
      label: field.label,
    };
```

cpvalente (approved):
good stuff

PR #781, opened 2024-02-18
Client remote control

kellhogs:
This aim's to add remote redirects for connected websocket clients from within ontime

PR #930, opened 2024-04-30
Client remote control 

alex-Arc:
#781 continued here

# TODO
- [x] Make nice overlay
- [x] styling the client list
- [x] Handle multiple tabs


apps/client/src/common/stores/clientStore.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const clientNameKey = 'ontime-client-name';
+
+function persistNameInStorage(newValue: string) {
+  localStorage.setItem(clientNameKey, newValue);
+}
+
+export const useClientStore = create<ClientStore>((set) => ({
+  self: localStorage.getItem(clientNameKey) ?? undefined,
```
cpvalente, line 18: why casting to undefined? could we not deal with the null?

apps/client/src/common/stores/clientStore.ts
```diff
@@ -0,0 +1,49 @@
+import { create } from 'zustand';
+
+interface ClientStore {
+  self?: string;
```
cpvalente, line 4: i think self is a loaded word in software and we should not use it here. name, setName seem to me like reasonable choices for the feature

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
```diff
@@ -0,0 +1,120 @@
...
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>Redirect Client</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <div>Redirect {clientName} to a new URL</div>
+          <InputGroup variant='ontime-filled' size='md'>
+            {/* TODO: better description */}
+            <InputLeftAddon>ontime:port/</InputLeftAddon>
+            <Input placeholder='newpath?and=params' value={path} onChange={(event) => setPath(event.target.value)} />
```
cpvalente, line 104: question: i wonder if we should use a full path like `http://localhost:4001/timer?hide=true`
The `pathandparams` was an attempt at simplifying the feature which i feel never really helped

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -76,8 +76,7 @@ export class SocketServer implements IAdapter {
 
       ws.on('message', (data) => {
         try {
-          // @ts-expect-error -- ??
-          const message = JSON.parse(data);
+          const message = JSON.parse(data.toString());
```
cpvalente, line 79: hmmm... I am unsure i follow the double casting here?

apps/client/src/AppRouter.tsx
```diff
@@ -36,6 +37,8 @@ const MessageControl = lazy(() => import('./features/control/message/MessageCont
 export default function AppRouter() {
   return (
     <Suspense fallback={null}>
+      <IdentToast />
```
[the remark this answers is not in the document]
alex-Arc: Yes I was being lazy
I'm also unsure if this is the right location for that module

apps/client/src/common/components/ident-toast/IdentToast.tsx
```diff
@@ -0,0 +1,34 @@
...
+        });
+      }
+      // you are asked to stop identifying
+      if (!identify[self] && toast.isActive(id)) {
+        toast.closeAll();
+      }
+    }
+  }, [identify, self, toast]);
+
+  return null;
```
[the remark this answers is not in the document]
alex-Arc: Agreed

apps/client/src/common/stores/clientStore.ts
alex-Arc: don't think I wrote this part, but yeah null wold be OK I think

apps/client/src/common/api/clientRemote.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const clientPath = `${serverURL}/api/wsclient`;
+
+/**
+ * HTTP GET ws client list
+ */
+export async function getClients(): Promise<string[]> {
+  const res = await axios.get(`${clientPath}/list`);
+  return res.data.payload;
+}
```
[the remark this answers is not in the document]
alex-Arc: moved it to http to easier manage the replies

apps/client/src/common/stores/clientStore.ts
cpvalente: hahaha, i guess it was likely me then :D 

self usually means the instance. in javascript, it was common to assign this to self to keep the value of this across scopes

in python self is also the reference to the instance

what I meant is that self tend to mean things in programming, none of which are the string reference of a connection

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
alex-Arc: hmm maybe
I would prevent some nice predefined API calls
where `/clock` would work on any ontime instance

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -49,8 +49,22 @@ export class SocketServer implements IAdapter {
...
+          ws.send(
+            JSON.stringify({
+              type: 'client-name',
+              payload: clientId,
+            }),
+          ),
+        100,
+      );
+      //then send the client list
+      setTimeout(() => this.sendClientList(), 120);
```
cpvalente, line 67: I think adding side effects this way could give us a chance for hard to follow bugs.
I wonder if it is best for us to live with the renaming happening straight after connection

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -49,8 +49,22 @@ export class SocketServer implements IAdapter {
 
     this.wss.on('connection', (ws) => {
       let clientId = getRandomName();
-      this.clientIds.add(clientId);
-      logger.info(LogOrigin.Client, `${this.clientIds.size} Connections with new: ${clientId}`);
+      this.clients.set(clientId, { type: ClientTypes.Unknown, identify: false, redirect: '' });
```
cpvalente, line 52: nice!

We could keep the key as a unique ID we generate, and add the preferred name as a property of the object.
I wonder if that would make it easier for us to go around the issue of multiple instances sharing a name and allow us to improve this in future by adding more properties

apps/client/src/common/utils/socket.ts
```diff
@@ -33,7 +36,8 @@ export const connectSocket = (preferredClientName?: string) => {
         console.warn('WebSocket: attempting reconnect');
         if (websocket && websocket.readyState === WebSocket.CLOSED) {
           reconnectAttempts += 1;
-          connectSocket();
+          //TODO: add some jitter and a bit of backoff
+          setTimeout(() => connectSocket());
```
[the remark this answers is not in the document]
alex-Arc: was trying to solve the multiple connections that was a result of vite HRM and forgot to remove it

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -40,70 +41,116 @@ export class SocketServer implements IAdapter {
...
-        logger.info(LogOrigin.Client, `${this.clientIds.size} Connections with disconnected: ${clientId}`);
+        this.clients.delete(clientId);
+        logger.info(LogOrigin.Client, `${this.clients.size} Connections with disconnected: ${clientId}`);
+        this.sendClientList();
       });
 
       ws.on('message', (data) => {
         try {
-          // @ts-expect-error -- ??
+          //@ts-expect-error ???
```
cpvalente, line 97: I am a bit confused about some of the changes here.
I think it is important to keep the changes meaningful

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -40,70 +41,116 @@ export class SocketServer implements IAdapter {
...
+
+          if (type === 'set-client-rename') {
+            if (payload) {
+              const targetClient = this.clients.get(payload.target);
+              const rename = payload.name;
+              this.clients.set(payload.target, { ...targetClient, rename });
+            }
+            this.sendClientList();
             return;
           }
```
cpvalente, line 155: suggestions: 
- Would it be easier to maintain this feature if we extracted the functions here?
- Can we have a timeout for the client identify to stop someone forgetting this on and a screen staying blinking? ie: after 10 seconds an identify command resets itself.

apps/client/src/common/utils/socket.ts
```diff
@@ -33,6 +36,7 @@ export const connectSocket = (preferredClientName?: string) => {
         console.warn('WebSocket: attempting reconnect');
         if (websocket && websocket.readyState === WebSocket.CLOSED) {
           reconnectAttempts += 1;
+          //TODO: add some jitter and a bit of backoff
```
cpvalente, line 39: question: can we clarify or remove the comment?

apps/client/src/common/stores/clientStore.ts
```diff
@@ -0,0 +1,62 @@
+import { Clients } from 'ontime-types';
+import { create } from 'zustand';
+
+interface ClientStore {
+  name?: string;
+  setMyName: (newValue: string) => void;
```
cpvalente, line 6: suggestion: the function name should be connected to the property it affects on. So for a property `name` we would prefer having a `setName` function

apps/client/src/common/hooks/useSocket.ts
```diff
@@ -3,6 +3,12 @@ import { RuntimeStore, SimpleDirection, SimplePlayback } from 'ontime-types';
 import { useRuntimeStore } from '../stores/runtime';
 import { socketSendJson } from '../utils/socket';
 
+export const setClientRemote = {
+  setIdentify: (payload: { target: string; state: boolean }) => socketSendJson('set-client-identify', payload),
+  setRedirect: (payload: { target: string; path: string }) => socketSendJson('set-client-redirect', payload),
+  setRename: (payload: { target: string; name: string }) => socketSendJson('set-client-rename', payload),
+};
+
```
cpvalente, line 11: suggestion: would it be better if this was next to the `setClientName` function?

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+export default function ClientList() {
+  const { id: myId, clients } = useClientStore();
+  const { isOpen: isOpenRedirect, onOpen: onOpenRedirect, onClose: onCloseRedirect } = useDisclosure();
+  const { isOpen: isOpenRename, onOpen: onOpenRename, onClose: onCloseRename } = useDisclosure();
+  const { setIdentify, setRedirect, setRename } = setClientRemote;
+
+  const [targetId, setTargetId] = useState('');
+
+  const openRename = (id: string) => {
```
cpvalente, line 36: would it be best to call this something else to avoid shadowing the `id` from `useClientStore`? in that case we wouldnt need to rename the `id` to `myId`

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
```diff
@@ -0,0 +1,205 @@
...
+          <tr>
+            <td className={style.fullWidth}>Client Name (Connection ID)</td>
+            <td />
+          </tr>
+        </thead>
+        <tbody>
+          {Object.entries(clients).map(([id, client]) => {
+            const { identify, redirect, name } = client;
+            const isCurrent = id === myId;
+            const isRedirecting = redirect != '';
```
cpvalente, line 77:
```suggestion
            const isRedirecting = Boolean(redirect);
```

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
```diff
@@ -0,0 +1,205 @@
...
+                    <Button isLoading={isRedirecting} onClick={() => openRedirect(id)}>
+                      Redirect
+                    </Button>
+                    {isRedirecting && (
+                      <IconButton
+                        aria-label='Cancel the redirect'
+                        icon={<IoClose />}
+                        onClick={() => setRedirect({ target: id, path: '' })}
+                      />
+                    )}
```
cpvalente, line 115: question: do you think the user would see this? I wonder if the action would be too quick (a few milliseconds)

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
```diff
@@ -0,0 +1,205 @@
...
+              Submit
+            </Button>
+          </div>
+        </ModalFooter>
+      </ModalContent>
+    </Modal>
+  );
+}
+
+function RenameModal(props: {
```
cpvalente, line 166: question: why dont we use the existing `RenameClientModal`?

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
```diff
@@ -0,0 +1,205 @@
...
+              </tr>
+            );
+          })}
+        </tbody>
+      </Panel.Table>
+    </>
+  );
+}
+
+function RedirectModal(props: {
```
cpvalente, line 127: suggestion: lets extract this into another file. I think it may be easier for us

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
cpvalente: I am not sure I understand. 
I wonder if the user should be able to input anything here, we leave it up to them to get the url right

apps/client/src/common/components/identify-overlay/IdentifyOverlay.tsx
```diff
@@ -0,0 +1,57 @@
...
+        socketSendJson('set-client-name', rename);
+      } else {
+        setRename({ target: id, name: '' });
+      }
+    }
+  }, [clients, id, name, setMyName, setRename]);
+
+  const showOverlay = useMemo(() => {
+    return id && clients[id] && clients[id].identify;
+  }, [clients, id]);
```
cpvalente, line 42: I think the `memo` is a bit overkill. is there an issue I am not aware of?

apps/client/src/common/components/identify-overlay/IdentifyOverlay.tsx
```diff
@@ -0,0 +1,57 @@
...
+    if (id && clients[id] && clients[id].redirect !== '') {
+      const { redirect } = clients[id];
+      if (redirect !== window.location.pathname) {
+        setRedirect({ target: id, path: '' });
+        navigate(redirect);
+      } else {
+        setRedirect({ target: id, path: '' });
+      }
+    }
+  }, [clients, id, navigate, setRedirect]);
```
cpvalente, line 27: I think we can simplify a bit
- avoid redundant calls
- exit early
It is also nice in `useEffects` to describe what they do

Something like (untested):
```suggestion
  // redirect if new target is set
  useEffect(() => {
    if (!id || !clients[id] || clients[id].redirect === '') {
      return;
    }

    const { redirect } = clients[id];
    if (redirect !== window.location.pathname) {
      setRedirect({ target: id, path: '' });
    }
  }, [clients, id, navigate, setRedirect]);
```

apps/client/src/common/components/identify-overlay/IdentifyOverlay.tsx
```diff
@@ -0,0 +1,57 @@
...
+          {name}
+          <Badge>{id}</Badge>
+        </h1>
+        <h3>Click to close</h3>
+      </div>
+    );
+  }
+
+  return null;
+}
```
cpvalente, line 57: how about inverting this so we have less stuff behind logic branches.

something like 
```ts
  if (!showOverlay) {
    return null;
  }

  return (
    <div className='overlay' data-testid='identify-overlay' onClick={() => setIdentify({ target: id, state: false })}>
      <h1>
        {name}
        <Badge>{id}</Badge>
      </h1>
      <h3>Click to close</h3>
    </div>
  );
```

apps/client/src/common/components/identify-overlay/IdentifyOverlay.tsx
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    if (id && clients[id] && clients[id].rename !== '') {
+      const { rename } = clients[id];
+      if (rename !== name) {
+        socketSendJson('set-client-name', rename);
+      } else {
+        setRename({ target: id, name: '' });
+      }
+    }
+  }, [clients, id, name, setMyName, setRename]);
```
cpvalente, line 38: question: I am not sure I follow the logic here.
Why would renaming so many steps? I think the request for renaming should rename the client and that is it.
No need for client involvement

apps/server/src/adapters/WebsocketAdapter.ts
alex-Arc: i'm thinking about how long to set the timeout for...
it would be anoying to set the idetify walk over to the screens and the it has reset befor you get there
maybe atleast 30sec or a whole minute

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
alex-Arc: the existig modal sets the local cliet name
should I rewirte that?

apps/server/src/adapters/WebsocketAdapter.ts
cpvalente: I am not sure, I think 1 minute sounds reasonable to me.

I think if the venue is so large that you need a long time to walk around, we likely want either another solution or some help

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -135,12 +180,26 @@ export class SocketServer implements IAdapter {
...
   }
 
+  sendClientList(): void {
+    const payload = Array.from(this.clients.entries()).reduce((acc, [key, data]) => {
+      //We have to remove the WS part from the list before sending it to the clients
+      const { ws: _, ...dataExcludingWs } = data;
+      return Object.assign(acc, { [key]: dataExcludingWs });
+    }, {});
+    this.sendAsJson({ type: 'client-list', payload });
+  }
```
cpvalente, line 190: I think this logic could be hard to maintain in the long run. Few suggestions:
- since the size of the thing coming in is the same as the ones going out, a map may be easier to leverage
- we dont need to cast the `ws` to `_` just to silence eslint
- could we add some comments about what does this do and why do we need to exclude ws?

Something like (untested)

```ts
    const payload = Object.fromEntries(
        Array.from(this.clients.entries()).map(([key, data]) => {
            // eslint-disable-next-line @typescript-eslint/no-unused-vars -- using destructuring to remove property
            const { ws, ...dataExcludingWs } = data;
            return [key, dataExcludingWs];
        })
    );
    this.sendAsJson({ type: 'client-list', payload });
```

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -24,14 +24,17 @@ import { IAdapter } from './IAdapter.js';
 import { eventStore } from '../stores/EventStore.js';
 import { logger } from '../classes/Logger.js';
 import { dispatchFromAdapter } from '../api-integration/integration.controller.js';
+import { generateId } from 'ontime-utils';
 
 let instance: SocketServer | null = null;
 
+type WsClient = Client & { ws: WebSocket };
```
cpvalente, line 31: question: I am not sure why do we keep the websocket object here? is it not easier to keep just the id of the client. the websocket connection has its own list of clients and I dont think we want the two to become coupled

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -81,29 +95,60 @@ export class SocketServer implements IAdapter {
...
-                payload: clientId,
-              }),
-            );
+            return;
+          }
+
+          if (type === 'set-client-rename') {
+            if (payload && this.clients.has(payload.target)) {
+              const targetID = payload.target;
+              const newName = payload.name;
```
cpvalente, line 143: question: can we assume that payload being truthy means it carries the correct values?
What do we do in the rest of the app? would it be worth wrapping this in a try-catch?

cpvalente (commented):
Good work! I made a few suggestions for logic cleanup but things look good. I will try to do some testing and cleanup the styles after

apps/server/src/adapters/WebsocketAdapter.ts
alex-Arc: to be able to send the rename from one ws client to another one if it should not be in the list object?

apps/client/src/features/app-settings/panel/client-control-panel/Modal.tsx
```diff
@@ -0,0 +1,97 @@
...
+              Submit
+            </Button>
+          </div>
+        </ModalFooter>
+      </ModalContent>
+    </Modal>
+  );
+}
+
+export function RenameModal(props: {
```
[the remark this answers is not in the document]
alex-Arc: the control methods a different, the old rename modal would have to use this new method as well, maybe that's fine?

apps/server/src/adapters/WebsocketAdapter.ts
cpvalente: I am not sure I understand, we can chat about it.
But we have a client ID, which is unique and doesnt change. As far as I can see, this ID is the only thing we need to share with the websocket object

apps/client/src/features/app-settings/panel/client-control-panel/Modal.tsx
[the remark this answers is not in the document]
cpvalente: yes, that is fine. the gist of it is that we should maybe end up with a single `RenameModal`.

apps/client/src/common/components/identify-overlay/IdentifyOverlay.tsx
```diff
@@ -0,0 +1,47 @@
...
+    console.log('clear');
+
+    clearTimeout(identifyTimeout);
+    return null;
+  }
+
+  identifyTimeout = setTimeout(() => {
+    console.log('timeout');
+    setIdentify({ target: id, state: false });
+  }, MILLIS_PER_MINUTE);
```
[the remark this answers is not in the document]
alex-Arc: i've tried doing this and it looks to work but usure if it is correct

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
```diff
@@ -0,0 +1,126 @@
...
+      </Panel.Table>
+    </>
+  );
+}
+
+const selectColorScheme: Record<ClientTypes, string> = {
+  [ClientTypes.Ontime]: 'blue',
+  [ClientTypes.Unknown]: 'gray',
+  [ClientTypes.Companion]: 'red',
+  [ClientTypes.Chataigne]: 'orange',
```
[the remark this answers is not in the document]
alex-Arc: hmm yeah maybe we should have them in a separate list

apps/client/src/common/components/client-modal/RedirectClientModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+import style from './ClientModal.module.scss';
+
+export function RedirectClientModal(props: {
+  onClose: () => void;
+  isOpen: boolean;
+  id: string;
+  clients: Clients;
+  onSubmit: (path: string) => void;
+}) {
+  const { onClose, isOpen, id, clients, onSubmit } = props;
```
cpvalente, line 26:
```suggestion
interface RedirectClientModalProps {
  onClose: () => void;
  isOpen: boolean;
  id: string;
  clients: Clients;
  onSubmit: (path: string) => void;
}

export function RedirectClientModal(props: RedirectClientModalProps) {
  const { onClose, isOpen, id, clients, onSubmit } = props;
```

packages/types/src/definitions/Clients.type.ts
```diff
@@ -0,0 +1,15 @@
+export type CustomFieldLabel = string;
+
+export enum ClientTypes {
+  Unknown = 'unknown',
+  Ontime = 'ontime',
+}
+
+export type Client = {
+  name: string;
+  type: ClientTypes | string;
```
cpvalente, line 10: question? why the string? dont we have control over this?

apps/client/src/common/components/client-modal/RenameClientModal.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export function RenameClientModal(props: {
+  onClose: () => void;
+  isOpen: boolean;
+  id: string;
+  clients: Clients;
+  onSubmit: (path: string) => void;
+}) {
+  const { onClose, isOpen, id, clients, onSubmit } = props;
+  const [name, setName] = useState('');
```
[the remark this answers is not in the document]
alex-Arc: we get it from the useEffect right below, it dose not exist at this point yet
or maybe there is abetter way?

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -135,12 +173,55 @@ export class SocketServer implements IAdapter {
...
+    const payload = Object.fromEntries(this.clients.entries());
+    this.sendAsJson({ type: 'client-list', payload });
+  }
+
+  public getClientList(): string[] {
+    return Array.from(this.clients.keys());
+  }
+
+  public renameClient(target: string, name: string) {
+    const previousData = this.clients.get(target);
```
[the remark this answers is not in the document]
alex-Arc: ahh yeah it dose mutate it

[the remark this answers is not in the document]
alex-Arc: should we then just not do the `this.clients.set(target, previousData);` or finde a different way?

packages/types/src/definitions/Clients.type.ts
alex-Arc: yes that was my thought
and I have the enum to make it easier to set in the server, but maybe that is bad?

apps/server/src/adapters/WebsocketAdapter.ts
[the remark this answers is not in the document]
cpvalente: I think the set call makes it explicit, without it wouldnt be immediately obvious that we are purposely changing the data in the origin Set.
we could just make a spread with the new property like

```ts
    this.clients.set(target, { ...previousData, name });
```

packages/types/src/definitions/Clients.type.ts
cpvalente: It is unclear what a best way would be, but I suggest keeping this simple for now.

how about a union like
```ts
export type ClientType = 'ontime' | 'unknown' | string;
```

I believe we dont get benefits from the enum since this is dynamic


Also, mega nitpick: the ClientType should not be plural
https://stackoverflow.com/questions/15755955/naming-of-enums-in-java-singular-or-plural

alex-Arc:
@jwetzell if you could give this feature a try that would be very welcome

i find it easiest to test by open one window to `http://localhost:3000/editor?settings=network__clients` and the another one in incognito mode

jwetzell:
I don't really know what is being added. I did pull this branch and run it. If I open 2 tabs (one in incognito) I see 2 entries in `Ontime client` and can interact with them and redirect and identify work.

jwetzell:
Seems to be, I didn't immediately understand what redirect would do but simple enough to just try it out

apps/client/src/common/components/client-modal/RedirectClientModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export function RedirectClientModal(props: RedirectClientModalProps) {
+  const { onClose, isOpen, id, clients, onSubmit } = props;
+  const [path, setPath] = useState('');
+
+  // when clientlist and id is populated get the relevant path to use a default value
+  useEffect(() => {
+    if (clients[id]?.path) setPath(clients[id].path);
+  }, [clients, id]);
```
cpvalente, line 34: question: usually combinations of useState and useEffect are incorrect.

can we combine this by initialising the state?

```ts
const [path, setPath] = useState(clients[id]?.path || '');
```

apps/client/src/common/components/client-modal/RenameClientModal.tsx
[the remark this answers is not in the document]
cpvalente: same as above, combining useState and useEffect is generally not what you want, you likely can combine this by setting the state from the prop, like

```ts
  const [name, setName] = useState(clients[id]?.name || '');
```

apps/client/src/common/components/client-modal/RenameClientModal.tsx
```diff
@@ -0,0 +1,63 @@
...
+  }, [clients, id]);
+
+  return (
+    <Modal isOpen={isOpen} onClose={onClose} variant='ontime'>
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>Rename {clients[id]?.name}</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <div>All connections in this client will have the same name after a reload</div>
```
cpvalente, line 41: question: I am not sure this feedback is useful. could we rephrase or remove?

apps/client/src/features/app-settings/panel/client-control-panel/ClientList.tsx
```diff
@@ -0,0 +1,150 @@
...
+                  >
+                    Redirect
+                  </Button>
+                </td>
+              </tr>
+            );
+          })}
+        </tbody>
+      </Panel.Table>
+      <Panel.Divider></Panel.Divider>
```
cpvalente, line 121:
```suggestion
      <Panel.Divider />
```

apps/client/src/common/components/navigation-menu/NavigationMenu.tsx
```diff
@@ -18,12 +18,13 @@ import { IoLockClosedOutline } from '@react-icons/all-files/io5/IoLockClosedOutl
 import { IoSwapVertical } from '@react-icons/all-files/io5/IoSwapVertical';
 
 import { navigatorConstants } from '../../../viewerConfig';
+import { RenameClientModal } from '../../components/client-modal/RenameClientModal';
```
cpvalente, line 21:
```suggestion
import { RenameClientModal } from '../client-modal/RenameClientModal';
```

apps/client/src/common/components/client-modal/RenameClientModal.tsx
alex-Arc: Yeah maybe just note it some where in the docs

cpvalente:
Seems like there is something broken ?

Screenshot 2024-06-12 at 21 34 57

alex-Arc:
yeah name and path are unknown at start up

PR #936, opened 2024-05-07
fix: await errors in sheet client

alex-Arc:


cpvalente (approved):
I havent been able to test this, but it looks correct

PR #943, opened 2024-05-09
fix: prevent mutation of initial object

### LINKED ISSUE #942: Planned start and end incorrect
### Discussed in https://github.com/cpvalente/ontime/discussions/941

There seems to be buggy behaviour around the calculation of planned start and end.
This seems to be related to the loading of an event

See the user report

cpvalente:
The issue is described in #941 and #942 
Fixes #942 

It seems that we were accidentally leaving the runtime object connected to its initial reference.
I believe the fix did not introduce unintended side effects, but it would be good to have some help with sanity check

### DIFF
--- a/apps/server/src/services/rundown-service/rundownCache.ts
+++ b/apps/server/src/services/rundown-service/rundownCache.ts
@@ -80,6 +80,8 @@ export function generate(
   links = {};
   firstStart = null;
   lastEnd = null;
+  totalDuration = 0;
+  totalDelay = 0;
 
   let accumulatedDelay = 0;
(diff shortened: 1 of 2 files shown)


alex-Arc (approved):
👍🏽

alex-Arc:
Planned end retains some value after clearing the rundown 
image

i corrects it self when a new event is added but it prober should show --:--:--

cpvalente:
Good catch @alex-Arc , appreciate your feedback

alex-Arc (approved):

PR #946, opened 2024-05-10
better logging from http integration

alex-Arc:


apps/server/src/services/integration-service/HttpIntegration.ts
```diff
@@ -51,7 +51,22 @@ export class HttpIntegration implements IIntegration<HttpSubscription, HttpSetti
...
+        logger.warning(LogOrigin.Tx, `HTTP Integration: DNS lookup failed`);
+        return;
+      }
+
+      if (err.code === 'ETIMEDOUT') {
+        logger.warning(LogOrigin.Tx, `HTTP Integration: The connection timed out`);
+        return;
+      }
+
+      logger.warning(LogOrigin.Tx, `HTTP Integration: ${err.code}`);
```
cpvalente, line 69: This looks good to me, but quick question: what would we get if we used `err.message` instead of `err.code`? Maybe we can let the library do the errors for us? it would be easier for the user to google

cpvalente (approved):

PR #954, opened 2024-05-11
refactor: improve payload type

### LINKED ISSUE #953: OSC Output "base 10:" Error
Hi,

first i have to say i really like ontime. Especially the new v.3 version.
But i have a problem since i have started using it (v.2)

When i use the OSC Output to trigger Ableton Live v.11 (via the plugin AbletonOSC) i get a base 10 error.

to trigger a cue in ableton i have to use the command: `/live/clip/fire track_id clip_id`

in the case of triggering the 1 column in the first 1 row i have to use: `/live/clip/fire 0 0` but it looks like if ontime is converting these integers to a string because of the space inbetween. 

When i just use `/live/clip/fire 0` i am only getting an out of range error but it leaves the command as an integer.

This occours for all inputs with more than one variable.
This is the exact error message: `/live/error,Error handling OSC message: invalid literal for int() with base 10: '0 0'`

When i use the console from the AbletonOSC plugin to send an OSC message to Ableton averything works as it should with exactly the same command.

Thanks for your help.

cpvalente:
This is a quick test to solve the issue described by the user

There are two points of improvement here
- we should attempt keeping the types defined by the user
- we should allow sending multiple commands

For the user defined case, where we want to send two integers `1` `1`, I believe the easiest case is to wrap this in an array  `[1,1]`, so that we can treat the arguments individually

We should take some time investigating to make sure there are no unintended side effects

Closes #953

### DIFF
--- a/apps/server/src/services/integration-service/OscIntegration.ts
+++ b/apps/server/src/services/integration-service/OscIntegration.ts
@@ -3,7 +3,6 @@ import { LogOrigin, MaybeNumber, MaybeString, OSCSettings, OscSubscription } fro
 
 import IIntegration, { TimerLifeCycleKey } from './IIntegration.js';
 import { parseTemplateNested } from './integrationUtils.js';
-import { isObject } from '../../utils/varUtils.js';
 import { logger } from '../../classes/Logger.js';
 import { OscServer } from '../../adapters/OscAdapter.js';
 
@@ -59,34 +58,26 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
         continue;
       }
       const parsedAddress = parseTemplateNested(address, state || {});
-      let parsedPayload = payload ? parseTemplateNested(payload, state || {}) : undefined;
+      const parsedPayload = payload ? parseTemplateNested(payload, state || {}) : undefined;
+
+      const parsedArguments = this.stringToOSCArgs(parsedPayload);
 
       try {
-        const maybePayload = JSON.parse(parsedPayload);
-        parsedPayload = maybePayload;
-      } catch (_) {
-        /* we dont handle errors */
-      }
-      try {
-        this.emit(parsedAddress, parsedPayload);
+        this.emit(parsedAddress, parsedArguments);
       } catch (error) {
         logger.error(LogOrigin.Tx, `OSC Integration: ${error}`);
       }
     }
   }
 
-  emit(address: string, payload?: ArgumentType) {
+  emit(address: string, args?: ArgumentType[]) {
     if (!this.oscClient) {
       return;
     }
 
     const message = new Message(address);
-    if (payload) {
-      if (isObject(payload)) {
-        message.append(JSON.stringify(payload));
-      } else {
-        message.append(payload);
-      }
+    if (args) {
+      message.append(args);
     }
 
     this.oscClient.send(message);
@@ -118,6 +109,44 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
       this.oscClient = null;
       throw new Error(`Failed initialising OSC client: ${error}`);
     }
+  }
+
+  private stringToOSCArgs(argsString): ArgumentType[] {
+
+    // NOTE: regex taken from https://stackoverflow.com/questions/4031900/split-a-string-by-whitespace-keeping-quoted-segments-allowing-escaped-quotes
+    const parseRegex = /\w+|"(?:\\"|[^"])+"/g;
+
+    const parsedArguments = argsString.match(parseRegex).map((argString: string) => {
+      const argAsNum = Number(argString);
+
+      // NOTE: number like: 1 2.0 33333
+      if (!Number.isNaN(argAsNum)) {
+        if (argString.includes('.')) {
+          return {
+            type: 'f',
+            value: argAsNum,
+          };
+        } else {
+          return {
+            type: 'i',
+            value: argAsNum,
+          };
+        }
+      }
+
+      if (argString.startsWith('"') && argString.endsWith('"')) {
+        // NOTE: "quoted string"
+        return {
+          type: 's',
+          value: argString.substring(1, argString.length - 1).replaceAll('\\"', '"'),
+        };
+      }
+      return {
+        type: 's',
+        value: argString,
+      };
+    });
+    return parsedArguments;
   }
 
   private initRX(enabledIn: boolean, portIn: number) {
--- a/apps/server/src/services/integration-service/OscIntegration.ts
+++ b/apps/server/src/services/integration-service/OscIntegration.ts
@@ -110,7 +110,7 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
     }
   }
 
-  private stringToOSCArgs(argsString): ArgumentType[] {
+  private stringToOSCArgs(argsString: string): ArgumentType[] {
 
     // NOTE: regex taken from https://stackoverflow.com/questions/4031900/split-a-string-by-whitespace-keeping-quoted-segments-allowing-escaped-quotes
     const parseRegex = /[\w.]+|"(?:\\"|[^"])+"/g;
--- a/apps/server/src/services/integration-service/OscIntegration.ts
+++ b/apps/server/src/services/integration-service/OscIntegration.ts
@@ -111,17 +111,16 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
   }
 
   private stringToOSCArgs(argsString: string): ArgumentType[] {
-
     // NOTE: regex taken from https://stackoverflow.com/questions/4031900/split-a-string-by-whitespace-keeping-quoted-segments-allowing-escaped-quotes
     const parseRegex = /[\w.]+|"(?:\\"|[^"])+"/g;
 
     const matches = argsString.match(parseRegex);
 
-    const parsedArguments = []
+    const parsedArguments = [];
 
-    if(!matches){
-      parsedArguments.push({type: 's', value: argsString})
-      return parsedArguments 
+    if (!matches) {
+      parsedArguments.push({ type: 's', value: argsString });
+      return parsedArguments;
     }
 
     matches.forEach((argString: string) => {
--- a/apps/server/src/services/integration-service/OscIntegration.ts
+++ b/apps/server/src/services/integration-service/OscIntegration.ts
@@ -111,10 +111,12 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
     }
   }
 
-  private stringToOSCArgs(argsString: string): ArgumentType[] {
-    const matches = splitWhitespace(argsString);
-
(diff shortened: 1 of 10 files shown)


alex-Arc:
I think it might be a bit to broad, it results i cue numbers becoming floats and so on

cpvalente:
Agreed, I dont suggest we merge this.
The PR is just to point to the place of improvement and the desired shape

I think we should try and find a way to keep the user desired format.
Perhaps just numbers and arrays of values?

apps/server/src/services/integration-service/OscIntegration.ts
```diff
@@ -113,6 +110,44 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
     }
   }
 
+  private stringToOSCArgs(argsString): ArgumentType[] {
```
cpvalente, line 113: could we type the argument here?

apps/server/src/services/integration-service/OscIntegration.ts
```diff
@@ -113,6 +110,44 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
...
+
+    // NOTE: regex taken from https://stackoverflow.com/questions/4031900/split-a-string-by-whitespace-keeping-quoted-segments-allowing-escaped-quotes
+    const parseRegex = /[\w.]+|"(?:\\"|[^"])+"/g;
```
cpvalente, line 116: Nice, could we extract this regex and add some simple tests to it?

apps/server/src/services/integration-service/OscIntegration.ts
```diff
@@ -113,6 +110,44 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
...
+
+    const parsedArguments = argsString.match(parseRegex).map((argString: string) => {
```
cpvalente, line 118: suggestion: would it be easier to extract the matching? 
I wonder if that would make the logic easier to leverage, we could also bail out early if there are no matches (null)

something like (untested):

```ts
  const matches = argsString.match(parseRegex);

  if (!matches) {
    return argsString 
  }
```

apps/server/src/services/integration-service/OscIntegration.ts
```diff
@@ -113,6 +110,44 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
...
+            type: 'f',
+            value: argAsNum,
+          };
+        } else {
+          return {
+            type: 'i',
+            value: argAsNum,
+          };
+        }
+      }
```
cpvalente, line 134: suggestion: I see that the only real difference is the type, it could make it easier to read if that was made explicit- Something like:

```ts
   const isValidNumber = !Number.isNaN(argAsNum)
    if (isValidNumber) {
      const argAsNum = Number(argString);
      const type = argString.includes('.') ? 'f' : 'i';
      return { type, value: argAsNum };
    }
```

apps/server/src/services/integration-service/OscIntegration.ts
```diff
@@ -113,6 +110,44 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
...
+        return {
+          type: 's',
+          value: argString.substring(1, argString.length - 1).replaceAll('\\"', '"'),
+        };
+      }
+      return {
+        type: 's',
+        value: argString,
+      };
+    });
```
cpvalente, line 147: suggestion: similar possible improvement here, since we are only changing the value we could make that change explicit.
Something like (untested):

```ts
  const isQuotedString = argString.startsWith('"') && argString.endsWith('"');
  const value = isQuotedString ? argString.substring(1, argString.length - 1).replaceAll('\\"', '"') : argString;
  return { type: 's', value };
```

cpvalente (commented):
Great stuff, just left some small suggestions for the code style. Behaviour seems good to me, lets see with @alex-Arc that is more in touch with this part of the application

apps/server/src/services/integration-service/OscIntegration.ts
cpvalente: we have a regex directory in the client and utils.
We just make simple test for the happy paths to provide documentation

See as example
https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/utils/regex.ts

And the tests
https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/utils/__tests__/regex.test.ts

alex-Arc:
looks to work good
I can setup the tests if you need help?

apps/server/src/services/integration-service/OscIntegration.ts
cpvalente: Nice. Let me know if you need help

Right - wrong might be all we need for now. 
The tests are a good place to show a few examples of expected payloads and something we don't handle at all.

If there are bug reports in future that we track down to this regex we can leverage the existing test to expose the bug

jwetzell: For this regex the application isn't so much interested in the regex pass/failing as it is the result of the regex groups. This regex will turn a string like `'1.1 "string with spaces" 3'` into an array `[1.1, "string with spaces", 3]` so the tests to add would more make sure it does that properly. I'll take a crack at the tests if alex doesn't get around to it by the time I am free this evening.

jwetzell: I'll leave this to @alex-Arc as the string parsing logic is now completely different then what I had originally committed so they will be able to write tests for what they are intending to parse.

apps/server/src/services/integration-service/OscIntegration.ts
```diff
@@ -111,12 +112,9 @@ export class OscIntegration implements IIntegration<OscSubscription, OSCSettings
...
 
   private stringToOSCArgs(argsString: string): ArgumentType[] {
-    // NOTE: regex taken from https://stackoverflow.com/questions/4031900/split-a-string-by-whitespace-keeping-quoted-segments-allowing-escaped-quotes
-    const parseRegex = /[\w.]+|"(?:\\"|[^"])+"/g;
+    const matches = splitWhitespace(argsString);
 
-    const matches = argsString.match(parseRegex);
-
-    const parsedArguments = [];
+    const parsedArguments = new Array<ArgumentType>();
```
jwetzell, line 118: @alex-Arc if you want to redo the individual arg type/value parsing as well go ahead. The splitWhitespace doesn't return things in the same way as my previous method did so there will need to be some changes here (probably just remove the quoted string stuff) once you are done with the splitWhitespace stuff.

alex-Arc: i'm was looking around to try and get a way for us to force floats for `0.0` and the like
but found that our OSC lib will do a check to see if `0.0 == 0` and then convert it to an integer
this all happens in the Message contractor and since we can't get rid of that for now I think we should just let it do most of the work

jwetzell: I did see that but I don't like that logic if the user types `1.0` it should be sent as a float NOT an integer. What am I doing right now bypasses all of that number parsing logic in `node-osc`.

jwetzell: Also that link is to the old `node-osc` repo the relevant part is [here](https://github.com/MylesBorins/node-osc/blob/e6ab7013fd7de599a07c60920aab0749aff0a837/lib/Message.mjs#L25) in the new repo. If we pass in the complete arg object `{ type: 'f', value: 'something'}` then the library doesn't try to guess what argument type it should use.

alex-Arc: ahh sorry you are way ahead of me
i totally agree, think my OSC monitor cut of the floats so I thought it was not working

apps/server/src/utils/oscArgParser.ts
```diff
@@ -0,0 +1,27 @@
...
+  if (typeof argsString === 'undefined') {
+    return new Array<ArgumentType>();
+  }
+  const matches = splitWhitespace(argsString);
+
+  const parsedArguments = matches.map((argString: string) => {
+    const maybeNumber = Number(argString);
+    if (!Number.isNaN(maybeNumber)) {
+      //NOTE: number like: 1 2.0 33333
+      //TODO: we cant foce a float with our current osc lib
```
jwetzell, line 14: This is not true. You removed all the code I had that forced types and gave us complete control over the OSC argument types.

alex-Arc: yeah sorry I totally broke your stuff in an attempt to fix something you already had working
I'm reverting some of it now

apps/server/src/services/integration-service/OscIntegration.ts
jwetzell: Yeah, I did do some research before doing it the way I did it. I saw how the library parsed raw values into OSC arg types and didn't think that it was that great (mainly that 1.0 would become an integer). I'll stick out of this for now and let you finish it.

apps/server/src/utils/oscArgParser.ts
jwetzell: No worries!

apps/server/src/utils/oscArgParser.ts
```diff
@@ -19,6 +19,16 @@ export function stringToOSCArgs(argsString: string | undefined): Argument[] {
...
     }
 
+    if (argString === 'TRUE') {
+      // NOTE: Boolean true
+      return { type: 'T', value: true };
+    }
+
+    if (argString === 'FALSE') {
+      // NOTE: Boolean false
+      return { type: 'F', value: false };
```
jwetzell, line 29: Does this actually work? The `node-osc` library [does not support those new types](https://github.com/MylesBorins/node-osc/blob/e6ab7013fd7de599a07c60920aab0749aff0a837/lib/Message.mjs#L1) as far as I can see.

jwetzell: Tested locally and it does send as boolean types.

alex-Arc: Thanks for checking 
Do you have any comments for the tests?

jwetzell: Nope looks pretty complete (or at least the cases I could think of). Just change the contents of `describe` at the beginning looks like it wasn't changed from copy/paste.

alex-Arc:
@jwetzell 
if the user puts something like `{{eventNow.title}}` in the payload and that returns `''` be course the title is empty
Should we drop that payload? We kind of have to be course `''` is what we get if the user manually types an empty payload

then what if the payload is `{{eventNow.title}} somthingElse` then somethingElse will move it's position in the array depending on the title

jwetzell:
@alex-Arc is this templating specific to OSC arguments or is it used in multiple places?

I'm thinking if the templating could be setup (maybe a boolean passed into it) to return empty string results as a double quoted empty string `""`.

alex-Arc:
we use the same template parser for all the integrations, but we could maybe make a version for OSC

Issue #940, opened 2024-05-08
Clock fixable?


pbelbin:
I'm trying out 'ontime'.

Looks interesting and useful.

Here it comes:

But, during testing, I'm finding that when I have the Docker image (latest build) on a windows machine, and I have NTP (Chrony) configured within, and seemingly the terminal shows a fairly good correlation with what the https://www.time.is shows, that ontime's browser clock is not keeping good sync.

Seems like it's kinda free-running and periodically resyncs, and between those resyncs, the time slips further and further behind.

![GetOnTimeClockNotGood](https://github.com/cpvalente/ontime/assets/10079005/72e4ed73-6857-46d9-b18a-3fdf5e0a1105)

So, the clock kinda 'jumps' back into sync for a moment, and thereafter, it progressively falls behind, until the next 'jump' back to correctness.

For a tool that tries to keep everything running like clockwork, this seems to be an odd behavior.

Can it be fixed somehow?  Is there something I'm supposed to do?  It kinda looks broken to me.

cpvalente:
Hi @pbelbin , thank you for reaching out and for your very detailed specification.

I just want to send you a message to say that we are looking into it. 
I am not entirely sure what the ideal behaviour is here.

Few things at play:
- Ontime uses your system clock. Usually operating systems are pretty good at learning their internal deviation and correcting it. However I am unsure on how docker containers handle this. 
- The window from Ontime to be out of date is 32ms + network time. This just to update the UI, all the network traffic for integrations comes out before in a higher priority update. I will try to analyse if what we are seeing here is correct.

I appreciate your feedback, we will investigate and get back to you.

pbelbin:
Hey there.  Thank you for letting me know you're taking a look.

I'd be very interested in what you find, and what the solution is.

cpvalente:
HI @pbelbin,

I took some time to evaluate the issue and agreed that, even if this does not affect the functionality of the application, it does affect the confidence users take from it.
Asides from that, there is no good reason for us not to be more accurate.

You can see part of the conversation in the pull request at #955 

I have made an initial proposal which improves the stability of the timer. This has been merged and is part of the new release `v3.0.1`. It would be great to have your thoughts on it

Meanwhile, we have a few more ideas to explore and we will likely improve this further in the near releases

Thank you again for your help

cpvalente:
Closing this issue for now, please feel free to reopen

pbelbin:
[...]

PR #955, opened 2024-05-11
refactor: leading updates in timers

cpvalente:
This PR offers a solution to improve the described in #940 

The existing implementation assumes the following:
- we update the clocks once a second, not on the turn of the second
- clock updates are considered low priority
- clock updates mentioned here are just for the WebSocket message, the integrations are checked at intervals of 32ms, apart from events such as `end` which are scheduled for extra accuracy

From previous conversations, we felt this was sufficient, but it is understandable that it may cause suspicion and loss of confidence from users side.

The proposed solution implements changes so that we have a leading update on clock and timer.
ie: we update the clock on the turn of the second, not once a second

See below the rate of updates in websocket. The clock message comes within 32ms after the second turn.
Screenshot 2024-05-11 at 11 28 31

I believe we can further improve this by
a) schedule individual updates instead of scheduling the interval update
b) increasing the standard refresh rate (currently 32ms)

However I am unsure if the changes are necessary. 
The refresh rate is currently a setting (which is kept away from the users). We could also expose this as an application setting which I am unsure if it is a good idea.

### DIFF
--- a/apps/server/src/config/config.ts
+++ b/apps/server/src/config/config.ts
@@ -2,5 +2,5 @@ export const timerConfig = {
   skipLimit: 1000, // threshold of skip for recalculating
   updateRate: 32, // how often do we update the timer
   notificationRate: 1000, // how often do we notify clients and integrations
-  triggerAhead: 16, // how far ahead do we trigger the end event
+  triggerAhead: 10, // how far ahead do we trigger the end event
 };
--- a/apps/server/src/services/TimerService.ts
+++ b/apps/server/src/services/TimerService.ts
@@ -2,6 +2,7 @@ import { OntimeEvent } from 'ontime-types';
 
 import * as runtimeState from '../stores/runtimeState.js';
 import type { UpdateResult } from '../stores/runtimeState.js';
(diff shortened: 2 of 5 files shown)


alex-Arc:
I think this is a good solution to update on the second rollover

we could look in to the scheduling, but maybe it is more complexity than it is worth (if it's easy more precision is always nice).
I would suggest going with this for now.

increasing notification rate seams "dangerous" at least as a default, but exposing the setting could be nice (and clamp it to some sane values)

cpvalente:
> I think this is a good solution to update on the second rollover
> 
> we could look in to the scheduling, but maybe it is more complexity than it is worth (if it's easy more precision is always nice). I would suggest going with this for now.
> 
> increasing notification rate seams "dangerous" at least as a default, but exposing the setting could be nice (and clamp it to some sane values)

Agreed.
Scheduling shouldnt be too complex (in principle)

We would need some time to check that there are no unintended side effects.
I suggest merging this as a good improvement and continuing investigating

alex-Arc (approved):

Issue #971, opened 2024-05-14
Import from spreadsheet does not seem to parse smart entry times correctly


lukestein:
*Installation*: v3.0.2 via Docker (Synology NAS), Microsoft® Excel for Mac 16.84.

---
Import from spreadsheet does not seem to parse smart entry times correctly. This may be used error or an Excel version incompatibility.

For example, I can not get a one hour duration successfully parsed as "60" (string or number)

Warning times of "15" parse as 15 minutes, but "10" doesn't parse at all

(It seems like maybe actual Excel booleans also don't get parsed for e.g., Link start, Public, Skip… they have to be strings "false" vs. "true.)

cpvalente:
> *Installation*: v3.0.2 via Docker (Synology NAS), Microsoft® Excel for Mac 16.84.
> 
> 
> 
> ---
> 
> Import from spreadsheet does not seem to parse smart entry times correctly. This may be used error or an Excel version incompatibility.
> 
> 
> 
> For example, I can not get a one hour duration successfully parsed as "60" (string or number)
> 
> 
> 
> Warning times of "15" parse as 15 minutes, but "10" doesn't parse at all
> 
> 
> 
> (It seems like maybe actual Excel booleans also don't get parsed for e.g., Link start, Public, Skip… they have to be strings "false" vs. "true.)

Hi! 

I wonder if the issue might have to do with the formatting of the fields. 

Are you able to send me a template excel file with the issue to mail@getontime.no?

lukestein:
Here's an example. All the data in the yellow-shaded cells doesn't successfully for me.

[Demo import.xlsx](https://github.com/cpvalente/ontime/files/15313153/Demo.import.xlsx)

cpvalente:
Hi @lukestein , thank you for this.

I will do some investigation on our side and hopefully have a fix in a release soon

lukestein:
(Not sure whether better to comment here or at #973 ; sorry)

> **duration `"60"` gets parsed as `0`** the documentation suggests that the parsing of time strings is the same as the interface. this is almost correct, we first try to create a date from the string, if that fails we attempt parsing the string according to the UI rules. Unfortunately `"60"` is a valid input to the Date constructor and makes our logic fail. I have attempted to use a regex to look for ISO 8601 strings and parse the string using the `forgivingTimeString()` otherwise

[...]

PR #973, opened 2024-05-16
fix: excel parsing

cpvalente:
Investigating the issues raised in #971 

This far, the issue encountered were related to unexpected data types.
I have made some small code changes and a test to support the bug report

**time warning `15` gets parsed as `0`**
in the time fields, we expect a string which is either a ISO 8601 date string or a string which includes a timestamp (eg `"09:00:00"`). In some cases the value was a number such as `15`, which was ignored by our parser

**duration `"60"` gets parsed as `0`**
the documentation suggests that the parsing of time strings is the same as the interface. this is almost correct, we first try to create a date from the string, if that fails we attempt parsing the string according to the UI rules. Unfortunately `"60"` is a valid input to the Date constructor and makes our logic fail. I have attempted to use a regex to look for ISO 8601 strings and parse the string using the `forgivingTimeString()` otherwise

**boolean `true` gets parsed as `false`**
it seems that we never expected to receive a boolean value from excel

### DIFF
--- a/apps/server/src/utils/__tests__/parser.test.ts
+++ b/apps/server/src/utils/__tests__/parser.test.ts
@@ -19,7 +19,7 @@ import { dbModel } from '../../models/dataModel.js';
 import { createEvent, getCustomFieldData, parseExcel, parseJson } from '../parser.js';
 import { makeString } from '../parserUtils.js';
 import { parseRundown, parseUrlPresets, parseViewSettings } from '../parserFunctions.js';
-import { ImportMap } from 'ontime-utils';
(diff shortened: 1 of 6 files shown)


alex-Arc (approved):
Also works from google sheets as far as I can see

lukestein:
I'm not sure how easy it is (whenever appropriate) to push out a beta docker image to DockerHub, but I'm happy to try some testing whenever.

cpvalente:
> Also works from google sheets as far as I can see

thank you @alex-Arc, did you test with our google sheets template or with another source?

alex-Arc:
tried it from sheets with Luke's example

Issue #976, opened 2024-05-16
Background colors (`?key=`) in more views, and actually support transparency rather than drawing on opaque black canvas?


lukestein:
Related to #162 :

- The `?key=` argument takes an 8-digit hex code as if supporting an alpha/transparency, but seems to be drawn on top of a black background so the page doesn't wind up rendering transparent. My use case is that I like to overlay timers in OBS or Electron Capture as in [this demo](https://www.youtube.com/watch?v=2BvBerQfX_0) and this screenshot (not ontime)
<img width="956" alt="image" src="https://github.com/cpvalente/ontime/assets/44452336/012ad3d6-a733-4ca4-94f3-ffbad990d04c">

- The `?key=` argument is only supported in `/minimal` and `/clock`. Could it be easily added to at least `/timer`? (Honestly wouldnt' look bad for public, countdown, backstage for visual consistency)

cpvalente:
Thank you @lukestein , this is a good suggestion.

Meanwhile to get the key in OBS you need to override the CSS to set the body background color to transparent
See the docs https://docs.getontime.no/quick-tips/lower-thirds-in-obs/

lukestein:
> Meanwhile to get the key in OBS you need to override the CSS to set the body background color to transparent
> 
> See the docs https://docs.getontime.no/quick-tips/lower-thirds-in-obs/

Thanks for this pointer (which I'd somehow missed)

However, I can't get obs to successfully eliminate the background with a CSS override. May be user error.

cpvalente:
Hm,

I did a quick test now in `/minimal` and the combination of setting both the background in the body and the key value to `#00000000` seemed to result in a transparent background

Having said this, I havent tested this in OBS and I am not sure this still works.
@alex-Arc , do you have time to sanity check?

lukestein:
Nothing doing for me (again, may be user error):
https://github.com/cpvalente/ontime/assets/44452336/e2dd0bf6-24bb-407a-9367-62e2a8711d0c

cpvalente:
> Nothing doing for me (again, may be user error):
> 
> https://github.com/cpvalente/ontime/assets/44452336/e2dd0bf6-24bb-407a-9367-62e2a8711d0c

I will take a look. 

A workaround for now could be to set the key colour to something you can key (green or full black)

But this should work and I will see what is happening

alex-Arc:
the bacgound color line in index.scss is preventing the transparent background

[...]

PR #978, opened 2024-05-17
style: simplify background override

cpvalente:
Improves the keying of the views as described in #976 

The app has two levels of background, one set at the body level, and one set at route (page) level

These are effectively the same. It seems the reason the body background was introduced was to prevent a white flash while navigating
This had the unintended consequence of complicating the keying of views.

This PR introduces some changes to improve the ease of configuration:
- use `--background-color-override` with a fallback in body
- simplify selectors in `body` to avoid issues with specificity
- set background and page level in cuesheet and editor to avoid issues with overrides

This means that we avoid the issues with flashing of background colour.
The users only need to override a single variable `--background-color-override` to see their changes.

Note that `--background-color-override` is overriden by the `key` parameter. 
That means that the users can use `--background-color-override` without the `key` to key any view

We will need a follow up in the docs

### DIFF
--- a/apps/client/src/features/cuesheet/CuesheetWrapper.module.scss
+++ b/apps/client/src/features/cuesheet/CuesheetWrapper.module.scss
@@ -1,4 +1,5 @@
 .tableWrapper {
+  background-color: $ui-black;
   width: 100%;
   height: 100vh;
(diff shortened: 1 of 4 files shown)


alex-Arc (approved):
🟢

PR #952, opened 2024-05-10
refactor: improve detection of day after

cpvalente:
Tweaks to how we infer gaps in the rundown
Fixes #949 

It also starts some philosophical questions which I added to #951 

I believe the issue with the planned end being the day after is resolved in #952
We can merge that one first and double-check

PR #980, opened 2024-05-17
refactor: recover day after indicator

cpvalente:
This recovers the day after indicator by fixing the out of sync logic in UI and server regarding what makes something being the day after.

It is a follow up of #952  and is discussed in #951


alex-Arc (approved):
Nice
at some point we should also consider what happens if we mark the event as skip should we then recalculate the start and end time

PR #982, opened 2024-05-18
refactor: cap timer values

cpvalente:
Currently, users can add whatever they want in the time fields.
eg: 1000:01:01 is totally acceptable as 1000 hours 1 minute and 1 second

This has the potential of causing issues for us, at the same time, we dont see that it brings value to users

This PR caps all timer values for a single event to 24 hours, both in the interface and using the change endpoint

Note to reviewer:
I have opted for leave the parser code untouched. This is because the logic is complex and I would like to find a good way to handle edge cases
I suggest moving time parsing improvements to another PR

### DIFF
--- a/apps/client/src/common/hooks/useEventAction.ts
+++ b/apps/client/src/common/hooks/useEventAction.ts
@@ -1,7 +1,14 @@
 import { useCallback } from 'react';
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { isOntimeEvent, MaybeString, OntimeEvent, OntimeRundownEntry, RundownCached } from 'ontime-types';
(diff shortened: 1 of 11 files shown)


alex-Arc:
should we also cap the add time?
both in the UI sense and in how much the maximum added time can be
image

alex-Arc (approved):

cpvalente:
> should we also cap the add time?
both in the UI sense and in how much the maximum added time can be


Yes, that is a good idea.
What do you think should be the maximum time to add? maybe 1 hour?

alex-Arc:
> Yes, that is a good idea. What do you think should be the maximum time to add? maybe 1 hour?

maybe also just 24h ?
but 1h is also fine for me

PR #983, opened 2024-05-19
Multistage docker build

thelan:
Hello,

This is a simple change to do a multi-stage build.

This might be useful to avoid installing node / pnpm on the host and running build commands before packaging the result to the image. 

It will be helpful for CI/CD pipelines

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,3 +12,7 @@
 # Ignore git and cache folders
 .git
 .cache
+
+# Ignore default volumes created by running docker compose up
+ontime-db
+ontime-styles
--- a/.dockerignore
+++ b/.dockerignore
@@ -11,7 +11,12 @@
 
 # Ignore git and cache folders
 .git
+.gitignore
 .cache
+
+# Ignore build folders
+node_modules
+dist
 
 # Ignore default volumes created by running docker compose up
 ontime-db
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -88,8 +88,6 @@ While it should allow for a generic setup, it might need to be modified to fit y
 
 From the project root, run the following commands
 
-- __Install the project dependencies__ by running `pnpm i`
-- __Build packages__ by running `pnpm build:localdocker`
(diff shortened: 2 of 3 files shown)


cpvalente:
Hi @thelan, thank you for this.
A multi-stage build can definitely help the image be slimmer and this is a great idea

I am not very strong in this field and will take some time to test the image to make sure everything is fine.
Meanwhile would you mind providing some extra information as to the impact you have observed in image size and build run time?

thelan:
Actually my first goal was to provide a way to build the image simply without installing node on the 'host' computer.

After checking it made the new image slightly heavier (2MB)😔

```
REPOSITORY                              TAG                   IMAGE ID       CREATED          SIZE
getontime/ontime                        multistage            96258552ec50   53 minutes ago   145MB
getontime/ontime                        latest                106787ad4478   6 days ago       143MB
```

Maybe it's possible to tune what's copied from the "build" container inside the Dockerfile

cpvalente:
Thank you for the clarification @thelan.

I am not entirely sure who the change is helping.
Users who are building Ontime themselves rather than consuming images, will need node and pnpm do be able to develop.

I may be missing something, again, this is not my strong side. What do you think?

jwetzell:
This looks pretty close. There are a few things I would clean up. You don't need to download pnpm, just follow the [instructions from pnpm](https://pnpm.io/docker) on how to use in docker. Mainly these lines which can go right after the `FROM` line of the builder
```
ENV PNPM_HOME="/pnpm"
ENV PATH="$PNPM_HOME:$PATH"
RUN corepack enable
```


The `NODE_ENV` should probably be `production`.

I would also model the build steps after the current github action that builds the docker image. So  a `pnpm install --frozen-lockfile` followed by `pnpm turbo build:docker`.

thelan:
@jwetzell Thanks for the feedback I've updated the Dockerfile accordingly.

The image seems to drop back to 143MB like the original one

jwetzell:
Look good to me @thelan

PR #985, opened 2024-05-20
fix #984 Expected End of timeline is not calculated correctly

### LINKED ISSUE #984: Expected End of timeline is not calculated correctly
Installation: current master(dae6f84af0dd25c390b9e762ab963f118ca0c0e1) on linux

Steps to Reproduce:
- Create one event starting at 20:00 with a duration of 1hour
- Start the Event

Now the Expected End in the Editor Header will display wrong times
Starting the event at 19:45 will result in a Expected End of 21:15
Starting the event at 20:15 will result in a Expected End of 20:45

The expected End of the current Event in the playback area is displayed correctly

ekordas:


ekordas:
One Todo left: Tigger a recalculation of Expected End continuously when an event timer goes overtime. I am not yet that familiar with the code, and dont know how to do that the best way

cpvalente:
> One Todo left: Tigger a recalculation of Expected End continuously when an event timer goes overtime. I am not yet that familiar with the code, and dont know how to do that the best way

There is a function called `getExpectedEnd()` which takes the `runtimeState` and calculates the new `expectedEnd` value, we would need to add this to the update function in `runtimeState`

cpvalente:
Thank you @ekordas, this is a good catch and good job finding your way through the code.
Do you need help with that last todo?

ekordas:
I will give it a try

alex-Arc (approved):

cpvalente:
Nice @ekordas , this looks good to me, is it ok with you if I merge it to include in the next release?

ekordas:
I thought this is the point oft making Pull requests😂 Of course!

PR #988, opened 2024-05-21
Add option to display the full time (HH:mm:ss) in the timer view

thelan:
Hello,

Its just a small patch to add an view option to enable the timer to display the full-clock instead of being hardcoded to remove leading zeros.

Let me know if i need to change or optimize things

### DIFF
--- a/apps/client/src/common/components/view-params-editor/constants.ts
+++ b/apps/client/src/common/components/view-params-editor/constants.ts
@@ -30,6 +30,14 @@ const hideTimerSeconds: ParamField = {
   defaultValue: false,
 };
 
+const showLeadingZeros: ParamField = {
+  id: 'showLeadingZeros',
+  title: 'Show leading zeros in timer',
+  description: 'Whether to show leading zeros in the running timer',
+  type: 'boolean',
+  defaultValue: false,
+};
+
 export const getClockOptions = (timeFormat: string): ParamField[] => [
   getTimeOption(timeFormat),
   {
@@ -107,6 +115,7 @@ export const getTimerOptions = (timeFormat: string, customFields: CustomFields):
   return [
     getTimeOption(timeFormat),
     hideTimerSeconds,
+    showLeadingZeros,
     {
       id: 'hideClock',
       title: 'Hide Time Now',
--- a/apps/client/src/features/viewers/common/viewUtils.ts
+++ b/apps/client/src/features/viewers/common/viewUtils.ts
@@ -84,7 +84,7 @@ export function getFormattedTimer(
   }
 
   let display = millisToString(timeToParse);
-  if(options.removeLeadingZero) {
+  if (options.removeLeadingZero) {
     display = removeLeadingZero(display);
   }
 
--- a/apps/client/src/features/viewers/common/viewUtils.ts
+++ b/apps/client/src/features/viewers/common/viewUtils.ts
@@ -84,7 +84,9 @@ export function getFormattedTimer(
   }
 
   let display = millisToString(timeToParse);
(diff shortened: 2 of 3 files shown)


apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -30,6 +30,14 @@ const hideTimerSeconds: ParamField = {
   defaultValue: false,
 };
 
+const removeLeadingZeros: ParamField = {
+  id: 'hideTimerSeconds',
```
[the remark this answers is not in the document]
thelan: my mistake. Fixing this

thelan:
Ok so i fixed the typos, also inverted the logic to keep the default setting as it use to be to be.

I think it's fully functional now.
Let me know if i need to change anything more.

PS: I'll look forward to setup a proper environment to deal with node (actually using vim to edit files is pretty limited)

apps/client/src/features/viewers/common/viewUtils.ts
```diff
@@ -84,7 +84,9 @@ export function getFormattedTimer(
   }
 
   let display = millisToString(timeToParse);
-  display = removeLeadingZero(display);
+  if(options.removeLeadingZero) {
```
cpvalente, line 87:
```suggestion
  if (options.removeLeadingZero) {
```

apps/client/src/common/components/view-params-editor/constants.ts
```diff
@@ -30,6 +30,14 @@ const hideTimerSeconds: ParamField = {
   defaultValue: false,
 };
 
+const showLeadingZeros: ParamField = {
```
cpvalente, line 33: all other options are for hiding custom behaviour, should we keep it here as well? I guess we can just default it to true

cpvalente:
Thank you @thelan , left some small suggestions and questions for you.

Setting up an environment should be pretty straightforward, you can also use GitHub codespaces in your fork.
If you get into any trouble or the documentation for developers is insufficient please let me know, I would like this to be as easy as possible
https://github.com/cpvalente/ontime/blob/master/DEVELOPMENT.md

apps/client/src/common/components/view-params-editor/constants.ts
thelan: Honestly I don't know. Usually I prefer keeping the original behavior and use an option to change it.
User updating the app will probably be surprised that the default format of the timer has changed ?

But if you prefer i can reverse the condition to make it the new default

cpvalente: if the variable was to hide, and the default was on, the users wouldnt see a difference.

My suggestion was with the inclination that all the other options are to hide things, so this one is different.

But I guess you may be correct in the idea that it doesnt matter if it is hide / show, what we are doing is changing the default behaviour. In which case your change makes more sense than my proposal. What do you think?

thelan: Changing the text and the default value of the option should be nice to have all the option text to do a "hide function" but i struggle with the boolean parser here:
https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/viewers/common/viewUtils.ts#L32-36

When the parameter is absent from the parameters part in url it will be parsed to "false" thus breaking the default value.
I don't know how to fix this so that's why I inverted the logic from "hide" to "show"

cpvalente:
Thank you @thelan , apologies it took so long.
We have been considering whether we should be looking at cleanup the the options for time formatting altogether, but that wont happen at this time

PR #990, opened 2024-05-21
fix docker multi-architecture build

jwetzell:
I've tested this locally and all 4 platforms now build successfully.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,11 @@
-FROM node:18.18-alpine AS base
+FROM node:18.18-alpine AS builder
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 RUN corepack enable
-
-FROM base AS builder
 COPY . /app
 WORKDIR /app
-RUN pnpm install --frozen-lockfile \
-    && pnpm turbo build:docker
+RUN pnpm --filter=ontime-ui --filter=ontime-server --filter=ontime-utils install --config.dedupe-peer-dependents=false --frozen-lockfile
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,13 +1,11 @@
...
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 RUN corepack enable
-
-FROM base AS builder
 COPY . /app
 WORKDIR /app
-RUN pnpm install --frozen-lockfile \
-    && pnpm turbo build:docker
+RUN pnpm --filter=ontime-ui --filter=ontime-server --filter=ontime-utils install --config.dedupe-peer-dependents=false --frozen-lockfile
```
cpvalente, line 7: nice!

- Can you please provide some background as to the need of disabling deduplication? Is to to avoid installing the client dependencies?

cpvalente (approved):
🚀

Dockerfile
jwetzell: I don't know pnpm too well but I was trying to just install the dependencies that are needed using the `--filter` but it was still trying to install dependencies for electron which is what was causing the docker build to error out on arm/v7 and arm/v6 builds (after I removed turbo). [This open issue](https://github.com/pnpm/pnpm/issues/6300) on `pnpm` says filtering for install does not work unless `dedupe-peer-dependents` is false.

I can link the issue in a comment in the Dockerfile.

cpvalente: yes, the pnpm config makes sense.

In this case, the client is not built, just double-checking that this is the intended behaviour

PR #996, opened 2024-05-24
refactor: context menu actions

cpvalente:
adding delete action for multiple selection


alex-Arc:
nice
- could/should we make the shortcut work for this also?
- maybe add a popup waring or make it more clear in the context menu that multiple events are being changed

Issue #997, opened 2024-05-24
Ability to hide shortcuts list on the rundown webpage window 


cpvalente:
### Discussed in https://github.com/cpvalente/ontime/discussions/989

<div type='discussions-op-text'>

<sup>Originally posted by **johnman51** May 21, 2024</sup>
I'd like to know if it would be possible to add the ability to hide shortcuts list (on the right side) on the rundown webpage window ?
I can't shrink the window to hide it, it always resizes keeping the shortcut list visible…</div>

cpvalente:
This is done and was released in 3.0.4

PR #999, opened 2024-05-24
refactor: allow hiding sidebar

cpvalente:
Proposal implementation for hiding the side bar once the rundown is extracted. Here we do this by adding a parameter to the URL #997 #989 .

Screenshot 2024-05-24 at 13 23 16

This is a familiar pattern in the app, however I am not particularly happy here since it is not self documenting
ie: the user would have to know the parameter to use

I have also considered adding a toggle button, however this clutters the interface. I am not inclined to want a settings panel here as we have in cuesheet or on the views. Unsure on the best way forwards

Any ideas @alex-Ar @jwetzell @lukestein


alex-Arc:
hmm
yeah that might be more useful solution than what I trying with @media (max-width: 1050px) 
https://github.com/cpvalente/ontime/compare/master...hide-rundown-editor

Issue #991, opened 2024-05-22
Feature Request: Display multiple custom fields in operator view


Ashtlind:
Hi team!

We are really enjoying playing with v3. We would love to see the ability to show multiple custom fields in the operator view much like the cuesheet view. Feel this would be an excellent view option to display in a view switcher multiview box or something similar for quick glance info.

cpvalente:
> We are really enjoying playing with v3. We would love to see the ability to show multiple custom fields in the operator view much like the cuesheet view. Feel this would be an excellent view option to display in a view switcher multiview box or something similar for quick glance info.


Hi @Ashtlind 

Thank you for reaching out.

We try to approach the creation of views with a "different views for different jobs" mentality.
In this case, would you kind clarifying how the cuesheet view does not enable the workflow that you are after?

The operator view is not particularly well suited for multiple subscriptions (custom fields) but if we can assert that there is a gap in the feature set I am happy to consider an extension or a new view

Ashtlind:
Hi @cpvalente 

Thanks so much for your reply.

Totally understand that. We are hoping for something similar to the shoflo display view, this is ideal for using within a camera multiview layout. (Example of this being used below)
It makes a really easy point of reference for the whole team on where we are in the rundown and the next few items (not just the next one but looking a few ahead also is ideal)

We are looking for something that can display the relevant info in a really easy to read and un-cluttered format with quite a large font size. We have found the backstage view to have a bit much info on it and the cuesheet isn't super easy to read multiple columns at a glance in a small window. This would be super helpful as there would be multiple operators seeing this view to be able to have multiple custom fields shown.

Here is showflow's layout
https://shoflo.tv/display-view/
![image](https://github.com/cpvalente/ontime/assets/19161455/f8cc8969-6feb-45c9-9591-899d6c498c6f)


In this screenshot you can see shoflow's display view in the camera multiview underneath cam 13.

![image](https://github.com/cpvalente/ontime/assets/19161455/c5ff2d53-7d31-46b2-823b-1d696d28594d)



I really like the way the operater view would look for this, here is a mockup:
![image](https://github.com/cpvalente/ontime/assets/19161455/7732e356-6d73-4827-8938-805941b97366)


Compared to the current cuesheet look at a high zoom ratio:
![image](https://github.com/cpvalente/ontime/assets/19161455/07dd73c6-ad8e-44b6-a0ab-bb3afbc156bc)


Hopefully this helps clear up some of our thoughts

[...]

PR #1001, opened 2024-05-26
display multiple custom fields in operator view

alex-Arc:
implements #991

image

image


apps/client/package.json
```diff
@@ -19,6 +19,7 @@
     "@tanstack/react-table": "^8.11.3",
     "autosize": "^6.0.1",
     "axios": "^1.2.0",
+    "chakra-react-select": "^4.7.6",
```
cpvalente, line 22: I am not particularly confident on adding this as a dependency.

This seems to be maintained by a single person. From a quick review of the github repo it seems that the maintenance is not as good as we would like.

Chakra is about to release v3, we would be stuck waiting for this library to upgrade. We really are hoping that chakra will improve its performance or we would look to bail to another library.

Either way, we need to have independence on this task.

How about an array of checkboxes like we do in the cuesheet? not as nice, but simple enough?
We could look into improving later on

alex-Arc: sounds good

alex-Arc:
@cpvalente I could use some help with the styling

cpvalente:
Hi @alex-Arc , just finish what you can and I can do the styling.
would you mind adding a screenshot highlighting the issue?

alex-Arc:
I have gotten as far as i can. It works as before,
it's mainly just getting it to look like the other fileds
image
image

alex-Arc:
@cpvalente I tried again to get the theming to work, but no luck
https://v2.chakra-ui.com/docs/components/menu/usage

I would like it to look somewhat like the select element or at least be the same colour

apps/client/src/features/operator/edit-modal/EditModal.tsx
```diff
@@ -16,33 +16,45 @@ export default function EditModal(props: EditModalProps) {
...
-    }
+    inputRef.current?.forEach(async (element) => {
+      if (element.dataset.field && element.defaultValue != element.value) {
+        await updateCustomField(event.id, element.dataset.field, element.value);
+      }
+    });
+    // const newValue = inputRef.current?.value;
+    // if (newValue === undefined) {
+    //   return;
+    // }
```
cpvalente, line 31: ??

apps/client/src/common/hooks/useEventAction.ts
```diff
@@ -174,6 +174,7 @@ export const useEventAction = () => {
 
   const updateCustomField = useCallback(
     async (eventId: string, field: string, value: string) => {
+      console.log({ id: eventId, custom: { [field]: value } });
```
cpvalente, line 177: can we remove this?

cpvalente (commented):
I think this looks OK. 

It seems a bit annoying that the multi select menu does not keep between versions.

1. open the menu and multi-select some stuff
2. press save and marvel
3. open the menu again and change another setting
4. press save and see all your multi-select is gone

Can we also have a sanity check for performance?
Since subscriptions is no longer a string, we lost our stable reference

We could split the ontime event block for some fewer calculations around mapping around subscriptions, but dealing with objects makes things less performant and more complex

apps/client/src/features/operator/operator-event/OperatorEvent.tsx
```diff
@@ -89,12 +89,19 @@ function OperatorEvent(props: OperatorEventProps) {
...
-          <>
-            <span className={style.field}>{subscribedAlias}</span>
-            <span className={style.value}>{subscribed}</span>
-          </>
-        )}
+        {subscribed &&
+          subscribed
+            .filter((field) => field.value)
+            .map((field, index) => (
+              <div key={index}>
```
cpvalente, line 96: tip: it is good to avoid using indexes as keys since they dont accurately distinguish the dataset (example: if the dataset reorders, a different piece of data can be at this dataset)

```suggestion
            .map((field) => (
              <div key={field.id}>
```

apps/client/src/features/operator/edit-modal/EditModal.tsx
```diff
@@ -16,33 +16,45 @@ export default function EditModal(props: EditModalProps) {
...
 
   const handleSave = async () => {
     setLoading(true);
-    const newValue = inputRef.current?.value;
-    if (newValue === undefined) {
-      return;
-    }
+    inputRef.current?.forEach(async (element) => {
+      if (element.dataset.field && element.defaultValue != element.value) {
+        await updateCustomField(event.id, element.dataset.field, element.value);
```
cpvalente, line 25: I fear this may be a bit too complicated. some tips:

- check if inputRef.current is null and short circuit, like
```ts
  if (!inputRef.current) return;
```

- we likely dont want to make several requests to update all the promises. could we generate a single patch object and patch at once?

apps/client/src/features/operator/edit-modal/EditModal.tsx
alex-Arc: yes 
was just keeping the original for reference

apps/client/src/common/hooks/useEventAction.ts
alex-Arc: yep 😉

alex-Arc:
i fixed the disappearing values when editing again
but they still disappear when looking at the dropdown twice

https://github.com/cpvalente/ontime/assets/19875125/f99bf2ba-f7f0-41a6-85e1-eb5fe33c18ce

I think be course we don't have any where to keep the string array
Should I add one more useRef for this or a useState?

cpvalente:
> I think be course we don't have any where to keep the string array
> 
> Should I add one more useRef for this or a useState?
> 
> 

To have a piece of data which causes re-renders you need usaste-te

cpvalente:
> I needed a useEffect to make it stable is this ok ? @cpvalente [01cf0b0](https://github.com/cpvalente/ontime/commit/01cf0b08a53deda619a7357e0581abd7aa3ec2cf)

you usually dont need to combine both the `useEffect` and a `useState`
would the following not work?
```ts
  const [paramState, setParamState] = useState<string>(setParamState(defaultOptionValue));
````


> should I make the other types into components as well?

I would be of the opinion that things that are "as well" are generally out of scope. For now lets focus on merging this feature and keep the improvements in mind

cpvalente (approved):
Works well and code looks good, good work

PR #1004, opened 2024-05-26
add notes to secondaryOptions in timer view

### LINKED ISSUE #1002: Timer: Unable to Select Secondary Text
Server: Version 3.0.3 on Docker
Client: Safari (17.5) and Firefox  (126.0) on MacOS 14.5

In the settings for Timer display, unable to select any options for Secondary Text, only see 'Select an Option'

Works in Backstage View.

alex-Arc:
closes #1002

### DIFF
--- a/apps/client/src/common/components/view-params-editor/constants.ts
+++ b/apps/client/src/common/components/view-params-editor/constants.ts
@@ -103,7 +103,7 @@ export const getClockOptions = (timeFormat: string): ParamField[] => [
 ];
 
 export const getTimerOptions = (timeFormat: string, customFields: CustomFields): ParamField[] => {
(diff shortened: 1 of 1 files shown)


cpvalente:
hey @alex-Arc , do you think this should be added?

The reason it doesnt exist is that Note is usually an operator dataset, which we consider private to the production team and out of scope for the timer views.

Users have custom fields to add whatever they want. 

What do you think?

alex-Arc:
I don't have a strong opinion here
We show all if the custom fileds even if the the user intends for them to be for production, so I don't see a big reason to treat the notes differently?

cpvalente:
Ship it 🚀

PR #1005, opened 2024-05-26
refactor: extract parseUserTime

cpvalente:
This PR extracts the recent logic for parsing time strings in frontend to be shared in server and UI.
This means that in Excel, we use the same parsing of time strings as we do in the UI inputs

There arent many changes otherwise, but
- rename to parseUserTime 
- simplify AM / PM logic to expect that to always be the last thing in a string
- use `Number()` instead of `parseInt()` so that it will fail if the string is not a parseable number

Note for reviewer:
The first commit is where the large change is, which is just moving things from one package to another and can be ignored
The change itself is pretty minimal


jwetzell (approved):
Nothing stands out to me here.

alex-Arc (commented):
most functionality seams preserved

This now somewhat parses the`p+30` syntax during import and `p+30` will become `00:30:00` before there is no reference to what came before `p` is always 0
dont know it that is a good or bad thing

also it seams to accept a lot of stuff after the number
`15whatever` -> `15:00:00`
but
`5whatever` -> `00:00:00`

alex-Arc (approved):

PR #1009, opened 2024-05-27
Allow time change from api

### LINKED ISSUE #970: Feature Request: Whitelist allow changing end time of event
I have been using ontime for a while now to manage clocks and timers around a large orchestral recording studio and loving the new workflow improvements in V3!

We use roll mode and often need to update the end time of an event if a break is started early. It would be incredibly helpful if this was possible to do over the APIs as we could then press one button on companion to set the end time of the current event to the current time.

alex-Arc:
#1017 is fixed and sending a time to the change API that is outside the allowed range will no longer be corrected just dropped

closes #970
allowed properties for change API is now: 
- title
- note
- cue
- isPublic
- skip
- colour
- custom
- timeWarning 
- timeDanger
- endAction 
- timerType 
- duration 
- timeStart 
- timeEnd 

skip, duration, timeStart, timeEnd is throttled to 1 call per 20ms

the behaviour is:
- msg 1 <-- applied
- msg 2
- msg 3
- msg 4 <-- timeout reached <-- applied
- mgs 5
- msg 6
- msg 7
- (nothing sendt from user)
- (nothing sendt from user)
- (nothing sendt from user)
- (nothing sendt from user) <-- timeout reached <-- msg 7 applied

### DIFF
--- a/apps/client/src/common/models/OntimeSettings.ts
+++ b/apps/client/src/common/models/OntimeSettings.ts
@@ -8,5 +8,4 @@ export const ontimePlaceholderSettings: Settings = {
   operatorKey: null,
   timeFormat: '24',
   language: 'en',
-  apiAllowTimeChange: false,
 };
--- a/apps/client/src/common/models/OntimeSettings.ts
+++ b/apps/client/src/common/models/OntimeSettings.ts
@@ -8,4 +8,5 @@ export const ontimePlaceholderSettings: Settings = {
   operatorKey: null,
   timeFormat: '24',
   language: 'en',
+  apiAllowTimeChange: false,
 };
--- a/apps/server/src/api-integration/integration.controller.ts
+++ b/apps/server/src/api-integration/integration.controller.ts
@@ -14,7 +14,7 @@ import { throttle } from '../utils/throttle.js';
 import { regenerateWhitelist } from '../services/rundown-service/rundownCacheUtils.js';
 
 const propertiesWhitelist = Object.freeze(Object.values(regenerateWhitelist));
-const throttledUpdateEvent = throttle(updateEvent, 1000);
+const throttledUpdateEvent = throttle(updateEvent, 20);
 
 export function dispatchFromAdapter(type: string, payload: unknown, _source?: 'osc' | 'ws' | 'http') {
   const action = type.toLowerCase();
--- a/apps/server/src/api-integration/integration.controller.ts
+++ b/apps/server/src/api-integration/integration.controller.ts
@@ -13,7 +13,7 @@ import { parseProperty, updateEvent } from './integration.utils.js';
 import { throttle } from '../utils/throttle.js';
 import { regenerateWhitelist } from '../services/rundown-service/rundownCacheUtils.js';
 
-const propertiesWhitelist = Object.freeze(Object.values(regenerateWhitelist));
+const propertiesWhitelist = Object.values(regenerateWhitelist);
 const throttledUpdateEvent = throttle(updateEvent, 20);
 
 export function dispatchFromAdapter(type: string, payload: unknown, _source?: 'osc' | 'ws' | 'http') {
--- a/apps/server/src/api-integration/integration.controller.ts
+++ b/apps/server/src/api-integration/integration.controller.ts
@@ -11,9 +11,9 @@ import * as assert from '../utils/assert.js';
 import { isEmptyObject } from '../utils/parserUtils.js';
 import { parseProperty, updateEvent } from './integration.utils.js';
 import { throttle } from '../utils/throttle.js';
-import { isDataStale } from '../services/rundown-service/rundownCacheUtils.js';
+import { willCauseRegeneration } from '../services/rundown-service/rundownCacheUtils.js';
 
-const throttledUpdateEvent = throttle(updateEvent, 2000);
+const throttledUpdateEvent = throttle(updateEvent, 20);
 
 export function dispatchFromAdapter(type: string, payload: unknown, _source?: 'osc' | 'ws' | 'http') {
   const action = type.toLowerCase();
@@ -47,12 +47,17 @@ const actionHandlers: Record<string, ActionHandler> = {
     const data = payload[id as keyof typeof payload];
     const patchEvent: Partial<OntimeEvent> & { id: string } = { id };
 
+    let shouldThrottle = false;
+
     Object.entries(data).forEach(([property, value]) => {
       if (typeof property !== 'string' || value === undefined) {
         throw new Error('Invalid property or value');
       }
 
       const newObjectProperty = parseProperty(property, value);
+
+      const key = Object.keys(newObjectProperty)[0] as keyof OntimeEvent;
+      shouldThrottle = willCauseRegeneration(key) || shouldThrottle;
 
       if (patchEvent.custom && newObjectProperty.custom) {
         Object.assign(patchEvent.custom, newObjectProperty.custom);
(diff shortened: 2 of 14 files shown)


apps/client/src/common/models/OntimeSettings.ts
```diff
@@ -8,4 +8,5 @@ export const ontimePlaceholderSettings: Settings = {
   operatorKey: null,
   timeFormat: '24',
   language: 'en',
+  apiAllowTimeChange: false,
```
cpvalente, line 11: hmmm, not super eager on this setting

I think changing too much behaviour on settings is a bad practice.
In my opinion, other software that does this ends up being difficult to use and to predict the behaviour of.
You get the situations where there are 5 technicians crunched over a laptops going through the multiple settings that affect something to try and explain some unexpected behaviour.

I think apps (and Ontime) should have reasonable standards.
We can choose to allow this or not, but that should be it.

The issue that I see, is that a bad integration (ie: someone using a module or a script which has issues) could make Ontime unusable.
The user would not see the connection since the problem surfaces in Ontime. Our user base tends to be unforgiving with runtime crashes 😅 

Do we have an expectation on how much of the users this whitelisting would improve the lives of?
Maybe the way forwards is to whitelist everything and do some rate limiting?

apps/server/src/api-integration/integration.utils.ts
```diff
@@ -6,19 +6,40 @@ import { editEvent } from '../services/rundown-service/RundownService.js';
...
+
+let rateLimitTimeout: NodeJS.Timeout | null = null;
+
+function rateLimit(value: unknown) {
+  if (rateLimitTimeout) {
+    throw new Error('time endport rate limited');
+  }
+  rateLimitTimeout = setTimeout(() => (rateLimitTimeout = null), 1000);
+  return value;
+}
```
alex-Arc, line 26: something like this?

cpvalente: could we use one of our debounce function?
We want our rate limiting to be much smaller, like 20 calls per second or something

apps/server/src/utils/coerceType.ts
```diff
@@ -1,5 +1,16 @@
 import { isColourHex } from 'ontime-utils';
 
+//TODO: write tests
```
cpvalente, line 3: :D

I see from the other functions here that this TODOS get forgotten, do you mind adding some simple tests?

apps/server/src/api-integration/integration.utils.ts
alex-Arc: I don't think the debounce or throttle dose what we want, it looks to they would potently set the value with a delay
here I think we should do it immediately or drop it, not buffer it?

I would think it is good to send a rejected message back so the call knows why it did not work?

cpvalente: I might be misunderstanding, but I believe your assumptions are not correct

I don't think the debounce or throttle dose what w

> e want, it looks to they would potently set the value with a delay
> here I think we should do it immediately or drop it, not  buffer it?

Debounce does not buffer function calls, it will always ensure that the function is eventually called, while maintaining a call rate. which means it may drop intermediate values
I am not sure if our debounce function fits the bill
https://css-tricks.com/debouncing-throttling-explained-examples/

> I would think it is good to send a rejected message back so the call knows why it did not work?

I would generally agree but in this case i feel its unnecessary and we are just contributing to the network noise

Lets say the user does something weird in touchdesigner and starts pushing data to ontime at 60fps.
we would activate our rate limit and ignore all the changes that do not fit into our (eg) 20 calls per second.
In this case, the user would still see ontime update, we internally just decided that there is no point updating so often. In the same light, there is no point sending 40 error messages per second to the client saying that the change did not apply while, in fact, the user can see the applied value

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -43,22 +48,33 @@ const actionHandlers: Record<string, ActionHandler> = {
...
+    let throttleProperty = false;
+
     Object.entries(data).forEach(([property, value]) => {
       if (typeof property !== 'string' || value === undefined) {
         throw new Error('Invalid property or value');
       }
 
       const newObjectProperty = parseProperty(property, value);
 
+      if (!throttleProperty && !propertiesWhitelist.includes(Object.keys(newObjectProperty)[0])) {
```
cpvalente, line 60: Few suggestions:
1. I think the logic for checking if a property is not whitelisted should be migrated to the scope of in rundownCacheUtils.

2. I also wonder if we could simplify with simplify the throttle assignment to simply 
`throttleProperty = throttleProperty || notWhitelisted(property)`

3. the user will always see their change (eventually) in that case, does the throttled message achieve anything?

apps/server/src/api-integration/integration.utils.ts
```diff
@@ -1,24 +1,45 @@
...
-import { coerceBoolean, coerceColour, coerceNumber, coerceString } from '../utils/coerceType.js';
+import { coerceBoolean, coerceColour, coerceEnum, coerceNumber, coerceString } from '../utils/coerceType.js';
 
-const whitelistedPayload = {
+/**
+ *
+ * @param {number} value time amount in seconds
+ * @returns {number} time in milliseconds clamped to 0 and max duration
+ */
+function clampToMaxDuration(value: number) {
```
cpvalente, line 14: Suggestion: Could we simplify the function to basically say "do the duration value clamping thing" without having to be explicit about the rules?

like:
```suggestion
function clampDuration(value: number) {
```

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -10,6 +10,11 @@ import { eventStore } from '../stores/EventStore.js';
 import * as assert from '../utils/assert.js';
 import { isEmptyObject } from '../utils/parserUtils.js';
 import { parseProperty, updateEvent } from './integration.utils.js';
+import { throttle } from '../utils/throttle.js';
+import { regenerateWhitelist } from '../services/rundown-service/rundownCacheUtils.js';
+
+const propertiesWhitelist = Object.freeze(Object.values(regenerateWhitelist));
```
cpvalente, line 16: not sure we need the `Object.freeze`. This hasnt been something we use in the rest of the codebase
If, for some reason, we expect that this might be mutated, we can throw a typescript `Readonly` on it so we get told off :D 

You can check the current enum validation in the `validateEvent.tsx` file

apps/server/src/api-integration/integration.controller.ts
alex-Arc: i seams that `isDataStale` is doing what we want but the name dose not imply that 
should we maybe rename it?

alex-Arc: > does the throttled message achieve anything

it's very helpful for testing, I could remove it now and return `undefined` so that the ws connation dosn't generate a reply

cpvalente: i think isDataStale is checking two objects against each other. in our case it would mean we iterate through the patch object twice.

Since we are already iterating, it may be simpler to extract the code that checks if a key is in the enum

alex-Arc: https://github.com/cpvalente/ontime/blob/8d526c7d31bd3bb765a633ba6030638b43758617/apps/server/src/services/rundown-service/rundownCacheUtils.ts#L121-L123

jwetzell (commented):
nothing sticks out to me!

cpvalente (approved):
Thank you. lets try this

PR #1011, opened 2024-05-27
fix: column reordering

cpvalente:
Fixes an issue with column reordering not working in cuesheets


alex-Arc (changes requested):
There seam to still be some problems

PR #1012, opened 2024-05-27
Clear upload folder on startup

alex-Arc:
should I incorporate some of the stuff from #904 ?


cpvalente:
> should I incorporate some of the stuff from #904 ?

I dont think so. That is a complex task that should be done on its own.
We can do it, but we should not attempt to do two things at once

cpvalente (approved):
🚀

PR #1018, opened 2024-05-29
fix: correct key on shortcuts

cpvalente:
- fixes an issue with wrong arrow direction in shortcuts
- fixes issues with using arrows instead of shift key

I have decided to spell out the keys rather than having symbols. I believe this makes it clearer even it it is more noisy (maybe)

Screenshot 2024-05-29 at 21 59 39

@lukestein what do you think?

### DIFF
--- a/apps/client/src/common/utils/deviceUtils.ts
+++ b/apps/client/src/common/utils/deviceUtils.ts
@@ -3,6 +3,6 @@ export function isMacOS() {
   return userAgent.includes('macintosh') || userAgent.includes('mac os');
 }
 
-export const deviceAlt = isMacOS() ? '⌥' : 'Alt';
+export const deviceAlt = isMacOS() ? 'Option' : 'Alt';
 
-export const deviceMod = isMacOS() ? '⌘' : 'Ctrl';
+export const deviceMod = isMacOS() ? 'Cmd' : 'Ctrl';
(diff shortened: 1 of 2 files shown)


lukestein:
Personally, I prefer the symbols to the words

However, it seems like the old code was perhaps using the incorrect glyph for the shift key, which is why it looked particularly confusing with the arrows.

Up arrow should be `↑` (U+2191 Up Arrow)
Shift modifier should be `⇧` (U+21E7 Upwards White Arrow)

cpvalente:
> seems like the old code was perhaps using the incorrect glyph for the shift key, which is why it looked particularly confusing with the

@lukestein 
You are correct, this is what I was initially fixing.

I have tried using the shift key but the symbols seemed very similar to me
Screenshot 2024-05-29 at 22 23 36

What do you think?

apps/client/src/common/utils/deviceUtils.ts
```diff
@@ -3,6 +3,6 @@ export function isMacOS() {
   return userAgent.includes('macintosh') || userAgent.includes('mac os');
 }
 
-export const deviceAlt = isMacOS() ? '⌥' : 'Alt';
+export const deviceAlt = 'Alt';
```
lukestein, line 6: export const deviceAlt = isMacOS() ? 'Option' : 'Alt';

lukestein (commented):
Mac refers to Alt key as "Option"

lukestein:
Oh wow. Those shift symbols do render ugly in your screenshot. I note this is how Apple labels a keyboard shortcut with shift and an up arrow:
image

Text probably best in this case. But Mac calls it "option" not "Alt."

apps/client/src/common/utils/deviceUtils.ts
cpvalente: Good catch, thank you!

lukestein (approved):

alex-Arc (approved):

PR #1019, opened 2024-05-29
remove redundant path evaluation

jwetzell:
having `productionPath` and `dockerPath` is not necessary as they both resolve to the same path relative to `srcDirectory`

### DIFF
--- a/apps/server/src/setup/index.ts
+++ b/apps/server/src/setup/index.ts
@@ -59,16 +59,12 @@ const currentDir = dirname(__dirname);
 export const srcDirectory = isProduction ? currentDir : path.join(currentDir, '../');
 
 // resolve path to external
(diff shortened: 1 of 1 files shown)


cpvalente (approved):
Thank you @jwetzell 

The path setup has gotten out of hand and it would be good for everyone to clean this up.

There was a previous attempt here https://github.com/cpvalente/ontime/pull/904/commits/1abedd1e2765a567d380f68b258b2eacf6e33adb but I believe there was some issues that the change introduced.

Ontime really only has two main directories, the local path relative to the src and the user directory that is given by the OS. All other paths are relative to those

It would make sense to me that we clarify this situation and organise the paths so developers have a clearer picture of where the assets are.

As for this change, I am unclear on whether the paths are the same but I trust that you have verified. 

Thank you, good work 🚀

Issue #900, opened 2024-04-16
CSS: Improve style of MultiProgressBar


cpvalente:
We use a `MultiProgressBar` in several views.
https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.tsx

This component allows showing a progress bar which contains 3 subsections (default, warning, danger)

![Screenshot 2024-04-16 at 22 41 24](https://github.com/cpvalente/ontime/assets/34649812/17c84019-2322-4cfd-8a21-7e094a6769fc)

Unfortunately, because of the way the bar is composed with 3 elements, there seems to be some issue with the rendering of the element where all the borders align (right handside)
![Screenshot 2024-04-16 at 22 41 38](https://github.com/cpvalente/ontime/assets/34649812/962e8312-43fc-41bc-b1ed-71855f7e0fb4)

The issue does not impede any functionality, but it is ugly.
We hope that someone with better CSS skills than myself can help us find a solution

EDIT:
It could be that the composition of the components is making this unnecessarily difficult.
We could consider a larger change to the progress bar so that it is composed of different parts (which should get rid of the problem)
Something similar to what JIRA does
![Screenshot 2024-05-06 at 11 16 09](https://github.com/cpvalente/ontime/assets/34649812/c9788a25-add3-4897-8206-a73befbe6852)

cpvalente:
fixed and merged

PR #1022, opened 2024-05-30
fix for MultiPartProgressBar styling

hingobway:
**this is a fix for #900**

Hey, new user here really enjoying the app! I saw this issue was very similar to something I ran into recently, so throwing in this potential solution. As I understand it this has to do with sub-pixel rendering issues in the browser when using `overflow: hidden` to clip corners, where multiple elements being clipped always come out slightly differently due to anti-aliasing.

I reworked the HTML a little bit so that the colored backgrounds don't overlap, which solves most of this. But the indicator has to overlap with them to work, so I made it match the background color and go 1px beyond the edge of the status bar. This works, but should be noted it would look terrible on any other background color unless the component was fed the new background color as well. Hope that's still helpful!

### DIFF
--- a/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
+++ b/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
@@ -7,10 +7,6 @@ $progress-bar-br: 3px;
   position: relative;
   width: 100%;
   height: $progress-bar-size;
-  border-radius: $progress-bar-br;
-  background-color: var(--timer-progress-bg-override, $viewer-card-bg-color);
-  display: flex;
-  overflow: hidden;
 
   &--hidden {
     display: none;
@@ -18,31 +14,36 @@ $progress-bar-br: 3px;
   }
 }
 
+.multiprogress-bar__bg {
+  position: absolute;
+  inset: 0;
+  overflow: hidden;
+  border-radius: $progress-bar-br;
+
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+}
+
 .multiprogress-bar__indicator {
   position: absolute;
-  height: inherit;
-  background-color: black;
+  inset: 0;
+  margin: -1px;
+  margin-left: 0;
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+}
+
+.multiprogress-bar__indicator-bar {
+  background-color: $ui-black;
   opacity: 0.8;
   transition: 1s linear;
   transition-property: width;
-  right: 0;
 }
 
 .multiprogress-bar__bg-normal {
-  position: absolute;
-  height: inherit;
-  right: 0;
-  width: 100%;
+  flex: 1;
 }
-
-.multiprogress-bar__bg-warning {
-  position: absolute;
-  height: inherit;
-  right: 0;
-}
-
-.multiprogress-bar__bg-danger {
-  position: absolute;
-  height: inherit;
-  right: 0;
-}
+// .multiprogress-bar__bg-warning {}
+// .multiprogress-bar__bg-danger {}
--- a/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
+++ b/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
@@ -8,7 +8,7 @@ $progress-bar-br: 3px;
   width: 100%;
   height: $progress-bar-size;
 
-  &.hidden {
+  &--hidden {
     display: none;
     transition: display 0.5s;
   }
@@ -18,7 +18,7 @@ $progress-bar-br: 3px;
   position: absolute;
   inset: 0;
   overflow: hidden;
-  border-radius: $progress-bar-br;
+  border-radius: var(--progress-bar-br, $progress-bar-br);
 
   display: flex;
   flex-direction: row;
@@ -41,7 +41,7 @@ $progress-bar-br: 3px;
   transition: 1s linear;
   transition-property: width;
 
-  .ignore-css-override & {
+  .multiprogress-bar--ignore-css-override & {
     background-color: $ui-black;
   }
 }
@@ -49,8 +49,6 @@ $progress-bar-br: 3px;
 .multiprogress-bar__bg-normal {
   flex: 1;
 }
-
-// .multiprogress-bar__bg-warning {}
 
 .multiprogress-bar__bg-danger {
   flex-shrink: 0;
--- a/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
+++ b/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
@@ -45,5 +45,9 @@ $progress-bar-br: 3px;
 .multiprogress-bar__bg-normal {
   flex: 1;
 }
+
 // .multiprogress-bar__bg-warning {}
-// .multiprogress-bar__bg-danger {}
+
+.multiprogress-bar__bg-danger {
+  flex-shrink: 0;
+}
--- a/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
+++ b/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
@@ -8,7 +8,7 @@ $progress-bar-br: 3px;
   width: 100%;
   height: $progress-bar-size;
 
-  &--hidden {
+  &.hidden {
     display: none;
     transition: display 0.5s;
   }
@@ -40,6 +40,10 @@ $progress-bar-br: 3px;
   opacity: 0.8;
   transition: 1s linear;
   transition-property: width;
+
+  .ignore-css-override & {
+    background-color: $ui-black;
+  }
 }
 
 .multiprogress-bar__bg-normal {
--- a/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
+++ b/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
@@ -36,7 +36,7 @@ $progress-bar-br: 3px;
 }
 
 .multiprogress-bar__indicator-bar {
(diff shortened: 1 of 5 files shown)


alex-Arc:
Hey @hingobway nice work
It looks good in the default timer, but I think there are some problems when the the css overide is on

image
image

hingobway:
Aha, that’s what I meant by the background color thing. I’ll take a better look tomorrow when I have more time. Do you know the way you usually reference the current background color that includes the override? Fixing [this line](https://github.com/cpvalente/ontime/blob/e50d228718fbabd50f9f1b686fd61cb81a2a6181/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss#L39) to match should solve the problem.

cpvalente:
> Aha, that’s what I meant by the background color thing. I’ll take a better look tomorrow when I have more time. Do you know the way you usually reference the current background color that includes the override? Fixing [this line](https://github.com/cpvalente/ontime/blob/e50d228718fbabd50f9f1b686fd61cb81a2a6181/apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss#L39) to match should solve the problem.
> 
> (edit) I hadn’t read much about the CSS overrides, I assume it’s this? Will hold off on committing until I can properly test it
> 
> ```css-scss
> background: var(--background-color-override, $ui-black);
> ```

yes, this is correct, the variable `--background-color-override` will override our preset if it exists.
You should be able to set these in your browser

hingobway:
This should fix the override issue. Just want to flag for you that it does look a little different, but I think probably more aligned with what would be expected.

**old override look:** (black shading for non-indicator is forced)
old style

**new override look:** (shading follows background color)
new style

(the new override image is after setting a dark normalColor in settings as well.)

alex-Arc:
Very nice
This has now exposed another problem, in view that do not respect overrides
image

apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
```diff
@@ -7,42 +7,47 @@ $progress-bar-br: 3px;
...
-  height: inherit;
-  background-color: black;
+  inset: 0;
+  margin: -1px;
+  margin-left: 0;
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+}
+
```
alex-Arc, line 37: maybe it could be solved with something like
```css
.ignore-css-override {
  background-color: $ui-black;
}
```

apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.tsx
```diff
@@ -22,22 +22,26 @@ export default function MultiPartProgressBar(props: MultiPartProgressBar) {
...
+              className='multiprogress-bar__bg-warning'
+              style={{ width: `${warningWidth}%`, backgroundColor: warningColor }}
+            />
+            <div
+              className='multiprogress-bar__bg-danger'
+              style={{ width: `${dangerWidth}%`, backgroundColor: dangerColor }}
+            />
+          </div>
+          <div className='multiprogress-bar__indicator'>
+            <div className='multiprogress-bar__indicator-bar' style={{ width: `${percentRemaining}%` }} />
```
alex-Arc, line 43:
```suggestion
            <div className={`multiprogress-bar__indicator-bar ${ignoreCssOverride ?? 'ignore-css-override'}`} style={{ width: `${percentRemaining}%` }} />
```

hingobway:
That looks good to me. Is `ignoreCssOverride` something that is already accessible in context? Should it just be a component prop?

alex-Arc (approved):

apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
```diff
@@ -7,42 +7,51 @@ $progress-bar-br: 3px;
   position: relative;
   width: 100%;
   height: $progress-bar-size;
-  border-radius: $progress-bar-br;
-  background-color: var(--timer-progress-bg-override, $viewer-card-bg-color);
-  display: flex;
-  overflow: hidden;
 
-  &--hidden {
+  &.hidden {
```
cpvalente, line 11: question: is there a reason for this name change?
the CSS properties that we expose to client use the BEM convention.

Changing it will break users setup and our convention

apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
```diff
@@ -7,42 +7,51 @@ $progress-bar-br: 3px;
...
-  width: 100%;
+  flex: 1;
 }
 
-.multiprogress-bar__bg-warning {
-  position: absolute;
-  height: inherit;
-  right: 0;
-}
+// .multiprogress-bar__bg-warning {}
```
cpvalente, line 53: cleanup: can we remove the commented code?

apps/client/src/features/operator/status-bar/StatusBar.module.scss
```diff
@@ -105,5 +105,7 @@
 }
 
 .progressOverride {
-  border-radius: 0;
+  & :global(.multiprogress-bar__bg) {
+    border-radius: 0;
+  }
```
cpvalente, line 110: I am not sure breaking the encapsulation of the CSS modules is a good pattern to fall into. is there another alternative for us here?

cpvalente (changes requested):

apps/client/src/features/operator/status-bar/StatusBar.module.scss
hingobway: Good question, I wasn't sure the best way to expose border radius now that it is further into the DOM. In my testing it seemed pretty unavoidable that it had to be a layer down or the corner issue would come back.

As far as leakage goes the scope is still technically maintained here as the above compiles to 
```scss
.progressOverride-abc123 .multiprogress-bar__bg {}
``` 
but it's definitely not ideal so here are a few alternatives, tell me if you like any better or if anyone has a better idea.


1. Just make border radius a prop, I thought best to avoid that but could go that way if you want.

1. How about a css variable at the parent? Not sure if this is just more confusing to chase down, but could definititely go for something like...
    
  ```scss
  // StatusBar.module.scss
  .progressOverride {
    --border-radius: 0;
  }
  
  // MultiPartProgressBar.scss
  .multiprogress-bar__bg {
    border-radius: var(--border-radius, $progress-bar-br);
  }
  ```

apps/client/src/common/components/multi-part-progress-bar/MultiPartProgressBar.scss
hingobway: Nope I think I was just matching with what @alex-Arc wrote above but will switch both to BEM style

apps/client/src/features/operator/status-bar/StatusBar.module.scss
cpvalente: since the problem we are trying to solve is of styling, i believe the solution should not leak into the application code.
In that light, the CSS variable seems more indicated to me.

I am unsure on whether the common issues with CSS module loading order would end up being problematic for us, but assume that it may be a small edge case. Do you have experience with this?

On a side note, I am not actually sure what this code achieve, is it just a cleanup element for when the users have overridden the default colours?

We would ideally like to keep away from unnecessary complexity here. Do you think a solution that separates the elements as in the screenshot from JIRA in the ticket beneficial?
https://github.com/cpvalente/ontime/issues/900#issue-2246855661

hingobway: > since the problem we are trying to solve is of styling, i believe the solution should not leak into the application code. In that light, the CSS variable seems more indicated to me.

Sounds good!

> I am unsure on whether the common issues with CSS module loading order would end up being problematic for us, but assume that it may be a small edge case. Do you have experience with this?

For the variable method, my solution above would leave the css variable unset except when overridden like for the operator page, ie border radius would be set by the sass variable unless the css variable is defined. So I don't think module order should make a difference there as there will be no conflict and it should follow inheritance. I'll see if I can manage to test it.

> On a side note, I am not actually sure what this code achieve, is it just a cleanup element for when the users have overridden the default colours?

The CSS override code is just about matching the shading of the inactive part of the bar to the background behind, even when overridden. I didn't originally realize there were pages where the CSS override color was ignored, so that was the added fix. Even without the clipping issue, I would think it would be better for the bar to follow the background color this way. Would it make more sense to _opt in_ to the background override instead of opting out? (i.e. `allowCssOverrides`)

> We would ideally like to keep away from unnecessary complexity here. Do you think a solution that separates the elements as in the screenshot from JIRA in the ticket beneficial? [#900 (comment)](https://github.com/cpvalente/ontime/issues/900#issue-2246855661)

The Jira method is interesting, but I think it is more complex for this use case than needed because of the need for active time on top of the warning states. In order to avoid any div overlaps with this method, I think you'd have to further subdivide the sections (normal:past, normal:upcoming, warning:past, warning:upcoming, danger:past, danger:upcoming) and calculate the width for all 6 on each render. It is possible, but I think the current method is somewhat more flexible and future-proofed to allow for additional states or colors if needed. Does that sound sensible?

cpvalente: Hi @hingobway , I apologise for the slow replies and hope this does not make it seem that I am not appreciative of your contribution.

From what I gather, your solution with CSS custom properties is solid and you have a good understanding of the reasonings and the technology.
I would suggest we follow with that solution 

> Would it make more sense to opt in to the background override instead of opting out? (i.e. allowCssOverrides)

Honestly, it likely doesnt matter. The component has limited use cases and I fail to see how the choice of one or the other has a meaningful impact on the ergonomics of using it. If I were to flip a coin, I would be inclined to say that options should modify behaviour (since CSSOverride is a natural behaviour, we would disable it instead)

hingobway:
Busy few days, sorry! Code review changes are in, let me know if there's anything else!

cpvalente (approved):

Issue #1025, opened 2024-05-31
v 3.0.4 countdown skips the first second


craigreilly:
When starting a countdown, it seems to skip the first second, ie 10:00 to 9:58.

lukestein:
Replication:

https://github.com/cpvalente/ontime/assets/44452336/edb9a22c-13a6-4899-8d8d-03960b48e8b5

cpvalente:
Thank you @lukestein and @craigreilly 

I believe this was introduced in our recent attempts to make the timer more accurate. I am not immune to the irony here...

I will take a look at improving this, meanwhile:
Ontime updates at a minimum rate of 32ms (when nothing is happening), then we go through a bunch of filter to decide when to notify the views. We dont want to do this too much so try and minimise it to once a second.

What is likely happening is that we got some overzealous filter for the first update, should be a simple fix.
However, I believe this does not have any implications to the usage of the app, just looks strange..

cpvalente:
Hi @craigreilly , the new release v3.1.0 includes changes to address this issue.
Do you mind having a look so we can close the ticket?

craigreilly:
It does fix the previous error, though introduces a smaller concern.
If I am in an "event" and it is running and I click "Next", it almost skips the first second of the next timer.
ie: 10:00 and a 20:00 event.
Start running the 10.
Click "Next"
Display shows 20:00 for a millisecond before showing 19:59.

I am not sure this is really a problem. More cosmetic perhaps. I do not think anyone will notice except maybe the Op sitting in front of the laptop.

craigreilly:
This is a solid write up.I was ok with the fix today... but thought I should bring up the behavior just in case.

lukestein:
I think underlying design choice sounds solid and would **not** show subsecond frequencies anywhere.

Even if it's what's used internally, I think we don't want to give users the impression that they should be able to expect high-frequency accuracy (or even enter partial-second timers). And I wouldn't suggest increasing network traffic for this, either.

cpvalente:
Thank you both @craigreilly and @lukestein for your help and for your comments.

@craigreilly , could we close this ticket?

craigreilly:
Please do…

PR #1031, opened 2024-05-31
refactor: fix missed first update

cpvalente:
Fix issue described in #1025

### DIFF
--- a/apps/server/src/services/runtime-service/rundownService.utils.ts
+++ b/apps/server/src/services/runtime-service/rundownService.utils.ts
@@ -26,8 +26,7 @@ export function getShouldTimerUpdate(previousValue: number, currentValue: MaybeN
     return false;
   }
   // we avoid trigger ahead since it can cause duplicate triggers
-  // we force the timer value to be negative because we need a ceiling reduction
-  const shouldUpdateTimer = millisToSeconds(-currentValue) !== millisToSeconds(-previousValue);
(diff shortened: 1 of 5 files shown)


kev-ac:
I merged in these changes locally and it did fix the bug. One thing to note is that when coming from a stopped state the timer very very briefly shows the duration (ex. 00:40:00) and then jumps to 00:39:59. 
If a event is loaded and then played it immediately jumps to the next second.

While this possibly can be correctly calculated it looks weird in my opinion. It should either start on 00:39:59 or on 00:40:00 if this "second" stays for a whole second. 


https://github.com/cpvalente/ontime/assets/60135953/08a6a27b-983b-40ea-b7a2-f2fb9e8ed6be

apps/server/src/services/runtime-service/rundownService.utils.ts
```diff
@@ -26,8 +26,7 @@ export function getShouldTimerUpdate(previousValue: number, currentValue: MaybeN
     return false;
   }
   // we avoid trigger ahead since it can cause duplicate triggers
-  // we force the timer value to be negative because we need a ceiling reduction
-  const shouldUpdateTimer = millisToSeconds(-currentValue) !== millisToSeconds(-previousValue);
```
alex-Arc, line 30: with the way we push updates now we end up with a timer that is almost constantly 980ms ahead of time
as @kev-ac  notes this is the correct rounding
but since we trigger updates using `millisToSeconds`  that means that we now send updates exactly when the error is the biggest

PR #1032, opened 2024-05-31
fix: osc new project

### LINKED ISSUE #1026: OSC issue when creating a new project
If OSC was already on and accepting commands from previous project, then the state continues to be on and accepting commands when you start a new project. Then, when you close the software (QUIT) and Reopen it, it opens the previous project but the OSC status is not accepting commands because it is off.

1) Open a new project.
2) Enable QSC
3) Start a new Project
4) OSC appears to be off in the project settings, but will continue to accept commands
5) Quit the program completely
6) Restart ontime
7) OSC is OFF and must be turned on

Perhaps starting a new project would close the OSC port until it is enabled again.

alex-Arc:
closes #1026

i'm unsure on how the ProjectService works but I think this is the missing step?

### DIFF
--- a/apps/server/src/services/project-service/ProjectService.ts
+++ b/apps/server/src/services/project-service/ProjectService.ts
@@ -15,6 +15,7 @@ import { ensureDirectory, getFilesFromFolder, removeFileExtension } from '../../
 import { dbModel } from '../../models/dataModel.js';
 import { deleteFile } from '../../utils/parserUtils.js';
 import { switchDb } from '../../setup/loadDb.js';
+import { startIntegrations } from '../../app.js';
 
 // init dependencies
 init();
@@ -209,7 +210,7 @@ export async function applyDataModel(data: Partial<DatabaseModel>, _options?: Op
   // TODO: allow partial project merge from options
   const { rundown, customFields, ...rest } = data;
   const newData = await DataProvider.mergeIntoData(rest);
-
+  startIntegrations();
   if (rundown != null) {
     initRundown(rundown, customFields ?? {});
   }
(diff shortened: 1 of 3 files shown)


apps/server/src/services/project-service/ProjectService.ts
```diff
@@ -209,7 +210,7 @@ export async function applyDataModel(data: Partial<DatabaseModel>, _options?: Op
   // TODO: allow partial project merge from options
   const { rundown, customFields, ...rest } = data;
   const newData = await DataProvider.mergeIntoData(rest);
-
+  startIntegrations();
```
cpvalente, line 213: hmm, we cant really do this... or shouldnt at least.

inner service layers should not have dependencies on the outer layers

alex-Arc: Hmm
So we should trigger it from the project controller?

Or?

cpvalente: I think that would work. At least to maintain the flow inwards into the app

I think we already had similar problems that we solved? Like with restarting some services when the options change?

It sounds to me that when we apply a new project there will be several things to reset, so it could make sense to expose a pattern to reset the services

cpvalente:
> @cpvalente there seams to also be something wrong with the save button it stays available even after the settings are applied https://github.com/cpvalente/ontime/assets/19875125/47e2a2b2-b72f-4ab1-a62c-c6b0b0bb50e7

good catch, I believe I have made a fix

alex-Arc:
cool let's merge then?

PR #1036, opened 2024-06-02
refactor: declare timer phase

cpvalente:
This PR exposes a timer phase property on the running timer.

```ts
export enum TimerPhase {
  None = 'none',
  Default = 'default',
  Warning = 'warning',
  Danger = 'danger',
  Negative = 'negative',
}
```

Since we want to make the Danger and Warning thresholds events, it makes sense to calculate this in the server. Might as well expose it.

By exposing it we get the added advantage of providing better feedback to integrations.

Note to reviewer: `TimerPhase` is not a great name, suggestions are welcome


alex-Arc (approved):

alex-Arc:
TimerPhase is better than TimerZone that will be confused with time zone eg. UTC

PR #1037, opened 2024-06-02
feat:toggle index column

### LINKED ISSUE #899: Option to hide index column in cuesheet
In the current implementation of the cuesheet view, the first cue shows the event index.
This is the order of the event in the rundown

Screenshot 2024-04-16 at 22 31 39

This dataset is not used by all users and can be confused with the cue field

The issue task is to add an option in the cuesheet settings to hide the index column, similarly to how the other options are hidden.
Follow here for the entrypoint of the relevant code
https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/cuesheet/cuesheet-table-elements/EventRow.tsx#L58

asharonbaltazar:
closes #899

This PR adds an option to toggle the visibility of the index column:

https://github.com/cpvalente/ontime/assets/58940073/db306797-df4e-4eb7-ab3a-ba52487ed1f5

This approach doesn't use React Table's toggling implementation due to the difficulty of calculating the `eventIndex`. 

The `eventIndex` is incremented upon render and skips delaying or blocking blocks, making it hard to calculate the _ontime event index_ sum in the `cell` property when instantiating columns. I think to circumvent this we'd have to return `BlockRow`/`DelayRow` within each `cell`. 

Therefore, I've extended the existing `CuesheetSettings` store (which I see enables functionality outside of React Table's "control") to manage the toggle state

Would love to hear alternative approaches 👀

### DIFF
--- a/apps/client/src/features/cuesheet/cuesheet-table-settings/CuesheetTableSettings.tsx
+++ b/apps/client/src/features/cuesheet/cuesheet-table-settings/CuesheetTableSettings.tsx
@@ -41,7 +41,8 @@ function CuesheetTableSettings(props: CuesheetTableSettingsProps) {
         <div className={style.sectionTitle}>Toggle column visibility</div>
         <div className={style.options}>
           <label className={style.option}>
-            <Checkbox variant='ontime-ondark' defaultChecked={showIndexColumn} onChange={() => toggleIndexColumn()} />#
+            <Checkbox variant='ontime-ondark' defaultChecked={showIndexColumn} onChange={() => toggleIndexColumn()} />
+            Event Order
           </label>
           {columns.map((column) => {
             const columnHeader = column.columnDef.header;
--- a/apps/client/src/features/cuesheet/cuesheet-table-settings/CuesheetTableSettings.tsx
+++ b/apps/client/src/features/cuesheet/cuesheet-table-settings/CuesheetTableSettings.tsx
@@ -24,11 +24,13 @@ function CuesheetTableSettings(props: CuesheetTableSettingsProps) {
   const { columns, handleResetResizing, handleResetReordering, handleClearToggles } = props;
   const {
     followSelected,
+    showIndexColumn,
     toggleFollow,
     showPrevious,
     togglePreviousVisibility,
     showDelayBlock,
     showDelayedTimes,
+    toggleIndexColumn,
     toggleDelayedTimes,
     toggleDelayVisibility,
   } = useCuesheetSettings();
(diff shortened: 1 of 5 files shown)


apps/client/src/features/cuesheet/cuesheet-table-settings/CuesheetTableSettings.tsx
```diff
@@ -38,6 +40,9 @@ function CuesheetTableSettings(props: CuesheetTableSettingsProps) {
       <div className={style.leftPanel}>
         <div className={style.sectionTitle}>Toggle column visibility</div>
         <div className={style.options}>
+          <label className={style.option}>
+            <Checkbox variant='ontime-ondark' defaultChecked={showIndexColumn} onChange={() => toggleIndexColumn()} />#
```
cpvalente, line 44: suggestion: maybe we call this Index instead of `#`? what do you think?

cpvalente (commented):
Code looks good and the solution is appropriate. 

As you mention, the index column is currently handled outside the context of react-table, so this is what needs to be done

@alex-Arc is the one who requests the feature so I will let him take a look, but good work, happy to have your collaboration again! ⭐

alex-Arc:
very nice!
would it be possible to extend the full colour from the index onto the cue? 
image
This way we don't  loos the colour information on the playing event
It could be something that only happens when the index is hidden, but I think it should maybe be that way always since that is also the look in the editor rundown
image

apps/client/src/features/cuesheet/cuesheet-table-settings/CuesheetTableSettings.tsx
asharonbaltazar: I didn't mind, of course. I chose that name because the column header is named `#`, so inconsistent naming may be confusing for users

cpvalente: I agree. The # convention is common as an index in other apps, we chose it here as a way to keep the text small since we want the column to take as least space as possible.

I wonder whether having a name here would clarify the dataset for users. Potentially we can do `Index #`

I am not too sure either way so will leave up to your judgement

asharonbaltazar: > Index #

That's a good compromise :) 

Pushing a commit

asharonbaltazar: Resolving with https://github.com/cpvalente/ontime/pull/1037/commits/4961949ea0b0cb1a5fe2e3afd50eb9148a61399c

PR #1038, opened 2024-06-03
refactor: supress modifiers in count-up timers

cpvalente:
When a timer is marked as count-up I believe it makes no sense to show the warning, danger times and negative times modifier since we made we made a choice to hide this from the user

This PR changes the minimal and timer views to make sure that count-up timers dont have the colour modifiers

Screenshot 2024-06-03 at 20 44 56

@lukestein what do you think?

### DIFF
--- a/apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
+++ b/apps/client/src/features/viewers/minimal-timer/MinimalTimer.tsx
@@ -126,13 +126,14 @@ export default function MinimalTimer(props: MinimalTimerProps) {
 
   const isPlaying = time.playback !== Playback.Pause;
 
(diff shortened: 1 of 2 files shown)


lukestein:
Sorry I was slow on this

I can see the case for this change (I realize it was already pushed), but to be honest I somewhat preferred the old way. Just my one vote in favor of either undoing this or perhaps better making it a customization option on the `timer`.

This is actually a normal use case for me: I don't always want to tell a speaker—especially a senior one—"you have 15 minutes." Also, some people find countdown timers super stressful. But the color evolution from white to yellow to red signals that it's time things wrap up.

I'll also regularly run a count-up timer in meetings that don't necessarily have a pre-scheduled end time. If it goes yellow/red, that's a signal that we should find a good time to take a coffee/bathroom break.

PR #1046, opened 2024-06-04
Validate filename

alex-Arc:


apps/client/src/features/app-settings/panel/project-panel/ProjectCreateForm.tsx
```diff
@@ -50,7 +51,12 @@ export default function ProjectCreateForm(props: ProjectCreateFromProps) {
   const handleSubmitCreate = async (values: ProjectCreateFormValues) => {
     try {
       setError(null);
-      const filename = values.title?.trim();
+
+      const maybeFileName = values.title === '' || values.title === undefined ? 'untitled' : values.title.trim();
+
+      const filename = filenamify(`${maybeFileName}.json`, {
+        replacement: '_',
+      });
```
cpvalente, line 59: I am not really sure about the strategy here, few thoughts:
- we want to avoid adding libraries in the UI side. we can do some rough validation and let the backend deal with it
- why are we accepting falsy title values? I may be missing something, but is there a case where we want to let users not give us a name?

if we are to maintain this logic, since we are not interested in the specifics of the false state, we can simplify the logic to say "if no value is given, we call it untitled", like

```ts
const fileName = !values.title  ? 'untitled' : values.title.trim();
```

Also, since the variable always contains a filename, I wouldnt say it makes sense to call it maybeFileName

apps/server/package.json
```diff
@@ -21,6 +21,7 @@
     "ontime-utils": "workspace:*",
     "steno": "^3.1.0",
     "ts-essentials": "^9.4.1",
+    "valid-filename": "^4.0.0",
```
cpvalente, line 24: My initial suggestion was to use filenamify instead. I am skeptic of the small usage of this library.

express has millions of weekly downloads, how is everyone doing file name validation?

dependabot recommends https://www.npmjs.com/package/sanitize-filename

apps/server/src/api-data/db/db.validation.ts
```diff
@@ -1,16 +1,24 @@
...
   body('backstageUrl').optional().isString().trim(),
   body('backstageInfo').optional().isString().trim(),
   body('endMessage').optional().isString().trim(),
+  body('filename')
+    .exists()
+    .withMessage('Filename is required')
+    .isString()
+    .withMessage('Filename must be a string')
+    .custom(throwIsValidFilename)
+    .withMessage('Invalid file name'),
```
cpvalente, line 21: we can also add the trim() here

apps/server/package.json
alex-Arc: valid-filename and filenamify  are from the same people
 but [sanitize-filename](https://www.npmjs.com/package/sanitize-filename) looks good also, I will give that a try

apps/client/src/features/app-settings/panel/project-panel/ProjectCreateForm.tsx
alex-Arc: > why are we accepting falsy title values? I may be missing something, but is there a case where we want to let users not give us a name?

that was the previous implementation where users could leave the Title blank, I'm fine with forcing a file name

We need to do some sort of transformation if we want to allow users to use symbols like `>` in the title

cpvalente (approved):
Good work, thank you!

PR #1049, opened 2024-06-04
consider custom fields when loading operator main field

### LINKED ISSUE #1047: Operator view allows custom field as "main data field" but doesn't show it
Installation: 3.1.1 via Docker on Synology NAS

---

It seems that the customize panel for operator view allows you to select Title, Note, or a custom field as the "main data field."

However, if I choose the custom field, the Operator view actually shows the Title as the main text.

May want to remove this option and only allow Title or Note as main data field. (Or actually show chosen custom field?)

image

jwetzell:
The operator view had some different logic for loading the main field that did not consider custom fields. I have aligned the main field population with the secondary field. If this is not the desired behavior then a solution that handles custom fields should be proposed.

Closes #1047

### DIFF
--- a/apps/client/src/features/operator/Operator.tsx
+++ b/apps/client/src/features/operator/Operator.tsx
@@ -171,7 +171,7 @@ export default function Operator() {
               return null;
             }
 
-            const mainField = main ? entry?.[main] || entry.title : entry.title;
+            const mainField = getPropertyValue(entry, main) ?? '';
             const secondaryField = getPropertyValue(entry, secondary) ?? '';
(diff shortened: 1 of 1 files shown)


alex-Arc (approved):
I think this is correct

Should we maybe add similar to the timer view,
I don't see a big reason to prevent users from not using the custom fields there

jwetzell:
I don't see any issue it only adds functionality. The timer view does not currently have a main field just a secondary from what I can see.

cpvalente (approved):
good stuff 🚀

PR #1051, opened 2024-06-05
add a headless version of ontime

jwetzell:
- files are stored in the same location as the electron version
- currently bundles binaries for win/mac/linux and x64 architecture
- github action to build binaries and attach to release

### DIFF
--- a/.github/workflows/build_cli.yml
+++ b/.github/workflows/build_cli.yml
@@ -1,0 +1,49 @@
+name: Ontime CLI build
+
+on:
+  push:
+    tags: [ "*" ]
+  workflow_dispatch:
+
+jobs:
+  build_cli:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18.18.2
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v3
+        with:
+          version: 8
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build project packages
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+        run: pnpm build
+
+      - name: Copy server
+        run: mkdir -p apps/cli/server && cp apps/server/dist/index.cjs apps/cli/server/index.cjs
+      
+      - name: Copy client
+        run: cp -R apps/client/build apps/cli/client
+
+      - name: Copy external
+        run: cp -R apps/server/src/external apps/cli/external
+      
+      - name: Build CLI
+        run:  pnpm build:cli
+
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: './apps/cli/dist/*'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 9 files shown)


alex-Arc (approved):
LGTM
I don't have much to add to this
But let me know if there is a test build you want me to run

jwetzell:
@alex-Arc I did publish a test version to NPM so on a system with Node >= 18 you can run `npx @getontime/cli@dev` to test.

.github/workflows/build_cli.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          version: 8
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build project packages
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+        run: pnpm build
```
cpvalente, line 34: question: what do you think about adding an environment variable here so that we know that the app is running from a CLI build

apps/cli/main.js
```diff
@@ -0,0 +1,20 @@
...
+  await startServer();
+
+  await startIntegrations();
+}
+
+startOntime();
+
+process.on('SIGINT',()=>{
+  shutdown()
+})
```
cpvalente, line 20: questions: 
- I think the formatting here is not following the prettier rules
- should we consider other shutdown signals? the app handles `SIGHUP` `SIGINT` and `SIGTERM`
- should we improve log to user (simple console log for now like:
```suggestion
process.on('SIGINT', () => {
  console.log('User requested Ontime shutdown');
  shutdown();
})
```

apps/cli/main.js
```diff
@@ -0,0 +1,20 @@
...
+
+const ontimeServer = require('./server/index.cjs');
+const { initAssets, startServer, startIntegrations, shutdown } = ontimeServer;
+
+async function startOntime() {
+  await initAssets();
+
+  await startServer();
+
+  await startIntegrations();
```
cpvalente, line 13: question: should we provide some feedback here?
This is what we get in electron
```ts
    console.log('Request: Initialise assets...');
    await initAssets();
    console.log('Request: Start server...');
    await startServer();
    console.log('Request: Start integrations...');
    await startIntegrations();
```

jwetzell: All logging should go through the logger in the server so I would add these there and go through the logger. No more console.log

.github/workflows/build_cli.yml
jwetzell: What would be consuming the env variable?

cpvalente: We would use it in our setup file to produce a isCLI constant. That would be used by services that benefit from knowing this. For now, the logger

jwetzell: I think you would want something that isn't override-able by the person running ontime.

cpvalente: If we can find another way that is as convenient and safer it's all good. I don't have strong opinion don't he details of how this should work 

However, I would not rate the fact that users can override it as an important requirement

jwetzell: I just meant that the const's like isDocker/isProduction/etc. are not baked in and I could start up the electron app in a way that it thinks it "is Docker" just by setting an env variable on my system. Using an env variable means we would have to get the user to run ontime with that env set. I'll try and see if there is something easy I can do to know if the server is running headless or not.

apps/cli/main.js
jwetzell: I will leave this for the broader logging cleanup. I think these should be on the server to log out.

.github/workflows/build_cli.yml
jwetzell: I think I will again leave this to the console logger phase. Still not sure of a good way for the headless version to know it is headless. Maybe a optional argument to `startServer` or something.

PR #1052, opened 2024-06-05
fix and test only space osc payload

alex-Arc:


jwetzell (approved):
Good edge case catch. Looks good to me!

PR #1053, opened 2024-06-06
Fix: link events in sheet when there is nothing to link

### LINKED ISSUE #1048: Link start can cause sheet import to fail
"Link start" is True but there's no (unskipped) event to set the link start to, the import gives a confusing error:
image

_Originally posted by @lukestein in https://github.com/cpvalente/ontime/issues/971#issuecomment-2127675647_

alex-Arc:
closes #1048

### DIFF
--- a/apps/server/src/utils/__tests__/parserFunctions.test.ts
+++ b/apps/server/src/utils/__tests__/parserFunctions.test.ts
@@ -1,5 +1,21 @@
-import { CustomFields, HttpSubscription, OscSubscription } from 'ontime-types';
-import { sanitiseCustomFields, sanitiseHttpSubscriptions, sanitiseOscSubscriptions } from '../parserFunctions.js';
+import {
+  CustomFields,
+  DatabaseModel,
+  EndAction,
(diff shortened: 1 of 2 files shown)


apps/server/src/utils/__tests__/parserFunctions.test.ts
```diff
@@ -155,3 +171,119 @@ describe('sanitiseCustomFields()', () => {
...
+
+  it('returns linked events', () => {
+    const data: Partial<DatabaseModel> = {
+      rundown: [
+        // @ts-expect-error - only the necessary parts
+        {
+          id: '1',
+          type: SupportedEvent.Event,
+          skip: false,
+        },
```
cpvalente, line 206: suggestion: using casting would at least make sure you dont pass wrong data. I think it is more productive, like

```suggestion
        {
          id: '1',
          type: SupportedEvent.Event,
          skip: false,
        } as OntimeEvent,
```

cpvalente (approved):
looks good, lets just cleanup the ts directives and merge

PR #1058, opened 2024-06-08
refactor: improve error states in spreadsheet flow

cpvalente:
This PR adds better handling of error states in the google sheet flow


alex-Arc:
good
But we then need to clear the errors, so that when the user redose the steps the old errors are not blocking them

alex-Arc (approved):

PR #1059, opened 2024-06-08
refactor: improve console visibility

cpvalente:
since so many users are consuming Ontime headless, I thought we should look into improving the visibility of messages in terminal

In this PR I am just adding visibility for critical operations
- where is ontime running
- did ontime crash

TODO:
- [x] style console to help visibility 
- [x] create a SEVERE log level for unrecoverable datasets
- [x] SEVERE logs should be escalated to electron process
- [x] clarify app init in console to distinguish across boots

Suggestions for improvement are welcome

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -81,6 +81,9 @@ From the project root, run the following commands
 
 The build distribution assets will be at `.apps/electron/dist`
 
+Note: The MacOS build will only work in CI, locally it will fail due to notarisation issues.
+Use the `turbo dist-mac:local` command to build a MacOS distribution locally.
+
 ## DOCKER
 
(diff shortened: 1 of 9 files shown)


alex-Arc (approved):
Good start
I guess we will the just slowly move stuff over to these

jwetzell (approved):
Yep good stuff. I haven't explored the code base but utilizing different log levels would be nice to support. The ability to start it up with more or less logging for troubleshooting or other reasons.

cpvalente:
> Yep good stuff. I haven't explored the code base but utilizing different log levels would be nice to support. The ability to start it up with more or less logging for troubleshooting or other reasons.

Thank you,

I have created infrastructure to add log levels, I suggest we add an environment level that activates the `canLog` flags.
We should look into it with the CLI tasks.

PR #1062, opened 2024-06-10
Corrupted parser

cpvalente:
# THIS PR WILL NOT BE MERGED

This PR tries to improve the use case where Ontime project files are corrupt.

Currently we read what we can and override the rest. We propose that on encountering a a) file with corrupt data or b) a file that cant be read, we move that file to a corrupted files folder. This aims to avoid losing user data, but maintaining safety in the data Ontime consumes

These are the following handled use cases:

### App state does not exist
- we assume that it is the first load
- - create a new project file from demo
- - point `AppStateService` to the new created file

### Last loaded project cannot be found
- Ontime creates a new file "new project.json" from the empty data model. 
- The `AppStateService` is updated to the new created file

### Project cannot be loaded (eg. invalid JSON)
- Ontime creates a new file "new project.json" from the empty data model. 
- we copy the corrupt file to `corrupted files` directory
- The `AppStateService` is updated to the new created file

### Project contains corrupt data
- Ontime parses the data for the individual objects. If corrupted data is found:
- - loaded file is copied to the `corrupted files` directory
- - a new copy of the file is created with the recovered data
- - `AppStateService` shows the loaded file is the recovered file


apps/server/src/setup/loadDb.ts
```diff
@@ -44,35 +47,49 @@ const populateDb = (directory: string, filename: string): string => {
 };
 
 /**
- * @description parses a json file to the adapter
- * It will create an empty file from the model if the parsing fails
+ * Handles a corrupted fle by copying it to a corrupted folder
```
alex-Arc, line 50:
```suggestion
 * Handles a corrupted file by copying it to a corrupted folder
```

apps/server/src/setup/loadDb.ts
```diff
@@ -44,35 +47,49 @@ const populateDb = (directory: string, filename: string): string => {
...
+    const result = parseJson(maybeProjectFile);
+    newData = result.data;
+    errors = result.errors;
+  } catch (error) {
+    consoleError(`Unable to parse project file: ${getErrorMessage(error)}`);
+    handleCorruptedDb(dbInDisk);
+  } finally {
+    if (errors.length > 0) {
+      handleCorruptedDb(dbInDisk);
+    }
```
alex-Arc, line 86: I think somewhere here we should the load the recovered file?
Right now it looks like we stay on the broken one

https://github.com/cpvalente/ontime/assets/19875125/c784a92a-57d0-4e43-a85a-a84c047ed600

apps/server/src/setup/loadDb.ts
```diff
@@ -44,35 +47,49 @@ const populateDb = (directory: string, filename: string): string => {
...
-    // this will throw if file is not valid
-    parseProjectFile(fileToRead);
-    await adapterToUse.read();
-  } catch (error) {
-    adapterToUse.data = dbModel;
+    const fileName = getFileNameFromPath(filePath);
+    const newFilePath = join(resolveCorruptedFilesDirectory, fileName);
+
+    ensureDirectory(resolveCorruptedFilesDirectory);
+    copyFileSync(filePath, newFilePath);
```
alex-Arc, line 59: Should we maybe move the broken file or delete it after copy so it doesn't stay in the list?

alex-Arc (changes requested):
Could we maybe also add a flag to the return message of `loadProject` 
that could trigger an alert popup in the UI to say "Hey this file is broken we moved a copy to ../.. and recovered as much as we could"

apps/server/src/setup/loadDb.ts
cpvalente: We basically continue working on the "broken" file.
We rename it to have the (recovered) append and overwrite its contents

alex-Arc (approved):

apps/server/src/services/project-service/ProjectService.ts
```diff
@@ -89,17 +94,29 @@ export async function getProjectFiles(): Promise<ProjectFile[]> {
...
+  const appState = await appStateService.get();
+
+  if (appState.lastLoadedProject) {
+    const filePath = join(resolveProjectsDirectory, appState.lastLoadedProject);
+    if (!checkIfFileExists(filePath)) {
+      await appStateService.updateDatabaseConfig('');
+      return '';
+    }
+  }
+  return appState.lastLoadedProject;
```
cpvalente, line 107: question: i wonder if we can stop it from happening here. 
Checking for loaded project should happen a) when a project is loaded b) when the app is started

From there we just get the state from the appStateService.
I dont see a good scenario for receiving an empty string here

apps/server/src/services/project-service/projectFileUtils.ts
```diff
@@ -14,18 +14,18 @@ export function filterProjectFiles(files: Array<string>): Array<string> {
...
+ * @returns
+ */
 export function parseProjectFile(filePath: string): object {
   if (!filePath.endsWith('.json')) {
     throw new Error('Invalid file type');
   }
 
   const rawdata = readFileSync(filePath, 'utf-8');
-  const uploadedJson = JSON.parse(rawdata);
+  const jsonData = JSON.parse(rawdata);
```
cpvalente, line 28: I wonder if we will think many times that this needs to be caught, we need a comment here to mention that the try-catch is on the consumer. maybe add an @throws directive in JSDocs

apps/server/src/setup/index.ts
```diff
@@ -81,9 +84,25 @@ export const resolveExternalsDirectory = join(isProduction ? getAppDataPath() :
...
+export const pathToStartDb = isTest
+  ? join(srcDirectory, '..', config.database.testdb, config.database.filename)
+  : join(srcDirectory, '/preloaded-db/', config.database.filename);
+export const resolveDbDirectory = join(testDbStartDirectory, isTest ? `../${config.database.testdb}` : config.projects);
+
+const ensureAppState = (firstStartup = false) => {
   ensureDirectory(getAppDataPath());
-  fs.writeFileSync(appStatePath, JSON.stringify({ lastLoadedProject: 'db.json' }));
+  if (firstStartup) {
+    fs.writeFileSync(appStatePath, JSON.stringify({ lastLoadedProject: DEMO_PROJECT }));
```
cpvalente, line 96: cleanup: lets import just the functions instead of the entire fs module

apps/server/src/setup/index.ts
```diff
@@ -93,24 +112,21 @@ const getLastLoadedProject = () => {
       ensureAppState();
     }
     return appState.lastLoadedProject;
-  } catch {
+  } catch (_) {
     if (!isTest) {
-      ensureAppState();
+      consoleHighlight('No app state found, most lilky first startup');
```
cpvalente, line 117: cleanup, mind typo, also this log likely doesnt need to exist outside debugging

apps/server/src/setup/loadDb.ts
```diff
@@ -21,58 +24,68 @@ const populateDb = (directory: string, filename: string): string => {
...
       }
-
-      const newFileDirectory = join(dbDirectory, startDbName);
-
-      copyFileSync(pathToStartDb, newFileDirectory);
+      writeFileSync(appStatePath, JSON.stringify({ lastLoadedProject: 'new empty project.json' }));
       dbPath = newFileDirectory;
     } catch (_) {
       /* we do not handle this */
     }
```
cpvalente, line 40: question: if we dont catch this, can the app work at all? should we instead just let the error bubble up so the app can crash with an uncaught error?

cpvalente:
Feature has been merged, closing for now

PR #1063, opened 2024-06-11
None colour from API

alex-Arc:
Allow the api to set the event colour to '' by calling 'none'
Otherwise there is no way for the api to match the UI button

image

### DIFF
--- a/apps/server/src/utils/coerceType.ts
+++ b/apps/server/src/utils/coerceType.ts
@@ -90,7 +90,12 @@ export function coerceColour(value: unknown): string {
     if (!isColourHex(lowerCaseValue)) {
       throw new Error('Invalid hex colour received');
     }
-  } else if (!(lowerCaseValue in cssColours)) {
+    return lowerCaseValue;
+  }
+  if (lowerCaseValue === 'none') {
+    return ''; // None colour the same as the UI 'Ø' button
+  }
+  if (!(lowerCaseValue in cssColours)) {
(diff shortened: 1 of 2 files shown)


apps/server/src/utils/coerceType.ts
```diff
@@ -90,7 +90,12 @@ export function coerceColour(value: unknown): string {
     if (!isColourHex(lowerCaseValue)) {
       throw new Error('Invalid hex colour received');
     }
-  } else if (!(lowerCaseValue in cssColours)) {
+    return lowerCaseValue;
+  }
+  if (lowerCaseValue === 'none') {
+    return ''; // None colour the same as the UI 'Ø' button
+  }
```
cpvalente, line 97: why use a 'none' string instead of accepting a falsy value like `null` or `''`

alex-Arc: We could do that 
was thinking this was good be course it would reject accidental empty values
Do you think it's better to leave it loose?

cpvalente: You are correct that you should be careful with undefined, since that would mean that we could accidentally override the value if property is not in the patch

But if we can verify that the property exists, I think it is best to accept `''` since that is what our data looks like. I believe that translating `'none'` to meaning `''` becomes a bit of a secret handshake and adds more complexity

cpvalente (approved):

PR #1064, opened 2024-06-11
Fix coerce enum

alex-Arc:


apps/server/src/utils/__tests__/coerceType.test.ts
```diff
@@ -22,9 +22,9 @@ describe('parses a colour string that is', () => {
 
 describe('match a string to an enum that is', () => {
   enum testEnum {
-    'abc',
-    'def',
-    'ghi',
+    ABC = 'abc',
```
cpvalente, line 25: just need to fix the expected error
i wonder if the error is important at all, since it is not escalated i think that expecting `toThrow` is good enough

cpvalente (approved):

PR #1071, opened 2024-06-13
add warning and danger transitions as timer life cycle events

jwetzell:
- onWarning when timer transitions to warning phase
- onDanger when timer transitions to danger phase

### DIFF
--- a/apps/server/src/services/runtime-service/RuntimeService.ts
+++ b/apps/server/src/services/runtime-service/RuntimeService.ts
@@ -94,7 +94,7 @@ class RuntimeService {
       this.eventTimer.roll(rundown);
     }
 
-    const timerPhaseChanged = RuntimeService.previousState.timer.phase !== newState.timer.phase;
+    const timerPhaseChanged = RuntimeService.previousState.timer?.phase !== newState.timer.phase;
 
     if (timerPhaseChanged) {
       switch (newState.timer.phase) {
--- a/apps/server/src/services/runtime-service/RuntimeService.ts
+++ b/apps/server/src/services/runtime-service/RuntimeService.ts
@@ -1,4 +1,13 @@
-import { EndAction, LogOrigin, MaybeNumber, OntimeEvent, Playback, RuntimeStore, TimerLifeCycle, TimerPhase } from 'ontime-types';
+import {
+  EndAction,
+  LogOrigin,
+  MaybeNumber,
+  OntimeEvent,
+  Playback,
+  RuntimeStore,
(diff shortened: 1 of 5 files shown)


apps/server/src/services/runtime-service/RuntimeService.ts
```diff
@@ -93,6 +93,25 @@ class RuntimeService {
       const rundown = getPlayableEvents();
       this.eventTimer.roll(rundown);
     }
+
+    const timerPhaseChanged = RuntimeService.previousState.timer.phase !== newState.timer.phase;
```
cpvalente, line 97: On the first time the app runs, `RuntimeService.previousState` is an empty object

we need to guard this check otherwise the app will crash on the first start
something like

```suggestion
    const timerPhaseChanged = RuntimeService.previousState.timer?.phase !== newState.timer.phase;
```

You can recreate this by navigating to the user install folder (for me in macos `/Users/carlos/Library/Application Support/Ontime/`) and deleting the ontime.restore file

On crash a crash log is made in the same folder, mine contains

```ts
  "error": {
    "message": "Cannot read properties of undefined (reading 'phase')",
    "stack": 

"TypeError: 
    Cannot read properties of undefined (reading 'phase')\n    
    at RuntimeService.checkTimerUpdate (file:///Users/carlos/ontime/apps/server/src/services/runtime-service/RuntimeService.ts:107:66)\n    <---------
    at RuntimeService.descriptor.value (file:///Users/carlos/ontime/apps/server/src/services/runtime-service/RuntimeService.ts:554:35)\n   
    at TimerService.onUpdateCallback (file:///Users/carlos/ontime/apps/server/src/services/runtime-service/RuntimeService.ts:134:48)\n    
    at TimerService.update (file:///Users/carlos/ontime/apps/server/src/services/TimerService.ts:92:10)\n   
    at Timeout._onTimeout (file:///Users/carlos/ontime/apps/server/src/services/TimerService.ts:35:12)\n   
    at listOnTimeout (node:internal/timers:569:17)\n    
    at processTimers (node:internal/timers:512:7)"
  },
```

cpvalente (commented):
Good work, just need to clean up a small guard and I think it seems to work well otherwise

apps/server/src/services/runtime-service/RuntimeService.ts
jwetzell: Good catch I trusted the typing too much 😂

alex-Arc (approved):
Nice

PR #1099, opened 2024-06-23
events with 0 duration dose not contribute to next day calculation

alex-Arc:


apps/server/src/services/rundown-service/__tests__/rundownCache.test.ts
```diff
@@ -170,6 +170,19 @@ describe('generate()', () => {
     expect(initResult.totalDuration).toBe(500 - 100);
   });
 
+  it('calculates total duration with 0 durration events not caussing a next day', () => {
```
cpvalente, line 173:
```suggestion
  it('calculates total duration with 0 duration events without causing a next day', () => {
```

cpvalente (approved):
Good stuff, we just clarify the test name and are good to merge

PR #1100, opened 2024-06-23
add relevantBlock function

alex-Arc:
Adds  block info to the runtimeStore

```ts
{
"block": {"title":"<Title of the current block>","type":"block","id":"<id of the current block>"} || null,
"startedAt": <time in ms whan thsi block was enterd> || null
}
```


packages/utils/src/rundown-utils/rundownUtils.ts
```diff
@@ -326,3 +326,29 @@ export const swapEventData = (eventA: OntimeEvent, eventB: OntimeEvent): { newA:
...
+
+/**
+ * Gets relevant block element for a given ID
+ * @param rundown
+ * @param order
+ * @param {string} currentId
+ * @return {OntimeBlock | null}
+ */
+export function getRelevantBlock(rundown: OntimeRundown, currentId: string): OntimeBlock | null {
+  const index = rundown.findIndex(({ id }) => id === currentId);
```
cpvalente, line 338: we can avoid iterating twice through the data by doing this in the for loop

packages/utils/src/rundown-utils/rundownUtils.ts
```diff
@@ -326,3 +326,29 @@ export const swapEventData = (eventA: OntimeEvent, eventB: OntimeEvent): { newA:
...
+ * @param rundown
+ * @param order
+ * @param {string} currentId
+ * @return {OntimeBlock | null}
+ */
+export function getRelevantBlock(rundown: OntimeRundown, currentId: string): OntimeBlock | null {
+  let now = false;
+  for (let i = rundown.length - 1; i >= 0; i--) {
+    const entry = rundown[i];
+    if (!now && entry.id == currentId) {
```
cpvalente, line 341: Tip: remember the ===, if nothing else, just to keep convention

packages/utils/src/rundown-utils/rundownUtils.ts
```diff
@@ -326,3 +326,29 @@ export const swapEventData = (eventA: OntimeEvent, eventB: OntimeEvent): { newA:
...
+      continue;
+    }
+    //the first block before 'now' is the relevant one
+    if (now && isOntimeBlock(entry)) {
+      return entry;
+    }
+  }
+
+  //no blocks exist before given id
+  return null;
```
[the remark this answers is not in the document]
cpvalente: I think the most relevant improvement from this version is that we simplify the code path without the continue

apps/client/src/common/utils/socket.ts
```diff
@@ -150,6 +150,12 @@ export const connectSocket = () => {
           updateDevTools({ eventNow: payload });
           break;
         }
+        case 'ontime-blockState': {
+          console.log(payload)
```
cpvalente, line 154: Remember to check and remove console logs

apps/client/src/common/stores/runtime.ts
```diff
@@ -38,6 +38,10 @@ export const runtimeStorePlaceholder: RuntimeStore = {
     actualStart: null,
     expectedEnd: null,
   },
+  blockState: {
```
cpvalente, line 41: Our names here may not be ideal, but I am not sure `blockState` really works with the rest 

How about `currentBlock`?

apps/server/src/services/TimerService.ts
```diff
@@ -96,8 +96,8 @@ export class TimerService {
    * Loads roll information into timer service
    * @param {OntimeEvent[]} rundown -- list of events to run
    */
-  roll(rundown: OntimeEvent[]) {
-    runtimeState.roll(rundown);
+  roll(playableEvents: OntimeEvent[], rundown: OntimeRundown) {
```
cpvalente, line 99: Question: I am not convinced this is something we should keep.

Historically we had problems with having the wrong dataset.

The playable events were supposed to be a cached, filtered version of the rundown.
At this point, I wonder if it is causing more issues than the convenience it brings.
This is especially true if we start using both the filtered (playableEvents) and unfiltered (rundown) datasets

I think the best is to extract utilities to filter the rundown and always pass the entire rundown around 

What do you think?

apps/client/src/common/stores/runtime.ts
alex-Arc: Do you have a good name for the type containing the block and extra info
maybe the type can stay `BlockState`  ?

cpvalente: I think it is easier if the type name matches the object, so for `currentBlock` we would have `CurrentBlock` or `CurrentBlockState`

alex-Arc: I what to avoid the confusion that BurrentBlock is only the block
as there is also other data in there

apps/server/src/services/TimerService.ts
alex-Arc: yeah that is a bit harry.
Maybe we could use the Normalised Rundown 
and maybe extend the it with a list of timedEntrys 
like:
```ts
type RundownCache = {
  rundown: NormalisedRundown;
  order: string[];
  timedEvents: string[];
  revision: number;
  totalDelay: number;
  totalDuration: number;
};

``` 

or something like it?

cpvalente: I am not sure. Let's chat

cpvalente:
Hi @alex-Arc , what is the status here?

apps/server/src/stores/runtimeState.ts
```diff
@@ -165,8 +190,14 @@
...
 
-export function loadNow(event: OntimeEvent, playableEvents: OntimeEvent[]) {
+export function loadNow(event: OntimeEvent, rundown: OntimeRundown) {
   runtimeState.eventNow = event;
+  runtimeState.currentBlock.block = getRelevantBlock(rundown, event.id);
+
+  //if we are still in the same block keep the startedAt time
+  if (runtimeState._prevCurrentBlock?.block?.id === runtimeState.currentBlock.block.id) {
+    runtimeState.currentBlock.startedAt = runtimeState._prevCurrentBlock.startedAt;
+  }
```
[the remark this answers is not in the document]
cpvalente: this seems correct, we dont necessarily have a block in state

cpvalente (approved):
I think this looks good, we need to find a way to show this in the interface (maybe the overview panel?) but lets do that in another PR

PR #1106, opened 2024-06-26
add freeze feature to rundown

jwetzell:


apps/server/src/app.ts
```diff
@@ -188,6 +188,7 @@ export const startServer = async (
       playback: SimplePlayback.Stop,
       direction: SimpleDirection.CountDown,
     },
+    frozen: false,
```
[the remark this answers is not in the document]
jwetzell: LOL what?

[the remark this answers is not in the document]
jwetzell: Silly rabbit that variable has not been declared yet. `startServer` doesn't have a parameter called `frozen`.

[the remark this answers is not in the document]
jwetzell: no thanks

apps/server/src/api-data/rundown/rundown.controller.ts
```diff
@@ -69,6 +70,21 @@ export async function rundownBatchPut(req: Request, res: Response<MessageRespons
   }
 }
 
+export async function rundownFrozenPost(req: Request, res: Response<MessageResponse | ErrorResponse>) {
+  if (failEmptyObjects(req.body, res)) {
+    return res.status(404);
+  }
```
cpvalente, line 76: question: would this not have been taken care of by the validator?

apps/server/src/api-data/rundown/rundown.controller.ts
```diff
@@ -69,6 +70,21 @@ export async function rundownBatchPut(req: Request, res: Response<MessageRespons
...
+
+  try {
+    const { state } = req.body;
```
cpvalente, line 79: question: could this name be more descriptive? even if it is just `frozen`?

apps/server/src/api-data/rundown/rundown.controller.ts
```diff
@@ -69,6 +70,21 @@ export async function rundownBatchPut(req: Request, res: Response<MessageRespons
...
+    setFrozenState(state);
+    res.status(200).send({ message: `Rundown frozen state updated.` });
```
cpvalente, line 81: suggestion: no need for template literals here

cpvalente (commented):
Code looks good, I made some suggestions that you can consider

I will apply the frontend part and see how the feature behaves.
Thank you for your help

apps/server/src/api-data/rundown/rundown.controller.ts
jwetzell: Most of the other functions here had that check at the top so I just followed suit.

cpvalente: ah I see.
There is a case where the other code is also wrong,

But in some cases, all the elements of the body are optional, so the validation could potentially pass with an empty object

jwetzell: Fair. I've cleaned that out and this should be good to go now.

cpvalente:
Thank you @jwetzell , I will continue the PR with the client implementation

PR #1115, opened 2024-07-02
switch workflows to release trigger

### LINKED ISSUE #1107: Dockerhub "latest" includes alpha release
I'm not sure what the process is that gets new GitHub releases up to dockerhub, but would suggest setting tags so that `latest` points to the newest production release, not (as is currently the case) the alpha build

jwetzell:
- published type is any published release (pre-release/latest)
- the release [object](https://docs.github.com/en/webhooks/webhook-events-and-payloads#release) includes the tag name so no need to extract it anymore

Closes #1107

### DIFF
--- a/.github/workflows/build_cli.yml
+++ b/.github/workflows/build_cli.yml
@@ -1,8 +1,8 @@
 name: Ontime CLI build
 
 on:
-  push:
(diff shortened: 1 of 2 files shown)


cpvalente (approved):
Good work

Issue #1007, opened 2024-05-27
Improvements to countdown view


cpvalente:
**User request**

Currently the main timer is the time to the scheduled time, I would propose the following, where the main countdown shows the time to projected start, and the projected start and end are shown as well. The offset itself would not have to be shown, but be used in calculation.

![Screenshot 2024-05-27 at 08 57 19](https://github.com/cpvalente/ontime/assets/34649812/2d6d36cb-9c31-4572-8ec0-bc125f37df61)

The task is to add `Projected Start` and `Projected End` to the `/countdown` view.
We would also need to modify the timer to account for the offsets

This is a good opportunity to look into the bigger picture for this view. I believe @alex-Arc had some suggestions here

cpvalente:
Hi @calebwest-SS 

I know this has been a while in the making, we are reviewing the view features and have made changes to the countdown following this feedback

I am attaching here a design reference, your feedback is welcome
![Image](https://github.com/user-attachments/assets/2cad2160-d9bb-408a-bfd8-5eba3990b2e2)

cpvalente:
Closing this as completed and released with Ontime v4

PR #1124, opened 2024-07-07
Countdown

cpvalente:
This PR adds the ability to show projected times in the countdown

- The projected times are disabled by default but can be enabled in the menu
- The new section (Projected times) shows the time + offset + delay
- The existing section (Scheduled times) shows time + delay
- if the projected times option is enabled, the main countdown adds the offset to the timers

I have also cleaned up the effects to allow users to clear the selection by refreshing the page

Partially implements #1007 
The part described as Kitchen view is not implemented and should be extracted to a new ticket

Note to reviewer:
There are a lot of changes here, mostly in code organisation and style.
We just need to assert that applying the setting shows the offset times


alex-Arc (approved):
seams to work good 👍🏽

PR #1129, opened 2024-07-10
Rundown api

cpvalente:
This PR exposes the list of changes in the refetch message.
This means that consumers can know which events are affected by a change

In addition, we also add a few endpoints for granular rundown getting
- paginated query params for rundown
- getEventById


jwetzell (approved):
I tested the pagination/limit stuff and then fetching by id and everything seemed to behave as expected and return the correct data

PR #1134, opened 2024-07-13
switch Dockerfile build stage to non-alpine image

jwetzell:
Seems to be some [known issue](https://github.com/nodejs/docker-node/issues/1946) with alpine images and node >= 20 this is breaking the docker builder stage. This switches the builder stage to a debian based node image. I have tested this is working now with multi-platform builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20-alpine AS builder
+FROM node:20-bullseye AS builder
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 RUN corepack enable
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:20-bullseye AS builder
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
-RUN corepack enable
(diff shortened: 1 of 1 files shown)


cpvalente (approved):
It is still unclear to me what are we gaining with corepack.

Does it mean we don't need to have NPM as a dependency? 
Can we make our bundle somehow smaller with taking advantage of this? Or maybe it doesn't actually do anything and we could disable it?


Either way, this change looks good, let's try it!
Thank you for looking into it

jwetzell:
@cpvalente I tried keeping the builder image on `-alpine` image and switching the `pnpm` install method to `npm install -g pnpm@9.5.0` and that still hung things up so it wasn't specific to corepack and alpine. It's node >=20 and alpine in general it seems. I do like the npm install method as it's easier to specify a version so I'll switch it to that here.

cpvalente:
> @cpvalente I tried keeping the builder image on `-alpine` image and switching the `pnpm` install method to `npm install -g pnpm@9.5.0` and that still hung things up so it wasn't specific to corepack and alpine. It's node >=20 and alpine in general it seems. I do like the npm install method as it's easier to specify a version so I'll switch it to that here. 

Sounds good. Let's try and merge and make a new beta release to test it out. Are you able to do that?

PR #1143, opened 2024-07-16
refactor: dev label

cpvalente:
the main objective of this PR is to simplify the code around updating the timer by removing code paths specific to roll

Making the functions agnostic of the play mode means
- easier maintenance of timer logic
- integration events "just work"

on the downside, we would need to consider that there is more logic involve for any of the code paths

We would need a separate PR to tidy up the behaviour around roll. I am not aware of any change of behaviour introduced by this PR


alex-Arc (commented):
Should roll trigger a load and start dispatch when it rolls an event?
Because that seams to not happen right now

apps/server/src/services/timerUtils.ts
```diff
@@ -88,6 +93,12 @@ export function getCurrent(state: RuntimeState): number {
  * @returns {boolean}
  */
 export function skippedOutOfEvent(state: RuntimeState, previousTime: number, skipLimit: number): boolean {
+  // eslint-disable-next-line no-unused-labels -- dev code path
+  DEV: {
+    if (state.timer.expectedFinish === null || state.timer.startedAt === null) {
```
alex-Arc, line 98: I'm also hitting this 
just by creating an event 1 minnut from now and then clicking roll

cpvalente:
> Should roll trigger a load and start dispatch when it rolls an event?
> 
> Because that seams to not happen right now

I havent added any new behaviour. Just removed the paths specific to roll. 

We would need a follow up ticket to improve roll

apps/server/src/services/timerUtils.ts
alex-Arc: @cpvalente 
how about this is it intended?

cpvalente: This should no longer be called before an event is loaded. Is that not the case?

alex-Arc: looks to work fine after the recent force push
https://github.com/cpvalente/ontime/commit/d721d550b48e66b65ee358275776151bb6827b60

alex-Arc (approved):

PR #1145, opened 2024-07-16
remove husky and pre-commit scripts

jwetzell:
per our conversation.

I think looking at the test workflow (or making a separate workflow) and making sure that is setup to lint/format checking will be more worthwhile than having these pre-commit hooks.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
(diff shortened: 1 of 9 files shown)


cpvalente (approved):
Thank you for this. 

I agree with your observation, linting and type-checking is already running in the pipeline, we just need to add formatting

You are welcome to add that here or chose do it it in another PR

jwetzell:
I'll get it in another PR. Like to get that test workflow behaving before adding to it.

PR #1146, opened 2024-07-17
convert shared dependencies to pnpm catalog

jwetzell:
I tried to get all the dependencies that were in more than one `package.json`. If you can think of what that might be missed let me know and I can add it.


cpvalente:
This looks like a good start, and probably as much as we should do for now

I wonder if we should split things a bit more with the namespaces? 

Maybe `catalog:typescript` `catalog:lint` and `catalog:format`?
What do you think? Either way, it is easy enough to change later if we feel like it

cpvalente (approved):

jwetzell:
@cpvalente I thought about that but didn't have a great idea for dealing with the overlap. Does eslint-config-prettier go under format or lint? @typescript-eslint/eslint-plugin under lint or typescript?

PR #1150, opened 2024-07-19
test failing

alex-Arc:
test failing in github actions


apps/server/src/stores/__tests__/runtimeState.test.ts
```diff
@@ -159,7 +159,8 @@ describe('mutation on runtimeState', () => {
     const event1 = { ...mockEvent, id: 'event1', timeStart: 0, timeEnd: 1000, duration: 1000 };
     const event2 = { ...mockEvent, id: 'event2', timeStart: 1000, timeEnd: 1500, duration: 500 };
     // force update
-    initRundown([event1, event2], {});
+    withFlushTimers(initRundown, [[event1, event2], {}]);
```
cpvalente, line 162: question: dont we need to await the init rundown?

alex-Arc: yeah you might be right, initRundown is async...
I will try and see if I can adapt the function

cpvalente (approved):
Good work

Issue #1154, opened 2024-07-23
Display version in help menu


epnlespieux:
Hi,

Could you add the version of the running ontime software in the help menu or somewhere else easily visible ?

As your very good software is often updated, it will be cool !

Cheers

cpvalente:
> Hi,
> 
> 
> 
> Could you add the version of the running ontime software in the help menu or somewhere else easily visible ?
> 
> 
> 
> As your very good software is often updated, it will be cool !
> 
> 
> 
> Cheers

Hi, thank you for your suggestion. The app version is currently in the settings. Do you think this is hard to find?

If we were to copy the version into the menu, there wouldn't be a button for you to check for updates, I am unsure on whether knowing the version on its own is a good improvement since you would still need to open GitHub to check the latest version

What do you think?

epnlespieux:
![image](https://github.com/user-attachments/assets/d1db21bf-e237-4005-a7b4-545cd9ea830a)

No it's not hard to find once you know it's there :-). But even without a button to check for updates in the help menu, it can be faster to see.

Cheers

epnlespieux:
![image](https://github.com/user-attachments/assets/dd1e183d-9663-481d-a41a-612cf74a7bab)

epnlespieux:
> Meanwhile, are there any other items in these menus which you would propose improvements to?

you said it's not possible but a line to tell us "a new version is available" is there is a newer version.
But we can easily click on "See on github".
Thanks for your work 

ps : do you know the french bd author Lewis Trondheim ? His artist name comes from your town name (https://www.instagram.com/lewistrondheim/) ?

alex-Arc:
@epnlespieux you can look forward to it in the next release

PR #1155, opened 2024-07-23
about in electron help menu

alex-Arc:
#1154 

unsure if `win.loadUrl` is the best way


apps/electron/src/menu/applicationMenu.js
```diff
@@ -150,6 +150,12 @@ function getApplicationMenu(isMac, askToQuit, urlBase) {
     {
       role: 'help',
       submenu: [
+        {
+          label: 'About',
+          click: async () => {
+            redirectCb('editor?settings=about');
+          },
+        },
```
cpvalente, line 158: could we show the app version here somewhere? we would need to retrieve it from the package.json in a similar way we do in the client / server

cpvalente (approved):

PR #1156, opened 2024-07-23
FIX: remove leading '/' from path in client-control on first connect

alex-Arc:


apps/client/src/common/utils/socket.ts
```diff
@@ -36,7 +36,10 @@ export const connectSocket = () => {
     }
 
     socketSendJson('set-client-type', 'ontime');
-    socketSendJson('set-client-path', location.pathname);
+
+    //remove leading '/' from path
+    const fullPath = location.pathname.startsWith('/') ? location.pathname.slice(1) : location.pathname;
+    socketSendJson('set-client-path', fullPath);
```
cpvalente, line 42: question: as far as I understand, the issue is only in the settings popup.
Could we not just remove one of the slashes there? I dont think the fix is to cover up the issue, but to find where it is introduced

alex-Arc: The popup gets its data from what is sent in `'set-client-path`
We set the client path from 2 place
I changed this one to match the other one
https://github.com/cpvalente/ontime/blob/c7893938eb22a8f5ef73e6820f90247997623b52/apps/client/src/common/hooks/useClientPath.ts

cpvalente: hmm, I still wonder if there is a way to stop the extra / from making it to the path to begin with.
as of now, just looks like we are patching it up in two places

alex-Arc: [pathname](https://developer.mozilla.org/en-US/docs/Web/API/Location/pathname) includes a staring slash
but we don't have to remove it the, the redirect works in both cases

alex-Arc: yes

apps/client/src/common/utils/socket.ts
```diff
@@ -36,6 +36,8 @@ export const connectSocket = () => {
     }
 
     socketSendJson('set-client-type', 'ontime');
+
+    //remove leading '/' from path
```
cpvalente, line 40: todo: do we need to remove this comment?

apps/client/src/common/hooks/useClientPath.ts
```diff
@@ -13,7 +13,7 @@ export const useClientPath = () => {
   // notify of client path changes
   useEffect(() => {
     //remove leading '/' from path
-    const fullPath = (pathname.startsWith('/') ? pathname.slice(1) : pathname) + search;
+    const fullPath = pathname + search;
```
cpvalente, line 16: todo: do we need to remove this comment?
also, should we use a the URL object to create a path that we know is valid?

cpvalente (approved):

apps/client/src/common/hooks/useClientPath.ts
alex-Arc: I think we are already getting these from a URL object 
and  it only provides pathname and search separately

PR #1162, opened 2024-07-29
restore rundownGetAll

alex-Arc:
`ip:port/data/rundown` -> array with rundown
`ip:port/data/rundown/paginated?offset=0&limit=10` -> object with rundown section and total event number


cpvalente (approved):
Looks good, thank you!

PR #1164, opened 2024-07-30
Merge import

alex-Arc:
image

TODO:
- [x] some css formatting 

---
optional:
- [ ] have the option for each part to eitheroverride or append


apps/client/src/common/utils/mergeProjects.ts
```diff
@@ -0,0 +1,19 @@
+import { DatabaseModel } from 'ontime-types';
+
+import { getDb, patchData } from '../api/db';
+
+export async function mergeProjects(fileName: string, mergeKeys: object) {
```
cpvalente, line 5: I think DataProvider already exposes a function to do this?

alex-Arc: yeah that is what the `patchPartialProjectFile` route uses but we need to create the partial object first

apps/client/src/common/utils/mergeProjects.ts
```diff
@@ -0,0 +1,19 @@
...
+  const { data } = await getDb(fileName);
+  const patchObject: Partial<DatabaseModel> = {};
+
+  for (const key in mergeKeys) {
+    if (isValidKey(key, data)) {
+      Object.assign(patchObject, { [key]: data[key] });
+    }
+  }
+  await patchData(patchObject);
+}
```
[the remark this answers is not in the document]
alex-Arc: the error catching is handled in `ProjectMergeForm.tsx` in the function `handleSubmitCreate`

alex-Arc:
@cpvalente could you perhaps help with some CSS

cpvalente:
Hi @alex-Arc , I was unable to verify that the feature worked.

I opened up a project, titled Test 1
I then partial merged the rundown of a new project "Eurovision"

I see that the title of my project in projectData is now "Eurovision", however, I would have expected that only the rundown was merged.

I can tidy up the CSS once the feature is finished

alex-Arc:
> I was unable to verify that the feature worked.

should be fixed now

apps/client/src/features/app-settings/panel/project-panel/ProjectPanel.module.scss
```diff
@@ -50,3 +54,9 @@
   flex-direction: column;
   gap: 1em;
 }
+
+.toggleOption {
+  display: flex;
+  flex-direction: row;
+  gap: 1em;
+}
```
cpvalente, line 62: we can remove this and add the class `toggleOption` to the CSS above that contains equivalent styling

cpvalente (approved):
Thank you for your work here, I suggested some small cleanup for now but generally, it seems to work well and the code looks good

cpvalente (approved):
I havent tested all variants but as far as I can see everything is fine with the existing backend code

Thank you Alex

Issue #1139, opened 2024-07-15
Display Delay at every Event


AWi92:
First of all, I would like to say that I really like your software and that this is exactly what I wanted to program some time ago. How quickly the delay (over / undertime) feature was implemented was impressive and that's what makes the software really useful for me.

For the analysis during the event and especially after the event, I would find it practical if it could be displayed per event how much delay this event has produced (positive and negative). 
Especially for events with the same crew, this information could help to organize processes differently / better for the future. For example if the opening allways produces -3 min delay maybe we should change this duration from scratch or find out why it takes so much extra time.

If this is not a feature that is relevant for other users, I think that your extensive API gives me enough data to log this by myself.

Here is a image that shows what I mean:
![image](https://github.com/user-attachments/assets/bd9235ca-dec1-4559-895e-67acf007ce63)

cpvalente:
Hi! Thank you for taking the time to write this proposal.

We have investigated a similar feature in the past which we called the reporter.
The initial specification was written in a comment here
https://github.com/cpvalente/ontime/issues/165#issuecomment-1556218244

We ended up making a demo implementation but hit some roadblocks on deciding how the feature should behave, at the time, we could not find users with a workflow that would make use of the feature so I took it as a sign that the feature had little value to the users. Perhaps I was wrong or this has changed?

In my proposed implementation, the report was something you could download after a show was done. We saw that this could be useful for users that need to keep track of deviations on the time something took (like in repertoire theater). Your proposal takes it a (good) extra step of pushing it one extra level by showing it as the rundown progresses.

A good first step could be revisiting the initial specification. If you would like to collaborate with us on the creation of this feature by providing testing and guidance on the specification, it would be much appreciated. If so, please do get in touch at mail@getontime.no

cpvalente:
Hi @AWi92 , we have added this as part of the new Report feature in v3.13.0, could you please double check whether the implementation satisfies the requirements so we can close this ticket as completed?

cpvalente:
Closing this as completed and released

PR #1166, opened 2024-07-31
rundown reporter

alex-Arc:
#1139


alex-Arc:
@coderabbitai summary

apps/server/src/services/report-service/ReportService.ts
```diff
@@ -21,13 +21,25 @@ export function clear(id?: string) {
...
-  if (state.eventNow?.id === undefined) {
-    return;
+  // eslint-disable-next-line no-unused-labels -- dev code path
+  DEV: {
+    if (state.eventNow === null) {
+      throw new Error('report.eventStop: called without eventNow present');
+    }
   }
+
+  //TODO: How should pauses be calculated into the report?
```
cpvalente, line 42: I think the reporter doesnt need to know the details of what happened inside a timer.
It would be sufficient to have startedAt and finishedAt timestamps

alex-Arc:
@cpvalente 
could use some help checking that that change to the `RuntimeService.ts` is ok
and that I'm using the debounce hook correctly here [EventBlockReporter.tsx](https://github.com/cpvalente/ontime/pull/1166/commits/252a66165e18b5ffad969a764363e4a8e11ba317#diff-d46ef9220f743d994191314f2282c1833ea5f497d4ab85b0c7c011d6a2835a94

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { tooltipDelayFast } from '../../../../ontimeConfig';
+import { getTimeToStart } from '../../../../views/timeline/timeline.utils';
+
+import style from './EventBlockReporter.module.scss';
+
+export default function EventBlockReporter(props: { id: string; timeStart: number; isPastOrLoaded: boolean }) {
+  const { id, timeStart, isPastOrLoaded } = props;
+  const { data } = useReport();
+  const status = useReportStatus();
+  const [debouncedStatus] = useDebouncedValue(status, 200);
```
cpvalente, line 18: What are you trying to achieve with this hook? The 200ms debounce is lower than the update rate of clock (1s)
Is this because the values are out of sync?
If that is the case, I don't think the clients should be patching up, and we rather that the server syncs up the state it sends 

Regardless:

We may want the leading flag on so we always get an update at start

I also think we want to isolate this in a separate hook. As it is, we are re-rendering the component in every change, even though we may not update the value we show

apps/client/src/features/rundown/event-block/EventBlockInner.tsx
```diff
@@ -94,6 +97,9 @@ const EventBlockInner = (props: EventBlockInnerProps) => {
           timerType={timerType}
         />
       </div>
+      <div className={style.reportSection}>
```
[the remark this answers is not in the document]
alex-Arc: I don't know how to remove it..
It needs a placement in the `grid-template-areas` just like the ` <TimeInputFlow>`

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { clock, offset, playback } = debouncedStatus;
+  const report: ReportData | undefined = data[id];
+
+  const showReport =
+    (playback !== Playback.Play && playback !== Playback.Roll) || //If we are not playing never show until time
```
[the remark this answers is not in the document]
alex-Arc: I can't find it, but could make one as we do the  in a lot of places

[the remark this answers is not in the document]
cpvalente: this is the one I meant https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/timerUtils.ts#L195

apps/client/src/common/utils/socket.ts
```diff
@@ -143,8 +143,8 @@ export const connectSocket = () => {
           break;
         }
         case 'ontime-timer': {
-          patchRuntime('timer', payload);
-          updateDevTools({ timer: payload });
+          runtimeStore.setState({ timer: payload, clock } as Partial<RuntimeStore>);
```
cpvalente, line 146: using the patchRuntime provides some advantages, including us knowing that the state is up to date, and all state changes go through the same place
I wonder if it would be preferable:
- a) to make two calls
- b) modify the patch function to receive a patch object

what do you think?

apps/client/src/features/overview/Overview.tsx
```diff
@@ -125,7 +125,7 @@ function RuntimeOverview() {
   const { clock, offset } = useRuntimePlaybackOverview();
 
   const offsetText = getOffsetText(offset);
-  const offsetClasses = offset === null ? undefined : offset <= 0 ? style.behind : style.ahead;
+  const offsetClasses = offset === null ? undefined : offset >= 0 ? style.ahead : style.behind;
```
cpvalente, line 128: good fix, thank you

apps/client/src/features/rundown/event-block/EventBlock.module.scss
```diff
@@ -139,7 +143,7 @@ $skip-opacity: 0.2;
 
   .nextTag {
     font-size: 1rem;
-    color: $orange-500;
+    color: $warning-orange;
```
cpvalente, line 146: I am not sure this is a good change
it is unfortunate that we use the labelled warning colour for this. but that doesnt mean that the next tag is a warning

ie: if we decide to say that all warnings will become purple, should the next tag also become purple?

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { usePlayback, useTimeUntil } from '../../../../common/hooks/useSocket';
+import useReport from '../../../../common/hooks-query/useReport';
+import useRundown from '../../../../common/hooks-query/useRundown';
+import { cx } from '../../../../common/utils/styleUtils';
+import { formatDuration, getTimeToStart } from '../../../../common/utils/time';
+import { tooltipDelayFast } from '../../../../ontimeConfig';
+
+import style from './EventBlockReporter.module.scss';
+
+interface EventBlockReporterProps {
```
cpvalente, line 14: cleanup: looks like this is used in `EventBlockChip`, so it should be called EventBlockChipProps and be next to the function that uses it

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  id: string;
+  timeStart: number;
+  isPast: boolean;
+  isLoaded: boolean;
+  className: string;
+}
+
+function EventReport(props: { id: string; className: string }) {
```
cpvalente, line 22: cleanup: lets make an interface for this for sake of consistency

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const timeDisplay = isDue ? 'DUE' : `${formatDuration(Math.abs(timeUntil), timeUntil > MILLIS_PER_MINUTE * 2)}`;
+  return (
+    <Tooltip label='Expected time until start' openDelay={tooltipDelayFast}>
+      <div className={cx([style.chip, isDue ? style.due : null, className])}>{timeDisplay}</div>
+    </Tooltip>
+  );
+}
+
+//TODO: what about gaps and overlaps
+export default function EventBlockChip(props: EventBlockReporterProps) {
```
cpvalente, line 74: nitpick: if we need to have multiple components in one file, I think it is nicer to have the main one at the top so we can find it straight away
also, our convention is that the default component matches the file name

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  //show seconds if the amount is less than 2 minutes as it could then represent up to 50% of the actual value?
+  const timeDisplay = isDue ? 'DUE' : `${formatDuration(Math.abs(timeUntil), timeUntil > MILLIS_PER_MINUTE * 2)}`;
+  return (
+    <Tooltip label='Expected time until start' openDelay={tooltipDelayFast}>
+      <div className={cx([style.chip, isDue ? style.due : null, className])}>{timeDisplay}</div>
+    </Tooltip>
+  );
+}
+
+//TODO: what about gaps and overlaps
```
cpvalente, line 73: good point, it makes sense we would need to account for these, is there a challenge here?

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }
+  return null;
+}
+
+function EventUntil(props: { className: string; timeStart: number }) {
+  const { timeStart, className } = props;
+  const { clock, offset } = useTimeUntil();
+
+  if (offset === null) {
+    return null; //TODO: this partially fixes the DUE flashing, but instead hides everything
```
cpvalente, line 58: if the fix is not good enough we can look into it, otherwise it shouldnt be a todo since it indicates that there is code change needed

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,93 @@
...
+function EventReport(props: { id: string; className: string }) {
+  const { id, className } = props;
+  const { data } = useReport();
+  const {
+    data: { rundown },
+  } = useRundown();
+
+  const report: ReportData | undefined = data[id];
+
+  if (report && report.startAt !== null && report.endAt !== null) {
```
cpvalente, line 31: tip:
could we invert this if statement to return null earlier and reduce the nesting?
also, could we have a label for it, like

`const hasReport = report?.startAt !== null && report?.endAt !== null`

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const thisEvent = rundown[id] as OntimeEvent;
```
cpvalente, line 32: nit: the word `this` carries to much baggage in javascript
can we call it something else, maybe `currentEvent`?

apps/client/src/features/rundown/event-block/composite/EventBlockReporter.tsx
```diff
@@ -0,0 +1,93 @@
...
+        </Tooltip>
+      );
+    }
+  }
+  return null;
+}
+
+function EventUntil(props: { className: string; timeStart: number }) {
+  const { timeStart, className } = props;
+  const { clock, offset } = useTimeUntil();
```
cpvalente, line 55: it would be good to try and stabilise this value to match our rules

something like (I didnt test it)
```ts
function useLessChangingTimeUntil(timeStart: number) {
  const { clock, offset } = useTimeUntil();

  const lazyTimeToStart = useMemo(() => {
    // we dont show anything if there is no offset
    if (offset === null) {
      return;
    }

    const timeToStart = getTimeToStart(clock, timeStart, 0, offset);
    // if the event is due, we dont need need the accurate value
    if (timeToStart <= MILLIS_PER_SECOND) {
      return 0;
    }

    // if the timer is over 2 minutes, we just show minutes
    if (timeToStart > MILLIS_PER_MINUTE * 2) {
      return formatDuration(timeToStart);
    }

    // at this point, timeToStart is positive and larger than 2 minutes
    return formatDuration(timeToStart, false);
  }, [clock, offset, timeStart]);

  return lazyTimeToStart
}
```

PR #1170, opened 2024-08-05
Fix current block

alex-Arc:


apps/server/src/stores/runtimeState.ts
```diff
@@ -183,7 +181,7 @@ export function load(
   }
 
   // load events in memory along with their data
-  loadNow(timedEvents, eventIndex);
+  loadNow(rundown, timedEvents, eventIndex, prevBlock);
```
cpvalente, line 184: Instead of passing conditional data. How about creating a loadCurrentBlock function to encapsulate the logic?

apps/server/src/stores/runtimeState.ts
```diff
@@ -654,5 +650,33 @@ export function roll(rundown: OntimeRundown): { eventId: MaybeString; didStart:
...
   // update runtime
   runtimeState.runtime.actualStart = runtimeState.clock;
+  runtimeState.currentBlock.startedAt = runtimeState.clock;
   return { eventId: runtimeState.eventNow.id, didStart: true };
 }
+
+function loadBlock(
+  rundown: OntimeRundown,
+  prevBlockId: string | undefined = runtimeState.currentBlock.block?.id,
+  prevBlockStartAt: number | null = runtimeState.currentBlock.startedAt,
```
cpvalente, line 660: question: I am not sure I understand this condition.

loadBlock has access to the state, why not consume it directly?

apps/server/src/services/runtime-service/RuntimeService.ts
```diff
@@ -698,11 +696,16 @@ function broadcastResult(_target: any, _propertyKey: string, descriptor: Propert
...
 
+    let syncBlockStartAt = false;
+
     if (!deepEqual(RuntimeService?.previousState.currentBlock, state.currentBlock)) {
       eventStore.set('currentBlock', state.currentBlock);
       RuntimeService.previousState.currentBlock = { ...state.currentBlock };
+      syncBlockStartAt = true;
     }
 
+    const shouldUpdateClock = getShouldClockUpdate(RuntimeService.previousClockUpdate, state.clock) || syncBlockStartAt;
```
cpvalente, line 707: tip: how about having `syncBlockStartAt` first? if that is `true`, we want dont need to run the function `getShouldClockUpdate`

cpvalente (approved):
small suggestion, thank you alex!

PR #1172, opened 2024-08-09
fix: stop playback when event is deleted

cpvalente:
fixes issue with playback continuing when an event is deleted
this also means that playback is stopped if user changes the current event to skip (not possible in app UI)

### DIFF
--- a/apps/server/src/services/runtime-service/RuntimeService.ts
+++ b/apps/server/src/services/runtime-service/RuntimeService.ts
@@ -231,19 +231,23 @@ class RuntimeService {
     // 1. we are not confident that changes do not affect running event (eg. all events where changed)
     const safeOption = typeof affectedIds === 'undefined';
     // 2. the edited event is in memory (now or next) running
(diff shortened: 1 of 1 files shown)


apps/server/src/services/runtime-service/RuntimeService.ts
```diff
@@ -231,19 +231,23 @@ class RuntimeService {
...
     let isNext = false;
 
-    // TODO: review logic
+    // if we are not sure, or the event is in memory, we reload
     if (safeOption || eventInMemory) {
       if (state.eventNow !== null) {
         // load stuff again, but keep running if our events still exist
         const eventNow = getEventWithId(state.eventNow.id);
         if (!isOntimeEvent(eventNow) || !isPlayableEvent(eventNow)) {
+          // maybe the file was deleted or the skip state was changed
```
alex-Arc, line 245:
```suggestion
          // maybe the event was deleted or the skip state was changed
```

alex-Arc (approved):

PR #1177, opened 2024-08-10
refactor: review styles and implement BEM for user override

cpvalente:
Updating the styles of the timeline
Also made the top level overridable with BEM, however, I have decided to protect the styles of the timeline for the time being
Maybe it is a bit ugly, but it would prevent the view from being unusable

Screenshot 2024-08-10 at 16 37 17
Screenshot 2024-08-10 at 16 37 33


alex-Arc (commented):
can't really get anything to work 
witch overrides should work?

cpvalente:
> can't really get anything to work 
> 
> witch overrides should work?

I normally just use the ones from the css that is included with Ontime

Remember to activate it in the settings

alex-Arc (approved):

PR #1179, opened 2024-08-12
fix TimerPahse difference from load and reload

alex-Arc:
TimerPahse when loading and reloading from a playing event is different

uses the same function path for load and reload

optionally, also renamed runtimeState functions to better reflect their use, but this could be rolled back
 `runtimeState.reload` -> `runtimeState.updateLoaded`   (maybe refresh is better?)
`runtimeState.reloadAll` -> `runtimeState.updateAll`


cpvalente (approved):
Looks good, thank you!

PR #1181, opened 2024-08-16
add shortcut to jump between blocks

alex-Arc:


apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -110,10 +118,25 @@ export default function Rundown({ data }: RundownProps) {
         newCursor =
           (direction === 'up' ? getLastNormal(rundown, order)?.id : getFirstNormal(rundown, order)?.id) ?? null;
         newIndex = direction === 'up' ? order.length : 0;
+
+        if (jumpBlock && newCursor) {
+          const selected =
+            direction === 'up'
+              ? getRelevantBlockNormal(rundown, order, newCursor)
```
cpvalente, line 125: suggestion: I think the function name is showing that is unfit as `getRelevantBlock` no longer describes its usage. How about `getPreviousBlock`?

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -110,10 +118,25 @@ export default function Rundown({ data }: RundownProps) {
         newCursor =
           (direction === 'up' ? getLastNormal(rundown, order)?.id : getFirstNormal(rundown, order)?.id) ?? null;
         newIndex = direction === 'up' ? order.length : 0;
+
+        if (jumpBlock && newCursor) {
```
cpvalente, line 122: Question: I think the logic here is becoming too nested and hard to follow.
We should look for ways to improve this.

With the current structure, we assign the values at the end, meaning we need to mentally hold the values and execution of the entire function to understand what it does.
Itw also seems that there is little overlap

I would suggest over the blocks and event logic

- add early returns for logic blocks
- split the block logic to a new function (may need to rename the functions to `selectBlock` and `selectEvent`)

cpvalente (approved):
I would consider renaming the getRevelantBlock function to getPreviousBlock

This makes it easier to relate that `getPreviousBlock` and `getPreviousBlockNormal` are the same functions but working on a different dataset

PR #1183, opened 2024-08-19
Fix: change api endpoint

alex-Arc:


apps/server/src/api-integration/integration.controller.ts
```diff
@@ -54,18 +54,15 @@ const actionHandlers: Record<string, ActionHandler> = {
...
-
-        if (patchEvent.custom && newObjectProperty.custom) {
-          Object.assign(patchEvent.custom, newObjectProperty.custom);
-        } else {
-          Object.assign(patchEvent, newObjectProperty);
-        }
-      });
+      const newObjectProperty = parseProperty(property, value);
+      const key = Object.keys(newObjectProperty)[0] as keyof OntimeEvent;
+      shouldThrottle = willCauseRegeneration(key) || shouldThrottle;
```
cpvalente, line 60: suggestion: we can swap these around to avoid unnecessary operations

shouldThrottle = willCauseRegeneration(key) || shouldThrottle;

```suggestion
      shouldThrottle = shouldThrottle || willCauseRegeneration(key);
```

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -54,18 +54,15 @@ const actionHandlers: Record<string, ActionHandler> = {
...
+      if (patchEvent.custom && newObjectProperty.custom) {
+        Object.assign(patchEvent.custom, newObjectProperty.custom);
+      } else {
+        Object.assign(patchEvent, newObjectProperty);
+      }
     });
 
     if (shouldThrottle) {
```
cpvalente, line 68: suggestion: i dont think the consumer needs to know about throttling. we can simplify like

```ts
    if (shouldThrottle) {
      throttledUpdateEvent(patchEvent)
    } else {
      updateEvent(patchEvent);
    }
    return { payload: 'success' };
```

alex-Arc: I would like to keep it, knowing that the API is limited is a good thing and a good user might even back off a bit

PR #1185, opened 2024-08-22
fix: stop log incorrectly

alex-Arc:


apps/server/src/services/runtime-service/RuntimeService.ts
```diff
@@ -525,8 +525,8 @@ class RuntimeService {
     }
     const didStop = this.eventTimer?.stop();
     if (didStop) {
-      const newState = state.timer.playback;
-      logger.info(LogOrigin.Playback, `Play Mode ${newState.toUpperCase()}`);
+      const newPlaybackState = runtimeState.getState().timer.playback;
+      logger.info(LogOrigin.Playback, `Play Mode ${newPlaybackState.toUpperCase()}`);
```
cpvalente, line 529: this was copy-pasted so we need to make the same change on the other functions
also, how about simplifying like 

```suggestion
      const newState = runtimeState.getState();
      logger.info(LogOrigin.Playback, `Play Mode ${newState.timer.playback.toUpperCase()}`);
```

alex-Arc: think it was the only one that did not get new state after the playback action

cpvalente (approved):

packages/utils/src/validate-action/validatePlayback.ts
```diff
@@ -5,7 +5,7 @@ import { Playback } from 'ontime-types';
  */
 export function validatePlayback(currentPlayback: Playback) {
   return {
-    start: currentPlayback !== Playback.Stop,
+    start: currentPlayback !== Playback.Stop && currentPlayback !== Playback.Play,
```
alex-Arc, line 8: @cpvalente are you OK with this?
I saw that it was possible to click the play button in the UI while something was playing and that would generate error messages in the Log, nothing was breaking but....

cpvalente: yeah, I cant remember if there was a use case for still letting it through. The change sounds good

PR #1192, opened 2024-08-27
add time in block to restore point

alex-Arc:


apps/server/src/stores/runtimeState.ts
```diff
@@ -193,15 +193,16 @@ export function load(
...
   if (initialData) {
     patchTimer(initialData);
-
     const firstStart = initialData?.firstStart;
     if (firstStart === null || typeof firstStart === 'number') {
       runtimeState.runtime.actualStart = firstStart;
       runtimeState.runtime.offset = getRuntimeOffset(runtimeState);
       runtimeState.runtime.expectedEnd = getExpectedEnd(runtimeState);
     }
+    if (typeof initialData.blockStartAt === 'number') {
```
cpvalente, line 202: Question: don't we need to keep the id? How would we know when we go to the next event whether the block has changed?

alex-Arc: I did save it in the beginning 
but saw that we will calculate the block id when the restored event is loaded, and there should be no difference here
we could try to detect a mismatch at that point but we can't do anything with it...
if there was it would mean that the rundown changed, and we would just have to restore as best we can

cpvalente (approved):

Issue #1168, opened 2024-08-04
Disabling of Schedule and ON Air in Studio Clock


fiete96:
Hey,
Would be nice to have a Switch to turn of the Schedule and On Air Sign in the StudioClock Display!
I would  really like to customize the View a bit more besides the toggle of the seconds.

Thanks!

cpvalente:
Hi @fiete96 , thank you for this.

Would you mind making a list of things you would like to see and what is the desired workflow?

Once we have a specification and understand the use cases, I am happy to take a look

fiete96:
Hey!
I would like to use the "Clock" of the StudioClock in a standalone display without showing the schedule and On air sign. As I understand it, the StudioClock contains three "modules": the clock itself, the ON AIR sign, and the schedule.
I would like to toggle the visibility of these three modules. An ON AIR sign would be nice on a little spare tablet. ;)
Maybe this is possible with custom URLs? I don't know...
I hope you understand what I mean.

cpvalente:
Sure, I just wondered if there were more customisations you were looking after from your comment 
> I would really like to customize the View a bit more besides the toggle of the seconds.

I will add the task to our improvement list. Small improvements like these usually come out pretty quick.

Meanwhile, if you want to hide stuff from the views, you can inspect the page to find the necessary selector and add a CSS override
https://docs.getontime.no/features/custom-styling/
It sounds more complicated than it actually is.

fiete96:
Oh wow! Thanks a lot! I got it working with the CSS override! I learned something! 


For everyone who is interested, here is the CSS override to disable the ON AIR sign and schedule, and to center the clock:



```
.schedule-container {
    display: none; /* Hide the schedule container */
}

.clock-container {
    display: flex;
    flex-direction: column;
    align-items: center;
    justify-content: center;
    height: 100vh; /* Use the full height of the viewport */
}

.studio-clock {
    display: flex;
    justify-content: center;
    align-items: center;
    height: 100%; /* Ensure the clock takes up the full height available */
}
```

fiete96:
But a seperate On Air View would be nice to have!

cpvalente:
HI @fiete96 the newest release v3.5.0 contains the option to remove the right elements in the Studio view.
Could you please confirm that this works as you need?

cpvalente:
Closing this as completed

fiete96:
Thanks a lot! Will check it in the next days

PR #1203, opened 2024-08-31
refactor: add options to hide schedule

cpvalente:
Adds option to disable right section in studio clock so we only get the clock view in the center
As described in #1168


alex-Arc (approved):
Works as expected 👍🏽

PR #1206, opened 2024-09-02
feat (ui): implement speed controls

cpvalente:


apps/server/src/stores/runtimeState.ts
```diff
@@ -360,6 +361,33 @@ export function updateAll(rundown: OntimeRundown) {
...
+
+export function resetSpeed() {
+  runtimeState.timer.speed = 1.0;
+}
+
+export function getSpeed() {
+  return runtimeState.timer.speed;
+}
+
+export function calculateSpeed() {
```
cpvalente, line 376: please dont see these as answers, but as suggestions

Right now, we are attaching the functionality of time manipulation to a target time. I believe that we should not make the feature in a way that would stop us from doing some time manipulation on its own (ie, set timer to 1.2x) 

With this assumption, this is a utility function that gives me the speed that I would need to finish on time. without making mutations to state.
With this in mind

> // TODO: what should this return if no timer is running?

Good question. 
Considering this as a utility function, I believe the appropriate behaviour is to give me a result, or a reason not to do so. In that case, it is appropriate to throw when the operation is invalid

>// TODO: this can produce negative speeds (i.e. the desired finish time is in the past)
// TODO: should there be some clamping or rounding on this number?

I dont think we should do rounding.
Following the previous rule, I believe that we should throw if the desired finish time is in the past

jwetzell: `calculateSpeed` does exactly that, simply returns the speed factor necessary to finish on the timer scheduled end. The user would then need to apply that speed factor (or any other speed factor they want). As far as the clamping/rounding the function can return something like `1.234567898765432` I guess the precision is probably "free" to carry around just might only need truncated in displaying to the user.

PR #1208, opened 2024-09-03
feat: timer preview

### LINKED ISSUE #1024: Count Up and Clock not clear in the editor timer
Just installed 3.10-alpha1 (also tried 3.04), and when i select Count Up or Clock the timer acts as a countdown. 

Skærmbillede 2024-05-31 kl  13 58 57

cpvalente:
WIP

The idea is to help users manage the timer view and make things friendlier
https://www.figma.com/design/TOjSQ4e8q4dhNDNN5DElD3/ontime?node-id=1722-6&t=9OAa5afIUFi2Ule1-1

This also helps resolve #1024 

Screenshot 2024-09-20 at 22 34 11

- expose button to show auxiliary as secondary
- expose button to show external as secondary (previously disabled)
- create preview of timer
- show timer type with tooltips in preview
- show timer phase (warning / danger) in preview

Questions:

The button for external visibility is duplicated, I experimented without it but it felt strange to have two different inputs for External Message and Timer Message. I am not sure what the best way is here

I opted for not showing the real message and time to be conservative of performance. While this is a nice feature, I didnt feel it was worth the performance investment. However we could do better at showing the current formatting from settings (eg: HH:MM), any ideas for this?

### DIFF
--- a/apps/client/src/common/hooks/useSocket.ts
+++ b/apps/client/src/common/hooks/useSocket.ts
@@ -1,4 +1,4 @@
-import { RuntimeStore, SimpleDirection, SimplePlayback } from 'ontime-types';
+import { RuntimeStore, SimpleDirection, SimplePlayback, TimerMessage } from 'ontime-types';
 
 import { useRuntimeStore } from '../stores/runtime';
 import { socketSendJson } from '../utils/socket';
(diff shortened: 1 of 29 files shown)


alex-Arc (commented):
The legacy traslation works
haven't tryed the new stuff yet

cpvalente:
> The legacy traslation works
> 
> haven't tryed the new stuff yet

Nice, I have created a documentation page that hopefully helps

alex-Arc:
for this
image
to work we need to implement some url encode/decodeing
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI

PR #1216, opened 2024-09-18
Shutdown

cpvalente:
small change to allow shutting down from localhost outside electron as discussed in #1215

This will also impact the users in node and docker environments.
As far as I can see, there are no downsides here, but thought of double checking @alex-Arc @jwetzell. Am I maybe missing something? In those environments, is there a reason for us not to want users to shutdown the process from the interface?


jwetzell:
I can't think of any "gotchas" on the headless/docker side of things.

alex-Arc (approved):

PR #1225, opened 2024-09-25
handle timer type whitespace in sheet import

alex-Arc:
add trim to the `makeString` parser utility functions
and use it for timer type

### DIFF
--- a/apps/server/src/utils/parser.ts
+++ b/apps/server/src/utils/parser.ts
@@ -195,11 +195,12 @@ export const parseExcel = (excelData: unknown[][], options?: Partial<ImportMap>)
       const column = row[j];
       // 1. we check if we have set a flag for a known field
       if (j === timerTypeIndex) {
-        if (column === 'block') {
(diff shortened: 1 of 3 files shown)


apps/server/src/utils/parser.ts
```diff
@@ -195,11 +195,15 @@ export const parseExcel = (excelData: unknown[][], options?: Partial<ImportMap>)
       const column = row[j];
       // 1. we check if we have set a flag for a known field
       if (j === timerTypeIndex) {
-        if (column === 'block') {
+        if (typeof column !== 'string' || column !== undefined) {
+          return;
+        }
+        const maybeTimeType = typeof column === 'string' ? column.trim() : '';
```
cpvalente, line 201: question: should this be something we do to all string imports? I cant imagine a scenario where we would want whitespace in a string

alex-Arc: that sounds reasonable

cpvalente (approved):

Issue #1069, opened 2024-06-11
Freeze data at runtime


cpvalente:
We want to add to Ontime a feature which allows freezing the data during runtime.
This is to safeguard users from accidental writes

To clarify
- What should be in the covered by freeze? I assume that the custom fields should still be editable since operators would need these notes during the show? 

We will need some feedback rounds before writing a full spec for the work to be done

cpvalente:
We have decided that this feature did not bring enough user value and have abandoned its development

PR #1228, opened 2024-09-27
feat(ui): implement `Freeze` for Rundown UI

asharonbaltazar:
#### Feature
Customers want a way to view the Rundown safely without any danger of making changes. Refs #1069 

#### Changes
- create new freeze request functions
- create a third option to the `RundownMenu` toggles called `Freeze`. Toggling this option sends a request to `/frozen` and effectively prevents any backend changes as well
- add a few `disabled` props to disable inputs 
- add `isRundown` flag in components and pass it accordingly


apps/client/src/common/stores/appModeStore.ts
```diff
@@ -9,7 +9,18 @@ export enum AppMode {
...
+  switch (appModeFromSessionStorage) {
+    case AppMode.Run:
+      return AppMode.Run;
+    case AppMode.Edit:
+      return AppMode.Edit;
+    case AppMode.Freeze:
+      return AppMode.Freeze;
+    default:
+      return AppMode.Run;
+  }
```
cpvalente, line 23: in the domain, run and freeze are not mutually exclusive. It is not ideal to build our logic in a way that goes agains the domain

Another challenge I see here, is that the data provenance between the AppMode and the Freeze is different. AppMode is a local setting for the user, while freeze is a session setting. 
I am unsure it makes sense for us to mix the concerns here, although I understand the convenience

I wonder if the correct place to consume this is from the runtime store. we could create a [feature selector](https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/hooks/useSocket.ts) for it and consume it in the components instead of passing it as props. 

The other data is passed to the blocks as props, since we maintain the relationship of list > item, this is not the case here. Performance wise we would also prefer consuming this data as low down in the tree as possible

apps/client/src/features/rundown/rundown-header/RundownHeader.module.scss
```diff
@@ -1,7 +1,8 @@
 .header {
   padding-inline: calc(2rem - 6px + 0.5rem) calc(1rem + 2px);
-  
+
   display: flex;
   align-items: center;
   justify-content: space-between;
+  column-gap: 0.5rem;
```
cpvalente, line 7: in which case did you see the need for a column gap? I couldnt find a rundown small enough for it to be necessary?

and why using column-gap instead of gap since we expect no wrap?

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
```diff
@@ -64,9 +68,14 @@ const TimeInputFlow = (props: EventBlockTimerProps) => {
...
         hasDelay={hasDelay}
         placeholder='Start'
-        disabled={Boolean(linkStart)}
+        disabled={isRundownFrozen || Boolean(linkStart)}
       >
-        <InputRightElement className={activeStart} onClick={() => handleLink(!linkStart)}>
+        <InputRightElement
+          className={activeStart}
+          onClick={() => handleLink(!linkStart)}
+          as='button'
```
cpvalente, line 76: 👍

cpvalente (changes requested):
We need to create a handler in the `socket.ts` file which assigns data from the frozen message to the local store.
Once we have done that, we have real-time change of the data and this should be our source of truth

We also need to make some UI tweaks
- maybe a line between Run | Edit | Freeze?
- Frozen fields should not have the disabled opacity since it make is difficult to read
- Should the frozen field be indicated somewhere? Maybe the input should become of the "ghosted type" or perhaps a different colour and indicator in the label, like
`Blocked field label ❄`

apps/client/src/features/rundown/rundown-header/RundownHeader.module.scss
asharonbaltazar: Resolved with https://github.com/cpvalente/ontime/pull/1228/commits/1fde566b5d537a76a7402efb7d9eaa3eefeba47a

apps/client/src/common/stores/appModeStore.ts
asharonbaltazar: Resolving with https://github.com/cpvalente/ontime/pull/1228/commits/e7ce68b39c1cd938a98b80b0a180d6d9acafe521

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
```diff
@@ -88,7 +88,7 @@ const TimeInputFlow = (props: EventBlockTimerProps) => {
         disabled={isLockedDuration}
         placeholder='End'
       >
-        <Tooltip label='Lock end' openDelay={tooltipDelayMid}>
+        <Tooltip label='Lock end' openDelay={tooltipDelayMid} as='button'>
```
cpvalente, line 91: question: should the as be in the `InputRightElement`?

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
```diff
@@ -73,7 +73,7 @@ const TimeInputFlow = (props: EventBlockTimerProps) => {
         disabled={Boolean(linkStart)}
       >
         <Tooltip label='Link start to previous end' openDelay={tooltipDelayMid}>
-          <InputRightElement className={activeStart} onClick={() => handleLink(!linkStart)}>
+          <InputRightElement className={activeStart} onClick={() => handleLink(!linkStart)} as='button'>
```
cpvalente, line 76: I see. Nice 👍

apps/client/src/features/rundown/rundown-header/RundownHeader.tsx
```diff
@@ -9,16 +9,22 @@ import style from './RundownHeader.module.scss';
...
-  const setEditMode = () => setAppMode(AppMode.Edit);
 
   return (
     <div className={style.header}>
       <ButtonGroup isAttached>
-        <Button size='sm' variant={appMode === AppMode.Run ? 'ontime-filled' : 'ontime-subtle'} onClick={setRunMode}>
+        <Button
+          size='sm'
+          variant={appMode === AppMode.Run ? 'ontime-filled' : 'ontime-subtle'}
+          onClick={() => setAppMode(AppMode.Run)}
```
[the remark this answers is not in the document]
asharonbaltazar: Resolving with https://github.com/cpvalente/ontime/pull/1228/commits/b63d05cf9e6880ad06a88ddca192e6d2a918f280

apps/client/src/common/stores/appModeStore.ts
```diff
@@ -9,7 +9,9 @@ export enum AppMode {
 const appModeKey = 'ontime-app-mode';
 
 function getModeFromSession() {
-  return sessionStorage.getItem(appModeKey) === AppMode.Run ? AppMode.Run : AppMode.Edit;
+  const appModeFromSessionStorage = sessionStorage.getItem(appModeKey);
+
+  return appModeFromSessionStorage === AppMode.Run ? AppMode.Run : AppMode.Edit;
```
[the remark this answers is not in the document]
asharonbaltazar: Resolving with https://github.com/cpvalente/ontime/pull/1228/commits/b63d05cf9e6880ad06a88ddca192e6d2a918f280

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
asharonbaltazar: Resolving with https://github.com/cpvalente/ontime/pull/1228/commits/3b948e6b29a2dffc36a6e0ce2a878bdbcd378272

cpvalente:
Hey @alex-Arc , I have been working on this with @asharonbaltazar and the code seems solid
Could you please double-check the functionality here?

Issue #1230, opened 2024-09-30
Delay - Make Permanent has no impact?


juzz197:
When adding a delay, I see the amended, new times above the scheduled times in rundown.  

When clicking 'Make Permanent' I was expecting the scheduled times to update to include the delay but nothing seems to happen - is this the correct behaviour?

Thanks as always!

juzz197:
I see that when creating the delay, everything works as expected, e.g.

1) Event starts at 20:00:00
2) Create delay of 00:05:00 immediately before
3) New Start shows as 20:05:00 and this is reflected on the backstage view

However, when I then click 'Make Permanent', in the delay, the delay disappears and my times revert to their originals i.e. 20:00:00 start in the example and not 20:05:00 as I was expecting.  

Am I misunderstanding the the 'Make Permanent' function?

I can do a screen recording if that'll help.

Thanks!

cpvalente:
> I see that when creating the delay, everything works as expected, e.g.
> 
> 1. Event starts at 20:00:00
> 2. Create delay of 00:05:00 immediately before
> 3. New Start shows as 20:05:00 and this is reflected on the backstage view
> 
> However, when I then click 'Make Permanent' in the delay, the delay disappears and my times revert to their originals i.e. 20:00:00 start in the example and not 20:05:00 as I was expecting.
> 
> Am I misunderstanding the the 'Make Permanent' function?
> 
> I can do a screen recording if that'll help.
> 
> Thanks!

I see, that is not expected to happen. The screenshot could be helpful

If you prefer you can drop into the discord server and I can try to help you there later today 
https://discord.com/invite/eje3CSUEXm

juzz197:
Here's the screen recording.  I have some commitments this evening but will try and drop into the discord if I can.

cpvalente:
Thank you @juzz197 , I found the issue.

In v2, you could use blocks to prevent a delay from propagating. This was changed in v3 in a way that you need a gap in scheduling or a positive delay to do so

It seems that we forgot to update the logic to apply delays.
I will open a task and we will make sure this is fixed in the next release

Meanwhile, you can add the delay after the block and everything should work as expected

juzz197:
I've run a few tests and everything is working as expected now.

Thank you!

PR #1232, opened 2024-10-01
fix: apply delay

cpvalente:
Fixes issue described in #1230

I tried testing that all the delays are the same when applied and when they are visible (orange indicators), would be good to sanity check

### DIFF
--- a/apps/server/src/services/rundown-service/__tests__/delayUtils.test.ts
+++ b/apps/server/src/services/rundown-service/__tests__/delayUtils.test.ts
@@ -1,249 +1,243 @@
 import { OntimeBlock, OntimeDelay, OntimeEvent, OntimeRundown, SupportedEvent } from 'ontime-types';
 import { apply } from '../delayUtils.js';
+import { MILLIS_PER_HOUR } from 'ontime-utils';
 
-describe('apply() ', () => {
-  describe('in a rundown without the delay field, persisted rundown', () => {
-    it('applies delays', () => {
-      const delayId = '1';
-      const testRundown: OntimeRundown = [
-        { id: delayId, type: SupportedEvent.Delay, duration: 10 } as OntimeDelay,
-        { id: '2', type: SupportedEvent.Event, timeStart: 0, timeEnd: 10, duration: 10, revision: 1 } as OntimeEvent,
-        { id: '3', type: SupportedEvent.Event, timeStart: 0, timeEnd: 10, duration: 10, revision: 1 } as OntimeEvent,
-        { id: '4', type: SupportedEvent.Block } as OntimeBlock,
-        { id: '5', type: SupportedEvent.Event, timeStart: 0, timeEnd: 10, duration: 10, revision: 1 } as OntimeEvent,
-      ];
+/**
+ * Small utility to fill in the necessary data for the test
+ */
+function makeOntimeEvent(event: Partial<OntimeEvent>): OntimeEvent {
+  return { ...event, type: SupportedEvent.Event, revision: 1 } as OntimeEvent;
+}
 
-      const expected = [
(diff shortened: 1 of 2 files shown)


apps/server/src/services/rundown-service/delayUtils.ts
```diff
@@ -109,27 +105,69 @@ export function apply(eventId: string, rundown: OntimeRundown): OntimeRundown {
...
+    const currentEntry = updatedRundown[i];
+
+    // we need to remove the link in the first event to maintain the gap
+    let shouldUnlink = i === delayIndex + 1;
+
+    // we don't do operation on other event types
+    if (!isOntimeEvent(currentEntry)) {
+      continue;
+    }
+
```
alex-Arc, line 131: something like this is needed

```suggestion
  let lastEntry: OntimeEvent | null = null;
  let nonEventAccumilator = 0;

  for (let i = delayIndex + 1; i < updatedRundown.length; i++) {
    const currentEntry = updatedRundown[i];

    // we don't do operation on other event types
    if (!isOntimeEvent(currentEntry)) {
      nonEventAccumilator++;
      continue;
    }
    
    // we need to remove the link in the first event to maintain the gap
    let shouldUnlink = i === delayIndex + 1 + nonEventAccumilator;

```

apps/server/src/services/rundown-service/__tests__/delayUtils.test.ts
```diff
@@ -1,249 +1,221 @@
...
+    expect(updatedRundown).toMatchObject(expected);
+  });
+
+  it('unlinks events to maintain gaps when applying positive delays', () => {
+    const testRundown = [
+      makeOntimeEvent({ id: '1', timeStart: 0, timeEnd: 100, duration: 100, revision: 1 }),
+      makeOntimeDelay(50),
+      makeOntimeEvent({ id: '2', timeStart: 100, timeEnd: 150, duration: 50, revision: 1, linkStart: '1' }),
+    ];
+
```
alex-Arc, line 91: test for falling case
```suggestion
  it('unlinks events to across blocks is it is the first event after the delay', () => {
    const testRundown = [
      makeOntimeEvent({ id: '1', timeStart: 0, timeEnd: 100, duration: 100, revision: 1 }),
      makeOntimeDelay(50),
      makeOntimeBlock('BLOCK'),
      makeOntimeEvent({ id: '2', timeStart: 100, timeEnd: 150, duration: 50, revision: 1, linkStart: '1' }),
    ];

    expect(apply('delay', testRundown)).toMatchObject([
      { id: '1', type: SupportedEvent.Event, timeStart: 0, timeEnd: 100, duration: 100, revision: 1 } as OntimeEvent,
      { id: 'block', type: SupportedEvent.Block, title: 'BLOCK' },
      {
        id: '2',
        type: SupportedEvent.Event,
        timeStart: 150,
        timeEnd: 200,
        duration: 50,
        linkStart: null,
        revision: 2,
      } as OntimeEvent,
    ]);
  });
```

apps/server/src/services/rundown-service/__tests__/delayUtils.test.ts
```diff
@@ -1,249 +1,221 @@
...
+  return { ...event, type: SupportedEvent.Event, revision: 1 } as OntimeEvent;
+}
+
+/**
+ * Small utility to make a delay event
+ */
+function makeOntimeDelay(duration: number): OntimeDelay {
+  return { id: 'delay', type: SupportedEvent.Delay, duration } as OntimeDelay;
+}
+
```
alex-Arc, line 18:
```suggestion
/**
 * Small utility to make a block
 */
function makeOntimeBlock(title: string): OntimeBlock {
  return { id: 'block', type: SupportedEvent.Block, title } as OntimeBlock;
}

```

alex-Arc (changes requested):
The spcific case where the delay is before a block and the event after is linked needs fixing

alex-Arc (approved):

PR #1238, opened 2024-10-03
Translation pt br

Cicatelli:
Added Brazilian Portuguese language to the translation system and added translator codes to the operator view

Adicionado ao sistema de tradução o idioma português brasileiro e adicionado códigos do tradutor a visualização do operador


cpvalente:
Hi @Cicatelli, thank you for your work here

You did not ask for a review, but from your inactivity I assume that you are finished with the code you are proposing to merge

Unfortunately we cannot consider merging the PR in the current state.
The current code mixes two proposals
- creating a brazilian portuguese translation
- moving the entire application to being translated

The first one is completely fine, and we could isolate the change and merge without issues. I currently did not see that much difference between the existing portuguese and would be more inclined to revise the existing portuguese translation

As for the second, this is something that we would need to discuss as it is a major project change.

It currently achieves so at the cost of every other translatio being unimplemented. 

It also commits us to maintain translations for the entire interface
This means that: every time we want to add a feature, tweaks some text or label, we would need to look for translations in all the languages users have requested.

This is not sustainable and likely something we dont want to do.

The current decision is that the signage views are localisable, while the production views are always in english. Changes to this principle would need to be properly investigated and likely supported by a group of users who could provide translations on demand

I hope this does not discourage your contribution and is taken in the way that is meant.
In short: If you would extract the translation changes, those can easily be reviewed. The remaining changes need to be discussed

Cicatelli:
Hi @cpvalente!

The changes presented were a personal need and I made them available since a complete translation of the interfaces could help other people. I understand the big change to implement translation variables for the entire code.

I did not close the translation variables in the entire code, only for the EDITOR view. There is still a lot to do in the other views, however, due to work I have not been able to continue yet...

Regarding the PT vs PT-BR language, there are some terms that sound strange due to cultural differences, and just to avoid changing a language that already exists in the system, I decided to create another one.

I would like to thank you for the excellent work you have been doing with this code, which has helped me a lot!

I hope that within the scope of my knowledge, I can help the growth of a system that has helped me so much. I know little about high-level programming, my area is electrical engineering, the little I know about JS is as a hobby.

cpvalente:
Hi @Cicatelli , thank you for this.

I appreciate your involvement and your help.
If you want to make a separate PR for the the PT-BR translation we can look into it

Meanwhile, if you would like to contribute just reach out and we can find a task that aligns with the roadmap.
A good first start would be to evaluate how you use Ontime in your workflow and where is there friction that we could remove

The easiest way to reach out is likely [on discord](https://discord.com/invite/eje3CSUEXm)

cpvalente:
Hi, I will close this PR for now
If you want to create a smaller one tweaking the text it would be welcome

PR #1244, opened 2024-10-06
refactor: allow messages not starting with http

cpvalente:
We have got a user report that they need the HTTP integration to allow a target that who is not behind an HTTP protocol
In the user example, we would like to have `kmtrigger://macro=...`

Our development policy is often to not interfere with users setup as long as we can make it safe in Ontime.
In this case, I am unsure if we can get into trouble by being too permissive with the HTTP target

I couldn't find any reason for issues other than https requests will be rejected, https://httpwg.org/specs/rfc9110.html#field.host
Any ideas @alex-Arc or @jwetzell? Is there maybe another type of validation we should be doing?


jwetzell:
I haven't dug into what is actually making the calls for the HTTP Integration but as long as it supports the use of schemes other than `http` and `https` I don't see an issue with letting the user input it.

PR #1247, opened 2024-10-08
pop-out-clock proof of concept

Haavard15:
Very rough proof of concept for Video PiP that can pop out on top of presentation notes etc.


cpvalente:
Thank you @Haavard15 , we will plug in the data from here

PR #1262, opened 2024-10-13
Refactor setup

cpvalente:
Cleaning up the setup process

This aims to clarify that there are only two sources for the directories in the project
- relative to source directory
- relative to user directory

On the electron side is simply a renaming


apps/server/src/app.ts
```diff
@@ -82,17 +75,16 @@ app.use(express.json({ limit: '1mb' }));
 app.use('/data', appRouter); // router for application data
 app.use('/api', integrationRouter); // router for integrations
 
-// serve static - css
-app.use('/external/styles', express.static(resolveStylesDirectory));
-app.use('/external/', express.static(resolveExternalsDirectory));
+// serve static external files
+app.use('/external/', express.static(publicDir.externalDir));
```
cpvalente, line 79: here we replace the redundant serving or `/external` and `/external/styles` for a simpler root pointer to `/external`

alex-Arc (approved):

PR #1268, opened 2024-10-16
add more shortcuts for zooming in

### LINKED ISSUE #1256: Shortcut Zoom In doesn't work
Pressing Ctrl + + in the editor in Ontime doesn't zoom in, while zooming out (Ctrl + -) works fine. 
The option in the views menu and clicking on Zoom In works as well.

jwetzell:
As noted I still contend this is a bug with electron. I still haven't found a single app where pressing`ctrl` plus the key that `+` is on does not zoom in. 

Closes #1256


cpvalente (approved):
I think this is what we can do from our side

As long as we feel like zooming is important to our users, we just need to commit to patch this up.

Thank you for the investigation and the fix 👍

PR #1269, opened 2024-10-16
fix external path and remove unneeded stuff

alex-Arc:


apps/server/src/setup/index.ts
```diff
@@ -109,13 +109,9 @@ export const publicDir = {
   /** path to uploads folder */
   uploadsDir: join(resolvePublicDirectory, config.uploads),
   /** path to external folder */
-  externalDir: externalsStartDirectory,
+  externalDir: join(resolvePublicDirectory, config.external),
```
cpvalente, line 112: I am assuming that the fix is related to this somehow,

However, I think we still want to use `externalsStartDirectory`
In development mode, that provides us the folder at src, which makes it easier for us to work with

Maybe I am missing something?

apps/server/src/setup/index.ts
```diff
@@ -109,13 +109,9 @@ export const publicDir = {
...
   /** path to demo project folder */
-  demoDir: join(
-    externalsStartDirectory,
-    isProduction ? '/external/' : '', // move to external folder in production
-    config.demo.directory,
-  ),
+  externalDemoDir: join(externalsStartDirectory, config.externalDemo),
```
cpvalente, line 114: as above: I believe the previous logic provided value we are losing here

apps/server/src/setup/index.ts
```diff
@@ -73,7 +73,7 @@ export const srcDir = {
   /** Path to the react app */
   clientDir: isProduction ? join(srcDirectory, 'client/') : join(srcDirectory, '../../client/build/'),
   /** Path to the demo app */
-  demoDir: join(srcDirectory, 'external', config.demo.directory),
+  externalDemoDir: join(srcDirectory, 'external', config.externalDemo),
```
cpvalente, line 76: suggestion: can we rename this back to `demoDir`? The change doesnt really make too much of a difference, but it would be easier to see what the fix actually is if we removed unnecessary changes

PR #1271, opened 2024-10-17
Improve xlsx import

alex-Arc:
- switch from `node-xlsx` to `xlsx`
- evaluate excel formulars functions
- allow space and upper case in custom import

### DIFF
--- a/apps/client/src/features/app-settings/panel/sources-panel/import-map/importMapUtils.ts
+++ b/apps/client/src/features/app-settings/panel/sources-panel/import-map/importMapUtils.ts
@@ -1,4 +1,3 @@
-import { CustomFields } from 'ontime-types';
 import { ImportCustom, ImportMap } from 'ontime-utils';
 
 export type NamedImportMap = typeof namedImportMap;
@@ -82,24 +81,4 @@ export function getPersistedOptions(): NamedImportMap {
   } catch {
     return namedImportMap;
   }
-}
-
-// add existing custom fields to an importmap if they don't already exist
-export function addExistingCustomFields(importMap: NamedImportMap, customFields: CustomFields): NamedImportMap {
-  const customFieldsInImportMap = importMap.custom;
-
-  // keep all customfieds that are already diffind in the importmap
-  const newCustomFields = new Array(...customFieldsInImportMap);
-  Object.entries(customFields).forEach(([key, { label }]) => {
-    if (
-      !customFieldsInImportMap.some((customFieldFromImportMap) => {
-        return customFieldFromImportMap.ontimeName === key;
-      })
-    ) {
-      // If the custom field dosn't exist in the import map we add it
-      newCustomFields.push({ ontimeName: key, importName: label });
-    }
-  });
-
-  return { ...importMap, custom: newCustomFields };
 }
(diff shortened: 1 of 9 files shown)


apps/server/src/api-data/excel/excel.service.ts
```diff
@@ -23,32 +24,33 @@ export async function saveExcelFile(filePath: string) {
   if (extname(filePath) != '.xlsx') {
     throw new Error('Wrong file format');
   }
-  excelData = xlsx.parse(filePath, { cellDates: true });
+  excelData = xlsx.readFile(filePath, { cellDates: true, cellFormula: false });
```
cpvalente, line 27: question: did you try to see what happens if we pass `cellFormula: true`?

apps/client/src/features/app-settings/panel/sources-panel/import-map/importMapUtils.ts
```diff
@@ -82,3 +83,19 @@ export function getPersistedOptions(): NamedImportMap {
...
+        return ic.ontimeName === key;
+      })
+    ) {
+      // const ontimeName = key as unknown as string;
+      custom.push({ ontimeName: key, importName: label });
+    }
+  });
+
+  return importMap;
+}
```
cpvalente, line 101: question: could you please add some comments as to what this does?
We prefer explicit in favour of terser logic. Terser is a job for the builder

apps/client/src/features/app-settings/panel/sources-panel/import-map/importMapUtils.ts
```diff
@@ -82,3 +83,19 @@ export function getPersistedOptions(): NamedImportMap {
     return namedImportMap;
   }
 }
+
+export function addExistingCustomFields(importMap: NamedImportMap, customFields: CustomFields): NamedImportMap {
+  const custom = importMap.custom;
```
cpvalente, line 88: question: is this just a convenience rellabelling? 
did you check that we are mutating the original `importMap`?

apps/server/src/api-data/excel/excel.service.ts
alex-Arc: yes then we get the raw excel formular

apps/client/src/features/app-settings/panel/sources-panel/import-map/importMapUtils.ts
```diff
@@ -82,3 +83,23 @@ export function getPersistedOptions(): NamedImportMap {
...
+        return customFieldFromImportMap.ontimeName === key;
+      })
+    ) {
+      // If the custom field dosn't exist in the import map we add it
+      newCustomFields.push({ ontimeName: key, importName: label });
+    }
+  });
+
+  return { ...importMap, custom: newCustomFields };
+}
```
cpvalente, line 105: I am not entirely confident on this logic here. I could be missing something.
Could be also nice to add a single happy path test just as documentation

`const customFieldsInImportMap = importMap.custom;`

This reassignemnt doesnt seem to do much for us, but could provide the indirection of us thinking that the object is different, I suggest we remove it

`const newCustomFields = new Array(...customFieldsInImportMap);`

I am not sure why do we need to call the Array constructor, is spreading into a new array not enough? `newCustomFields = [...customFieldsInImportMap];`

`return { ...importMap, custom: newCustomFields };`
why do we need to spread in the end? the `importMap` is given by the consumer and used as a reference here.
Would it make more sense for the consumer to get the `importMap` and pass only the `custom` part here?

cpvalente:
@coderabbitai full review

cpvalente (approved):
Thank you alex, good work

PR #1272, opened 2024-10-17
hide seconds in cuesheet

alex-Arc:
truncates seconds in cuesheet
image
image


apps/client/src/features/cuesheet/cuesheetCols.tsx
```diff
@@ -18,21 +17,32 @@ function makePublic(row: CellContext<OntimeRundownEntry, unknown>) {
 }
 
 function MakeTimer({ getValue, row: { original } }: CellContext<OntimeRundownEntry, unknown>) {
-  const showDelayedTimes = useCuesheetSettings((state) => state.showDelayedTimes);
+  const { showDelayedTimes, showSeconds } = useCuesheetSettings((state) => state);
```
cpvalente, line 20: tip: make sure to consume the memoised version of this hook

```ts
const showDelayedTimes = useCuesheetSettings((state) => state.showDelayedTimes);
const showSeconds = useCuesheetSettings((state) => state.showSeconds);

```
It is more verbose, but it will prevent unnecessary re-renders

PR #1276, opened 2024-10-19
refactor: add nif to navigation

cpvalente:
Adding navigate and copy for network interfaces in the navigation menu

Note that this is only available for the localhost (ie: machine running Ontime)

Screenshot 2024-10-19 at 10 41 41


alex-Arc (approved):
nice
cool with both the external link and copy
it shows properly multiple nic's

PR #1278, opened 2024-10-19
feat: event finder

cpvalente:
feature to pair up with reporter.
It provides an interface for searching an event in the rundown

Screenshot 2024-10-19 at 23 15 41

I guess this could be easily extendable to have a list of settings shortcuts like network view settings
With the current setup this would mean typing `settings network`or `settings view`.
Perhaps we would prefer having a character to signify our tags? like `>cue` or `>settings`?

We will be a bit stuck with whatever we decide now

### DIFF
--- a/apps/client/package.json
+++ b/apps/client/package.json
@@ -11,7 +11,7 @@
     "@emotion/react": "^11.10.6",
     "@emotion/styled": "^11.10.6",
     "@fontsource/open-sans": "^5.0.28",
-    "@mantine/hooks": "^7.6.2",
+    "@mantine/hooks": "^7.13.3",
     "@react-icons/all-files": "^4.1.0",
     "@sentry/react": "^8.19.0",
     "@tanstack/react-query": "^5.17.9",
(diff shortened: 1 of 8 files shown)


cpvalente:
@coderabbitai full review

alex-Arc (commented):
this is a bit forced, but if we could find a way to accommodate it, that would be good
Screenshot 2024-10-21 102541


I like having it function on keyboard only but maybe it would be ok to make the filter buttons instead off keywords only?
to get around some of these issues, we could then add it later when we have a good solution

cpvalente:
> this is a bit forced, but if we could find a way to accommodate it, that would be good
> 
> Screenshot 2024-10-21 102541
> 
> 
> 
> 
> 
> I like having it function on keyboard only but maybe it would be ok to make the filter buttons instead off keywords only?
> 
> to get around some of these issues, we could then add it later when we have a good solution

This is a good point. Maybe we should informe a tagging system like `:index`?

I forgot about mouse, but I guess it is ok as is?

alex-Arc (approved):

PR #1295, opened 2024-10-24
Better update rate and batch in client

alex-Arc:


cpvalente:
@alex-Arc , should we get this ready to merge?
the next release will likely have a few rounds of beta which is a great opportunity to add this

cpvalente (commented):
@alex-Arc, can we split this PR into 2?

one thing is to do client side timer progress, we should definitely look into that at a later stage

but the rounding behaviour changes have been a point of contention and we can merge that earlier independently

Could we get the second part ready to merge?

alex-Arc:
Yes, will look at this after #1406

Issue #1289, opened 2024-10-23
Excel spreadsheet upload error


maurice941:
Once a Server Error shows up you can't import spreadsheets.
The error is caused by an incorrect sheet, which is missing a name or other identifying information. I'm not exactly sure what triggers the error message itself, but the key issue is that as long as the error message is present, you won't be able to import a new (correct) sheet.

https://github.com/user-attachments/assets/cc876247-c551-45b0-97cf-b8af23675b7b

PR #1296, opened 2024-10-25
refactor: clear state on sheet flows

cpvalente:
This PR resets the form state on cancel and fixes the issue described in #1289

I am also revoking authentication on google sheet cancel, that may be incorrect

### DIFF
--- a/apps/client/src/features/app-settings/panel/sources-panel/SourcesPanel.tsx
+++ b/apps/client/src/features/app-settings/panel/sources-panel/SourcesPanel.tsx
@@ -70,6 +70,17 @@ export default function SourcesPanel() {
     fileInputRef.current?.click();
   };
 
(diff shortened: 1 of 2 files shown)


alex-Arc (approved):

apps/client/src/features/app-settings/panel/sources-panel/SourcesPanel.tsx
```diff
@@ -89,7 +100,9 @@ export default function SourcesPanel() {
   };
 
   const cancelGSheetFlow = () => {
-    setImportFlow('none');
+    resetFlow();
+    revoke();
```
alex-Arc, line 104: we don't want revoke here

alex-Arc (approved):

PR #1299, opened 2024-10-25
Timer speed UI

cpvalente:
adds a slider and connects endpoints so we can test the timer-speed feature


jwetzell:
Looks good enough for testing

alex-Arc (approved):

PR #1300, opened 2024-10-25
[DEMO] Native html pip elemnt

alex-Arc:
located at `/pop`

as opposed to #1274 this dose not need to create a video element to accomplish this
and can reuse out html as is (looks to, not all react stuff will be available)

to down side are that the feature is `Experimental` and not fully supported
[DocumentPictureInPicture](https://developer.mozilla.org/en-US/docs/Web/API/DocumentPictureInPicture)
from my testing it works in chrome but not in firefox

and there is a small header for the pip element that i'm trying to see if can be removed

image

@Haavard15 could you have a look


cpvalente:
Hi @alex-Arc , could you please compile a small list of bullet point pros and cons?

jacinlowe:
Take my upvote on this feature

cpvalente:
> Take my upvote on this feature

Hi @jacinlowe , this feature is stagnant because we dont have the capacity to finish it up. Perhaps you would consider creating contributing with some code to see it through?

alex-Arc:
work continued here #1862

PR #1301, opened 2024-10-26
UI time to end

cpvalente:
This PR makes small UI changes to improve the visibility of time to end
1. show icon with active style in preview
2. add a "time to end" label to time preview
3. add a warning with tooltip for timers of type time-to-end
4. add a active style to the icon of time-to-end (flag)

time-to-end

We also received some feedback that our extract button didnt have enough visual distinction from the icons, so I add a quick change to match its style to our buttons


alex-Arc (commented):
I think it is good
the only think i', unsure about is the preview
image
as that line is "not true" maybe it should just replace the "Timer" text

PR #1303, opened 2024-10-26
Quick start

cpvalente:
Creating a wizard to initialise selected properties of a project

The options mostly follow the suggestions by @RasterboxOff , however, I have decided to omit the OSC settings.
As I understand, the premise on including this is to allow companion control. The companion module does not use OSC and will work regardless

I believe this wizard should be either a) quick things you want to setup for every project or b) some more comprehensive list of important settings.
In the second case, we will likely want to make a multi-step form that goes through the app settings.

Currently we have implemented a). I believe this means that we need to keep it to simple settings and thus the decision to remove the OSC setup.

Screenshot 2024-10-27 at 10 01 52


alex-Arc (approved):
I have not tested the functionality 
but it looks correct

PR #1314, opened 2024-10-31
Upload corporate logo

arymoraes:


apps/client/src/common/api/project.ts
```diff
@@ -19,3 +19,25 @@ export async function getProjectData(): Promise<ProjectData> {
 export async function postProjectData(data: ProjectData): Promise<AxiosResponse<ProjectData>> {
   return axios.post(projectPath, data);
 }
+
+/**
+ * HTTP request to upload a project logo
+ */
+export async function uploadProjectLogo(file: File): Promise<AxiosResponse<ProjectLogoResponse>> {
```
cpvalente, line 26: suggestion: I am not sure if we should have so many endpoints specific to the project logo

how about creating creating a patch project data endpoint?
We could add the logo here, if no logo is given, we delete existing logos.

What do you think?

apps/client/src/features/app-settings/panel/project-panel/ProjectData.tsx
```diff
@@ -34,6 +38,36 @@ export default function ProjectData() {
     }
   }, [data, reset]);
 
+  const handleUploadProjectLogo = async (event: ChangeEvent<HTMLInputElement>) => {
```
cpvalente, line 41: question: would it not be preferrable that the logo field is handled lile any other fields? ie: changes must be saved to take effect. so we dont call endpoints until the whole form is submitted

apps/client/src/features/app-settings/panel/project-panel/ProjectData.tsx
```diff
@@ -86,6 +120,31 @@ export default function ProjectData() {
               {...register('title')}
             />
           </label>
+          <label>
```
cpvalente, line 123: can we use an upload card similar to what we have in the sheets page?

apps/server/src/services/project-service/projectServiceUtils.ts
```diff
@@ -17,6 +17,21 @@ export async function handleUploaded(filePath: string, name: string) {
...
+  await rename(filePath, newFilePath);
+
+  return name;
+}
+
+export async function handleImageDelete(name: string): Promise<void> {
+  const filePath = join(publicDir.logoDir, name);
+  if (existsSync(filePath)) {
+    await unlink(filePath);
+  }
```
cpvalente, line 32: we have a `deleteFile` utility that handles this

packages/types/src/definitions/core/ProjectData.type.ts
```diff
@@ -5,4 +5,5 @@ export type ProjectData = {
   publicInfo: string;
   backstageUrl: string;
   backstageInfo: string;
+  projectLogo?: string;
```
cpvalente, line 8: question: we seem to have some varying approaches here. We define projectLogo as `string | undefined`, but use `empty string` to signify no logo

Would it be better to have this field be an union `string | null` so that the value is more explicit?

apps/client/src/common/api/project.ts
arymoraes: Kept the `upload` endpoint, but removed the `delete` endpoint.

When we upload the image, it saves to the `logos` folder, then returns the path to it. We can see the preview, and if we `Save`, the `projectLogo` gets updated to the logo path.

apps/client/src/features/app-settings/panel/project-panel/ProjectData.tsx
arymoraes: image

packages/types/src/definitions/core/ProjectData.type.ts
arymoraes: I changed it to `string | null`, but would this break existing projects that do not have that property on the json?

cpvalente: thank you for bringing this up. added a test to cover this case

cpvalente:
@coderabbitai full review

PR #1329, opened 2024-11-13
fix: custom fields subscription

cpvalente:
A few issues here
- if a custom field contained upper case letters, the subscription did not work
- if a custom field contained spaces, it would break the parsing of multiple fields
- adding a chevron to the menu to symbolise that the item is a combobox

### DIFF
--- a/apps/client/src/features/operator/Operator.tsx
+++ b/apps/client/src/features/operator/Operator.tsx
@@ -114,10 +114,8 @@ export default function Operator() {
   // get fields which the user subscribed to
   const shouldEdit = searchParams.get('shouldEdit');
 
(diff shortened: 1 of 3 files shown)


apps/client/src/features/operator/Operator.tsx
```diff
@@ -114,9 +114,10 @@ export default function Operator() {
   // get fields which the user subscribed to
   const shouldEdit = searchParams.get('shouldEdit');
 
+  // we use '-' to separe multiple fields
+  // the fields may contain '_' if there are spaces in the name
```
jwetzell, line 118: This comment mentions an `_` but the following code splits on `-` instead of the old `_` split

cpvalente:
@coderabbitai full review

PR #1330, opened 2024-11-14
feat(ui): Rundown `Freeze` UX

asharonbaltazar:


apps/client/src/features/rundown/event-editor/EventEditor.module.scss
```diff
@@ -41,7 +41,9 @@
 }
 
 .inputLabel {
-  display: block;
+  display: flex;
+  align-items: center;
+  column-gap: 0.3rem;
```
asharonbaltazar, line 46: Change to align `FrozenIcon` with the label text

cpvalente (approved):
great work, I suggest we merge this into the feature branch. I will likely just make a few tweaks to the style and double-check the fields that are disabled

good to have you back

PR #1332, opened 2024-11-17
feat(ui): disable menu buttons with `locked` param

asharonbaltazar:
Screenshot 2024-11-17 at 6 48 58 PM

I attempted to lock navigation with `useBlocker`, but there are two roadblocks to implement this feature:

- the `react-router-dom` package needs an upgrade. I was able to do so without any problems, FYI
- the hook is a data hook and thus only works with the `RouterProvider` configuration, which means re-doing `AppRouter.tsx` to use `createBrowserRouter`

### DIFF
--- a/apps/client/src/common/components/navigation-menu/FloatingNavigation.tsx
+++ b/apps/client/src/common/components/navigation-menu/FloatingNavigation.tsx
@@ -1,8 +1,7 @@
-import { useEffect, useState } from 'react';
 import { IoApps } from '@react-icons/all-files/io5/IoApps';
 import { IoSettingsOutline } from '@react-icons/all-files/io5/IoSettingsOutline';
 
-import { debounce } from '../../utils/debounce';
+import { useFadeOutOnInactivity } from '../../../common/hooks/useFadeOutOnInactivity';
 
 import style from './NavigationMenu.module.scss';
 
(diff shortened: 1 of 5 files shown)


apps/client/src/common/components/navigation-menu/FloatingNavigation.tsx
```diff
@@ -14,6 +16,9 @@ interface FloatingNavigationProps {
 export default function FloatingNavigation(props: FloatingNavigationProps) {
   const { toggleMenu, toggleSettings } = props;
   const [showButton, setShowButton] = useState(false);
+  const [searchParams] = useSearchParams();
+
+  const isViewLocked = searchParams.get('locked') ? true : false;
```
cpvalente, line 21: we have a utility for this `isStringBoolean`

apps/client/src/common/components/navigation-menu/FloatingNavigation.tsx
```diff
@@ -39,22 +44,30 @@ export default function FloatingNavigation(props: FloatingNavigationProps) {
...
-      </button>
-      <button
-        className={style.button}
-        onClick={toggleSettings}
-        aria-label='toggle settings'
-        data-testid='navigation__toggle-settings'
-      >
-        <IoSettingsOutline />
-      </button>
+      {isViewLocked ? (
```
cpvalente, line 47: I am not sure we can do it globally here
we have two Navigation implementations
`ProductionNavigationMenu` and `ViewNavigationMenu`

The feature here is applicable to the views that consume `ViewNavigationMenu`, I suggest that we encapsulate this logic there
this also allows us some DOM cleanups, since we know that we wont navigate, we just need to show the <FloatingNavigation /> element containing our lock icon

cpvalente (commented):
Thank you!

I believe we can cleanup the DOM a bit by re-arranging things. We also need to move the logic to another level of the app tree to affect the right views

Moving to `createBrowserRouter` should be straightforward, but we would need to investigate what are the implications of upgrading to the newest react-router. 
I am under the impression that it makes some assumptions about the type of optimisations you want and we will need to make sure we stay in some SPA mode

cpvalente (approved):
👍 
good work

PR #1343, opened 2024-11-23
automations

cpvalente:
This is an attempt at re-writing the integrations service

For simplicity, I am renaming this to automation, this means that integrations becomes an internal name used for data into ontime

There are also some dependency upgrades which make the PR larger

The code parsing logic is the same as before, so hopefully there are not that many core changes to the outputs, however, the filtering logic is completely new

I am also proposing a new structure to our backend, it is annoying that I only did it in automations and - if we decide to go forwards with this - I will create a follow up PR to apply everywhere
- in the same directory, I have collocated all the router and services. the service file, only has business logic. All crud logic is moved out of the way to a `dao` file
- in the same way, all utilities like parsing are in that directory. this means the `automation` directory, contains all logic related to that domain concept
- i also considered merging the `controller` and `router` files, but didnt see an obvious improvement. in a way, it is nice to be able to read all the available routes in a small file without worrying about what they do

In this PR
- [x] expose an integration service which allows filtering logic before sending a message
- [x] expose implementations for OSC, HTTP and Companion
- [x] expose testing function
- [x] create supporting UI
- [x] implement triggering of the new service
- [x] implement template parsing
- [x] implement clients
- [ ] ~~create migration code for old subscriptions~~ (decided to migrate OSC data and leave the rest)
- [x] persist data

TODO
- [ ] OSC input port and toggle migrate to settings
- [ ] consider UI for writing templates
- [ ] update docs
- [x] double-check strict typing in new code
- [x] double-check deepsource


apps/client/src/common/models/AutomationSettings.ts
```diff
@@ -0,0 +1,9 @@
+import { AutomationSettings } from 'ontime-types';
+
+export const automationPlaceholderSettings: AutomationSettings = {
+  enabledAutomations: false,
+  enabledOscIn: false,
```
cpvalente, line 5: @alex-Arc i am wondering if the OSC settings should be in the app settings instead of here

What do you think?

apps/server/src/api-data/automation/automation.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  return filters.some((filter) => evaluateCondition(filter));
+
+  function evaluateCondition(filter: AutomationFilter): boolean {
+    const { field, operator, value } = filter;
+    const fieldValue = state[field];
+
+    // TODO: if value is empty string, the user could be meaning to check if the value does not exist
+    switch (operator) {
```
cpvalente, line 70: @alex-Arc i have been wondering if we could come up with a few examples and fine tune the operators?

apps/server/src/api-data/automation/clients/osc.client.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { logger } from '../../../classes/Logger.js';
+import { type RuntimeState } from '../../../stores/runtimeState.js';
+import { parseTemplateNested, stringToOSCArgs } from '../automation.utils.js';
+
+/**
+ * Expose possibility to send a message using OSC protocol
+ */
+export function emitOSC(output: OSCOutput, state: RuntimeState) {
+  const message = preparePayload(output, state);
```
cpvalente, line 13: Skip sending in cloud

apps/client/src/common/models/AutomationSettings.ts
alex-Arc: yes 
at least it should be separate from the automations

apps/client/src/features/app-settings/panel/automations-panel/BlueprintForm.tsx
```diff
@@ -0,0 +1,484 @@
...
+  };
+
+  const handleAddNewHTTPOutput = () => {
+    appendOutput({ type: 'http', url: '' });
+  };
+
+  const handleTestOSCOutput = async (index: number) => {
+    try {
+      const values = getValues(`outputs.${index}`) as OSCOutput;
+      if (!values.targetIP || !values.targetPort || !values.address || !values.args) {
```
jwetzell, line 100: This if statement currently prevents testing OSC output if no args are specified.

cpvalente:
@coderabbitai full review

cpvalente:
@alex-Arc , I believe this is ready to merge, would you mind sanity checking?

apps/client/src/common/utils/__tests__/regex.test.ts
```diff
@@ -26,8 +26,8 @@ describe('simple tests for regex', () => {
   });
 
   test('startsWithHttp', () => {
-    const right = ['http://test'];
-    const wrong = ['https://test', 'testing', '123.0.1'];
+    const right = ['https://test', 'http://test'];
```
alex-Arc, line 29: Did you get https working?
or did it always work we just did not allow it

apps/server/src/app.ts
```diff
@@ -283,7 +257,6 @@ export const shutdown = async (exitCode = 0) => {
 
   expressServer?.close();
   runtimeService.shutdown();
```
alex-Arc, line 259: do we not still need the osc shutdown here?

apps/client/src/common/utils/__tests__/regex.test.ts
cpvalente: i have been testing with ontime cloud and it seems fine. can you verify in one of your containers?

apps/server/src/app.ts
cpvalente: yes we do, good catch

PR #1345, opened 2024-11-25
Small fixes

alex-Arc:
- use UI index in finder
- remove unintended(?) console log


cpvalente:
It seems to me that the logic with the indexes is incorrect since it does not account for blocks and delays.

see the screenshots
Screenshot 2024-11-28 at 22 13 36
Screenshot 2024-11-28 at 22 13 30

PR #1348, opened 2024-11-28
detect `EADDRINUSE` on server start

alex-Arc:


apps/electron/src/main.js
```diff
@@ -109,8 +109,13 @@ function askToQuit() {
  * Allows processes to escalate errors to be shown in electron
  * @param {string} error
  */
-function escalateError(error) {
-  dialog.showErrorBox('An unrecoverable error occurred', error);
+function escalateError(error, unrecoverable = true) {
+  if (unrecoverable) {
+    dialog.showErrorBox('An unrecoverable error occurred', error);
+    //TODO: should we quit here
```
cpvalente, line 115: I dont have a strong opinion.
If I had to throw an opinion dice I would say: "yes if it is not too much work"

apps/server/src/app.ts
```diff
@@ -176,12 +176,20 @@ export const initAssets = async () => {
  * Starts servers
  */
 export const startServer = async (
-  escalateErrorFn?: (error: string) => void,
+  escalateErrorFn?: (error: string, unrecoverable: boolean) => void,
 ): Promise<{ message: string; serverPort: number }> => {
   checkStart(OntimeStartOrder.InitServer);
-  const { serverPort } = getDataProvider().getSettings();
+  // initialise logging service, escalateErrorFn is only exists in electron
```
cpvalente, line 182:
```suggestion
  // initialise logging service, escalateErrorFn only exists in electron
```

apps/server/src/app.ts
```diff
@@ -176,12 +176,20 @@ export const initAssets = async () => {
...
+  logger.init(escalateErrorFn);
+  const settings = getDataProvider().getSettings();
+  const { serverPort: desiredPort } = settings;
 
   expressServer = http.createServer(app);
+
+  // the express server must be started before the socket otherwise the on error eventlissner will not attach properly
```
cpvalente, line 189:
```suggestion
  // the express server must be started before the socket otherwise the on error event listener will not attach properly
```

apps/server/src/app.ts
```diff
@@ -324,3 +330,46 @@ process.on('uncaughtException', async (error) => {
 process.once('SIGHUP', async () => shutdown(0));
 process.once('SIGINT', async () => shutdown(0));
 process.once('SIGTERM', async () => shutdown(0));
+
+/**
+ * @description tries to open the server with the desired port, and if getting a `EADDRINUSE` will change to an efemeral port
```
cpvalente, line 335:
```suggestion
 * @description tries to open the server with the desired port, and if getting a `EADDRINUSE` will change to an ephemeral port
```

Questions: 
- what do we mean by ephemeral port? The app will move to a new port for its entire session. 
- would we want to persist the changes of the port?

apps/server/src/app.ts
```diff
@@ -324,3 +330,46 @@ process.on('uncaughtException', async (error) => {
...
+ * @returns {number} the resulting port number
+ * @throws any other server errors will result in a throw
+ */
+async function serverTryDesiredPort(server: http.Server, desiredPort: number): Promise<number> {
+  return new Promise((res) => {
+    expressServer.once('error', (e) => {
+      if (isDocker) throw e; // we should only move ports if we are in a desktp environment
+      if (testForPortInUser(e)) {
+        server.listen(0, '0.0.0.0', () => {
+          // @ts-expect-error TODO: find proper documentation for this api
```
cpvalente, line 347: can i help?

apps/server/src/classes/Logger.ts
```diff
@@ -103,8 +103,11 @@ class Logger {
    * @param origin
    * @param text
    */
-  error(origin: string, text: string) {
+  error(origin: string, text: string, notifyElectron = false) {
```
cpvalente, line 106: suggestion: we dont need to couple this logic to electron, should we instead call it `escalate` as in "you should call the escalate error callback?

apps/server/src/app.ts
```diff
@@ -324,3 +330,46 @@ process.on('uncaughtException', async (error) => {
...
+ * @description tries to open the server with the desired port, and if getting a `EADDRINUSE` will change to an efemeral port
+ * @param {http.Server}server http server object
+ * @param {number}desiredPort the desired port
+ * @returns {number} the resulting port number
+ * @throws any other server errors will result in a throw
+ */
+async function serverTryDesiredPort(server: http.Server, desiredPort: number): Promise<number> {
+  return new Promise((res) => {
+    expressServer.once('error', (e) => {
+      if (isDocker) throw e; // we should only move ports if we are in a desktp environment
```
cpvalente, line 344:
```suggestion
      if (isDocker) throw e; // we should only move ports if we are in a desktop environment
```

Also, question: will Ontime Cloud be have the NODE_ENV set to `docker`? I could actually not find where do we set this 😅 . I wonder if we need a more reliable way to check if we are in electron

apps/server/src/app.ts
```diff
@@ -324,3 +330,46 @@ process.on('uncaughtException', async (error) => {
...
 process.once('SIGTERM', async () => shutdown(0));
+
+/**
+ * @description tries to open the server with the desired port, and if getting a `EADDRINUSE` will change to an efemeral port
+ * @param {http.Server}server http server object
+ * @param {number}desiredPort the desired port
+ * @returns {number} the resulting port number
+ * @throws any other server errors will result in a throw
+ */
+async function serverTryDesiredPort(server: http.Server, desiredPort: number): Promise<number> {
```
cpvalente, line 341: cleanup: these functions should be extracted somewhere, maybe a new `network.ts` file in the utils directory?

apps/server/src/app.ts
```diff
@@ -324,3 +330,46 @@ process.on('uncaughtException', async (error) => {
...
+  return new Promise((res) => {
+    expressServer.once('error', (e) => {
+      if (isDocker) throw e; // we should only move ports if we are in a desktp environment
+      if (testForPortInUser(e)) {
+        server.listen(0, '0.0.0.0', () => {
+          // @ts-expect-error TODO: find proper documentation for this api
+          const port: number = server.address().port;
```
cpvalente, line 348: nice, i did not know about this API

apps/server/src/app.ts
```diff
@@ -324,3 +330,46 @@ process.on('uncaughtException', async (error) => {
...
+          logger.error(
+            LogOrigin.Server,
+            `Failed open the desired port: ${desiredPort} \nMoved to an Ephemeral port: ${port}`,
+            true,
+          );
+
+          res(port);
+        });
+      } else {
+        throw e;
```
cpvalente, line 359: question: should we reject the promise instead of throwing?

apps/server/src/app.ts
alex-Arc: it is set in the docker build file
yes, a isElectron would be good, where could we set that?

apps/server/src/app.ts
alex-Arc: there we reach the limit of my knowledge
the goal is to halt the program, how we do that best?

apps/server/src/app.ts
alex-Arc: as far as i understand that is the name of the range of random ports provided by the OS - [Ephemeral port](https://en.wikipedia.org/wiki/Ephemeral_port)
 a better message is welcome

as written now it is saved to the settings... I think we talked about that both approaches has upsides, now that it only targets electron I think it leans a bit more to the, "it should be saved" sider

apps/server/src/app.ts
alex-Arc: ts dosn't like the type returned
[serveraddress](https://nodejs.org/api/net.html#serveraddress)
it might be just be a case of having to narrow it down

apps/server/src/app.ts
cpvalente: the promise constructor gives you two callbacks, reject and resolve (see [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise))
you are using the resolve one, you could just use the reject instead of throwing

something like (untested)

```ts
async function serverTryDesiredPort(server: http.Server, desiredPort: number): Promise<number> {
  return new Promise((resolve, reject) => {
    expressServer.once('error', (error) => {
      if (isDocker) reject({message: error}); // <-----
      if (testForPortInUser(error)) {
        server.listen(0, '0.0.0.0', () => {
          const port: number = server.address().port;

          logger.error(
            LogOrigin.Server,
            `Failed open the desired port: ${desiredPort} \nMoved to an Ephemeral port: ${port}`,
            true,
          );

          resolve(port);
        });
      } else {
        reject({message: error});  // <-----
      }
    });
    server.listen(desiredPort, '0.0.0.0', () => {
      const port: number = server.address().port;
      resolve(port);
    });
  });
}
```

apps/server/src/app.ts
cpvalente: I see, this makes sense. 
From the naming, I would assume that the ports are short lived, which may not be true since a) they last the whole session and b) we may even be saving it

apps/server/src/app.ts
cpvalente: tricky, from the docs, I am not sure what this means

> For a server listening on a pipe or Unix domain socket, the name is returned as a string.

so we have two options:
- a) create a type guard to check what the response is
- b) if we are sure that in our cases we always get the AddressInfo, we can just cast the response like

```ts
          const {port} = server.address() as AddressInfo;
```

b) could be fine, as if we get a string in the first example we would need to decide what to do

apps/electron/src/main.js
```diff
@@ -109,8 +109,13 @@ function askToQuit() {
  * Allows processes to escalate errors to be shown in electron
  * @param {string} error
  */
-function escalateError(error) {
-  dialog.showErrorBox('An unrecoverable error occurred', error);
+function escalateError(error, unrecoverable = true) {
```
cpvalente, line 112: question: I am not sure we can have the unrecoverable defaulting to true now that we shutdown the app

apps/server/src/utils/network.ts
```diff
@@ -0,0 +1,83 @@
...
+          logger.error(
+            LogOrigin.Server,
+            `Failed open the desired port: ${desiredPort} \nMoved to an Ephemeral port: ${port}`,
+            true,
+          );
+
+          resolve(port);
+        });
+      } else {
+        throw e;
```
cpvalente, line 63: question: could we use the promise rejection here?

apps/server/src/utils/network.ts
```diff
@@ -0,0 +1,83 @@
...
+ * @description tries to open the server with the desired port, and if getting a `EADDRINUSE` will change to an efemeral port
+ * @param {http.Server}server http server object
+ * @param {number}desiredPort the desired port
+ * @returns {number} the resulting port number
+ * @throws any other server errors will result in a throw
+ */
+export async function serverTryDesiredPort(server: http.Server, desiredPort: number): Promise<number> {
+  return new Promise((resolve, reject) => {
+    server.once('error', (e) => {
+      if (isDocker) reject('test rejection'); // we should only move ports if we are in a desktop environment
```
cpvalente, line 45:
```suggestion
      if (isDocker) reject({ message: 'Unable to change ports in docker' });
```

apps/server/src/utils/network.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    });
+    server.listen(desiredPort, '0.0.0.0', () => {
+      const address = server.address();
+      if (typeof address !== 'object') {
+        reject('unknown port type, can not proceed');
+        return; // the return is needed here to let TS know that we wont continue
+      }
+      const port = address.port;
+      resolve(port);
```
cpvalente, line 73: cleanup: we dont need the intermediate assignment since we are not doing manipulations or relabelling

```suggestion
      resolve(address.port);
```

cpvalente (commented):
I think this is OK, although I didnt test, did we have some pending tasks here?
Meanwhile I left some small cleanup comments for you, we should be able to merge after that

alex-Arc:
> I think this is OK, although I didnt test, did we have some pending tasks here?

yeah it is difficult to test the electron <-> server interaction in dev, one more thing that could be nice to improve at some point

cpvalente (approved):

PR #1350, opened 2024-11-30
fix: prevent stale data in finder

cpvalente:
This PR fixes an issue where the finder would keep the previous filtering results even if the data set changed


apps/client/src/features/editors/finder/useFinder.tsx
```diff
@@ -29,6 +29,12 @@ export default function useFinder() {
   const [results, setResults] = useState<FilterableEntry[]>([]);
   const [error, setError] = useState<MaybeString>(null);
 
+  /** clear results when source data changes */
+  useEffect(() => {
+    setResults([]);
```
alex-Arc, line 34: this clears out the result if you are searching while an update arrives
could we maybe hold the search query in state also and then rerun the search if the data changes

cpvalente: good catch

Issue #1356, opened 2024-12-01
Auxiliary Timer Add-Subtract Time


STEMgrh:
I'm working on creating an Ontime workflow with OSC for a tv show I work on, and I would love to see add and subtract commands in the API for the Aux timer!

cpvalente:
Thank you for taking the time to reach out
The Auxiliary timer current allows setting the time but not increments / decrements

Do I understand correctly that you would like to add or remove an amount of time to a running auxiliary timer?
Are there other API requirements you have for the Auxiliary timer?

It would be good to understand your use case, if you are willing to share:
- what is the use case in which you miss the ability to add / remove time?
- what are you using to control ontime?
- how do you use the auxiliary timer?

cpvalente:
Implemented and released as part of 3.9.0
@STEMgrh , can you please double-check that this is what you meant?

See the docs for the implementation API
https://docs.getontime.no/api/protocols/osc/#add--remove-time-to-auxiliary-timer

PR #1357, opened 2024-12-01
feat: allow add time to aux timer

cpvalente:
exposes an endpoint for adding time to the aux timer. as described in #1356


7omas:
Saw this commit and started adding this functionality to the Chataigne-ontime-module.

When trying to add or remove time while the timer is running it changes way more than requested. The requested addtime is applied but then more time gets changed seemingly arbitraty. (sometimes seconds, sometimes minutes)

Expected behaviour: The same as sending a change on the main timer during running event.

@cpvalente @alex-Arc

PR #1359, opened 2024-12-02
refactor: add prefix to relative assets

cpvalente:


apps/client/index.html
```diff
@@ -1,31 +1,27 @@
...
+    <meta name="ontime" content="ontime - time keeping for live events" />
+    <link rel="apple-touch-icon" href="/ontime-logo.png" />
+    <link rel="icon" type="image/png" href="/ontime-logo.png" />
+    <link rel="manifest" href="/site.webmanifest" />
+    <link rel="manifest" href="/manifest.json" />
+    <title>ontime</title>
+    <style>
+      body,
+      html {
+        background-color: #101010 !important;
```
alex-Arc, line 18: I think there was a reason this was all black, but I don't remember
maybe leave it for now?
 Unless you know

cpvalente: good call, I seem to remember this was to avoid colour flashes on load, but I may be remembering wrong, anyway, it is not worth changing

PR #1360, opened 2024-12-02
Calculate time util all events start

alex-Arc:
Ignore all the elements in the client it is only to see the reasults


packages/utils/src/time-utils/timeUtils.ts
```diff
@@ -0,0 +1,59 @@
+import type { MaybeNumber, RundownCached } from 'ontime-types';
+import { isOntimeEvent } from 'ontime-types';
+
+import { dayInMs } from '../date-utils/conversionUtils';
+
+export function calculateExpectedStart(
```
cpvalente, line 6: this is a lot of work!
i wonder if our current normalised rundown should be doing this instead?

PR #1363, opened 2024-12-04
create addTime function in simple timer

alex-Arc:
@7omas thanks for finding this
#1357

### DIFF
--- a/apps/server/src/classes/simple-timer/SimpleTimer.ts
+++ b/apps/server/src/classes/simple-timer/SimpleTimer.ts
@@ -37,10 +37,12 @@ export class SimpleTimer {
     return this.state;
   }
 
-  public addTime(millis: number, timeNow: number): SimpleTimerState {
+  public addTime(millis: number): SimpleTimerState {
     this.state.duration += millis;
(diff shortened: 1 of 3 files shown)


apps/server/src/classes/simple-timer/SimpleTimer.ts
```diff
@@ -37,6 +37,12 @@ export class SimpleTimer {
     return this.state;
   }
 
+  public addTime(millis: number, timeNow: number): SimpleTimerState {
+    this.state.duration += millis;
+    this.state.current += millis; // optimistic update of curret value in case of pause or stop
```
cpvalente, line 42:
```suggestion
    this.state.current += millis; // optimistic update in case of pause or stop
```

apps/server/src/classes/simple-timer/SimpleTimer.ts
```diff
@@ -37,6 +37,12 @@ export class SimpleTimer {
...
+    return this.update(timeNow);
```
cpvalente, line 43: i dont think these function should do multiple things
if what we want is to add time and update, then we should call the two things

cpvalente (approved):
great work

PR #1368, opened 2024-12-06
fix: secure protocol resolution

cpvalente:


apps/client/src/externals.ts
```diff
@@ -34,9 +34,12 @@ export const websocketUrl = resolveUrl('ws', 'ws');
 function resolveUrl(protocol: 'http' | 'ws', path: string) {
   const url = new URL(window.location.origin);
 
-  // check if protocol URL is secure
-  url.protocol = protocol === 'http' ? 'http' : 'ws';
-  url.protocol += window.location.protocol === 'https:' ? 's' : '';
+  // generate ws url
+  if (protocol === 'ws') {
+    // ensure we remain in a secure context
```
cpvalente, line 39: the previous code grabbed a protocol like `ws:` and added an s to make `ws:s` which the URL instance seemed to sanitise for us

alex-Arc (approved):

PR #1372, opened 2024-12-10
Edit in cuesheet

alex-Arc:


apps/client/src/views/cuesheet/CuesheetPage.tsx
```diff
@@ -53,29 +53,44 @@ export default function CuesheetPage() {
...
+        // skip if there is no value change
+        if (previousValue === payload) {
+          return;
+        }
+
+        // check if value is valid
+        // in anticipation to different types of event here
+        if (typeof payload !== 'string') {
+          return;
+        }
```
cpvalente, line 69: question: can you explain this guard? I am not sure what we expect here?

apps/client/src/views/cuesheet/CuesheetPage.tsx
```diff
@@ -53,29 +53,44 @@ export default function CuesheetPage() {
...
+
+        // submit
+        try {
+          const key = accessor.split('custom_')[1];
+          await updateCustomField(event.id, key, payload);
+        } catch (error) {
+          console.error(error);
+        }
```
cpvalente, line 77:
```suggestion
        try {
          const key = accessor.split('custom_')[1];
          await updateCustomField(event.id, key, payload);
        } catch (_error) {
          /** unhandled error */
        }
```

apps/client/src/views/cuesheet/cuesheet-table-elements/EditableCell.tsx
```diff
@@ -36,18 +45,35 @@ const EditableCell = (props: EditableCellProps) => {
...
+        spellCheck={false}
+        style={{ padding: 0 }}
+      />
+    );
+  }
+
   return (
-    <AutoTextArea
+    <Input
       size='sm'
```
cpvalente, line 67: tweak: the input needs to be `xs` to match the size of the text area `sm` 🤷‍♂️

apps/client/src/views/cuesheet/CuesheetPage.tsx
```diff
@@ -53,29 +53,44 @@ export default function CuesheetPage() {
         return;
       }
 
-      const previousValue = event.custom[accessor];
-
-      if (previousValue === payload) {
+      // is it a custom fields
```
cpvalente, line 56: question: this function is becoming a bit hairy, do you think we could clean improve this?

apps/client/src/views/cuesheet/cuesheet-table-elements/EditableCell.tsx
```diff
@@ -1,27 +1,36 @@
...
 
 interface EditableCellProps {
   value: string;
   handleUpdate: (newValue: string) => void;
+  isMultiLine: boolean;
 }
 
 const EditableCell = (props: EditableCellProps) => {
-  const { value: initialValue, handleUpdate } = props;
+  const { value: initialValue, handleUpdate, isMultiLine } = props;
```
cpvalente, line 14: question: did you look into using the `useReactiveInput` hook? it could help with the submit logic

apps/client/src/views/cuesheet/cuesheetCols.tsx
```diff
@@ -11,9 +11,25 @@ import { useCuesheetSettings } from './store/cuesheetSettingsStore';
...
+
+  const event = row.original;
+  if (!isOntimeEvent(event)) {
+    return null;
+  }
+
+  const isChecked = event.isPublic;
+
+  //TODO: css styleing
+  return <Checkbox onChange={update} isChecked={isChecked} />;
```
cpvalente, line 32: I think it looks ok, maybe just add
```ts
style={{ verticalAlign: 'middle' }}
```

Also, this field is a <Toggle> everywhere else, do we prefer a checkbox here?

apps/client/src/views/cuesheet/cuesheet-table-elements/EditableCell.tsx
```diff
@@ -36,18 +45,35 @@ const EditableCell = (props: EditableCellProps) => {
     setValue(initialValue);
   }, [initialValue]);
 
+  if (isMultiLine) {
```
cpvalente, line 48: question: do you think it is worth making two components? it seems like we want quite a bit of logic to be different?

cpvalente (commented):
Good stuff, I think in general it looks good, but we may want to clean up since we know there will be a lot of changes here

Also, take a look at how difficult would it be to allow editing times as well, we can do it in another PR

apps/client/src/views/cuesheet/cuesheetCols.tsx
alex-Arc: maybe, was trying to keep it looking as close to the old cuesheet

apps/client/src/views/cuesheet/CuesheetPage.tsx
alex-Arc: left over from the original logic, I can clean it up a bit

e2e/tests/features/202-cuesheet.spec.ts
```diff
@@ -1,11 +1,20 @@
...
-  await page.getByRole('cell', { name: 'Lunch break' }).click();
-  await page.getByRole('cell', { name: 'Albania' }).click();
-  await page.getByRole('cell', { name: 'Latvia' }).click();
-  await page.getByRole('cell', { name: 'Lithuania' }).click();
+  await expect(page.getByText('Eurovision Song Contest')).toBeVisible();
+  await expect(page.getByRole('row', { name: 'Lunch break' })).toBeVisible();
+
+  await expect(
+    page.getByRole('row', { name: 'SF1.01 10:00:00 10:20:00 00:20:00' }).getByRole('textbox').first(),
+  ).toHaveValue('Albania');
```
cpvalente, line 11: any changes to the data would make this tests break. we need something a bit more generic that wont be so painful to maintain

apps/client/src/views/cuesheet/CuesheetPage.tsx
```diff
@@ -106,6 +120,7 @@ export default function CuesheetPage() {
         data={flatRundown}
         columns={columns}
         handleUpdate={handleUpdate}
+        handleUpdateCustom={handleUpdateCustom}
         selectedId={featureData.selectedEventId}
         currentBlockId={featureData.currentBlockId}
```
cpvalente, line 125: these will change pretty often, I wonder if this is an easy point for improvement

The `currentBlockId` is only used in the rows, we could create a row element that gets this data from the socket
We could further stabilise this by creating a selector on whether the `currentBlockIsSelected`

the `selectedId` is used to make `isPast` and a bit trickier to stabilise but it could be worth it

We shouldnt do that now but keep in mind for a near future improvement task

cpvalente (approved):

PR #1374, opened 2024-12-11
chore: note service limitations

cpvalente:
adding a small note that OSC doesnt work
I couldnt think of anything else that would not work

do you think the note is good enough?
Screenshot 2024-12-11 at 21 25 03


alex-Arc (approved):

cpvalente:
@alex-Arc , made some more changes to prevent the services from starting if we know they cant work
it would be great if you could take an extra look

alex-Arc (approved):

Issue #1376, opened 2024-12-12
Paste not possible in electron Interface


RasterboxOff:
A client is figuring out Ontime and he can't paste text into the electron interface. That makes creating presets not really viable. 
It works when pasting into an editor that runs in a browser. 

Apple M1
MacOS Sonoma 14.6.1
MacBook Pro 17,1

Ontime 3.9.4

cpvalente:
Thank you,

it turned out this was introduced when we removed the edit menu from the electron app
We will have a fix in the next release

PR #1378, opened 2024-12-12
fix: recover edit menu and shortcuts

cpvalente:
fixes the issue with some shortcuts not working in macOS as reported in #1376


alex-Arc (approved):
I can't test this one

PR #1384, opened 2024-12-13
feat(ui): add warning in view params when styles are overriden

asharonbaltazar:
Users enable their own custom CSS to style the views. Making any changes to views from the settings will potentially clash with their custom CSS settings. This PR adds a warning before saving their settings. 

Screenshot 2024-12-13 at 12 07 00 PM


cpvalente (approved):
Thank you @asharonbaltazar , this is simple and to the point

I am discussing alternatives for the presentation and will follow up on that. If you dont mind, we will tweak presentation and wording before merging

Meanwhile, this looks good to me, thank you for your work

PR #1390, opened 2024-12-15
refactor: extract time-to-end

cpvalente:
We can revise the wording if we get some feedback from users


apps/server/src/utils/parserFunctions.ts
```diff
@@ -349,3 +347,22 @@ export function sanitiseCustomFields(data: object): CustomFields {
 
   return newCustomFields;
 }
+
+/**
+ * Time to end was moved from a TimerType to a standalone boolean
+ * Released as part of v3.10.0
+ */
+function migrateTimeToEnd(event: any): OntimeEvent {
```
alex-Arc, line 355: Yeah we were going to have to do it at some point

alex-Arc (approved):

PR #1391, opened 2024-12-17
Cuesheet edit

cpvalente:
This is a refactor PR to tidy up the cuesheet
It also
- enable editing the cue property
- add action menu to every row

I will continue the missing work (editing times) in a separate PR

We can also consider run and edit mode for the change the displayed items (in a separate PR)


alex-Arc (changes requested):
There is something funky with the insert
video

and there is an offset between headers and cells

alex-Arc (approved):

PR #1392, opened 2024-12-17
feat: create offline indicator

cpvalente:
adding some simple styles for an offline indicator
dashboard has opacity and we show a small label in the bottom right corner

Screenshot 2024-12-17 at 21 43 14

### DIFF
--- a/apps/client/src/features/overview/Overview.module.scss
+++ b/apps/client/src/features/overview/Overview.module.scss
@@ -4,6 +4,23 @@
   display: flex;
 }
 
+.isOffline {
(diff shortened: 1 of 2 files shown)


alex-Arc:
Looks good
Should maybe the whole thing be greyed out?
We could so switch to our offline logo for the tab icon
The OP should also have this type

alex-Arc (commented):
Looks good
Should maybe the whole thing be greyed out?
We could so switch to our offline logo for the tab icon
The OP should also have this type

PR #1394, opened 2024-12-18
Display Time Until count down in editor UI

alex-Arc:


apps/client/src/common/hooks/useSocket.ts
```diff
@@ -250,3 +250,12 @@ export const useIsOnline = () => {
 
   return useRuntimeStore(featureSelector);
 };
+
+/** convert ping into a derived value which changes less often */
+export const usePlayback = () => {
```
cpvalente, line 255: I think this is in master, maybe you need to rebase?

apps/client/src/features/control/playback/add-time/AddTime.tsx
```diff
@@ -43,12 +43,22 @@ export default function AddTime(props: AddTimeProps) {
       <TimeInput name='addtime' submitHandler={handleTimeChange} time={time} placeholder='Add time' />
       <div className={style.addButtons}>
         <Tooltip label='Remove time' openDelay={tooltipDelayMid} shouldWrapChildren>
-          <TapButton onClick={() => handleAddTime('remove')} disabled={doDisableButtons} className={style.tallButtons}>
+          <TapButton
+            onClick={() => handleAddTime('remove')}
+            disabled={doDisableButtons}
+            className={style.tallButtons}
+            name='removetime'
```
cpvalente, line 50: suggestion: I think `aria-label` is the correct property for describing an element without a name

apps/client/src/features/control/playback/tap-button/TapButton.tsx
```diff
@@ -13,14 +13,15 @@ interface TapButtonProps {
...
-  const { children, disabled, onClick, theme = 'neutral', aspect = 'normal', active, className } = props;
+  const { children, disabled, onClick, theme = 'neutral', aspect = 'normal', active, className, name } = props;
   const classes = cx([style.tapButton, className, style[theme], style[aspect], active ? style.active : null]);
 
   return (
-    <button className={classes} disabled={disabled} type='button' onClick={onClick} ref={ref}>
+    <button className={classes} disabled={disabled} type='button' onClick={onClick} ref={ref} name={name}>
       {children}
     </button>
   );
```
cpvalente, line 27: suggestion: what if we modify the type signature to extend button? that would make it easier to consume and compose. 

like 
```ts

interface TapButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
  aspect?: 'normal' | 'square' | 'fill' | 'tight';
  free?: boolean;
  theme?: Playback | 'neutral';
  active?: boolean;
}
```

then we consume like
```ts
const TapButton = () => {
  const { children, theme = 'neutral', aspect = 'normal', active, className, ...elementProps } = props;
  ...
  return (
    <button {...elementProps} className={classes} type='button' ref={ref}>
      {children}
    </button>
  );
}
```

This code also makes it clearer for us to understand what are the properties we override

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -294,6 +298,10 @@ export default function Rundown({ data }: RundownProps) {
                 lastEvent = thisEvent;
 
                 if (isPlayableEvent(entry)) {
+                  if (!isPast) {
+                    //TODO: how to handle overlaps
+                    accumulatedGap = calculateAccumulatedGap(accumulatedGap, entry, lastEvent);
```
cpvalente, line 303: question: is this not coming from the other PR?

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
```diff
@@ -0,0 +1,85 @@
...
+        className={className}
+        timeStart={timeStart + (isNextDay ? dayInMs : 0)}
+        accumulatedGap={accumulatedGap}
+        isNextAndLinked={isNext && isLinked}
+      />
+    );
+  }
+
+  return null;
+}
```
cpvalente, line 51: question: maybe we dont need the extra element?

at this point, we know that the playback is active (we returned earlier in the guard before) so we could just have the logic of `EventUntil` here

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+    const consumedOffset = isNextAndLinked ? offset : Math.min(offset + gap, 0);
+    const offsetTimestart = timeStart - consumedOffset;
+    const timeUntil = offsetTimestart - clock;
+    const isDue = timeUntil < MILLIS_PER_SECOND;
+    return [isDue ? 'DUE' : `${formatDuration(Math.abs(timeUntil), timeUntil > 2 * MILLIS_PER_MINUTE)}`, isDue];
+  }, [accumulatedGap, clock, isNextAndLinked, offset, timeStart]);
+
+  if (offset === null) {
+    return null; //TODO: change offset to a maybe number
```
cpvalente, line 75: question: can offset be null?

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
alex-Arc: hmm can't  use react hooks conditionally I think?

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
alex-Arc: not yet, but I would like to make it so
it makes some things easier and I think it makes sense to say that nothing is playing so we don't have a valid offset value
and in this specific case is solves the issue of the playback playback triggerring a render befor offset has a valid value

apps/client/src/features/control/playback/tap-button/TapButton.tsx
```diff
@@ -1,11 +1,11 @@
-import { ForwardedRef, forwardRef, PropsWithChildren } from 'react';
+import { ButtonHTMLAttributes, ForwardedRef, forwardRef, PropsWithChildren } from 'react';
 import { Playback } from 'ontime-types';
 
 import { cx } from '../../../../common/utils/styleUtils';
 
 import style from './TapButton.module.scss';
 
-interface TapButtonProps {
+interface TapButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
```
cpvalente, line 8: cleanup: Since we extend the HTML Button attributes, we no longer need the normal stuff like `disabled` `onClick` and `className`

apps/client/src/features/control/playback/tap-button/TapButton.tsx
```diff
@@ -16,11 +16,20 @@ interface TapButtonProps {
...
+  const {
+    children,
+    disabled,
+    onClick,
+    theme = 'neutral',
+    aspect = 'normal',
+    active,
+    className,
+    ...elementProps
+  } = props;
```
cpvalente, line 28: cleanup: we no longer need to destructure the `disabled` and `onClick` props

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
cpvalente: I think we should only write code for a null offset when that is a valid case

As for making offset nullable, as I remember, the reason to make it always a number was to avoid a bunch of checks at runtime. but if you think this is better we can change it in a separate PR

apps/client/src/features/rundown/Rundown.tsx
alex-Arc: yes patially

apps/client/src/features/rundown/event-block/EventBlock.utils.ts
```diff
@@ -39,3 +40,29 @@ export function formatOverlap(timeStart: number, previousStart?: number, previou
   const overlapString = formatDuration(Math.abs(timeFromPrevious), false);
   return `${timeFromPrevious < 0 ? 'Overlap' : 'Gap'} ${overlapString}`;
 }
+
+export function calculateAccumulatedGap(
```
cpvalente, line 44: question: would it not be simpler for the accumulated gap to always be a number?

apps/client/src/features/rundown/event-block/EventBlock.utils.ts
```diff
@@ -39,3 +40,29 @@ export function formatOverlap(timeStart: number, previousStart?: number, previou
...
+  accGap: MaybeNumber,
+  currentEvent: PlayableEvent,
+  previousEvent?: PlayableEvent,
+): MaybeNumber {
+  if (previousEvent === undefined) return null;
```
cpvalente, line 49: suggestions: 

- we likely prefer spelled out variable names like `accumulatedGap` rather than `accGap`. Short names is the compilers job

- we would likely want some comments in this function to clarify what are the covered cases ie

`// check if there is an event before`
`// check if there is a gap from the previous event`
`// check if event is the day after`

apps/client/src/features/rundown/event-block/EventBlock.tsx
```diff
@@ -264,6 +267,11 @@ export default function EventBlock(props: EventBlockProps) {
     setSelectedEvents({ id: eventId, index, selectMode: editMode });
   };
 
+  const isNextDay =
+    previousStart === undefined || previousEnd === undefined
+      ? false
+      : checkIsNextDay(previousStart ?? 0, timeStart, calculateDuration(previousStart, previousEnd));
+
```
cpvalente, line 274: question: we calculated this when we calculated the gap, maybe we could return an object of metadata like
```ts

export function calculateAccumulatedGap(): { gap: number, isNextDay: boolean }

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
cpvalente: ah yes, good point, we should just add a comment for that like
/* we extracted the component to avoid unnecessary calculations and re-renders */

PR #1396, opened 2024-12-18
Dayoffset in rundown

alex-Arc:


apps/client/src/features/rundown/RundownEntry.tsx
```diff
@@ -75,33 +49,6 @@ export default function RundownEntry(props: RundownEntryProps) {
 
   const actionHandler = useMemoisedFn((action: EventItemActions, payload?: number | FieldValue) => {
     switch (action) {
-      case 'event': {
```
cpvalente, line 78: question: what was the motivation for removing these actions?

apps/client/src/features/rundown/event-block/EventBlock.utils.ts
```diff
@@ -18,24 +12,9 @@ export function formatDelay(timeStart: number, delay: number): string | undefine
   return `New start ${timeTag}`;
 }
 
-export function formatOverlap(timeStart: number, previousStart?: number, previousEnd?: number): string | undefined {
-  const noPreviousElement = previousEnd === undefined || previousStart === undefined;
-  if (noPreviousElement) return;
+export function formatOverlap(gap: MaybeNumber, isNextDay: boolean): string | undefined {
```
cpvalente, line 15: suggestion, could we add some jsdocs here to explain whats going on?
```suggestion
/**
 * Creates a string representation of the overlap or gap between two events.
 * @example formatOverlap(1000, false) => "Gap 1s"
 * @example formatOverlap(-1000, false) => "Overlap 1s"
 * @example formatOverlap(1000, true) => "Gap 1s (next day)"
 */
export function formatOverlap(gap: MaybeNumber, isNextDay: boolean): string | undefined {
```

apps/client/src/features/rundown/event-block/EventBlock.utils.ts
```diff
@@ -18,24 +12,9 @@ export function formatDelay(timeStart: number, delay: number): string | undefine
...
+  if (gap === null || gap === 0) return;
```
cpvalente, line 16: suggestion: is it more expressive to say that we want gap to have a value?

```suggestion
  // we only care if gap has a value non-zero and non-nullish
  if (!gap) return;
```

packages/types/src/definitions/core/OntimeEvent.type.ts
```diff
@@ -39,6 +39,7 @@ export type OntimeEvent = OntimeBaseEvent & {
   colour: string;
   revision: number;
   delay?: number; // calculated at runtime
+  dayOffset?: number; // calculated at runtime
```
cpvalente, line 42: suggestion: I wonder if we are starting to complicate our code base a bit with these types

would it be better to have a `OntimePersistedEvent` and an `OntimeEvent`, where the `OntimeEvent`contains all the stuff calculated in `rundownCache`?

packages/utils/src/date-utils/getTimeFromPrevious.ts
```diff
@@ -1,45 +1,28 @@
...
 /**
  * Utility returns the time elapsed (gap or overlap) from the previous
  * It uses deconstructed parameters to simplify implementation in UI
+ * @param currentStart the start time of the current event in ms
+ * @param currentDayOffset total number of days the current event is offset by
+ * @param previousStart the start time of the previous event in ms
+ * @param previousDuration the duration of the previous event in ms
+ * @param previousDayOffset total number of days the previous event is offset by
  */
 export function getTimeFromPrevious(
```
cpvalente, line 12: nice!

apps/server/src/services/rundown-service/rundownCache.ts
```diff
@@ -112,11 +115,18 @@ export function generate(
...
 
+        const isNextDay =
+          lastEntry === null ? 0 : checkIsNextDay(lastEntry.timeStart, currentEntry.timeStart, lastEntry.duration);
+
+        dayOffset += isNextDay ? 1 : 0;
+        currentEntry.dayOffset = dayOffset;
+
         const timeFromPrevious: number = getTimeFromPrevious(
           currentEntry.timeStart,
+          currentEntry.dayOffset,
```
cpvalente, line 126: cleanup: we need to make sure that these properties do not get persisted. I have flagged the entrypoint for the code in this comment
https://github.com/cpvalente/ontime/blob/master/apps/server/src/services/rundown-service/rundownCache.ts#L97

cpvalente (commented):
Thank you Alex, I agree that this changes will lead to a better experience dealing with this dataset

I think we are moving in a direction where the server is providing us more metadata. I suggest that we spend some time to cleanup typing and processes around this

apps/client/src/features/rundown/RundownEntry.tsx
alex-Arc: They are not used and allow us to remove quite a few props from the component

packages/types/src/definitions/core/OntimeEvent.type.ts
alex-Arc: Yeah tat sounds like a good idea
I've been struggling a bit with the optional so making a new type would be great 

Should we make a new type for pre or post cache.generate ?

cpvalente: I am not sure if I understood. but I think the two types we need are
- one type with the base data we need, this is stored in the json file
- one type with added metadata, this will be in all the runtime stores and generated in the rundownCache

alex-Arc: should the "normal" `OntimeEvent` type be the on with meta data since that is what is being operated on in most of the code base, and then we create a new `dbOntimeEvent` with out the metadata?

cpvalente: I am not sure if there is something definitely better. I think your suggestion is good and makes sense

PR #1398, opened 2024-12-20
Cuesheet times

cpvalente:
This PR allows editing duration in the cuesheet.
Once we are happy with the experience of this element, I will create a second PR that applies the same changes to the start and end times

The new implementation introduces a new Element `TextLikeInput` which handles the switch between a piece of text and an input
This will allow us to have different presentation and input elements and simplify things going forward

Also in the same change, we introduce a new flag to the time update function which allows locking the time strategy on update. This is to make the time input in the cuesheet simpler, without needing to use the lock icons and buttons

it also makes a few tweaks to the styles of the cuesheet
most significant:
- extract action button to visually outside the table (the idea is to be able to hide this later without the cuesheet changing)
- durations are shown in the smallest time unit like "10h2m"

### DIFF
--- a/apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/CuesheetHeader.tsx
+++ b/apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/CuesheetHeader.tsx
@@ -23,8 +23,8 @@ export default function CuesheetHeader(props: CuesheetHeaderProps) {
 
         return (
           <tr key={headerGroup.id}>
-            <th className={style.indexColumn}>{showIndexColumn && '#'}</th>
             <th className={style.actionColumn} />
+            <th className={style.indexColumn}>{showIndexColumn && '#'}</th>
             <SortableContext key={key} items={headerGroup.headers} strategy={horizontalListSortingStrategy}>
               {headerGroup.headers.map((header) => {
                 const width = header.getSize();
--- a/apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/CuesheetHeader.tsx
+++ b/apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/CuesheetHeader.tsx
@@ -3,6 +3,7 @@ import { flexRender, HeaderGroup } from '@tanstack/react-table';
 import { OntimeRundownEntry } from 'ontime-types';
 
 import { getAccessibleColour } from '../../../../common/utils/styleUtils';
+import { useCuesheetOptions } from '../../cuesheet.options';
 
 import { SortableCell } from './SortableCell';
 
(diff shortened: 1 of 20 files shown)


cpvalente:
@coderabbitai full review

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/CuesheetHeader.tsx
```diff
@@ -23,8 +23,8 @@ export default function CuesheetHeader(props: CuesheetHeaderProps) {
 
         return (
           <tr key={headerGroup.id}>
-            <th className={style.indexColumn}>{showIndexColumn && '#'}</th>
             <th className={style.actionColumn} />
+            <th className={style.indexColumn}>{showIndexColumn && '#'}</th>
```
alex-Arc, line 27: Is it intended that the column remains when the index is hidden?

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
```diff
@@ -27,18 +28,24 @@ function MakeTimer({ getValue, row: { original } }: CellContext<OntimeRundownEnt
   );
 }
 
-function MakeDuration({ getValue }: CellContext<OntimeRundownEntry, unknown>) {
-  const { hideTableSeconds } = useCuesheetOptions();
-  const cellValue = (getValue() as number | null) ?? 0;
+function MakeDuration({ getValue, row, table }: CellContext<OntimeRundownEntry, unknown>) {
```
alex-Arc, line 31: In `MakeTimer` moving the `DelayIndicator` to after the timer would look better i think

alex-Arc (approved):
Very nice

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/CuesheetHeader.tsx
cpvalente: I was wondering if we should always show the index column. What do you think?

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
cpvalente: Do you have an opinion on the delay format here?

> When an event contains delays, what should be displayed
- a) both the scheduled and delayed times in the table and the delay amount in the tooltip? (ie: "+10min")
- b) the scheduled time in the table and the new delayed time in the tooltip? (ie: "+10 min, new start 10:10:00")
- c) the delayed time in the table and the scheduled time in the tooltip? (ie: "+10min, scheduled start 10:00:00")

I would personally be inclined towards b)

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/CuesheetHeader.tsx
alex-Arc: well since I was the one who wanted it gone in the first place, I would say no
But if it can be shown in a way that is muted enough, like in the editor where it is clear that is it something other than the cue number

PR #1400, opened 2024-12-20
refactor: event editor color picker

### LINKED ISSUE #1397: create colour picker for event colour
In the `<EventEditor>` component, users can select a colour for the event
We currently allow the user a choice of pre-defined colours

The issue request is to create an option with a colour picker for users to be able to select a custom colour
The interface will likely look like the `<ViewParamsEditor>`

See the [entrypoint for the relevant code](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/rundown/event-editor/composite/EventEditorTitles.tsx#L49)

Shobhit-Nagpal:
closes #1397 

Attaching a video for the PR

https://github.com/user-attachments/assets/8c55d0a6-af3c-46d5-9a83-aaf3f0fff003

### Changes

- Made changes to `<PopoverPicker>` to take in `icon` and `hasInput` as optional props + added `<HexColorInput>` from react-colorful if hasInput is true
- Added `<EventColorPicker>` as a component which uses `<PopoverPicker>` passing in icon and hasInput props
- Added a `debounceWithValue` function under `utils/debounce`
- Debouncing changes on color picker, current wait time is set to 250ms (let me know if you'd like me to change this)
- Added styling on `<HexColorInput>` to look consistent with other input fields present

Hope this is what you had in mind, do let me know if there's any changes required :))

P.S: Super fun code base, super clean. Loved working on this


apps/client/src/features/rundown/event-editor/composite/EventColorPicker.tsx
```diff
@@ -0,0 +1,35 @@
...
+    },
+    [handleChange, name, value],
+  );
+
+  return (
+    <div className={style.inline}>
+      <PopoverPicker color={value} onChange={debouncedChange} icon={icon} hasInput={true} />
+      <input type='hidden' name={name} value={value} />
+    </div>
+  );
```
cpvalente, line 34: question: I dont think we need the inline styles or the wrapper element at all since the input is not visible..
As far as I can test, a fragment here can suffice

```ts
      <PopoverPicker color={value} onChange={debouncedChange} icon={icon} hasInput={true} />
```

apps/client/src/features/rundown/event-editor/composite/EventColorPicker.tsx
```diff
@@ -0,0 +1,35 @@
...
+      if (newValue !== value) {
+        handleChange(name, newValue);
+      }
+    },
+    [handleChange, name, value],
+  );
+
+  return (
+    <div className={style.inline}>
+      <PopoverPicker color={value} onChange={debouncedChange} icon={icon} hasInput={true} />
```
cpvalente, line 31: nitpick: as a project convention we tend not to omit boolean attributes, so 

```suggestion
      <PopoverPicker color={value} onChange={debouncedChange} icon={icon} hasInput />
```

apps/client/src/features/rundown/event-editor/composite/EventColorPicker.tsx
```diff
@@ -0,0 +1,35 @@
...
+  icon: ReactNode;
+  handleChange: (newValue: 'colour', name: string) => void;
+}
+
+export default function EventColorPicker(props: EventColorPickerProps) {
+  const { name, value, icon, handleChange } = props;
+
+  const debouncedChange = debounceWithValue((value: string) => {
+    setColour(value);
+  }, 250);
```
cpvalente, line 18: we should keep in mind that changing the element colour invalidates the entire array list of elements
unfortunately this amount of debouncing is doesnt seem sufficient to
a) prevent issues with re-renders
b) provide a good user experience with the picker

I wonder if it would be possible for us to only submit on `MouseUp` event

apps/client/src/features/rundown/event-editor/composite/EventEditorTitles.tsx
```diff
@@ -46,7 +48,10 @@ const EventEditorTitles = (props: EventEditorTitlesProps) => {
       </div>
       <div>
         <Editor.Label>Colour</Editor.Label>
-        <SwatchSelect name='colour' value={colour} handleChange={handleSubmit} />
+        <div className={style.colorSettings}>
```
cpvalente, line 51: unfortunately this still does not work for cases where the list must wrap 

I wonder if the easiest thing is to include `EventColorPicker` in the `SwatchSelect` list?

Screenshot 2024-12-21 at 10 12 10

apps/client/src/features/rundown/event-editor/composite/EventColorPicker.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { ReactNode, useCallback } from 'react';
+
+interface EventColorPickerProps {
+  name: 'colour';
+  value: string;
+  icon: ReactNode;
+  handleChange: (newValue: 'colour', name: string) => void;
+}
+
+export default function EventColorPicker(props: EventColorPickerProps) {
```
cpvalente, line 13: nice! I really appreciate the work you put into following project conventions

suggestion: I wonder if this would be better called `SwatchPicker` to signify its relationship to the other Swatch elements? We could also collocate this in the `components/input/colour-input directory`

apps/client/src/common/components/input/popover-picker/PopoverPicker.module.scss
```diff
@@ -14,3 +14,38 @@
...
+  &:hover {
+    background-color: $gray-1100;
+  }
+
+  &:focus {
+    background-color: $gray-1000;
+    color: $gray-50;
+    border: 1px solid $blue-500;
+  }
+}
```
cpvalente, line 51: great work, I appreciate that you went through the trouble of finding the styles

small correction: the `height` should be `2rem`
we should also remove the top `border-radius` in this context, so `border-radius: 0 0 3px 3px`

apps/client/src/common/components/input/popover-picker/PopoverPicker.module.scss
```diff
@@ -14,3 +14,38 @@
...
+  width: 30px;
+  height: 30px;
+  border-radius: 20px;
+  border: 0.2px solid white;
+  box-shadow: 0 0 0 0.2px $gray-300;
+  cursor: pointer;
+
+  transition-property: box-shadow;
+  transition-duration: $transition-time-action;
+}
```
cpvalente, line 30: cleanup: this styles should be mostly inherited from the `swatch` class. in fact, I believe we may not need the `icon` class

apps/client/src/features/rundown/event-editor/composite/EventEditorTitles.tsx
```diff
@@ -46,7 +48,10 @@ const EventEditorTitles = (props: EventEditorTitlesProps) => {
...
+          <SwatchSelect name='colour' value={colour} handleChange={handleSubmit} />
+          <EventColorPicker name='colour' value={colour} icon={<IoEyedrop size={15} />} handleChange={handleSubmit} />
```
cpvalente, line 53: questions: 

1. I believe the icon is at the wrong level of the abstraction.

I would think this should be included in the `EventColorPicker` element

We also dont need to inline the size for the icon, this should will come from the inherited styles or we can add it to the element styles in the CSS if necessary

2. The `Swatch` element compares the string to its constant to find out if a particular colour is the one selected. How could we write logic to know if the user has selected a custom colour and highlight the picker component? Maybe check the value against the colours array?

cpvalente (changes requested):
Thank you very much for this. The code is clean and it follows project conventions. I also really appreciate the demo video. This is a great first contribution

I have left some small cleanup notes for you to consider.
I also noticed that the changes did not run through linter and you likely are missing a setup on your side

Asides from the cleanup suggestions, we need to revisit the debouncing logic to make sure we can provide good performance and UX to the feature

Dec-21-2024 10-30-06

Feel free to reach out if I can help

Shobhit-Nagpal:
sure thing, i'll start off working on these changes and then ask for any opinion or guidance on the individual changes

apps/client/src/features/rundown/event-editor/composite/EventColorPicker.tsx
alex-Arc: Maybe adding simple "OK/Cancel" buttons would be good?

Shobhit-Nagpal: let me know which would be better. i've finished up on most of the tasks, was going to take this up next.

to use the `onMouseUp` event, will have to add the handler in the `<PopoverPicker>` component which is using `onChange` currently. If that's the behavior you're looking at other places, I'll add that

cpvalente: I believe both options have its merits, MouseUp may be simpler while the buttons may be more explicit. 

In my opinion, the UX of both options is equivalent and likely a matter of preference. For this reason I would favour the option that leads to the most elegant code

I will leave it up to you to make a proposal on this

Shobhit-Nagpal: I believe directly using the color picker without adding buttons specifically for the `<EventEditor>` seems like a good user experience. As for debouncing, I figured out why the issue was happening.

I've changed the function signature to this and it works well with parameters now:
```typescript
export function debounce(callback: (...args) => void, wait: number) {
  let timeout: NodeJS.Timeout | null;
  return (...args) => {
    if (timeout) {
      return;
    }
    timeout = setTimeout(() => {
      timeout = null;
      callback(...args);
    }, wait);
  };
}
```

The issue that's occurring when we try to debounce is that the function is more of a throttle than a debounce. Happens to be a small but significant difference.

Upon changing the condition where if there's a timeout then return, to clearing the timeout, it's working well with the color picker.

```typescript
export function debounce(callback: (...args) => void, wait: number) {
  let timeout: NodeJS.Timeout | null;
  return (...args) => {
    if (timeout) {
      clearTimeout(timeout);
    }
    timeout = setTimeout(() => {
      timeout = null;
      callback(...args);
    }, wait);
  };
}
```

Would you like me to change the `debounce` function to this and allow it to apply everywhere it's currently being used or add a `debounce` and `throttle` function separately and make the respective changes where `debounce` was previously being used to `throttle`?

cpvalente: Thank you for bringing this up.
I honestly always need to google throttle vs debounce as I tend to not be sure of the nuances

> **Throttle**: Ensures that a function is called at most once in a specified time period. 
**Debounce**: Ensures that a function is called only after a specified time period has passed since the last call.

In which case, you are correct that the existing function is a throttle with leading edge.

As for your question:

> Would you like me to change the debounce function to this and allow it to apply everywhere it's currently being used or add a debounce and throttle function separately and make the respective changes where debounce was previously being used to throttle?

I believe the second option is the correct. We do not want to make changes to the existing features. 
So we would prefer renaming the existing debounce to throttle and add a new debounce function as you suggest

Shobhit-Nagpal: >  I honestly always need to google throttle vs debounce as I tend to not be sure of the nuances

Haha, same. Had to google it myself to be sure.

I'll get on this task. To give an update:
- The `<EventColorPicker>` component is now `<SwatchPicker>` and I've put it under the respective directory too as specified.
- I've added the `<SwatchPicker>` component within the `<SwatchSelect>` component and it seems to have solved the wrapping issue too.
- Added and removed styles accordingly as mentioned
- Fixed the linting issue as well, I believe. Had to run a `pnpm lint` on my end. Thanks for letting me know about that.

Once I'm done with this, should send in another push soon

Shobhit-Nagpal:
Hey, added the changed mentioned.
- `<EventColorPicker>` is now `<SwatchPicker>` and is located under `components/input/color-input`.
- Shifted `<SwatchPicker>` within `<SwatchSelect>`
- Added a throttle function and made changes accordingly to where it was being used previously.
- Fixed debouncing on `<SwatchPicker>`
- Styling changes
- Updated the branch with latest changes

It was also mentioned that the icon class may not be required. I removed and tried to simply work with the swatch class itself but it wasn't giving the desired ui. The icon class mainly overrides some styles from the swatch class now.

Let me know if any further changes are required

Shobhit-Nagpal:
Attaching a video for the recent changes:

https://github.com/user-attachments/assets/71001905-4be7-4cde-9468-3828c3d3c5cb

cpvalente (changes requested):
Thank you for your work, really great.
I apologise for being slow in providing feedback at this time of year

I have left some small nitpicks for your consideration, only requested change is to fix the typing for the `throttle` and `debounce` functions.
It also seems that you have merge conflicts with master which we would need to resolve

Shobhit-Nagpal:
>Thank you for your work, really great.
>I apologise for being slow in providing feedback at this time of year

Not at all. I'm extremely grateful for the review and suggestions you're giving. I've learnt quite a few things through these changes, super grateful :)

I'll add in the recent changes suggested, fix the conflicts and send in another push

apps/client/src/common/components/input/popover-picker/PopoverPicker.module.scss
```diff
@@ -14,3 +17,35 @@
...
+  border: 0px solid black;
+  box-shadow: 0 0 0 0px;
+  width: 1.75rem;
+  height: 1.75rem;
+}
+
+.icon:hover {
+  box-shadow: 0 0 0 0px;
+}
+
```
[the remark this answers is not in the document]
Shobhit-Nagpal: I tried out some changes, and you were right about not needing the icon class in the first place. Removed that in the changes. Doing another push.

I believe the merge conflicts were sorted previously, in case another shows up, of course we'll fix that too

Shobhit-Nagpal:
Do let me know if there's any change to be done on the function signature. I believe with TypeScript that's the way to do it, in case there's a better way, would love to know about it

apps/client/src/common/components/input/colour-input/SwatchPicker.tsx
```diff
@@ -0,0 +1,41 @@
...
+  isSelected?: boolean;
+  onChange: (name: string) => void;
+}
+
+export default function SwatchPicker(props: SwatchPickerProps) {
+  const { color, onChange, isSelected } = props;
+
+  const classes = cx([style.swatch, isSelected ? style.selected : null, style.selectable]);
+
+  const iconColor = Color(color).isLight() && isSelected ? '#000000' : '#FFFFFF';
```
cpvalente, line 22: The tests are correctly catching an issue where the code will crash if the colour is not a valid colour string (ie: a empty string)

This process needs to be caught, similarly to what we do in the`getAccessibleColour` I linked earlier

cpvalente (changes requested):
Thank you, we need to resolve the uncaught error introduced with migrating the code to the color library
Otherwise the code looks good and I look forwards to merging it

apps/client/src/common/components/input/colour-input/SwatchPicker.tsx
Shobhit-Nagpal: ah, interesting. thanks, added this on the recent push

cpvalente (approved):
Nice 🎉 
Great work, lets fix the merge conflicts and we are ready to go

Shobhit-Nagpal:
sure thing. let me know how you'd like to proceed with that

cpvalente:
> sure thing. let me know how you'd like to proceed with that

Merged! I appreciate your contribution, thank you again.

Shobhit-Nagpal:
> Merged! I appreciate your contribution, thank you again.
I loved working on this. Huge thanks to you for the reviews and suggestions, they really helped.

>  Would you be interested in picking up a follow up task?
Yes, of course! I'll raise another PR for this. Anything else other than this?

cpvalente:
> else other than this

We have a large-ish backlog with lots of react task, but you would need to build some more familiarity with the code base to be able to be productive
But I am sure we can find some tasks going further, you are also welcome to reach out with some suggestions based on your interests

Shobhit-Nagpal:
> But I am sure we can find some tasks going further, you are also welcome to reach out with some suggestions based on your interests

Sure thing.

Regarding the change on `<InlineColourPicker>`, I noticed that we're passing in only the `name` and `value` as props and have the state for colour change in the component itself and I believe it's not saving changes currently. The `<SwatchPicker>` takes in a `onChange` as one of the props.

If we're thinking to replace the `<InlineColourPicker>` with the `<SwatchPicker>`, we'd probably need to:

1. Pass in (and maybe introduce) the `onChange` callback from the `<ParamInput>` component
2. Replace the `<PopoverPicker>` with the `<SwatchPicker>` within `<InlineColourPicker>`

Let me know which approach you'd like to take. Apologies for any delay on my end

cpvalente:
> Regarding the change on `<InlineColourPicker>`, I noticed that we're passing in only the `name` and `value` as props and have the state for colour change in the component itself and I believe it's not saving changes currently. The `<SwatchPicker>` takes in a `onChange` as one of the props.
> 
> If we're thinking to replace the `<InlineColourPicker>` with the `<SwatchPicker>`, we'd probably need to:
> 
> 1. Pass in (and maybe introduce) the `onChange` callback from the `<ParamInput>` component
> 2. Replace the `<PopoverPicker>` with the `<SwatchPicker>` within `<InlineColourPicker>`
> 
> Let me know which approach you'd like to take. Apologies for any delay on my end

The params editor is based on an HTML form element, on submit we convert the values into parameters and push it to the window location.
All the elements are uncontrolled and the layout is [built from data by a factory function](https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/components/view-params-editor/ParamInput.tsx#L27) 

I believe these design elements are good and worth maintaining.
A good solution here would avoid introducing state and callbacks into the factory

With this in mind, I believe that replacing the `PopoverPicker` with the `SwatchPicker` is the most elegant solution since `InlineColourPicker` already solves the issue of using a controlled component in an uncontrolled form

Shobhit-Nagpal:
> All the elements are uncontrolled and the layout is [built from data by a factory function](https://github.com/cpvalente/ontime/blob/master/apps/client/src/common/components/view-params-editor/ParamInput.tsx#L27)

Super interesting. I'm learning a lot honestly. My bad in case I've missed out something and thank you for telling me when I am missing out anything. Really grateful.

> With this in mind, I believe that replacing the PopoverPicker with the SwatchPicker is the most elegant solution since InlineColourPicker already solves the issue of using a controlled component in an uncontrolled form

Made this change and raised a PR under #1412

PR #1401, opened 2024-12-21
Cuesheet times

cpvalente:
This is a continuation PR that allows the user to edit event start and end times in the `cuesheet`
With this PR the `cuesheet` is fully editable

There is some code duplication, but I think it is best to leave things detached for now

Also includes a few decisions: 
- simplify the timer strategy properties by assuming the user will input the times that are relevant
- show delay indicator only on start time

### DIFF
--- a/apps/client/src/common/hooks/useEventAction.ts
+++ b/apps/client/src/common/hooks/useEventAction.ts
@@ -225,38 +225,32 @@ export const useEventAction = () => {
    */
   const updateTimer = useCallback(
     async (eventId: string, field: TimeField, value: string, lockOnUpdate?: boolean) => {
-      let newValMillis = 0;
-
-      // check for previous keyword
-      if (value === 'p' || value === 'prev' || value === 'previous') {
-        newValMillis = getPreviousEnd();
-
-        // check for adding time keyword
-      } else if (value.startsWith('+') || value.startsWith('p+') || value.startsWith('p +')) {
-        // TODO: is this logic solid?
-        const remainingString = value.substring(1);
-        newValMillis = getPreviousEnd() + parseUserTime(remainingString);
-      } else {
-        newValMillis = parseUserTime(value);
+      // an empty value with no lock has no domain validity
+      if (!lockOnUpdate && value === '') {
+        return;
       }
 
-      // dont allow timer values over 23:59:59
-      const cappedMillis = Math.min(newValMillis, dayInMs - MILLIS_PER_SECOND);
-      const newEvent = {
+      const newEvent: Partial<OntimeEvent> = {
         id: eventId,
-        [field]: cappedMillis,
       };
 
       // check if we should lock the field
       if (lockOnUpdate) {
         if (field === 'timeEnd') {
-          newEvent.timeStrategy = TimeStrategy.LockEnd;
+          // an empty value indicates that we should unlock the field
+          newEvent.timeStrategy = value === '' ? TimeStrategy.LockDuration : TimeStrategy.LockEnd;
+          newEvent.timeEnd = value === '' ? undefined : calculateNewValue();
         } else if (field === 'duration') {
-          newEvent.timeStrategy = TimeStrategy.LockDuration;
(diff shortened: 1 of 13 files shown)


apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
```diff
@@ -1,30 +1,65 @@
...
-    </span>
+    <TimeInput initialValue={startTime} onSubmit={update} lockedValue={isStartLocked}>
+      {formattedTime}
+      {delayValue !== 0 && showDelayedTimes && <DelayIndicator delayValue={delayValue} />}
+    </TimeInput>
+  );
+}
+
+function MakeEnd({ getValue, row, table }: CellContext<OntimeRundownEntry, unknown>) {
+  const { hideTableSeconds } = useCuesheetOptions();
```
alex-Arc, line 42: would it be better to pass this as part of `table.options.meta` to try a cut down on the amount of hooks in every cell?

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
```diff
@@ -1,30 +1,65 @@
...
+  const update = useCallback(
+    (newValue: string) => {
+      table.options.meta?.handleUpdateTimer(row.original.id, 'timeEnd', newValue);
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- we skip table.options.meta since the reference seems unstable
+    [row.original.id],
+  );
+
+  const endTime = (getValue() as number | null) ?? 0;
```
alex-Arc, line 51: would the type of the function not be `CellContext<OntimeRundownEntry, number>`
since we know that there is always a value here?
then we don't have to do the casting for `getValue()`

apps/client/src/common/hooks/useEventAction.ts
```diff
@@ -253,9 +253,13 @@ export const useEventAction = () => {
           newEvent.timeStrategy = TimeStrategy.LockEnd;
```
alex-Arc, line 253: right now when you submit an empty end time it results in it being set to 00:00:00
I think it should either move the lock to duration or not change the end valure

alex-Arc (commented):
Feels good
some comments about optimization and time end behaviour

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
cpvalente: i think that moves the typing issue further up and i am not sure how to type this correctly for any field
I think the correct typing would mean checking the value for the key of 

If you have any ideas on how to achieve this without too much typing gymnastics it would be welcome, I may have missed something in the docs

But you are right that we know there is always a value, so i might as well cast it correctly

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
cpvalente: yes it would

apps/client/src/common/hooks/useEventAction.ts
cpvalente: good catch

cpvalente:
@alex-Arc , thank you for the feedback.
What do you think about the behaviour now?

I am still uncertain about the delay behaviour, I wonder if we should always show the chevron and use the `showDelayedTimes` option to actually show the times including the delay?

alex-Arc:
I like the time end behaviour now
for the delay maybe
- always show the ^ in both end and start time
- when show delayed times is on- show the delayed time in both end and start and end, 
so if in the plan is 12:00:00 and the delay is 5m the field will show 12:05:00 and the tooltip will show 12:00:00 +5m

PR #1410, opened 2024-12-27
convert to and from database type

alex-Arc:


alex-Arc:
@coderabbitai review

alex-Arc:
@coderabbitai review

alex-Arc:
@coderabbitai review

apps/server/src/classes/data-provider/DataProvider.utils.ts
```diff
@@ -27,3 +35,22 @@ export function safeMerge(existing: DatabaseModel, newData: Partial<DatabaseMode
...
+    if (isOntimeBlock(entryClone)) {
+      return entryClone;
+    }
+    if (isOntimeDelay(entryClone)) {
+      return entryClone;
+    }
+  });
+
+  return databaseRundown;
+}
```
cpvalente, line 56: todo: 
- we need to handle unknown types, maybe by just throwing
- we likely dont need the structured clone AND the destructuring on the same dataset
- we likely do not need the flatmap
tip: remember to turn on strict mode when you are working

maybe like


```suggestion
export function rundownToDatabaseRundown(newData: OntimeRundown): DatabaseOntimeRundown {
  const databaseRundown = newData.map((entry) => {
    if (isOntimeEvent(entry)) {
      // eslint-disable-next-line @typescript-eslint/no-unused-vars -- removing the delay from the event
      const { delay, ...dbEvent } = entry;
      return dbEvent as DatabaseOntimeEvent;
    } else if (isOntimeBlock(entry) || isOntimeDelay(entry)) {
      return { ...entry };
    } else {
      throw new Error('Unknown entry type');
    }
  });

  return databaseRundown;
}
```

apps/server/src/classes/data-provider/__tests__/DataProvider.utils.test.ts
```diff
@@ -210,3 +220,18 @@ describe('safeMerge', () => {
...
+    fullEvent = { id: '1', cue: '#1', type: SupportedEvent.Event, ...event };
+    dbEvent = fullEvent;
+    delete (dbEvent as OntimeEvent).delay;
+  });
+  it('drops delay key', () => {
+    const testRundown: OntimeRundown = [{ ...fullEvent }];
+    const result = rundownToDatabaseRundown(testRundown);
+    expect(result).toEqual([{ id: '1', ...dbEvent }]);
+  });
+});
```
cpvalente, line 237: suggestion: this test seems a bit complicated for what it achieves
suggestions:
- we can work on making things reusable later
- remember to name the tests in a way that they tell us what is under test

maybe something like

```suggestion
describe('rundownToDatabaseRundown()', () => {
  it('converts the rundown to DAO type by removing the delay', () => {
    const testEvent: OntimeEvent = { ...event, id: '1', cue: '#1' };

    const testRundown: OntimeRundown = [testEvent];
    const result = rundownToDatabaseRundown(testRundown);
    // remove the delay from the event
    const { delay, ...matchObject } = testEvent;
    expect(result).toMatchObject([matchObject]);
  });
});
```

apps/server/src/services/rundown-service/rundownCache.ts
```diff
@@ -76,7 +78,7 @@ export async function init(initialRundown: Readonly<OntimeRundown>, customFields
  * @param rundown
  */
 export function generate(
-  initialRundown: OntimeRundown = persistedRundown,
+  initialRundown: DatabaseOntimeRundown | OntimeRundown = persistedRundown,
```
cpvalente, line 81: suggestion: `OntimeRundown` is structurally compatible to `DatabaseOntimeRundown`, so we dont need both. just adding the smaller one (`DatabaseOntimeRundown`) is sufficient

apps/server/src/services/rundown-service/rundownCache.ts
```diff
@@ -98,6 +100,7 @@ export function generate(
     // we assign a reference to the current entry, this will be mutated in place
     const currentEntry = initialRundown[i];
 
+    //inside this if block all conversions from the database type to the normal onime type should be handled
```
cpvalente, line 103: ??? I didnt quite understand

apps/server/src/services/rundown-service/rundownCache.ts
```diff
@@ -144,13 +147,19 @@ export function generate(
         if (isNewLatest(currentEntry.timeStart, currentEntry.timeEnd, lastEntry?.timeStart, lastEntry?.timeEnd)) {
           lastEntry = currentEntry;
         }
+      } else {
+        currentEntry.delay = 0; //FIXME: what should be done about skipped events
```
cpvalente, line 151: i think the 0 is good enough?

apps/server/src/services/rundown-service/rundownCacheUtils.ts
```diff
@@ -30,7 +31,7 @@ export function getLink(currentIndex: number, rundown: OntimeRundown): OntimeEve
  */
 export function handleLink(
   currentIndex: number,
-  rundown: OntimeRundown,
+  rundown: OntimeRundown | DatabaseOntimeRundown, //We can accept both types as we dont care about delay
```
cpvalente, line 34: suggestion: same as before, I believe we dont need both `OntimeRundown` and `DatabaseOntimeRundown`
The same is applicable to the other function signatures

cpvalente (commented):
thank you for doing this

There are some small cleanups we can do to improve the existing code
I also wonder if `OntimeEventDAO` and `RundownDAO` are better names and will provide us a better heuristic for this kinds of things in future

As of now, we seem to be struggling with naming, in `rundownCache`, the `persistedRundown` indicates to be a rundown as it is in the database, which will not be tree after these changes. `persistedRundown` will now be simply a not normalised rundown

apps/server/src/classes/data-provider/__tests__/DataProvider.utils.test.ts
alex-Arc: Yeah I thought I was going to make more tests and would need fullEvent multiple times, but can't come up with any more relevant test, can you?

apps/server/src/classes/data-provider/DataProvider.utils.ts
alex-Arc: I am sill a bit shaky on js copy semantics
but since our events contain an object in form of the custom fields would a destruction not only do a shallow copy and still point to the same custom object?

apps/server/src/classes/data-provider/__tests__/DataProvider.utils.test.ts
cpvalente: you could assert that the objects are not the same (ie, the function does not mutate the original object
but I am not sure this is necessary. it is probably fine as it is

apps/server/src/classes/data-provider/DataProvider.utils.ts
cpvalente: > I am sill a bit shaky on js copy semantics but since our events contain an object in form of the custom fields would a destruction not only do a shallow copy and still point to the same custom object?

yes, you are correct, destructuring creates shallow copies, so we would need to do something like (untested)

```ts
      const { delay, custom, ...dbEvent } = entry;
      return {...dbEvent, custom: {...custom}} as DatabaseOntimeEvent;
```

As for the other fields (delay and block, they do not have nested objects so the destructure is fine

apps/client/src/common/utils/eventsManager.ts
```diff
@@ -24,6 +24,7 @@ export const cloneEvent = (event: OntimeEvent): ClonedEvent => {
     skip: event.skip,
     colour: event.colour,
     revision: 0,
+    delay: 0, //force delay to zero, will be populated by the server
```
cpvalente, line 27: I am not sure this warrants the cognitive load of having even to think about it.
we could just 
a) remove the comment 
b) copy the delay of the cloned event

apps/server/src/classes/data-provider/DataProvider.utils.ts
```diff
@@ -27,3 +35,19 @@ export function safeMerge(existing: DatabaseModel, newData: Partial<DatabaseMode
     http: { ...existing.http, ...http },
   };
 }
+
+export function rundownToDAO(newData: OntimeRundown): OntimeRundownDAO {
+  const databaseRundown = newData.map((entry) => {
+    if (isOntimeEvent(entry)) {
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
cpvalente, line 42:
```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unused-vars -- we use destructuring to remove unwanted keys
```

apps/server/src/services/rundown-service/rundownCache.ts
```diff
@@ -31,14 +33,27 @@ type NormalisedRundown = Record<EventID, OntimeRundownEntry>;
...
+/**
+ * all mutating functions will set this value if there is a need for re-generation
+ * but will only be cleared by the generate function
+ */
 let isStale = true;
+
+/** Allows safely setting the stale state without accedentaly clearing it */
+const setIsStale = (newValue: boolean) => {
+  isStale = isStale || newValue;
+};
```
cpvalente, line 55: I didnt really understand what this achieves. Would this not mean that we cannot set isStale to false?

```suggestion
/** Allows safely setting the stale state without accidentaly clearing it */
function setIsStale(newValue: boolean) {
  isStale = isStale || newValue;
};
```

packages/types/src/definitions/core/Rundown.type.ts
```diff
@@ -1,7 +1,18 @@
...
+ * This is the Data Accese type for `OntimeRundownEntry`
+ * and is used when a rundonw is stored in the database
+ */
+export type OntimeRundownEntryDAO = OntimeDelay | OntimeBlock | OntimeEventDAO;
+/**
+ * This is the Data Accese type for `OntimeRundown`
+ * and is used when a rundonw is stored in the database
+ */
+export type OntimeRundownDAO = OntimeRundownEntryDAO[];
+
```
cpvalente, line 16:
```suggestion
/**
 * This is the Data Access type for `OntimeRundownEntry`
 * and is used when a rundown is stored in the database
 */
export type OntimeRundownEntryDAO = OntimeDelay | OntimeBlock | OntimeEventDAO;
/**
 * This is the Data Access type for `OntimeRundown`
 * and is used when a rundown is stored in the database
 */
export type OntimeRundownDAO = OntimeRundownEntryDAO[];

```

packages/types/src/definitions/core/OntimeEvent.type.ts
```diff
@@ -38,11 +38,17 @@ export type OntimeEvent = OntimeBaseEvent & {
...
-  delay?: number; // calculated at runtime
+  delay: number; // calculated at runtime
   timeWarning: number;
   timeDanger: number;
   custom: EventCustomFields;
 };
 
+/**
+ * This is the Data Accese type for `OntimeEvent`
+ * and is used when a rundonw is stored in the database
```
cpvalente, line 49:
```suggestion
 * This is the Data Access type for `OntimeEvent`
 * and is used when a rundown is stored in the database
```

cpvalente (approved):
nice! good work!

I just left some comments, mostly on typos. everything seems to work well and typescript is happy

apps/server/src/classes/data-provider/DataProvider.ts
```diff
@@ -86,7 +86,7 @@ function getCustomFields(): Readonly<CustomFields> {
 }
 
 async function setRundown(newData: OntimeRundown): ReadonlyPromise<OntimeRundown> {
-  db.data.rundown = [...newData];
+  db.data.rundown = newData;
```
alex-Arc, line 89: @cpvalente 
do we need to make a copy here?
we did not make a full copy befor?
and we don't do anything to this object

e2e/tests/features/209-rundown-shortcuts.spec.ts
```diff
@@ -1,5 +1,23 @@
...
+  await page.getByRole('button', { name: 'Create Event' }).click();
+  await page.getByTestId('block__title').click();
+  await page.getByTestId('block__title').fill('test');
+  await page.getByTestId('block__title').press('Enter');
+
+  //assert
+  await expect(page.getByTestId('entry-1')).toBeVisible();
+  await expect(page.getByTestId('entry-1').getByTestId('block__title')).toHaveValue('test');
+});
+
```
cpvalente, line 20: I am not sure this is testing a shortcut or otherwise giving us much?
we already have the `204-editor-crud` covering this. if anything we can improve that test instead

PR #1428, opened 2025-01-05
small `strict` improvements

alex-Arc:


apps/server/src/services/sheet-service/sheetUtils.ts
```diff
@@ -29,7 +29,16 @@ export type ClientSecret = {
  * @returns
  */
 export function validateClientSecret(clientSecret: object): clientSecret is ClientSecret {
-  return requiredClientKeys.every((key) => Object.keys(clientSecret['installed']).includes(key));
+  if (!('installed' in clientSecret)) {
+    return false;
+  }
+
+  const { installed } = clientSecret;
+  if (typeof installed !== 'object' || installed === null) {
```
cpvalente, line 37: I think we have an utility for this in the assert module?

apps/server/src/utils/__tests__/parser.test.ts
```diff
@@ -550,7 +550,7 @@ describe('test aliases import', () => {
           pathAndParams: 'testpathAndParams',
         },
       ],
-    } as DatabaseModel;
+    } as unknown as DatabaseModel;
```
cpvalente, line 553: I am not sure the downcasting is a good approach, at this point, the type is not doing anything for us anymore?

alex-Arc: it is to avoid filling out a full dataset for a test that doesn't need it right?
In that case I think it is ok to force the types to clear out some of the error messages

PR #1429, opened 2025-01-05
More rundown metadata

alex-Arc:


alex-Arc:
@coderabbitai review

apps/server/src/services/rundown-service/rundownCache.ts
```diff
@@ -117,50 +121,53 @@ export function generate(
...
       // 1. handle links - mutates updatedEvent
       handleLink(i, initialRundown, currentEntry, links);
 
       // 2. handle custom fields - mutates updatedEvent
       handleCustomField(customFields, customFieldChangelog, currentEntry, assignedCustomFields);
 
+      if (lastEntry) {
+        const lastEntryCrossedMidnight = lastEntry.timeStart + lastEntry.duration > dayInMs;
+        const isNextDay = checkIsNextDay(lastEntry.timeStart, currentEntry.timeStart, lastEntry.duration);
+        totalDays += isNextDay || lastEntryCrossedMidnight ? 1 : 0; //TODO: should we count in days or millis
```
alex-Arc, line 135: @cpvalente what do you think of this?
It feels a bit more easy to reason about and human to stor it as number of days, but most (if not all) of the places we end up using the data we multiply it with our `dayInMs` constant anyway...

apps/client/src/features/rundown/event-block/EventBlock.utils.ts
```diff
@@ -17,25 +10,15 @@ export function formatDelay(timeStart: number, delay: number): string | undefine
...
-    const normalisedPreviousEnd = previousCrossMidnight ? previousEnd + dayInMs : previousEnd;
-
-    const gap = dayInMs - normalisedPreviousEnd + timeStart;
-    if (gap === 0) return;
-    const gapString = formatDuration(Math.abs(gap), false);
-    return `Gap ${gapString} (next day)`;
+export function formatGap(gap: number, isNextDay: boolean) {
+  if (gap === 0) {
+    if (isNextDay) {
+      //We show a next day warning eaven if there is no gap
```
cpvalente, line 16:
```suggestion
      // We show a next day warning even if there is no gap
```

apps/client/src/features/rundown/event-block/RundownIndicators.tsx
```diff
@@ -1,24 +1,24 @@
...
 
-  const hasOverlap = formatOverlap(timeStart, previousStart, previousEnd);
+  const hasGap = formatGap(gap, isNextDay);
   const hasDelay = formatDelay(timeStart, delay);
 
   return (
     <div className={style.indicators}>
       {hasDelay && <div className={style.delay}>{hasDelay}</div>}
-      {hasOverlap && <div className={style.gap}>{hasOverlap}</div>}
+      {hasGap && <div className={style.gap}>{hasGap}</div>}
```
cpvalente, line 21: nice! our old ping-pong with overlap-gap was pretty confusing

apps/client/src/features/rundown/event-block/__tests__/EventBlock.utils.test.ts
```diff
@@ -13,66 +13,15 @@ describe('formatDelay()', () => {
 
 describe('formatOverlap()', () => {
```
cpvalente, line 14: Since these functions no longer contain logic, I am not sure how much we need to worry about the tests
either way, the test name should indicate what is under test

```suggestion
describe('formatGap()', () => {
```

apps/client/src/features/rundown/event-block/__tests__/EventBlock.utils.test.ts
```diff
@@ -13,66 +13,15 @@ describe('formatDelay()', () => {
...
-    expect(result).toBe('Gap 1h (next day)');
-  });
-
-  it('handles sequential events the day after, with previous ending over midnight', () => {
-    const previousStart = 23 * MILLIS_PER_HOUR;
-    const previousEnd = 1 * MILLIS_PER_HOUR;
-    const timeStart = 1 * MILLIS_PER_HOUR;
-    const result = formatOverlap(timeStart, previousStart, previousEnd);
-    expect(result).toBeUndefined();
+  it('recognises an gap between two times', () => {
```
cpvalente, line 19:
```suggestion
  it('formats time with gap', () => {
```

apps/client/src/features/rundown/event-block/__tests__/EventBlock.utils.test.ts
```diff
@@ -13,66 +13,15 @@ describe('formatDelay()', () => {
 
 describe('formatOverlap()', () => {
   it('recognises an overlap between two times', () => {
```
cpvalente, line 15:
```suggestion
  it('formats an overlapping time', () => {
```

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
```diff
@@ -45,7 +46,7 @@ function TimeInputFlow(props: EventBlockTimerProps) {
   };
 
   const warnings = [];
-  if (timeStart > timeEnd) {
+  if (timeStart + duration > dayInMs) {
```
cpvalente, line 49: I am not sure this is the same thing?

The previous logic said that if a time is scheduled like
22:00 - 09:00 , we say it crosses midnight. but it is not true that the event will have more than 1 day duration

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
```diff
@@ -129,7 +130,7 @@ function TimeInputFlow(props: EventBlockTimerProps) {
       </div>
 
       {warnings.length > 0 && (
-        <div className={style.timerNote}>
+        <div className={style.timerNote} aria-label='warning'>
```
cpvalente, line 133: I dont think the aria attribute is doing much for us here
Maybe we should reconsider our approach to accessibility, but we shouldnt add code that is not necessary for this PR

packages/utils/src/date-utils/isNewLatest.ts
```diff
@@ -1,24 +1,21 @@
-import { checkIsNextDay } from './checkIsNextDay.js';
+import type { OntimeEvent } from 'ontime-types';
+
+import { dayInMs } from './conversionUtils.js';
 
 /**
  * Checks whether a new element is the latest in the list
  */
-export function isNewLatest(timeStart: number, timeEnd: number, previousStart?: number, previousEnd?: number): boolean {
+export function isNewLatest(
```
cpvalente, line 8: nice!

apps/server/src/services/rundown-service/rundownCacheUtils.ts
```diff
@@ -147,3 +147,36 @@ export function hasChanges<T extends OntimeBaseEvent>(existingEvent: T, newEvent
     (key) => !Object.hasOwn(existingEvent, key) || existingEvent[key as keyof T] !== newEvent[key as keyof T],
   );
 }
+
+/**
+ * Utility for calculatig if the current events should have a day offset
```
cpvalente, line 152:
```suggestion
 * Utility for calculating if the current events should have a day offset
```

apps/server/src/services/rundown-service/rundownCacheUtils.ts
```diff
@@ -147,3 +147,36 @@ export function hasChanges<T extends OntimeBaseEvent>(existingEvent: T, newEvent
...
+export function calculateDayOffset(
+  current: Pick<OntimeEvent, 'timeStart'>,
+  previous?: Pick<OntimeEvent, 'timeStart' | 'duration'>,
+) {
+  // if there is no previous there can't be a day offset
+  if (!previous) {
+    return 0;
+  }
+
+  // if the previous events duration os zero it will push the current event to next day
```
cpvalente, line 166:
```suggestion
  // if the previous events duration is zero it will push the current event to next day
```

e2e/tests/features/401-report-data.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { expect, test } from '@playwright/test';
```
cpvalente, line 1: suggestion: the test numbers like `401` are to ensure execution order

In this case, the test does not start a new context, so I suggest we just continue the existing order and name it `211-report-data`

packages/utils/src/date-utils/checkIsNextDay.ts
```diff
@@ -1,8 +1,10 @@
+import type { OntimeEvent } from 'ontime-types';
+
 import { dayInMs } from './conversionUtils.js';
 
 /**
  * Utility function checks whether a given event is the day after from its predecessor
- * We consider an event to be the day after, if it begins before the start of the previous
+ * We consider an event to be the day after, if it begins on a new day
```
cpvalente, line 7: is there a reason for the change? this is not the existing logic or what the examples below indicate

packages/utils/src/date-utils/checkIsNextDay.ts
```diff
@@ -19,23 +21,23 @@ import { dayInMs } from './conversionUtils.js';
  * 22:00 - 02:00
  * 01:00 - 03:00
  */
-export function checkIsNextDay(previousStart: number, timeStart: number, previousDuration: number): boolean {
-  if (previousDuration === 0) {
+export function checkIsNextDay(
+  current: Pick<OntimeEvent, 'timeStart' | 'dayOffset'>,
+  previous?: Pick<OntimeEvent, 'timeStart' | 'duration' | 'dayOffset'> | null,
```
cpvalente, line 26: question: does previous need to be both `undefined` and `null`?

packages/utils/src/date-utils/checkIsNextDay.ts
```diff
@@ -19,23 +21,23 @@ import { dayInMs } from './conversionUtils.js';
...
-    const normalisedPreviousEnd = cappedStart + previousDuration;
-    if (normalisedPreviousEnd === dayInMs) {
-      return true;
-    }
-    // handle exception for an event that finishes exactly at midnight
-    if (normalisedPreviousEnd > dayInMs) {
-      return false;
-    }
-    return true;
+  // if the day offsets are the samme it can't be the next day
```
cpvalente, line 32:
```suggestion
  // if the day offsets are the same it can't be the next day
```

packages/utils/src/date-utils/checkIsNextDay.ts
```diff
@@ -19,23 +21,23 @@ import { dayInMs } from './conversionUtils.js';
...
+  if (current.dayOffset <= previous.dayOffset) {
+    return false;
+  }
+
+  // if the previous event crossed midnight then the current is the same day as that
```
cpvalente, line 37:
```suggestion
  // if the previous event crossed midnight then the current is the same day
```

packages/utils/src/date-utils/isNewLatest.test.ts
```diff
@@ -1,48 +1,55 @@
 import { MILLIS_PER_HOUR } from './conversionUtils';
 import { isNewLatest } from './isNewLatest';
 
+//TODO: this test should be updated if could be passed by a function always returning true
```
cpvalente, line 4: hahaha, good idea

how about an event that is contained in the previous? this is very common in our rundowns

```ts
    expect(
      isNewLatest(
        { timeStart: 21 * MILLIS_PER_HOUR, duration: MILLIS_PER_HOUR, dayOffset: 0 },
        { timeStart: 21 * MILLIS_PER_HOUR + 10 * MILLIS_PER_MINUTE, duration: 10 * MILLIS_PER_MINUTE, dayOffset: 0 },
      ),
    ).toBe(false);
```

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
alex-Arc: this was for easier access in the test. or maybe I should use the data-testid, but sometimes when I try that it gets removed by react, not sure when or why

packages/utils/src/date-utils/checkIsNextDay.ts
alex-Arc: in the the timeline the lastEvent is marked as OntimeEvent or null, os it is to accommodate that...
or should I rater change how the timeline dose it?

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
alex-Arc: in your example 
timeStart: 22h
timeEnd: 9h
duration: 11h

so the calculation 
timeStart + duration > dayInMs
           22 + 11 = 33  > 24h  

so it crossed midnight
the change is so that an event that ends exactly at midnight dose not count as crossing midnight

packages/utils/src/date-utils/checkIsNextDay.ts
alex-Arc: we now rely on the day offset to tell us this so `before the start of the previous` is no longer exactly what we are testing with

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
cpvalente: 👍

apps/client/src/features/rundown/time-input-flow/TimeInputFlow.tsx
cpvalente: yeah, if this is just for tests, data-testid is the right thing

apps/server/src/services/rundown-service/rundownCacheUtils.ts
```diff
@@ -147,3 +147,36 @@ export function hasChanges<T extends OntimeBaseEvent>(existingEvent: T, newEvent
     (key) => !Object.hasOwn(existingEvent, key) || existingEvent[key as keyof T] !== newEvent[key as keyof T],
   );
 }
+
+/**
+ * Utility for calculating if the current events should have a day offset
+ * @param current the current event under test
+ * @param previous the previous event
+ * @returns 0 or 1 for easy acumelation with the total days
```
cpvalente, line 155:
```suggestion
 * @returns 0 or 1 for easy accumulation with the total days
```

packages/utils/src/date-utils/checkIsNextDay.ts
cpvalente: in that case, i suggest we remove this sentence

> * We consider an event to be the day after, if it begins on a new day

i fear it doesnt really add anything other than potential for confusion
The examples are also incorrect since we are no longer inferring day after from the times, but from the `dayOffset` fields

packages/utils/src/date-utils/checkIsNextDay.ts
cpvalente: I think this would have to be your decision, ideally we want as few cases as possible

packages/utils/src/date-utils/getTimeFromPrevious.ts
```diff
@@ -1,45 +1,34 @@
-import { checkIsNextDay } from './checkIsNextDay.js';
+import type { OntimeEvent } from 'ontime-types';
+
 import { dayInMs } from './conversionUtils.js';
 
 /**
  * Utility returns the time elapsed (gap or overlap) from the previous
  * It uses deconstructed parameters to simplify implementation in UI
  */
 export function getTimeFromPrevious(
```
cpvalente, line 9: question: do we still need this function?

packages/utils/src/date-utils/getTimeFromPrevious.ts
```diff
@@ -1,45 +1,34 @@
-import { checkIsNextDay } from './checkIsNextDay.js';
+import type { OntimeEvent } from 'ontime-types';
+
 import { dayInMs } from './conversionUtils.js';
 
 /**
  * Utility returns the time elapsed (gap or overlap) from the previous
  * It uses deconstructed parameters to simplify implementation in UI
```
cpvalente, line 7: suggestion: this statement is no longer true and we would prefer removing the comment

cpvalente (approved):

packages/utils/src/date-utils/getTimeFromPrevious.ts
alex-Arc: used in `rundownCache` and `timeline.utils.ts`
might be able to remove it from timeline at some point...

PR #1430, opened 2025-01-05
fix: validateLinkStart

alex-Arc:


alex-Arc:
@cpvalente I can't reproduce this error locally
https://github.com/cpvalente/ontime/actions/runs/12623710716/job/35173015919?pr=1430

packages/utils/src/validate-events/validateEvent.ts
```diff
@@ -2,12 +2,15 @@ import type { MaybeString } from 'ontime-types';
 import { EndAction, TimerType, TimeStrategy } from 'ontime-types';
 
 /**
- * Check if a given value is a valid type of string, returns null otherwise
+ * Check if a given value is a valid type of string, returns the fallback otherwise
```
cpvalente, line 5:
```suggestion
 * Check if a given value is a valid type linkStart, returns the fallback otherwise
 * linkStart can be a string (id of an event to link) or null (unlinked)
```

cpvalente (approved):

PR #1448, opened 2025-01-12
Batch client updates

alex-Arc:
The goal here is to stabilize runtimestate so that calculations that combine data from runtime or timer and clock don't have incorrect data in the time between receiving the two datasets. Without affecting the mitigations we tried to put in place for #940

~~this is attempted to be done by allowing ws client to subscribe to either a key model or a patch model when connection~~

all data written to `eventStore.set` will be batch with using `setImmediate` so it should be sent next eventloop tick but all the data set will be sent together

### DIFF
--- a/apps/client/src/common/utils/socket.ts
+++ b/apps/client/src/common/utils/socket.ts
@@ -39,6 +39,7 @@ export const connectSocket = () => {
     }
 
     socketSendJson('set-client-type', 'ontime');
+    socketSendJson('set-client-use-patch', 'ontime');
     setOnlineStatus(true);
   };
 
@@ -64,7 +65,7 @@ export const connectSocket = () => {
   websocket.onmessage = (event) => {
     try {
       const data = JSON.parse(event.data);
-
+      console.log(data);
       const { type, payload } = data;
 
       if (!type) {
@@ -137,6 +138,11 @@ export const connectSocket = () => {
 
           patchRuntime(serverPayload);
           updateDevTools(serverPayload);
+          break;
+        }
+        case 'ontime-patch': {
+          patchRuntime(payload);
+          updateDevTools(payload);
           break;
         }
         case 'ontime-clock': {
--- a/apps/client/src/common/utils/socket.ts
+++ b/apps/client/src/common/utils/socket.ts
@@ -65,7 +65,6 @@ export const connectSocket = () => {
   websocket.onmessage = (event) => {
     try {
       const data = JSON.parse(event.data);
-      console.log(data);
       const { type, payload } = data;
 
       if (!type) {
@@ -145,6 +144,7 @@ export const connectSocket = () => {
           updateDevTools(payload);
           break;
         }
+        //TODO: remove all other types as they are now patched
         case 'ontime-clock': {
           patchRuntimeProperty('clock', payload);
           updateDevTools({ clock: payload });
--- a/apps/client/src/common/utils/socket.ts
+++ b/apps/client/src/common/utils/socket.ts
@@ -39,7 +39,6 @@ export const connectSocket = () => {
     }
 
     socketSendJson('set-client-type', 'ontime');
-    socketSendJson('set-client-use-patch', 'ontime');
     setOnlineStatus(true);
   };
 
@@ -65,6 +64,7 @@ export const connectSocket = () => {
   websocket.onmessage = (event) => {
     try {
       const data = JSON.parse(event.data);
+
       const { type, payload } = data;
 
       if (!type) {
@@ -139,12 +139,6 @@ export const connectSocket = () => {
           updateDevTools(serverPayload);
           break;
         }
-        case 'ontime-patch': {
-          patchRuntime(payload);
-          updateDevTools(payload);
-          break;
-        }
-        //TODO: remove all other types as they are now patched
         case 'ontime-clock': {
           patchRuntimeProperty('clock', payload);
           updateDevTools({ clock: payload });
--- a/apps/server/src/adapters/WebsocketAdapter.ts
+++ b/apps/server/src/adapters/WebsocketAdapter.ts
@@ -14,7 +14,7 @@
  * Payload: adds necessary payload for the request to be completed
  */
 
-import { Client, LogOrigin } from 'ontime-types';
+import { Client, LogOrigin, RuntimeStore } from 'ontime-types';
 
 import { WebSocket, WebSocketServer } from 'ws';
 import type { Server } from 'http';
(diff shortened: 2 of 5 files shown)


apps/server/src/services/runtime-service/RuntimeService.ts
```diff
@@ -678,7 +678,7 @@ function broadcastResult(_target: any, _propertyKey: string, descriptor: Propert
     // we do the comparison by explicitly for each property
     // to apply custom logic for different datasets
 
-    const shouldForceTimerUpdate = getForceUpdate(RuntimeService.previousTimerUpdate, state.clock);
+    const shouldForceTimerUpdate = getForceUpdate(RuntimeService.previousTimerUpdate, state.clock); //FIXME: pause causess an avlache of update if timer is negative
```
cpvalente, line 681: ?? lets chat

apps/server/src/stores/EventStore.ts
```diff
@@ -23,16 +25,13 @@ export const eventStore = {
...
-      store[key] = value;
-    });
-    this.broadcast();
+    if (!changedKeys.includes(key)) changedKeys.push(key);
+    if (!trigger) {
+      trigger = setImmediate(() => {
+        socket.sendRuntimeStoreUpdate(changedKeys, store);
+        trigger = null;
+      });
+    }
```
cpvalente, line 34: can you explain what this achieves?

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -256,6 +279,29 @@ export class SocketServer implements IAdapter {
     }
   }
 
+  public sendRuntimeStoreUpdate(keys: (keyof RuntimeStore)[], store: Partial<RuntimeStore>) {
```
cpvalente, line 282: suggestion: we would likely prefer not having distributed logic to achieve a task

ie: the Websocket adapter knows how to send payloads, but shouldnt care whats in it

if we want to make logic specific to the patch payloads, it should be in the store

apps/client/src/common/utils/socket.ts
```diff
@@ -139,6 +139,12 @@ export const connectSocket = () => {
           updateDevTools(serverPayload);
           break;
         }
+        case 'ontime-patch': {
+          patchRuntime(payload);
```
cpvalente, line 143: question: what can we expect of the patch? is it a deep patch or are all the objects replaced at top level?

apps/server/src/adapters/WebsocketAdapter.ts
cpvalente, line 282: this logic looks a bit complicated. 
Can you please explain what it does (in code)?

apps/client/src/common/utils/socket.ts
alex-Arc: that would be our decision, but right now it is only one key deep

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -36,6 +36,10 @@ export class SocketServer implements IAdapter {
   private lastConnection: Date | null = null;
   private shouldShowWelcome = true;
 
+  //should we be tacking versions
```
cpvalente, line 39: the clients should really have their own version as they get the UI from the server

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -257,6 +280,38 @@ export class SocketServer implements IAdapter {
...
+    const patch = {};
+    keysToUpdate.map((key) => {
+      Object.assign(patch, { [key]: store[key] });
+    });
+
+    //convert to JSON once and reuse
+    const stringifiedPatch = JSON.stringify({ type: 'ontime-patch', payload: patch });
+
+    // for each client that have subscribed to the patch method send the patch object
+    this.patchClients.forEach((ws) => {
```
cpvalente, line 294: do you think this is necessary?

I understand the motivation behind it, but I wonder if trying to keep backwards compatibility makes our code harder and more bug prone

I am ok with this as long as we are not trading the above for the compatibility

apps/server/src/stores/EventStore.ts
```diff
@@ -7,6 +7,8 @@ export type StoreGetter = <T extends keyof RuntimeStore>(key: T) => Partial<Runt
 
 let store: Partial<RuntimeStore> = {};
 
+const changedKeys = new Array<keyof RuntimeStore>();
+let isUpdatePending: NodeJS.Immediate | null = null;
```
cpvalente, line 11: this is an interesting way of batching updates in one event loop. nice!

Did you find this somewhere? I wonder whether `setImmediate` is more accurate to what we want instead of `setTimeout` or `nextTick`. Did you consider the alternatives?

I am especially wondering how this goes together with the scheduling of the store updates

alex-Arc: I tried reading through this

https://nodejs.org/en/learn/asynchronous-work/understanding-setimmediate

I'm not entirely sure what we actually want

I believe `setImmediate` runs at the end of the current event loop <---(I think this is a fitting spot)
and `setTimeout` may run on the start of the next loop
`nextTick` is the next tick in this current event loop <---(sounds like it would fire a bit to early)

apps/server/src/adapters/WebsocketAdapter.ts
alex-Arc: We would have to do a breaking version change right?
I would love to keep it simple, but I have no idea what things that are dependent one us?

apps/server/src/stores/EventStore.ts
cpvalente: sounds good

apps/server/src/adapters/WebsocketAdapter.ts
cpvalente: what if we continue sending the messages in serial?
the logic of the store should accumulate changes in the same event loop iteration, maybe the small difference in milliseconds on the message being received and parsed will not be visible?

alex-Arc: hmm yeah it's worth a try

alex-Arc: That seamed to work

apps/server/src/services/runtime-service/RuntimeService.ts
```diff
@@ -678,7 +678,8 @@ function broadcastResult(_target: any, _propertyKey: string, descriptor: Propert
     // we do the comparison by explicitly for each property
     // to apply custom logic for different datasets
 
-    const shouldForceTimerUpdate = getForceUpdate(RuntimeService.previousTimerUpdate, state.clock);
+    const shouldForceTimerUpdate =
+      state.timer.playback === Playback.Play && getForceUpdate(RuntimeService.previousTimerUpdate, state.clock);
```
cpvalente, line 682: suggestion: I think the playback decision should be inside the getForceUpdate, so that this function is a complete encapsulation of the logic

cpvalente (approved):
This looks nice to me, just left a small improvement suggestion

remember to squash all the fixup commits before merging

PR #1449, opened 2025-01-14
feat: add simple auth to server requests

cpvalente:
This is a proposal for protecting the HTTP connections with a password

Strategy
- requests go through authentication middleware
- if the password is active and the user has not provided a password in the cookie or params
- a) for the client assets, we serve a login page instead
- b) for the data endpoints, we refuse the request
- when the user sets the correct password in the login form, we set the cookie that will be with all requests

Notes:
- had to do some cheating to allow the client to access public files, which are requested before the connection is made
- need to test the token in query params solution, do we want UI to generate this? or pass the password in the websocket request? The token in query params will be important for generating an authenticated URL and for the companion module

Whis is a viable solution for ontime-cloud, but not for Ontime

### DIFF
--- a/apps/server/src/middleware/authenticate.ts
+++ b/apps/server/src/middleware/authenticate.ts
@@ -1,0 +1,141 @@
+import { LogOrigin } from 'ontime-types';
+
+import express, { type Request, type Response, type NextFunction } from 'express';
+import type { IncomingMessage } from 'node:http';
+import type { WebSocket } from 'ws';
+import { parse as parseCookie } from 'cookie';
+
+import { hashPassword } from '../utils/hash.js';
+import { srcFiles } from '../setup/index.js';
+import { logger } from '../classes/Logger.js';
+// import { password } from '../externals.js';
+const password = 'test';
+import { noopMiddleware } from './noop.js';
+
+export const hasPassword = Boolean(password);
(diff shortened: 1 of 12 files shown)


apps/server/src/middleware/authenticate.ts
```diff
@@ -0,0 +1,125 @@
...
+    return next();
+  }
+
+  const cookieString = req.headers.cookie;
+  if (typeof cookieString === 'string') {
+    const [key, value] = cookieString.split('=');
+    if (key === 'token' && value === hashedPassword) {
+      return next();
+    }
+  }
```
alex-Arc, line 122:
```suggestion
  }
  
  const url = new URL(`http://localhost${req.url}`);
  const maybeToken = url.searchParams.get('token');
  if (maybeToken === hashedPassword) {
    return next();
  }
```

allow ws clients without cookies to connect

cpvalente: how would the token come up here? do we need an interface in ontime to generate the token or should the websocket client include the raw password in the connection?
What is the normal use case?

alex-Arc: I was thinking for when companion connects to the ws endpoint directly 

Yeah some place to get the token would be needed, but that could be in the dashboard

cpvalente:
@coderabbitai full review

PR #1458, opened 2025-01-18
design review: timer

cpvalente:
Starting the design review with the timer and associated components. This is divided in two parts

Style and code tweaks
- [x] Simplify getting params
- [x] Ensure timer colour can be overridden in all phases
- [ ] Style timer type none
- [x] Clock seconds is substring
- [x] handle longer titles
- [x] mobile views
- [x] what should be displayed when nothing is playing
- [x] what should be displayed when waiting to start (roll mode)
- [x] what should be displayed when loaded but not started

Behaviour and structure changes
- [x] Timer type overrides per view
- [ ] Minimal and Clock are subviews


cpvalente:
@coderabbitai full review

PR #1467, opened 2025-01-24
Collapsible params

cpvalente:
This work is part of the design review effort

since we are getting more stuff here, I thought that making the sections collapsible can help us organise

The code changes look worse than it is, just a lot of moving around.
You can ignore the code and just evaluate the UX


alex-Arc (approved):
Nice!

could they default to closed?

cpvalente:
> Nice!
> 
> could they default to closed?

I think it would be ok to have them open the first time, this would also be the smallest change

PR #1468, opened 2025-01-25
More batch

alex-Arc:
- add batching to the client `runtime.ts`
- rearrange and explain the update logic `RuntimeService.ts`
- fix some strict stuff in `RuntimeService.ts`
- fix a issue where the paused state was not properly applied from a restore point in `runtimeState.ts`

### DIFF
--- a/apps/client/src/common/stores/runtime.ts
+++ b/apps/client/src/common/stores/runtime.ts
@@ -14,6 +14,18 @@ export const runtimeStore = createWithEqualityFn<RuntimeStore>(
 export const useRuntimeStore = <T>(selector: (state: RuntimeStore) => T) =>
   useStoreWithEqualityFn(runtimeStore, selector, deepCompare);
 
+let batchStore: Partial<RuntimeStore> = {};
+
+export function addToBatchUpdates<K extends keyof RuntimeStore>(key: K, value: RuntimeStore[K]) {
+  batchStore[key] = value;
+}
+
+export function flushBatchUpdates() {
+  const state = runtimeStore.getState();
+  runtimeStore.setState({ ...state, ...batchStore });
+  batchStore = {};
+}
+
 /**
  * Allows patching a property of the runtime store
  */
--- a/apps/client/src/common/utils/socket.ts
+++ b/apps/client/src/common/utils/socket.ts
@@ -14,7 +14,7 @@ import {
 } from '../stores/clientStore';
 import { addDialog } from '../stores/dialogStore';
 import { addLog } from '../stores/logger';
-import { patchRuntime, patchRuntimeProperty } from '../stores/runtime';
+import { addToBatchUpdates, flushBatchUpdates, patchRuntime, patchRuntimeProperty } from '../stores/runtime';
 
 export let websocket: WebSocket | null = null;
 let reconnectTimeout: NodeJS.Timeout | null = null;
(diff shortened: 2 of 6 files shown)


alex-Arc:
@coderabbitai review

apps/server/src/stores/runtimeState.ts
```diff
@@ -136,10 +136,12 @@ export function clear() {
  * Utility to allow modifying the state from the outside
  * @param newState
  */
-function patchTimer(newState: Partial<TimerState>) {
+function patchTimer(newState: Partial<TimerState & RestorePoint>) {
   for (const key in newState) {
     if (key in runtimeState.timer) {
       runtimeState.timer[key] = newState[key];
+    } else if (key in runtimeState._timer) {
+      runtimeState._timer[key] = newState[key];
```
cpvalente, line 144: the _timer tries to signify that this is a private property that we use to save some state. I dont think we should be passing it around or expose it externally

apps/server/src/services/runtime-service/RuntimeService.ts
```diff
@@ -681,71 +682,100 @@ function broadcastResult(_target: any, _propertyKey: string, descriptor: Propert
...
+     * - notification rate has been exceeded
+     *
+     * Then check if there is actually a change in the data
+     */
+    const shouldRuntimeUpdate =
+      (hasImmediateChanges || shouldUpdateTimer || getForceUpdate(RuntimeService.previousRuntimeUpdate, state.clock)) &&
+      !deepEqual(RuntimeService.previousState?.runtime, state.runtime);
+
+    /**
+     * the currentBlock object has no ticking values so we only need to check for equallety
```
cpvalente, line 724:
```suggestion
     * the currentBlock object has no ticking values so we only need to check for equality
```

apps/server/src/stores/runtimeState.ts
alex-Arc: I is strictly needed for the restore service to work properly
currently it restores to a strange state. If you pause a playback and force restart the server it will look like pause but continue playing

apps/client/src/common/stores/runtime.ts
```diff
@@ -14,13 +14,25 @@ export const runtimeStore = createWithEqualityFn<RuntimeStore>(
 export const useRuntimeStore = <T>(selector: (state: RuntimeStore) => T) =>
   useStoreWithEqualityFn(runtimeStore, selector, deepCompare);
 
+let batchStore: Partial<RuntimeStore> = {};
+let isUpdatePending: NodeJS.Timeout | null = null;
+
 /**
  * Allows patching a property of the runtime store
  */
 export function patchRuntimeProperty<K extends keyof RuntimeStore>(key: K, value: RuntimeStore[K]) {
```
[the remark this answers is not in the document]
alex-Arc: yeah, the slower the connection the longer timeout is needed

apps/server/src/services/runtime-service/RuntimeService.ts
```diff
@@ -181,10 +182,10 @@ class RuntimeService {
...
     const nextPublic = state.publicEventNext?.id;
     return (
-      affectedIds.includes(now) ||
-      affectedIds.includes(nowPublic) ||
-      affectedIds.includes(next) ||
-      affectedIds.includes(nextPublic)
+      (now !== undefined && affectedIds.includes(now)) ||
+      (nowPublic !== undefined && affectedIds.includes(nowPublic)) ||
+      (next !== undefined && affectedIds.includes(next)) ||
+      (nextPublic !== undefined && affectedIds.includes(nextPublic))
```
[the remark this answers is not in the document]
alex-Arc: strict mode still complains with the simple  `now && affectedIds`

should it just be ignored?

apps/server/src/stores/runtimeState.ts
cpvalente: Nah, sorry. Just leave it as you have it. Would be nice to enable strict mode in the server eventually

apps/server/src/services/runtime-service/RuntimeService.ts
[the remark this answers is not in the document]
cpvalente: Nah, sorry. Just leave it as you have it. Would be nice to enable strict mode in the server eventually

apps/server/src/stores/runtimeState.ts
cpvalente: Just add that explanation to the code

apps/client/src/common/utils/socket.ts
```diff
@@ -14,7 +14,7 @@ import {
 } from '../stores/clientStore';
 import { addDialog } from '../stores/dialogStore';
 import { addLog } from '../stores/logger';
-import { patchRuntime, patchRuntimeProperty } from '../stores/runtime';
+import { addToBatchUpdates, flushBatchUpdates, patchRuntime, patchRuntimeProperty } from '../stores/runtime';
```
cpvalente, line 17: Oh, I mean that the batch - flush could be on the server.

I assume that if the server sends all the messages when it is done, they will arrive within some milliseconds of each other

I wonder if this way we could have updates arriving out of order and not being flushed which would cause some weird state bugs?

cpvalente (approved):
Just left some small suggestions for comments. Looks good otherwise

PR #1469, opened 2025-01-25
Report

alex-Arc:
#1139

## now
- [x] record start and stop times of events
- [x] show event report in editor
- [x] settings panel for reporter
- [x] clear all reports

## later ?
- [ ] clear individual report in editor
- [ ] record start and stop times of blocks
- [ ] show report in cuesheet
- [ ] export a report
- [ ] intelligent clearing of reports (should it reports be cleared if the duration of an event is changed to avoid confusion) or on start/end of show


apps/client/src/features/rundown/event-block/EventBlock.tsx
```diff
@@ -177,6 +178,12 @@ export default function EventBlock(props: EventBlockProps) {
           },
           { withDivider: true, label: 'Clone', icon: IoDuplicateOutline, onClick: () => actionHandler('clone') },
           { withDivider: true, label: 'Delete', icon: IoTrash, onClick: () => actionHandler('delete') },
+          {
+            withDivider: true,
+            label: 'Clear Report',
+            icon: IoTrashBin,
```
cpvalente, line 184: I wonder if we want per event clear or clearing everything?

Otherwise:
how about just an X for symbol so it doesnt conflict with the delete?
Should it also be above delete?

```suggestion
            label: 'Clear report',
            icon: IoTrashBin,
```

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
```diff
@@ -75,3 +78,51 @@ function EventUntil(props: EventUntilProps) {
...
+    if (!startedAt || !endedAt) {
+      return [null, false];
+    }
+
+    const actualDuration = endedAt - startedAt;
+    const difference = actualDuration - duration;
+    const absDifference = Math.abs(difference);
+
+    if (absDifference < MILLIS_PER_SECOND) {
+      return ['==', false]; //TODO: how to symbolize ontime?
```
cpvalente, line 108: good question, how about just a checkmark? either unicode ✓ or a svg?

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
```diff
@@ -75,3 +78,51 @@ function EventUntil(props: EventUntilProps) {
...
+    const value = `${isOver ? '+' : '-'}${formatDuration(Math.abs(difference), difference > 2 * MILLIS_PER_MINUTE)}`;
+    return [value, isOver];
+  }, [currentReport, duration]);
+
+  if (!value) {
+    return null;
+  }
+
+  return (
+    <Tooltip label='Difference from the expected duration' openDelay={tooltipDelayFast}>
```
cpvalente, line 122: Maybe the tooltip for past events should explain the data like
- Event gained xx:xx time
- Event lost xx:xx time

I believe this may be safe since the past event times should no longer change

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
```diff
@@ -75,3 +78,51 @@ function EventUntil(props: EventUntilProps) {
...
+      <div className={cx([style.chip, isOver ? style.over : style.under, className])}>
+        <div>{value}</div>
```
cpvalente, line 124: suggestion, we dont need this extra div
```suggestion
        {value}
```

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
```diff
@@ -75,3 +78,51 @@ function EventUntil(props: EventUntilProps) {
...
   );
 }
+
+interface EventReportProps {
+  className: string;
+  id: string;
+  duration: number;
+}
+
+function EventReport(props: EventReportProps) {
```
cpvalente, line 88: I wonder if we can make this more efficient.

We know that, once an event has run, the report will never change until it runs again
Do you agree? if so, we can find a way together, to only run this once when the event finishes

apps/server/src/api-data/report/report.service.ts
```diff
@@ -0,0 +1,84 @@
+import { OntimeReport, OntimeEventReport } from 'ontime-types';
+import { RuntimeState } from '../../stores/runtimeState.js';
+
+//TODO: there seams to be some actions that should invalidate reports
+// events timer edits?
+// event delete
+// Also what about roll mode?
```
cpvalente, line 7: I think there are too many edge cases here and we should keep it simple

The report, should tell the user how long did something take last time it ran
I dont think it should matter what playback mode it is in

As for clearing:
- The user would chose to clear the entire report
- The user deletes an event, so we can clear the report for it

cpvalente (commented):
I think this is looking really good already and I am excited for this feature.
Great job with all the work to get this far

I would suggest that we simplify the code and consider the report as part of the user flow


I wonder if:
- User starts the app and clicks the `Start` -> `Next` button to go through the interface. lets call this the `progress button` just so we have a reference
- When the user is on the last event, the `progress button` button shows `End show`. Clicking end show will generate a report and terminate things
- while we have a report, the `progress button` shows `Clear report`. Clicking here will clear the report and the button will then show `Start`

Also, is it now the time for us to consider the rehearsal mode? (ie: offsets would ignore schedule and only care about the event duration)

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
alex-Arc: I will try that

apps/server/src/api-data/report/report.service.ts
alex-Arc: > I dont think it should matter what playback mode it is in

I agree we just have to find all the places to hook in.

as for the clearing
the way we display the report right now it is relative to the events duration, so of the user changes the duration of the event, dose the report still make sense?

apps/client/src/features/app-settings/panel/feature-settings-panel/ReportSettings.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const isLoading = status === 'pending';
+
+  const clear = useCallback(async () => {
+    console.log('cler')
+    await deleteAllReport();
+  }, []);
+
+  return (
+    <Panel.Section>
```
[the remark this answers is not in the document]
alex-Arc: we can do that
Was just thinking that this would also where you went to download the report

apps/client/src/common/hooks-query/useReport.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default function useReport() {
+  const { data, status } = useQuery<OntimeReport>({
+    queryKey: REPORT,
+    queryFn: fetchReport,
+    placeholderData: (previousData, _previousQuery) => previousData,
+    retry: 5,
+    retryDelay: (attempt) => attempt * 2500,
+    refetchInterval: queryRefetchIntervalSlow,
+    networkMode: 'always',
```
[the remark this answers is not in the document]
alex-Arc: yeah I think that would be better, I just can't quite figur out how it ties togetter

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
alex-Arc: i'm wondering if the words are clear
gained? 

did the event run shorter than expected so we gained some time back in a show running behind 
or
did the event it self gain time by taking to long

cpvalente: > i'm wondering if the words are clear gained?
> 
> did the event run shorter than expected so we gained some time back in a show running behind or did the event it self gain time by taking to long

hahaha, wording things is difficult. I suggest we leave it as is and wait for user feedback.
Either way, I hope the user will see that this looks like the offset and connect the dots

apps/server/src/api-data/report/report.service.ts
cpvalente: > I agree we just have to find all the places to hook in.

What if we use the same place where we do the integrations?
In that way, that becomes the side effects place

> the way we display the report right now it is relative to the events duration, so of the user changes the duration of the event, dose the report still make sense?

I believe the report should be a feature that says "last time this ran, this is what happened", in which case it doesnt matter if the user changes things post fact. This also has advantage of keeping the feature simpler and a more straightforward mental model

apps/client/src/common/hooks-query/useReport.ts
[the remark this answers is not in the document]
cpvalente: kinda like this
https://github.com/TanStack/query/discussions/1205

remember we have the messages from the server, perhaps in the ontime-refetch we could add one more case for the report?

apps/client/src/features/app-settings/panel/feature-settings-panel/ReportSettings.tsx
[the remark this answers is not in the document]
cpvalente: I agree, we can add this UI when we have the download report feature

apps/client/src/features/rundown/event-block/composite/EventBlockChip.module.scss
```diff
@@ -6,14 +6,18 @@
...
   &.over {
     color: $playback-negative;
   }
-
+  
   &.under {
     color: $playback-ahead;
   }
+  
+  &.ontime {
```
cpvalente, line 18: maybe avoid using this word? maybe we dont need this colour? it could be just the green?

apps/client/src/features/rundown/event-block/composite/EventBlockChip.module.scss
```diff
@@ -6,14 +6,18 @@
   color: $label-gray;
   padding: 0.125rem 0.5rem;
   border-radius: 2px;
-
+
```
cpvalente, line 9: tip: check your formatter, these spaces shouldnt be here

apps/client/src/features/rundown/event-block/composite/EventBlockChip.module.scss
alex-Arc: green - under time
red - over time
ontime ? blue/orange/purple ? something else? (felt like a nice place to sneak in the logo colour and i makes sense 😉  )

cpvalente: I would personally keep it simple

green: no overtime (on time or time ahead)
Orange: overtime 

this is consistent with the colours in other places

cpvalente (commented):
I am trying to catch up with code reviews and found this request
However, I realised halfway that maybe we spoke about it before and we are missing follow up?

Also: I speed ran through a loaded rundown to test out the logic and came into this situation which doesnt seem plausible
Screenshot 2025-02-09 at 19 40 48

apps/server/src/api-data/report/report.controller.ts
```diff
@@ -0,0 +1,18 @@
+import type { Request, Response } from 'express';
+import type { OntimeReport } from 'ontime-types';
+import * as report from './report.service.js';
+
+export async function getAll(_req: Request, res: Response<OntimeReport>) {
+  res.json(report.generate());
+}
+
+export async function deleteAll(_req: Request, res: Response<OntimeReport>) {
+  report.clear();
```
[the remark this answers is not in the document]
alex-Arc: Don't see how
it's a simple delete/clear of a map structure

apps/client/src/features/rundown/event-block/composite/EventBlockChip.tsx
```diff
@@ -30,47 +33,92 @@ export default function EventBlockChip(props: EventBlockChipProps) {
...
+
+  if (!value) {
+    return null;
+  }
 
   return (
-    <Tooltip label='Expected time until start' openDelay={tooltipDelayFast}>
-      <div className={cx([style.chip, isDue ? style.due : null, className])}>
-        <div>{timeUntilString}</div>
+    <Tooltip label={tooltip} openDelay={tooltipDelayFast}>
```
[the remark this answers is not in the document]
alex-Arc: It dosen't calculate other in the report, only in the report
and it currently uses date from the chip

apps/client/src/common/utils/socket.ts
```diff
@@ -196,19 +196,23 @@ export const connectSocket = () => {
...
+              ontimeQueryClient.invalidateQueries({ queryKey: CUSTOM_FIELDS });
+            }
+          } else if (target === 'REPORT') {
+            ontimeQueryClient.invalidateQueries({ queryKey: REPORT });
           }
           break;
         }
         case 'ontime-flush': {
-          flushBatchUpdates()
+          flushBatchUpdates();
```
cpvalente, line 215: question: it seems that we dont use the batch updates? maybe we can remove the logic?

cpvalente:
@coderabbitai full review

PR #1471, opened 2025-01-26
feat: generate authenticated url

cpvalente:
Adding a way to generate a link from Ontime

Screenshot 2025-01-26 at 13 15 31


alex-Arc:
do we need to contact the server for this?
we can get the token from the `document.cookie`

apps/client/src/features/app-settings/panel/network-panel/GenerateLinkForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+          </Select>
+        </Panel.ListItem>
+        <Panel.ListItem>
+          <Panel.Field
+            title='URL Preset'
+            description='Which preset will the link point to (will default to /timer if none is given)'
+          />
+          <Select variant='ontime' size='sm' {...register('path')}>
+            <option key='timer' value='timer'>
+              Timer
```
alex-Arc, line 93: here companion would need an option without a path

cpvalente:
> do we need to contact the server for this? we can get the token from the `document.cookie`

I believe we should

this allows us to keep the client having no knowledge of the cookie or the authentication process

It will also allow us to migrate into a session token process with not so much work 

is there a reason for us to want to avoid the server call?

apps/client/src/features/app-settings/panel/network-panel/GenerateLinkForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+    mode: 'onChange',
+    resetOptions: {
+      keepDirtyValues: true,
+    },
+  });
+
+  const onSubmit = async (options: GenerateLinkFormOptions) => {
+    try {
+      setFormState('loading');
+      const baseUrl = linkToOtherHost(options.baseUrl);
```
alex-Arc, line 50: dose this work past the proxy?
eg will it get the port proxy port or internal ontime port?

alex-Arc:
> is there a reason for us to want to avoid the server call?

not really just less code

apps/client/src/features/app-settings/panel/network-panel/GenerateLinkForm.tsx
cpvalente: can you please double-check whether the tests cover the possible cases?

alex-Arc (approved):

PR #1472, opened 2025-01-26
refactor: fix roll next day

cpvalente:


apps/server/src/stores/runtimeState.ts
```diff
@@ -497,7 +498,8 @@ export function update(): UpdateResult {
 
   // 2. are we waiting to roll?
   if (runtimeState.timer.playback === Playback.Roll && runtimeState.timer.secondaryTimer !== null) {
-    return updateIfWaitingToRoll();
+    const hasCrossedMidnight = previousClock > runtimeState.clock;
```
alex-Arc, line 501: yeah seams right,
should we guard a bit more?
like it should be less by at least 23 hours

cpvalente: I think you are right that we should have a safer code path. 

I suggest that instead of removing the dayInMs we recalculate the time to start. That way there is no destructive path on us getting this wrong

cpvalente: @alex-Arc I think the new logic covers this

if we think we crossed midnight, we will recalculate the target, which will again check the clock and potentially set the dayMs offset.
This could be a wasted operation (for example, if timer skipped back a little), but I believe it provides the safest outcome with the least amount of logic

alex-Arc (approved):

PR #1475, opened 2025-01-27
update types/node and osc-min

alex-Arc:
remove unused package


pnpm-workspace.yaml
```diff
@@ -6,7 +6,7 @@ catalog:
   typescript: 5.5.3
   "@typescript-eslint/eslint-plugin": 7.16.1
   "@typescript-eslint/parser": 7.16.1
-  "@types/node": 20.14.10
+  "@types/node": 22.10.10
```
cpvalente, line 9: I don’t think we can use the types for node 22. We need to find the new release of the 20

alex-Arc: hmm yeah
there is a v20 inside it that is updated, wonder how we acces that https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v20/dgram.d.ts

apps/server/src/adapters/OscAdapter.ts
```diff
@@ -28,8 +28,7 @@ class OscServer implements IAdapter {
       // args: extra data, only used on some API entries
 
       /**
-       * TODO: remove this type casting when mergend in deleration file
-       * https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71659
+       * TODO: fix ArrayBuffer casting
```
cpvalente, line 31: I am not entirely sure which casting that is, can we do it?

cpvalente (commented):
Do we not need our argument parser? I seemed to remember that the quoted string solution was specific for us?

Either way, should we maybe keep our tests to make sure that there is no change of behaviour?

apps/server/src/adapters/OscAdapter.ts
alex-Arc: just up above in the `.on('message'`

cpvalente (approved):
I think it would be helpful to have tests that expose the argument parsing logic, so it may be easier to migrate the existing tests than to write from scratch

but the code looks good

PR #1481, opened 2025-02-02
Lock views

cpvalente:
As part of the work with the URL sharing, this PR does the last work to make the cuesheet and operator views be lockable


alex-Arc (approved):
not a big issue but the lock could use some background
image

PR #1483, opened 2025-02-03
feat: add table navigation to cuesheet table

Shobhit-Nagpal:
### Changes:

- Installed [table-nav/react](https://www.npmjs.com/package/@table-nav/react) to add table navigation on `<CuesheetTable>`
- Added listeners and tab indices where required

Attaching 2 videos to show a preview (one includes editing an input and the other includes navigation through action menu if enabled):

#### Editing input with table navigation:

https://github.com/user-attachments/assets/a1ad9e54-a04f-4b9c-8f24-948dd7b29165

#### Navigating through action menu:
https://github.com/user-attachments/assets/d4b90ef3-0692-4996-93d4-bec2bb727d59

### Things to keep in mind:

- Currently, we can only navigate to a `BlockRow` from the first cell of the `EventRow` (if action menu is enabled, then we can navigate through action menu, else we navigate through the index as shown in the video)

Let me know if there's any changes required on my end


Shobhit-Nagpal:
hey, i believe the issue of losing focus after we submit the value is fixed now.

overview of the changes:
1. Refactored `useReactiveTextInput` hook to factor in keyboard navigation
2. Introduced `focusParentElement` method on the refs that is being passed into `SingleLineCell` and `TextLikeInput`

Here's a recording to show the working:


https://github.com/user-attachments/assets/1af839ec-232a-4a2b-8e16-461b8370d8d9

Let me know if there's any changes required

Shobhit-Nagpal:
Hey, I looked over a few issues we discussed in the channel and these are my findings:

- I went through the library's source code properly and turns out we're only restoring grid navigation when we press Escape. There's no case for Enter key here. The library is using parentElements for traversal and not for focusing on any parentElement itself.

- As for the race condition hypothesis, I'm mostly confirmed that there's a race condition on pressing Enter because the event listeners would execute in the order: Input -> Table -> setTimeout inside the input. However, considering there's no case of handling Enter in the library for any submission, we only would have to manually focus on the cell in this case.

- I've cleaned up the code a little bit where I've added a allowKeyboardNavigation option under the hook so as to not conflict with any other inputs using this hook.

- For the TimeInput, I believe we're using the right approach there considering the re-render from SingleLineCell to TextLikeInput


I went through html's specification on what happens when we call blur and focus, they end up calling a whole chain of elements and see whether they're focusable or not. So we're only calling focus on parentElement in cases of submitting an input, for the rest where we press Escape, the library restores grid navigation which is already working.

PR #1490, opened 2025-02-07
feat: add autocomplete to osc arguments

cpvalente:
this adds autocomplete for templates in the osc arguments

Screenshot 2025-02-07 at 19 40 10


alex-Arc (approved):
wow nice
would there be a way to remember to keep the templates up to date with the actual runtime state?

cpvalente:
> wow nice would there be a way to remember to keep the templates up to date with the actual runtime state?

I have started doing it, that is why the runtime state was refactored, but then I realised that the code necessary was just as likely to be forgotten since we need to pick individual keys of the state.
On the other hand, it added a lot of complexity

Lets keep this in mind and revisit if we ever plan to open up the entire runtime state

PR #1496, opened 2025-02-14
Backstage review

cpvalente:
style review for backstage
- [x] improve spacing
- [x] create empty states
- [x] show countdown to start in pending roll
- [x] create phone and tablet variants

improve schedule
- [x] schedule shows as many items as it fits
- [x] stop cycling option fixes on first page
- [x] show projected time (needs time until?)

I have also decided to remove the animation since it caused too large layout shifts
This may also mean we may be able to shake off motion as part of the chakra migration


apps/client/src/views/backstage/Backstage.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const clock = formatTime(time.clock);
+  const isPendingStart = getIsPendingStart(time.playback, time.phase);
+  const startedAt = isPendingStart ? formatTime(time.secondaryTimer) : formatTime(time.startedAt);
+  const scheduledStart =
+    hasEvents && showNow ? '' : formatTime(runtime.plannedStart, { format12: 'hh:mm a', format24: 'HH:mm' });
+  const scheduledEnd =
+    hasEvents && showNow ? '' : formatTime(runtime.plannedEnd, { format12: 'hh:mm a', format24: 'HH:mm' });
+
+  let stageTimer = millisToString(time.current, { fallback: timerPlaceholderMin });
+  stageTimer = removeLeadingZero(stageTimer);
```
alex-Arc, line 90: We call this stage timer but don't follow the same commends that the stage timer dose eg show message/aux/external

might be ok...

apps/client/src/views/backstage/Backstage.tsx
```diff
@@ -0,0 +1,178 @@
...
+      {showProgress && <ProgressBar className='progress-container' current={time.current} duration={time.duration} />}
+
+      {!hasEvents && <Empty text={getLocalizedString('common.no_data')} className='empty-container' />}
+
+      <div className='card-container'>
+        {showNow && (
+          <div className={cx(['event', 'now', blinkClass && 'blink'])}>
+            <TitleCard title={nowMain} secondary={nowSecondary} />
+            <div className='timer-group'>
+              <div className='aux-timers'>
```
alex-Arc, line 122: This has nothing to do with the actual aux timer right?
could it get a different name?

apps/client/src/views/backstage/Backstage.tsx
cpvalente: thats a good point

I think the showrunner may want the presenter to see different things now and then.
However, the crew backstage is mostly interested in how long its left for the next thing

However, this does not solve the case where the showrunner wants the people backstage to see an aux timer... Maybe we can take that another time once we have a clear use case?

alex-Arc: sounds good, but maybe just call it timer and not stage timer?

cpvalente: sounds good, called it `displayTimer`

apps/client/src/views/backstage/Backstage.tsx
cpvalente: done

alex-Arc (approved):

PR #1503, opened 2025-02-17
Cuesheet menu

cpvalente:
Trying to improve performance in the cuesheet by stabilising and sharing action menu


alex-Arc (approved):
look's good
just hope it doesn't block  #1483

PR #1507, opened 2025-02-18
L3 feat expantion

alex-Arc:
- refactor the animation system
- use transition instead of animations
- only animate in if playback is active


alex-Arc:
@coderabbitai review

cpvalente (commented):
There seems like we still have some issues with coordination:
- when the element animates in and out, the text reveal is not waiting for the line to finish. I think the line animation needs to be faster
- if you press play, wait for animation and and out, and then stop. the lower third animates an empty line

These are just small things and otherwise it looks good.
I also left small suggestions to help with the code

Good work

Issue #1451, opened 2025-01-14
Add Photo for presenters


hadphild:
Add a photo for the presenters.

hadphild:
![Screenshot 2025-01-15 195511](https://github.com/user-attachments/assets/660080a3-c7a8-4b05-8ec0-69ded495e6d8)

Something like this. Just so the crew could know who is onstage if photos are added. This could be good from getting in Slide preview or before the show.

cpvalente:
Thank you @hadphild, it is very helpful to have more context on the requests so we can be clearer on what we are discussing

As long as we can do it without negative impacting the other functions of the app I think this is a good idea, and I will investigate what that means for Ontime.

hadphild:
If we can add photos of who is in each section. This could be used in a lot of the views and help crew identify who is onstage.

cpvalente:
Hi @hadphild , this has been added in v3.13.0

We have decided to only allow images from an URL right now and hope this is a good start for users
Could you please give us some feedback so we can close this ticket?

hadphild:
Sounds good start and to see if people do use this idea.

PR #1510, opened 2025-02-19
feat: add custom field type of image

cpvalente:
This is a proof of concept for allowing users to add images in a special custom field type
The code should follow the current work in the cuesheet in #1483 

> Either way, there are still some decisions to be made
> - is it sufficient that the user gives us image links?
> - how about users leveraging ontime offline?
> - We could have a combined flow where users could upload a file or give us the link, what would happen to the file transfer flow?
> - what should happen in the operator view?
> - what should happen in all the views that consume images as a secondarySource?

We have decided to go with a simple implementation that only receives images as external URLs
Users should be able to consume and change this images in the editor and cuesheet
To answer the questions above:

>- is it sufficient that the user gives us image links?

Yes

>- how about users leveraging ontime offline?

We will need to look into this as a separate effort

>- We could have a combined flow where users could upload a file or give us the link, what would happen to the file transfer flow?

The initial suggestion is that we pack the project file with all its assets, this sounds like a v4 thing

>- what should happen in the operator view?

Nothing

>- what should happen in all the views that consume images as a secondarySource?

The fields should not be available 

Screenshot 2025-02-19 at 20 07 18

As mentioned in #1451

### DIFF
--- a/apps/client/src/features/rundown/event-editor/EventEditor.tsx
+++ b/apps/client/src/features/rundown/event-editor/EventEditor.tsx
@@ -8,9 +8,11 @@ import useCustomFields from '../../../common/hooks-query/useCustomFields';
 import { getAccessibleColour } from '../../../common/utils/styleUtils';
 import * as Editor from '../../editors/editor-utils/EditorUtils';
 
+import EventEditorImage from './composite/EventEditorImage';
 import EventEditorTimes from './composite/EventEditorTimes';
 import EventEditorTitles from './composite/EventEditorTitles';
 import EventTextArea from './composite/EventTextArea';
+import EventTextInput from './composite/EventTextInput';
 import EventEditorEmpty from './EventEditorEmpty';
 
 import style from './EventEditor.module.scss';
@@ -87,6 +89,7 @@ export default function EventEditor(props: EventEditorProps) {
             </Button>
           )}
         </Editor.Title>
+
         {Object.keys(customFields).map((fieldKey) => {
           const key = `${event.id}-${fieldKey}`;
           const fieldName = `custom-${fieldKey}`;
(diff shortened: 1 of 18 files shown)


apps/client/src/features/rundown/event-editor/EventEditor.tsx
```diff
@@ -87,24 +89,47 @@ export default function EventEditor(props: EventEditorProps) {
...
+                className={style.decorated}
+                style={{ '--decorator-bg': backgroundColor, '--decorator-color': color } as CSSProperties}
+              />
+            );
+          }
+
+          if (customFields[fieldKey].type === 'image') {
+            return (
+              <div key={key} className={style.customImage}>
+                <EventTextInput
```
alex-Arc, line 117: The length limit on the text input seams a bit to strict considering image url's

apps/client/src/features/rundown/event-editor/composite/EventEditorImage.tsx
```diff
@@ -0,0 +1,16 @@
...
+interface EventEditorImageProps {
+  src: string;
+}
+
+export default function EventEditorImage(props: EventEditorImageProps) {
+  const { src } = props;
+
+  return (
+    <div className={style.imageContainer}>
+      <img loading='lazy' src={src} alt='' crossOrigin='anonymous' />
```
alex-Arc, line 12:
```suggestion
      <img loading='lazy' src={src} alt={!src ? '' : ' '} crossOrigin='anonymous' />
```
if we did something like this the broken image icon would show if it fails to get the link

cpvalente: good call

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/EditableImage.tsx
```diff
@@ -0,0 +1,55 @@
+import { memo } from 'react';
+import { Input } from '@chakra-ui/react';
+
+import style from './EditableImage.module.scss';
+
+interface EditableImageProps {
+  initialValue: string;
```
alex-Arc, line 7: This value is undefined when the event has no data for the custom filed

it leads to it showing the delete button even though there is no value

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
```diff
@@ -105,6 +106,24 @@ function MakeMultiLineField({ row, column, table }: CellContext<OntimeRundownEnt
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps -- we skip table.options.meta since the reference seems unstable
+    [column.id, row.index],
+  );
+
+  const event = row.original;
+  if (!isOntimeEvent(event)) {
+    return null;
+  }
+
+  const initialValue = event.custom[column.id];
```
alex-Arc, line 123: could also be fixed here
```suggestion
  const initialValue = event.custom[column.id] ?? '';
```

alex-Arc (approved):

PR #1511, opened 2025-02-21
feat: allow redirecting to preset

cpvalente:
This is a small code change to allow the redirect modal to use URL presets

Screenshot 2025-02-21 at 14 40 16

### DIFF
--- a/apps/client/src/common/components/client-modal/RedirectClientModal.tsx
+++ b/apps/client/src/common/components/client-modal/RedirectClientModal.tsx
@@ -12,6 +12,7 @@ import {
 } from '@chakra-ui/react';
 import { IoArrowForward } from '@react-icons/all-files/io5/IoArrowForward';
 
+import { navigatorConstants } from '../../../viewerConfig';
 import { setClientRemote } from '../../hooks/useSocket';
 import useUrlPresets from '../../hooks-query/useUrlPresets';
 import Info from '../info/Info';
@@ -39,6 +40,7 @@ export function RedirectClientModal(props: RedirectClientModalProps) {
     if (newPath === '/' || newPath === currentPath) {
       return;
     }
+    console.log('----> redirect to', newPath);
     setRedirect({ target: id, redirect: newPath });
     onClose();
   };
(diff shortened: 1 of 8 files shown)


apps/client/src/common/components/client-modal/RedirectClientModal.tsx
```diff
@@ -1,64 +1,128 @@
...
+                <option value='/timer'>Timer</option>
+                <option value='/minimal'>Minimal Timer</option>
+                <option value='/clock'>Wall Clock</option>
+                <option value='/backstage'>Backstage</option>
+                <option value='/timeline'>Timeline</option>
+                <option value='/public'>Public</option>
+                <option value='/lower'>Lower Thirds</option>
+                <option value='/studio'>Studio Clock</option>
+                <option value='/countdown'>Countdown</option>
+                <option value='/info'>Project Info</option>
```
alex-Arc, line 83:
```suggestion
                <option value='timer'>Timer</option>
                <option value='minimal'>Minimal Timer</option>
                <option value='clock'>Wall Clock</option>
                <option value='backstage'>Backstage</option>
                <option value='timeline'>Timeline</option>
                <option value='public'>Public</option>
                <option value='lower'>Lower Thirds</option>
                <option value='studio'>Studio Clock</option>
                <option value='countdown'>Countdown</option>
                <option value='info'>Project Info</option>
```

apps/client/src/common/components/client-modal/RedirectClientModal.tsx
```diff
@@ -1,64 +1,128 @@
...
+            </a>
+          </Info>
+          <div>
+            <span className={style.label}>Select View or URL Preset</span>
+            <div className={style.textEntry}>
+              <Select
+                size='md'
+                variant='ontime'
+                isDisabled={enabledPresets.length === 0}
+                onChange={(event) => setSelected(`/${event.target.value}`)}
```
alex-Arc, line 71:
```suggestion
                onChange={(event) => setSelected(`${event.target.value}`)}
```

this was resulting in double slashes

alex-Arc (changes requested):

apps/client/src/common/components/client-modal/RedirectClientModal.tsx
cpvalente: good catch, I have fixed the issue

alex-Arc (approved):

PR #1512, opened 2025-02-21
Relative run mode

alex-Arc:


apps/server/src/stores/runtimeState.ts
```diff
@@ -93,7 +95,8 @@ export function clear() {
   runtimeState.publicEventNext = null;
 
   runtimeState.runtime.offset = 0;
-  runtimeState.runtime.actualStart = null;
+  runtimeState.runtime.relativeOffset = 0;
```
cpvalente, line 98: question: had we agreed to create two variants of the clear function like
softClear() and hardClear()?

apps/client/src/features/rundown/rundown-header/RundownHeader.tsx
```diff
@@ -22,6 +29,22 @@ export default function RundownHeader() {
           Edit
         </Button>
       </ButtonGroup>
+      <ButtonGroup isAttached>
+        <Button
+          size='sm'
+          variant={offsetMode === OffsetMode.Absolute ? 'ontime-filled' : 'ontime-subtle'}
+          onClick={setAbsoluteOffset}
```
cpvalente, line 36: if this continues being slow, how about using `onMouseDown`? that could help it feel more responsive

apps/client/src/common/api/settings.ts
```diff
@@ -26,3 +26,10 @@ export async function postSettings(data: Settings): Promise<AxiosResponse<Settin
 export async function postShowWelcomeDialog(show: boolean) {
   axios.post(`${settingsPath}/welcomedialog`, { show });
 }
+
+/**
+ * Allows setting offset mode
+ */
+export async function postOffsetMode(mode: OffsetMode) {
+  axios.post(`${settingsPath}/offsetmode`, { mode });
+}
```
cpvalente, line 35: our (soft) rules indicate that this feature, which is not persisted, should be communicated over websockets

cpvalente (commented):
I think this is looking good

alex-Arc:
@coderabbitai review

alex-Arc:
@cpvalente did you have an idea for the UI you could push?

apps/server/src/stores/runtimeState.ts
```diff
@@ -82,7 +84,32 @@ export function getState(): Readonly<RuntimeState> {
   };
 }
 
-export function clear() {
+//TODO: find other things that dose not need clearing
```
cpvalente, line 87: lets do it!

suggestion: could we have a single function resetState() which receives a patch of things to keep?

apps/server/src/stores/runtimeState.ts
```diff
@@ -155,9 +183,9 @@ export function load(
   initialData?: Partial<TimerState & RestorePoint>,
 ): boolean {
   // we need to persist the current block state across loads
-  const prevCurrentBlock = { ...runtimeState.currentBlock };
-  clear();
-  runtimeState.currentBlock = prevCurrentBlock;
+  // const prevCurrentBlock = { ...runtimeState.currentBlock };
+  softClear();
+  // runtimeState.currentBlock = prevCurrentBlock;
```
cpvalente, line 188: question: can we remove the commented out code?

cpvalente (commented):
I think this looks good, lets just double check the clearing logic

apps/server/src/stores/runtimeState.ts
alex-Arc: hmmm maybe
It looks to be difficult/compute intensive if we don't just what it to be a shallow patch...
that might be ok but right now we are already keeping one deep value `_time.totalDelay`

cpvalente: i dont think we should leave this comment behind as it indicates unfinished (and maybe buggy) behaviour

If we don't currently have a good idea on how to clean up lets leave as is, but rename the functions to give indication of what is happening. maybe like `clearState` and `clearEventData`?
Future us will also appreciate a comment as to what the new clear event scope is

apps/server/src/stores/runtimeState.ts
```diff
@@ -79,10 +85,36 @@ export function getState(): Readonly<RuntimeState> {
...
-export function clear() {
+//TODO: find other things that dose not need clearing
+export function softClear() {
+  runtimeState.eventNow = null;
+  runtimeState.publicEventNow = null;
+  runtimeState.eventNext = null;
+  runtimeState.publicEventNext = null;
+
+  // runtimeState.currentBlock.block = null;
+  // runtimeState.currentBlock.startedAt = null;
```
cpvalente, line 100: i believe this can bring confusion going forwards, lets comment its value or remove it

alex-Arc:
@coderabbitai review

apps/client/src/common/utils/__tests__/time.test.ts
```diff
@@ -40,82 +43,265 @@ describe('formatTime()', () => {
...
-      clock: 50,
-      offset: -20,
+      // totalGap: 10,
+      clock: 100,
+      actualStart: 100,
+      plannedStart: 0,
+      offset: 0, // relative
+      // offset: -100, // absolute
+      offsetMode: OffsetMode.Relative,
+      isLinkedToLoaded: false,
```
cpvalente, line 299: question: what do the commented out values mean?

cpvalente (approved):
Looks good, do you feel like we need a beta?

PR #1516, opened 2025-02-23
import id's from sheet

alex-Arc:
Allows also importing and exporting id's from sheets.

Adding other merge types is a bit more involved as we are currently using the same rundown parser for sheets that is also parsing our project file, I think they would need to be separated

### DIFF
--- a/apps/server/src/utils/parser.ts
+++ b/apps/server/src/utils/parser.ts
@@ -250,7 +250,7 @@ export const parseExcel = (
       } else if (j === colourIndex) {
         event.colour = makeString(column, '');
       } else if (j === entryIdIndex) {
-        event.id = makeString(column, undefined);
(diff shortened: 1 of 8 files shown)


apps/server/src/utils/parser.ts
```diff
@@ -185,11 +188,15 @@ export const parseExcel = (
       },
       [importMap.timeWarning]: (row: number, col: number) => {
         timeWarningIndex = col;
-        rundownMetadata['timeWarningIndex'] = { row, col };
+        rundownMetadata['timeWarning'] = { row, col };
```
cpvalente, line 191: i didnt understand this change, is this a bug that you are fixing?

cpvalente (approved):
didnt have an opportunity to test, but it looks fine code wise

questions:
- do we need to make changes to the template?
- how do we guarantee the IDs are unique?

PR #1527, opened 2025-03-02
feat: add table navigation to cuesheet table (#1483)

cpvalente:
* refactor: cuesheet table body
* feat: add table navigation


alex-Arc (approved):
I might not be something we want to spend time trying to fix it,
but if you click on a header and try to navigate you are stuck
but if you navigate from one of the fields up into the header you can continue freely

PR #1535, opened 2025-03-08
v4

cpvalente:


apps/client/src/common/utils/eventsManager.ts
```diff
@@ -23,6 +23,7 @@ export const cloneEvent = (event: OntimeEvent): ClonedEvent => {
     isPublic: event.isPublic,
     skip: event.skip,
     colour: event.colour,
+    currentBlock: event.currentBlock,
```
alex-Arc, line 26: Could we consider renaming it to something like parentBlock, since it hase nothing to to with the actually current block?

cpvalente: that is OK with me, lets do it once the groups code is merged so we don't have to deal with so many merge conflicts

PR #1542, opened 2025-03-16
feat: css editor

Shobhit-Nagpal:
Hey, opening a PR for implementing the CSS editor for custom styles building on top of #1514. Summary of changes:

- Compared between Prismjs and Shiki for syntax highlighting (Prismjs was giving a lesser bundle size)
- Shifted the CSS editor inside a modal which is being lazy loaded
- Implemented endpoints for posting and restoring css
- Added a prebuild script to write bundled css to the override.css file (do let me know if any changes are required here)
- Enabling the Edit CSS button when the override styles value is true

Here's a video to show the working:

https://github.com/user-attachments/assets/e9852081-7a3a-4fb8-853a-d5ba34b04693

Do let me know if any changes are required


apps/client/src/features/app-settings/panel/general-panel/StyleEditor.tsx
```diff
@@ -0,0 +1,43 @@
+import { useEffect, useState } from 'react';
+import Editor from 'react-simple-code-editor';
+import Prism from 'prismjs';
```
cpvalente, line 3: question: could we minimise the impact on bundle size by using the import 
`import Prism from 'prismjs/components/prism-core';`

apps/client/src/features/app-settings/panel/general-panel/StyleEditor.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import 'prismjs/components/prism-json';
+
+import 'prismjs/components/prism-css';
+import 'prismjs/themes/prism-funky.min.css';
```
cpvalente, line 8: Can we use the tomorrow night?

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+  }, []);
+
+  return (
+    <Modal isOpen={isOpen} onClose={onClose} variant='ontime' isCentered>
+      <ModalOverlay />
+      <ModalContent maxWidth='max(800px, 40vw)' padding='1rem'>
+        <ModalHeader>Edit CSS</ModalHeader>
+        <ModalCloseButton />
+        <ModalBody>
+          <CodeEditor onChange={(updatedCss: string) => setCSS(updatedCss)} initialValue={css} language='css' />
```
cpvalente, line 70: this structure would effectively cause the entire modal to re-render on every change of the code editor.

Can we isolate the `onChange` inside the `<CodeEditor>` component?

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
```diff
@@ -0,0 +1,105 @@
+import { useEffect, useState } from 'react';
+import {
+  Button,
+  Flex,
+  HStack,
```
cpvalente, line 5: cleanup: we dont use layout components from Chakra and have our own. You can search other interfaces in settings to see how things are composed or I can change if you prefer

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+  useEffect(() => {
+    async function fetchServerCSS() {
+      const css = await getCSSContents();
+      setCSS(css);
+    }
+    fetchServerCSS();
+  }, []);
+
+  return (
+    <Modal isOpen={isOpen} onClose={onClose} variant='ontime' isCentered>
```
cpvalente, line 64: question: what does the isCentered do? we dont have it in any other modals and they are all in the center

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
```diff
@@ -14,11 +14,14 @@ import { preventEscape } from '../../../../common/utils/keyEvent';
 import { isOntimeCloud } from '../../../../externals';
 import * as Panel from '../../panel-utils/PanelUtils';
 
+const CodeEditorModal = lazy(() => import('./StyleEditorModal'));
```
cpvalente, line 17: question: what do you think about lazy loading the `<CodeEditor>` instead?
I wonder if we can get it to load while we fetch the CSS?

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
```diff
@@ -115,6 +119,17 @@ export default function ViewSettingsForm() {
                   <Switch variant='ontime' size='lg' isChecked={value} onChange={onChange} ref={ref} />
                 )}
               />
+              <Panel.InlineElements>
```
cpvalente, line 122: question: why do we need the `<InlineElements>` component?

apps/client/src/features/app-settings/panel/general-panel/StyleEditor.tsx
```diff
@@ -0,0 +1,43 @@
...
+import style from './StyleEditor.module.scss';
+
+interface CodeEditorProps {
+  language: string;
+  initialValue: string;
+  onChange: (newValue: string) => void;
+}
+
+const CodeEditor = (props: CodeEditorProps) => {
```
cpvalente, line 17: questions: 
- we usually use the function keyword and export default inline, is there a reason not to do this here?
- is this component a good candidate for memoisation? it sounds like it can be slow to render

apps/server/src/api-data/db/db.controller.ts
```diff
@@ -311,3 +315,54 @@ export async function deleteProjectFile(req: Request, res: Response<MessageRespo
     res.status(500).send({ message });
   }
 }
+
```
cpvalente, line 318: it feels a bit strange to me to have this code here, but I dont know where to put it. What do you think of the following?

- the controllers shouldnt really be writing to disk, we would prefer having a very thin service to do the writing part
- how about creating an `asset` directory with `asset.router` + `asset.controller` + `assert.service` for this?

cpvalente (commented):
Good work, I left some small comments and suggestions for cleanup that you can consider

apps/client/src/features/app-settings/panel/general-panel/StyleEditor.tsx
Shobhit-Nagpal: optimized the bundle size. should be good on this end. working on other changes mentioned

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
Shobhit-Nagpal: in case the editor has no content, the modal is towards the top. the isCentered here is to keep the modal in center even if there's no content. attaching a preview for what happens when we remove this prop and the css is empty

image

apps/server/src/api-data/db/db.controller.ts
Shobhit-Nagpal: i like this idea, i'll make the changes accordingly

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
cpvalente: I see, thank you for clarifying.

I think what we need to do is to give the editor a fixed size and remove the is centered prop

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
Shobhit-Nagpal: without it, the button is taking full width. attaching a preview:

image

do you feel it's better to manipulate the width of the button through the button's props itself?

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
Shobhit-Nagpal: yes, this seemed to work. added a min height for the editor as you had suggested, works well now

apps/client/src/features/app-settings/panel/general-panel/StyleEditor.module.scss
```diff
@@ -0,0 +1,10 @@
+.code {
+  font-family: monospace;
+  font-size: 1rem;
+}
```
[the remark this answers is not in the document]
Shobhit-Nagpal: not in prism but the code editor seems to handle it. made the changes accordingly

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
cpvalente: yes, I think this is a CSS problem, not a layout one.
I dont think we should solve it by adding more elements to the DOM

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+                variant='ontime-ghosted'
+                onClick={handleReset}
+                isDisabled={saveLoading || resetLoading}
+                isLoading={resetLoading}
+              >
+                Reset to default
+              </Button>
+            </div>
+            <div>
+              <div className={style.buttonRow}>
```
cpvalente, line 90: we have the Panel.InlineElements which we should use instead

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
```diff
@@ -115,6 +119,16 @@ export default function ViewSettingsForm() {
                   <Switch variant='ontime' size='lg' isChecked={value} onChange={onChange} ref={ref} />
                 )}
               />
+              <Button
+                onClick={onCodeEditorOpen}
+                variant='ontime-subtle'
+                size='sm'
+                marginTop='1rem'
+                isDisabled={!data.overrideStyles}
+                width={24}
```
cpvalente, line 128: question: 
I wonder if we can improve the CSS here
- layout items like margin-top should come from the parents (ideally), is that possible here?
- the width 24 seems strange, we dont want to hard code a width, we just want to stop if from being 100%. So the correct value would be like `width: fit-contents`

apps/server/scripts/bundleCss.ts
```diff
@@ -0,0 +1,25 @@
...
+    const cssFile = path.resolve(stylesDir, 'override.css');
+
+    if (!existsSync(cssFile)) {
+      throw new Error('File does not exist');
+    }
+
+    await writeFile(cssFile, defaultCss, { encoding: 'utf8' });
+  } catch (error) {
+    console.error('Failed writing to CSS file: ', error);
+    process.exit(1);
```
cpvalente, line 21: cleanup: I dont think we need this, the program will terminate after this line anyway and there is nothing catching this directive

apps/server/src/api-data/assets/assets.router.ts
```diff
@@ -0,0 +1,8 @@
+import express from 'express';
+import { getCssOverride, postCssOverride, restoreCss } from './assets.controller.js';
+
+export const router = express.Router();
+
+router.get('/css', getCssOverride);
```
cpvalente, line 6: nice!

apps/server/src/api-data/assets/assets.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Reads the user's css file
+ * @returns css contents in the file
+ */
+export async function readCssFile(): Promise<string> {
+  const path = publicFiles.cssOverride;
+  if (!existsSync(path)) {
+    throw new Error('File not found');
+  }
```
cpvalente, line 13: question: if the path does not exist (ie: the user deleted the file mid session), should we write the contents of the bundled css to a new file?

apps/server/src/api-data/assets/assets.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Writes the user's css file
+ * @param css the updated css to write to file
+ */
+export async function writeCssFile(css: string) {
+  const path = publicFiles.cssOverride;
+  if (!existsSync(path)) {
+    throw new Error('File not found');
+  }
```
cpvalente, line 28: question: if the CSS file does not exist, should we make one to write the contents to?

apps/client/src/features/app-settings/panel/general-panel/StyleEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <div className={style.wrapper}>
+      <Editor
+        value={code}
+        padding={15}
+        onValueChange={handleChange}
+        highlight={highlight}
+        style={{
+          fontFamily: 'monospace',
+          fontSize: 12,
+          minHeight: 500,
```
cpvalente, line 49: question: does the minHeight mean that the contents can grow? we likely do not want that. we want the editor to have a fixed size with content scrolling if necessary

apps/client/src/features/app-settings/panel/general-panel/StyleEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+function CodeEditor(props: CodeEditorProps) {
+  const { language, initialValue, cssRef } = props;
+
+  const [code, setCode] = useState(initialValue);
+
+  useEffect(() => {
+    setCode(initialValue);
+  }, [initialValue]);
+
+  useEffect(() => {
```
cpvalente, line 24: cleanup: can you please write a comment to what these useEffects achieve? like

`sync editor contents to external source`
and
`add contents to editor on mount`

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export default function CodeEditorModal(props: CodeEditorModalProps) {
+  const { isOpen, onClose } = props;
+
+  const [css, setCSS] = useState('');
+  const [saveLoading, setSaveLoading] = useState(false);
+  const [resetLoading, setResetLoading] = useState(false);
+
+  const cssRef = useRef<string>(css);
```
cpvalente, line 31: question: I wonder if the CSS ref is misplaced

It feels strange for the parent to control the contents of the child this way. 
The UI should be a function of state.
When the reset function is called, the contents change, this state should trigger a re-render which updates the Editor without us explicitely doing so

apps/client/src/features/app-settings/panel/general-panel/StyleEditorModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const handleReset = async () => {
```
cpvalente, line 33: nitpick: should this be called `handleRestore`? resetting and restoring provide different expectations as to what will happen

cpvalente (commented):
Great work, I didnt get to test it but had a look at the code and I think we can do some minor improvements

I will try to find some time to test the usability of it, but assume everything is OK

apps/client/src/features/app-settings/panel/general-panel/StyleEditor.tsx
Shobhit-Nagpal: the minHeight is to give a minimum height in case there's no css so that the modal still retains its structure.

for the content growing, it's not growing with content. if there's more content than the height, then scrolling kicks in. attaching a preview below:


https://github.com/user-attachments/assets/5c0cbb09-8f83-4775-af23-ca8d77bddef8

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
Shobhit-Nagpal: > layout items like margin-top should come from the parents (ideally), is that possible here?

looked into this, one way this is possible is if we add put the button under a `div` and add margin-top to that, but that's rendering an extra node only to be able to put the margin.

Something similar is done in the `<GenerateLinkForm>` under the Network Panel with more styling [here](https://github.com/cpvalente/ontime/blob/73afbacdd76c2259c0b4caeebe852730a6ffc9ec/apps/client/src/features/app-settings/panel/network-panel/GenerateLinkForm.tsx#L128)

One option is to put a `row-gap` under the `.listItem` styling being applied on the `<Panel.ListItem>` component [here](https://github.com/cpvalente/ontime/blob/73afbacdd76c2259c0b4caeebe852730a6ffc9ec/apps/client/src/features/app-settings/panel-utils/PanelUtils.tsx#L90). This eliminates the need for adding a margin-top on the new rows that come in under `<Panel.ListItem>`

Shobhit-Nagpal:
hey, made changes according to the previous review:

- removed `process.exit` in catch block
- changed `handleReset` to `handleRestore`
- added comments on useEffect
- added modal button row under `<Panel.InlineElements>`
- handled modal render and the ref misplacement by using `useImperativeHandle` to get the changed css from the `StyleEditor` (UI should be a function of state now and not dependent on the ref for any changes on the editor)
- handled file not existing by creating the file with the bundledCss instead of throwing an error
- added disabling/enabling of save button according to the contents of the editor

let me know if any further changes are required

alex-Arc (approved):

PR #1548, opened 2025-03-18
Fix link

cpvalente:
Fixes issue with users not being able to create a link in Ontime cloud and copy it easily


cpvalente:
@coderabbitai full review

PR #1557, opened 2025-03-26
Event trigger

alex-Arc:


apps/client/src/features/rundown/event-editor/composite/EventTriggers.tsx
```diff
@@ -0,0 +1,159 @@
...
+      </Select>
+      <Select
+        size='sm'
+        variant='ontime'
+        value={cycleValue}
+        onChange={(e) => setCycleValue(e.target.value as TimerLifeCycle)}
+        defaultValue={TimerLifeCycle.onStart}
+      >
+        {/* Notice here that we don't support onStop */}
+        {['onLoad', 'onStart', 'onPause', 'onFinish', 'onWarning', 'onDanger'].map((cycle) => (
```
cpvalente, line 85: This sounds like a decision we should extract from the view. In general, views shouldnt contain many decisions other than presentation ones

cpvalente (commented):
Logic looks ok, lets just try to cleanup a bit since this is a relatively complex part of the app

I will review the designs slightly before we merge

apps/client/src/features/rundown/event-editor/composite/EventTriggers.tsx
cpvalente: I usually just make a auxiliary file like event trigger.constants.ts

alex-Arc:
@coderabbitai review

cpvalente (approved):

PR #1564, opened 2025-04-02
fix: revision inc on swap

alex-Arc:


cpvalente:
Hi @alex-Arc , I dont think this change is necessary.
The idea here was to keep this function free of side effects. the revision is bumped by the consumers
This also allow us to bump the revision just in the servers, so that the UI can optimistically update but still receive the update from the backend and recognise it as the new version

Maybe I missed something?

alex-Arc:
No sure then
The test was falling, but maybe the test should be changed then?

alex-Arc:
Here is how the revision bump was handled in master
https://github.com/cpvalente/ontime/blob/3ebb3451e4693d5ed7db4976ce4c27bcebbb7e84/apps/server/src/services/rundown-service/rundownCache.ts#L447-L450

but the test in master did not care what way round the revisions were incremented since they both stared at 0
https://github.com/cpvalente/ontime/blob/3ebb3451e4693d5ed7db4976ce4c27bcebbb7e84/apps/server/src/services/rundown-service/__tests__/rundownCache.test.ts#L532-L536

cpvalente:
> Here is how the revision bump was handled in master
> 
> https://github.com/cpvalente/ontime/blob/3ebb3451e4693d5ed7db4976ce4c27bcebbb7e84/apps/server/src/services/rundown-service/rundownCache.ts#L447-L450
> 
> but the test in master did not care what way round the revisions were incremented since they both stared at 0
> 
> https://github.com/cpvalente/ontime/blob/3ebb3451e4693d5ed7db4976ce4c27bcebbb7e84/apps/server/src/services/rundown-service/__tests__/rundownCache.test.ts#L532-L536

Makes sense

I still think it is best to keep the operation away from side effects
In which case updating the tests is the better option

PR #1565, opened 2025-04-08
Fix: time until calculation past cout-to-end events

alex-Arc:


apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -300,7 +300,7 @@ export default function Rundown({ data }: RundownProps) {
                   isNextDay = checkIsNextDay(entry, lastEvent);
                   if (!isPast) {
                     totalGap += entry.gap;
-                    isLinkedToLoaded = isLinkedToLoaded && entry.linkStart !== null;
+                    isLinkedToLoaded = isLinkedToLoaded && entry.linkStart !== null && !lastEvent?.countToEnd;
```
cpvalente, line 303: I am sure this is fine, but I did not understand the change 

How are these related? Maybe the name is just incorrect?

alex-Arc: They are related down stream in the time until calculation
since a unlinked event and a linked event coming after an count to end will have the same time until behaviour

i think... it feels correct when trying it out

PR #1570, opened 2025-04-11
refactor: simplify logic

cpvalente:


alex-Arc (approved):

packages/utils/src/common/arrayUtils.ts
```diff
@@ -31,7 +31,7 @@ export function insertAtIndex<T>(index: number, item: T, array: T[]): T[] {
  * @param array
  */
 export function deleteAtIndex<T>(index: number, array: T[]) {
-  return array.toSpliced(index, 1);
+  return array.filter((_, i) => i !== index);
```
alex-Arc, line 34: is `.splice` not more effective or do you want to avoid the in-place mutation

cpvalente: maybe, this was just the logic we had before. not sure why we used the filter method, but it is also returning a copy so it is probably fine

Issue #1585, opened 2025-04-18
Styling Bug on Studio Clock when seconds are hidden


dasrecht:
Hi there 👋 

I've noticed that the studio clock has a styling issue when the seconds are hidden. Which leads to the time overflowing the clock boundaries

<img width="1250" alt="Image" src="https://github.com/user-attachments/assets/50df6316-d4ad-4581-b9d4-425be6fe2a60" />

The CSS handles the `--with-seconds` already nicely ([SCSS](https://github.com/cpvalente/ontime/blob/ddd629d7db918d72f1608ca1fca29ebec340ff8d/apps/client/src/features/viewers/studio/StudioClock.scss#L101-L102)). As soon as the `Hide seconds in timer` is enabled.

I think there are now 2 ways of going about it:

- Removing seconds not only for the timer and additionally for the clock (I already committed that in my fork so that would be ready quickly - [yes it's a simple change](https://github.com/cpvalente/ontime/compare/master...dasrecht:ontime:master) :) )
- Adding an option which allows hiding seconds for the time and the clock independently

I'm torn between both options. Hiding the seconds in the entire timers makes it easier (and is also how it's done the timer view)

Happy to discuss.

/dasrecht

cpvalente:
Hi @dasrecht , thank you for bringing this up to my attention.

In some of the views, we use a naive approach to sizing by inferring it from the size available and the amount of characters we want to show. This works especially well for the clocks which have well known characters

As for the solution, there are a few things at play here:
- The studio clock is [being redesigned](https://github.com/cpvalente/ontime/pull/1517), and we likely don't want to make too large changes now
- The separation of options between timer and clock are global to all views, and we don't want to change it in a single place

With these in mind I would suggest:

**a) Tweak the [base multiplier](https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/viewers/studio/StudioClock.scss#L96) to fit**
Something like this seems to work well.
```css
font-size: calc(var(--clock-size) / 6); // <--- 6 instead of 5
```
This is a naive solution that gets us moving on without too much fuss. Has the downside of making the timers slightly smaller than they could be

**b) Ensure the `--with-seconds` is present** 
From what I can see, the modifier tries to communicate whether the timer contains more than 5 characters (hh:mm vs hh:mm:ss), and asks the UI to scale in response. The fact that this does not work, seems like an implementation bug and the real issue to fix.
[...]

PR #1590, opened 2025-04-20
fix - fix studio clock sizing when timer seconds are hidden  !1585

dasrecht:
Fix as discussed in https://github.com/cpvalente/ontime/issues/1585


cpvalente:
Nice and pragmatic, thank you @dasrecht 

I will add this to our next release, meanwhile, would you be interested in contributing with some code to Ontime?

PR #1593, opened 2025-04-20
Refactor: ms for api calls

alex-Arc:


apps/client/src/features/control/playback/add-time/AddTime.tsx
```diff
@@ -28,9 +28,9 @@ export default function AddTime(props: AddTimeProps) {
   const handleAddTime = (direction: 'add' | 'remove') => {
     // API expects input in seconds
```
cpvalente, line 29: cleanup: looks like this comment is no longer valid

apps/client/src/features/control/playback/add-time/AddTime.tsx
```diff
@@ -28,9 +28,9 @@ export default function AddTime(props: AddTimeProps) {
...
     if (direction === 'add') {
-      setPlayback.addTime(time / MILLIS_PER_SECOND);
+      setPlayback.addTime(time);
```
cpvalente, line 31: suggestion: what do you think about renaming this variable `timeInMs` to avoid questions in future?

apps/client/src/features/control/playback/aux-timer/AuxTimer.tsx
```diff
@@ -64,7 +64,7 @@ function AuxTimerInput() {
 
   const handleTimeUpdate = (_field: string, value: string) => {
     const newTime = parseUserTime(value);
```
cpvalente, line 66: suggestion: should we rename `newTime` to `newTimeInMs`

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -201,12 +201,11 @@ const actionHandlers: Record<string, ActionHandler> = {
       return { payload: 'success' };
     }
 
-    const timeToAdd = time * MILLIS_PER_SECOND; // frontend is seconds based
-    if (Math.abs(timeToAdd) > MILLIS_PER_HOUR) {
+    if (Math.abs(time) > MILLIS_PER_HOUR) {
```
cpvalente, line 204: suggestion: this is outside of your scope, but this logic is overly complex and I wonder if we should cleanup

I would suggest something simpler like 
```ts
  addtime: (payload) => {
    const time = (() => {
      if (typeof payload === 'object' && payload !== null) {
        if ('add' in payload) return numberOrError(payload.add);
        if ('remove' in payload) return numberOrError(payload.remove) * -1;
      }
      return numberOrError(payload);
    })();

    if (Math.abs(time) > MILLIS_PER_HOUR) {
      throw new Error(`Payload too large: ${time}`);
    }

    if (time !== 0) {
      runtimeService.addTime(time);
    }

    return { payload: 'success' };
  },
```

cpvalente (commented):
left some small suggestions for improvement.
Other than the old comment, the changes are optional for you to consider

cpvalente (approved):
🚀

PR #1594, opened 2025-04-20
Refactor: better rounding

alex-Arc:
The goal of this PR is to change the millis to seconds so that it rounds differently depending on it counting up or down

when counting down we round up so that the value will change at the exact point that the value is correct
and the opposite when counting up

also looks to help a bit with the wall clock skipping seconds sometimes

### DIFF
--- a/packages/utils/src/date-utils/conversionUtils.ts
+++ b/packages/utils/src/date-utils/conversionUtils.ts
@@ -1,4 +1,5 @@
 import type { MaybeNumber } from 'ontime-types';
+import { TimerType } from 'ontime-types';
 
 export const MILLIS_PER_SECOND = 1000;
 export const MILLIS_PER_MINUTE = 1000 * 60;
@@ -25,8 +26,22 @@ function convertMillis(millis: MaybeNumber, conversion: number): number {
  * @param millis
  * @returns
  */
-export function millisToSeconds(millis: MaybeNumber): number {
-  return convertMillis(millis, MILLIS_PER_SECOND);
+export function millisToSeconds(
+  millis: MaybeNumber,
+  direction: TimerType.CountDown | TimerType.CountUp = TimerType.CountDown,
+) {
+  if (millis === null) return 0;
+
+  if (direction === TimerType.CountDown) {
+    // 0 + is there to avoid result giving -0
+    return 0 + Math.ceil(millis / MILLIS_PER_SECOND);
+  }
+
+  if (direction === TimerType.CountUp) {
+    return Math.floor(millis / MILLIS_PER_SECOND);
+  }
+
+  throw new Error('?????');
 }
 
 /**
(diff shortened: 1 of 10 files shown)


packages/utils/src/date-utils/conversionUtils.ts
```diff
@@ -7,44 +8,27 @@ export const MILLIS_PER_HOUR = 1000 * 60 * 60;
...
-}
+export function millisToSeconds(
+  millis: MaybeNumber,
+  direction: TimerType.CountDown | TimerType.CountUp = TimerType.CountDown,
+) {
+  if (millis === null) return 0;
+
+  if (direction === TimerType.CountDown) {
+    // 0 + is there to avoid result giving -0
+    return 0 + Math.ceil(millis / MILLIS_PER_SECOND);
```
cpvalente, line 24: suggestion: what do you think about something like

```ts
const seconds = Math.ceil(millis / MILLIS_PER_SECOND);
return seconds === 0 ? 0 : seconds;
```

packages/utils/src/date-utils/conversionUtils.ts
```diff
@@ -7,44 +8,27 @@ export const MILLIS_PER_HOUR = 1000 * 60 * 60;
...
+    return Math.floor(millis / MILLIS_PER_SECOND);
+  }
 
-/**
- * Converts value in milliseconds to hours
- * @param millis
- * @returns
- */
-export function millisToHours(millis: MaybeNumber): number {
-  return convertMillis(millis, MILLIS_PER_HOUR);
```
cpvalente, line 47: question: dont we need these functions anymore?

cpvalente (commented):
The code seems to go in the right direction, but I have some reservations about the change. 
I am also skeptic that such a large change, produce so little code changes, and I fear that we may not be seeing all the ramifications the change has

I would suggest having a look through to make sure we are not forgetting anything. If we believe this is what we need, lets cleanup the code so that we can demo it to users. Maybe Erich can help us here

packages/utils/src/date-utils/conversionUtils.ts
```diff
@@ -7,44 +8,27 @@ export const MILLIS_PER_HOUR = 1000 * 60 * 60;
...
+  throw new Error('?????');
```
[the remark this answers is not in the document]
alex-Arc: yeah 😉

packages/utils/src/date-utils/conversionUtils.ts
alex-Arc: no could not find a use of it
I can revert this and we kan do a clean-up round later?

cpvalente: no, these functions are trivially written so if they are not in use lets bin it

PR #1595, opened 2025-04-20
Refactor: WebSocket from flush queue to one patch

alex-Arc:


apps/server/src/stores/EventStore.ts
```diff
@@ -32,10 +32,11 @@ export const eventStore = {
     //if there is already and update pending we don't need to schedule another one
     if (!isUpdatePending) {
```
cpvalente, line 33: I wonder what would be the design differences in letting the consumer control this

ie: event store could expose a batch function, which returns a closure with a send function.
in that way the consumer would do 

```ts
const batch = eventStore.createBatch()

batch.add(timer, {});
batch.add(clock, {});
...
batch.send()
```

The immediate advantage is that we are sure that the patches are under our control and we know what is in them, the current solution could mix patches between sources

I also wonder if the logic would be simpler and not rely on scheduling magic

What do you think?

apps/server/src/stores/EventStore.ts
```diff
@@ -32,10 +32,11 @@ export const eventStore = {
...
       isUpdatePending = setImmediate(() => {
+        const patch: Partial<RuntimeStore> = {};
         for (const dataKey of changedKeys) {
-          socket.sendAsJson({ type: `ontime-${dataKey}`, payload: store[dataKey] });
+          Object.assign(patch, { [dataKey]: store[dataKey] });
```
cpvalente, line 37: suggestion: as for adding the keys to the object. I think a direct assignment can lead to simpler code and it will also be more performant

All the structures here are read-only so we dont really need the immutability that comes with Object.assign

cpvalente (commented):
Good stuff, I left some comments to hopefully improve this logic

Meanwhile, as I was reviewing this, I also found a potential bug.
``ts
export function patchRuntimeProperty<K extends keyof RuntimeStore>(key: K, value: RuntimeStore[K]) {
  const state = runtimeStore.getState();
  state[key] = value;
  runtimeStore.setState({ ...state });
}
```

in this code, we mutate the state before creating a new object with the spread. I believe the correct logic should be 
```ts
  const state = runtimeStore.getState();
  runtimeStore.setState({ ...state, [key]: value });
```

apps/server/src/stores/EventStore.ts
alex-Arc: yeah agree, but could not get typescript to like this, might be misunderstanding something
```ts
patch[dataKey] = store[dataKey]
```

cpvalente: I think this problem will go away if we adopt the batch
the batch function will receive a key + value and we can just use that like

```ts
patch[dynamicKey] = value
````
in which chase the typing is in the function signature

cpvalente (approved):
With this logic, we assume that we we change top level properties of state, and that the top level objects must be complete (no deep patching)

I think that is OK but we need to keep this in mind

PR #1597, opened 2025-04-27
Fix project renumber

alex-Arc:
every project load was causing a renumber so I
moved the `generateUniqueFileName` logic from the ´loadProject´ function to the `createProject` function
and changed the deafult / new show funtions to use `create` instead of `load`

### DIFF
--- a/apps/server/src/services/project-service/ProjectService.ts
+++ b/apps/server/src/services/project-service/ProjectService.ts
@@ -82,11 +82,8 @@ export async function getCurrentProject(): Promise<{ filename: string; pathToFil
  * Private function loads a project file and handles necessary side effects
  */
 async function loadProject(projectData: DatabaseModel, projectName: string) {
-  // we need to make sure the file name is unique in the projects directory
-  const pathToNewFile = generateUniqueFileName(publicDir.projectsDir, projectName);
-
   // change LowDB to point to new file
-  await initPersistence(getPathToProject(pathToNewFile), projectData);
+  await initPersistence(getPathToProject(projectName), projectData);
   logger.info(LogOrigin.Server, `Loaded project ${projectName}`);
 
   // stop the runtime service
@@ -97,8 +94,8 @@ async function loadProject(projectData: DatabaseModel, projectName: string) {
   await initRundown(firstRundown, projectData.customFields);
 
   // persist the project selection
-  const newName = getFileNameFromPath(pathToNewFile);
-  await setLastLoadedProject(newName);
+  const newName = getFileNameFromPath(projectName);
+  await setLastLoadedProject(projectName);
 
   // update the service state
   currentProjectState = {
@@ -113,7 +110,7 @@ async function loadProject(projectData: DatabaseModel, projectName: string) {
  * Loads the demo project
  */
 export async function loadDemoProject(): Promise<string> {
-  return loadProject(demoDb, config.demoProject);
+  return createProject(config.demoProject, demoDb);
 }
 
 /**
@@ -121,7 +118,7 @@ export async function loadDemoProject(): Promise<string> {
  * to be composed in the loading functions
  */
 async function loadNewProject(): Promise<string> {
-  return loadProject(dbModel, config.newProject);
+  return createProject(config.newProject, dbModel);
 }
 
 /**
@@ -259,10 +256,9 @@ export async function renameProjectFile(originalFile: string, newFilename: strin
  */
 export async function createProject(filename: string, initialData: Partial<DatabaseModel>): Promise<string> {
   const data = safeMerge(dbModel, initialData);
-
-  const fileNameWithExtension = ensureJsonExtension(filename);
-  const newFilename = await loadProject(data, fileNameWithExtension);
-  return newFilename;
+  const fileNameWithExtension = generateUniqueFileName(publicDir.projectsDir, ensureJsonExtension(filename));
+  await loadProject(data, fileNameWithExtension);
+  return fileNameWithExtension;
 }
 
 /**
--- a/apps/server/src/services/project-service/ProjectService.ts
+++ b/apps/server/src/services/project-service/ProjectService.ts
@@ -80,6 +80,7 @@ export async function getCurrentProject(): Promise<{ filename: string; pathToFil
 
 /**
  * Private function loads a project file and handles necessary side effects
+ * @param projectData
  * @param fileName file name of the project including the extension
  */
 async function loadProject(projectData: DatabaseModel, fileName: string) {
@@ -125,8 +126,8 @@ async function loadNewProject(): Promise<string> {
 /**
  * Private function handles side effects on corrupted files
  * Corrupted files in this context contain data that failed domain validation
- * @param filePath file name of the project including the extension
- * @param filename file name of the project including the extension
+ * @param filePath path to the project type include the fileName and extension
+ * @param fileName as extracted from filePath, includes extension
  */
 async function handleCorruptedFile(filePath: string, fileName: string): Promise<string> {
   // copy file to corrupted folder
@@ -258,6 +259,7 @@ export async function renameProjectFile(originalFile: string, newFilename: strin
 /**
  * Creates a new project file and applies its result
  * @param fileName file name of the project including the extension
+ * @param initialData db to initialize the project with
  */
 export async function createProject(fileName: string, initialData: Partial<DatabaseModel>): Promise<string> {
   const data = safeMerge(dbModel, initialData);
--- a/apps/server/src/services/project-service/ProjectService.ts
+++ b/apps/server/src/services/project-service/ProjectService.ts
@@ -80,11 +80,12 @@ export async function getCurrentProject(): Promise<{ filename: string; pathToFil
 
 /**
  * Private function loads a project file and handles necessary side effects
+ * @param fileName file name of the project including the extention
  */
-async function loadProject(projectData: DatabaseModel, projectName: string) {
+async function loadProject(projectData: DatabaseModel, fileName: string) {
   // change LowDB to point to new file
-  await initPersistence(getPathToProject(projectName), projectData);
-  logger.info(LogOrigin.Server, `Loaded project ${projectName}`);
+  await initPersistence(getPathToProject(fileName), projectData);
+  logger.info(LogOrigin.Server, `Loaded project ${fileName}`);
 
   // stop the runtime service
   runtimeService.stop();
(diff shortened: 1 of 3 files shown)


apps/server/src/services/project-service/ProjectService.ts
```diff
@@ -80,14 +80,12 @@ export async function getCurrentProject(): Promise<{ filename: string; pathToFil
 
 /**
  * Private function loads a project file and handles necessary side effects
+ * @param fileName file name of the project including the extention
```
cpvalente, line 83: cleanup: I am not against adding JSDocs if we feel it adds value. Here the clarification that the name includes description could be worth it. But then we should add docs for the entire interface

Also, small typo
```suggestion
 * @param fileName file name of the project including the extension
```

apps/server/src/services/project-service/ProjectService.ts
```diff
@@ -97,8 +95,8 @@ async function loadProject(projectData: DatabaseModel, projectName: string) {
   await initRundown(firstRundown, projectData.customFields);
 
   // persist the project selection
-  const newName = getFileNameFromPath(pathToNewFile);
-  await setLastLoadedProject(newName);
+  const newName = getFileNameFromPath(fileName);
```
cpvalente, line 98: question: `getFileNameFromPath` retrieves the name from a path. However, we are already passing it a filename
This brings me to a few doubts:
- perhaps we should not be using this function
- `newName` provides incorrect expectations since there are no changes to the name

apps/server/src/services/project-service/ProjectService.ts
```diff
@@ -256,13 +257,13 @@ export async function renameProjectFile(originalFile: string, newFilename: strin
 
 /**
  * Creates a new project file and applies its result
+ * @param fileName file name of the project including the extention
```
cpvalente, line 260:
```suggestion
 * @param fileName file name of the project including the extension
```

apps/server/src/utils/fileManagement.ts
```diff
@@ -56,7 +56,7 @@ export function generateUniqueFileName(directory: string, filename: string): str
   const extension = extname(filename);
   const baseName = basename(filename, extension);
 
-  let counter = 0;
+  let counter = Number(baseName.match(/\((\d)\)$/)?.at(1) ?? 0);
```
cpvalente, line 59: suggestion: we could use the `getProjectNumber` function here to minimise the amount of times we need to maintain this regex and to simplify reading the code

apps/server/src/utils/fileManagement.ts
```diff
@@ -114,3 +114,11 @@ export async function dockerSafeRename(oldPath: PathLike, newPath: PathLike) {
   await copyFile(oldPath, newPath);
   await unlink(oldPath);
 }
+
+/**
+ * finds potential file index number in our (*) format
+ * @param baseName the base name of the project file, without extention af path
+ */
+export function getProjectNumber(baseName: string): number {
+  return Number(baseName.match(/ \((\d+)\)$/)?.at(1) ?? 0);
```
cpvalente, line 123: suggestion: I think regexes are a bit difficult to read and I wonder whether the corresponding code would be trivial enough that the regex is not worth it. The code might also be more robust. What do you think?

apps/server/src/utils/fileManagement.ts
```diff
@@ -114,3 +114,11 @@ export async function dockerSafeRename(oldPath: PathLike, newPath: PathLike) {
   await copyFile(oldPath, newPath);
   await unlink(oldPath);
 }
+
+/**
+ * finds potential file index number in our (*) format
+ * @param baseName the base name of the project file, without extention af path
```
cpvalente, line 120: question: does it matter if the path and extension are there? if so, would we be better off writing some simple logic to extract the number in parenthesis instead of regex?

```suggestion
 * @param baseName the base name of the project file, without extension or path
```

apps/server/src/services/project-service/ProjectService.ts
```diff
@@ -113,20 +111,22 @@ async function loadProject(projectData: DatabaseModel, projectName: string) {
...
 async function loadNewProject(): Promise<string> {
-  return loadProject(dbModel, config.newProject);
+  return createProject(config.newProject, dbModel);
 }
 
 /**
  * Private function handles side effects on corrupted files
  * Corrupted files in this context contain data that failed domain validation
+ * @param filePath file name of the project including the extention
+ * @param filename file name of the project including the extention
```
cpvalente, line 129:
```suggestion
 * @param filePath path to the project type include the fileName and extension
 * @param fileName as extracted from filePath, includes extension
```

cpvalente (commented):
I think the change looks good
I just left some small cleanup suggestions you can consider

Also, can we point this to master?
It sounds like there are no breaking changes and current users can already benefit from this change

apps/server/src/utils/fileManagement.ts
alex-Arc: ah yes, I just forgot to use it after extracting it

apps/server/src/utils/fileManagement.ts
alex-Arc: yes this only works without extension, it could be changed to work with it but
all that is already handled in the `generateUniqueFileName` where the information about the path and extension is needed anyway

apps/server/src/services/project-service/ProjectService.ts
alex-Arc: I am also a bit unsure of the `projectData` since it is not here we are parsing it and it seams to only be used as a fallback value in `initPersistence` ?

apps/server/src/utils/fileManagement.ts
cpvalente: I would suggest simplifying the regex to a simple function that looks for the last parenthesis. That should mean that it doesn’t care about extensions

apps/server/src/services/project-service/ProjectService.ts
cpvalente: it looks like we use the projectData as data source but I agree that it seems unnecessary and the filename should be enough. I suggest we move on so we dont risk breaking things unless we have a clear idea on how to re-write it

apps/server/src/utils/__tests__/fileManagement.test.ts
```diff
@@ -90,3 +95,25 @@ describe('generateUniqueFileName', () => {
...
+describe('file index', () => {
+  it('sets index to 1 when there is no index', () => {
+    expect(incrementProjectNumber('test file.json')).toBe('test file (1).json');
+  });
+
+  it('increments to 2 when index is 1', () => {
+    expect(incrementProjectNumber('test file (1).json')).toBe('test file (2).json');
+  });
+
+  it('dose not count number not warped in parenthesise', () => {
```
cpvalente, line 108:
```suggestion
  it('does not count number not wrapped in parenthesis', () => {
```

apps/server/src/utils/__tests__/fileManagement.test.ts
```diff
@@ -90,3 +95,25 @@ describe('generateUniqueFileName', () => {
...
+    expect(incrementProjectNumber('test file 1.json')).toBe('test file 1 (1).json');
+  });
+
+  it('dose not count number if there is not a space', () => {
```
cpvalente, line 112: Question: is there a reason we want to make this affordance in code? It seems like the logic could be simpler if we decided not to care

```suggestion
  it('does not count number if there is not a space', () => {
```

apps/server/src/utils/fileManagement.ts
```diff
@@ -114,3 +109,20 @@ export async function dockerSafeRename(oldPath: PathLike, newPath: PathLike) {
   await copyFile(oldPath, newPath);
   await unlink(oldPath);
 }
+
+/**
+ * finds potential file index number in our (*) format and increments
+ */
```
cpvalente, line 115: maybe we can describe a bit better the business decisions here

```suggestion
/**
 * finds potential file index number in our (*) format and increments
 * the number section (*) must be separated from the name by a space
 * @example incrementProjectNumber('test(1).json') -> 'test(1).json'
  * @example incrementProjectNumber('test (1).json') -> 'test(2).json'
 */
```

apps/server/src/utils/fileManagement.ts
```diff
@@ -114,3 +109,20 @@ export async function dockerSafeRename(oldPath: PathLike, newPath: PathLike) {
...
+export function incrementProjectNumber(path: string): string {
+  const { dir, name, ext } = parse(path);
+
+  if (!name.endsWith(')')) return join(dir, `${name} (1)${ext}`);
```
cpvalente, line 119: question: is there a reason why we cant just demand the existence of extension?

apps/server/src/utils/fileManagement.ts
```diff
@@ -114,3 +109,20 @@ export async function dockerSafeRename(oldPath: PathLike, newPath: PathLike) {
...
+
+  const openingParenIndex = name.lastIndexOf(' (');
+  if (openingParenIndex === -1) return join(dir, `${name} (1)${ext}`);
+
+  const maybeNumber = Number(name.slice(openingParenIndex + 2, -1));
```
cpvalente, line 124: question: what if we looked for the index of closing parenthesis and tried to parse the things in between?

apps/server/src/utils/fileManagement.ts
```diff
@@ -114,3 +109,20 @@ export async function dockerSafeRename(oldPath: PathLike, newPath: PathLike) {
   await copyFile(oldPath, newPath);
   await unlink(oldPath);
 }
+
+/**
+ * finds potential file index number in our (*) format and increments
+ */
+export function incrementProjectNumber(path: string): string {
```
cpvalente, line 116: suggestion: could we simplify this logic with a more generic algorithm?

- look for last opening parenthesis `(`
- if this exists, look for closing last parenthesis `)`

- try to parse the content inside the parenthesis as a number
- add `+1` to the number
- replace the new content where the old one was

- if any of the above fails or the order of parenthesis is invalid (opening is after closing) is invalid add the (1) before the last `.`

alex-Arc: if the `( )` is not at the end I dont think we can consider it ours?

apps/server/src/utils/fileManagement.ts
alex-Arc: not sure i understand here?
Am I not just assuming the extension?

cpvalente: yes, you are correct.

if the path is not really a path and there is no extension, the logic will break

PR #1601, opened 2025-05-01
fix test

alex-Arc:
the metadate is calculated as async so needs to be forced to run

### DIFF
--- a/apps/server/src/stores/__tests__/runtimeState.test.ts
+++ b/apps/server/src/stores/__tests__/runtimeState.test.ts
@@ -88,8 +88,13 @@ describe('mutation on runtimeState', () => {
   });
 
   describe('playback operations', async () => {
-    it('refuses if nothing is loaded', () => {
-      initRundown(makeRundown({}), {});
+    it('refuses if nothing is loaded', async () => {
+      // force update
+      vi.useFakeTimers();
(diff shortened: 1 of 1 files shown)


apps/server/src/stores/__tests__/runtimeState.test.ts
```diff
@@ -323,11 +339,17 @@ describe('roll mode', () => {
...
+      await initRundown(rundown, {});
+      vi.runAllTimers();
+
       load(rundown.entries[1] as PlayableEvent, rundown, rundown.order);
       start();
+      const currentOffset = getState().runtime.offset;
       let result = roll(rundown, rundown.order, getState().runtime.offset);
       expect(result).toStrictEqual({ eventId: '1', didStart: false });
-      expect(getState().runtime.offset).toBe(1000);
+      expect(getState().runtime.offset).toBe(currentOffset);
```
cpvalente, line 352: question: can you clarify why we are keeping the value of `currentOffset` and comparing it after the roll call?
Is the objective to make sure it didnt change?
If so: the name of the variable should be something like `offsetBeforeRoll`, we could also do with a small comment explaining that since it is not explicit in code

cpvalente (approved):
left a small suggestion for improvement

PR #1602, opened 2025-05-07
fix: relative offset with `count to end`

alex-Arc:


apps/server/src/services/__tests__/timerUtils.test.ts
```diff
@@ -741,7 +740,7 @@ describe('getRuntimeOffset()', () => {
       },
     } as RuntimeState;
 
-    const offset = getRuntimeOffset(state);
+    const offset = getRuntimeOffset(state).absoluteOffset;
```
cpvalente, line 743: nitpick: i think this whole test suite would become clearer if we used destructuring instead. Since we are introducing variants of offset, the word `offset` is no longer specific enough to describe a dataset

apps/server/src/stores/runtimeState.ts
```diff
@@ -482,8 +483,9 @@ export function addTime(amount: number) {
   runtimeState.timer.current += amount;
 
   // update runtime delays: over - under
-  runtimeState.runtime.offset = getRuntimeOffset(runtimeState);
-  runtimeState.runtime.relativeOffset = getRelativeOffset(runtimeState);
+  const { absoluteOffset, relativeOffset } = getRuntimeOffset(runtimeState);
+  runtimeState.runtime.offset = absoluteOffset;
```
cpvalente, line 487: question: should we name this absoluteOffset in v4?

cpvalente (approved):
looks good, thank you

do you want to reach out to erich for a quick test?

PR #1606, opened 2025-05-11
Fix: sheet export

alex-Arc:


cpvalente (approved):
🚀

apps/server/src/api-data/excel/excel.service.ts
```diff
@@ -43,9 +43,10 @@ export function generateRundownPreview(options: ImportMap): { rundown: OntimeRun
 
   const arrayOfData: unknown[][] = xlsx.utils.sheet_to_json(data, { header: 1, blankrows: false, raw: false });
 
-  const dataFromExcel = parseExcel(arrayOfData, getCustomFields(), options);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
cpvalente, line 46: suggestion: we could avoid these changes by just having the `parseRundown` function ignore the `rundownMetadata`

PR #1607, opened 2025-05-12
fix: import preview enable after error

Shobhit-Nagpal:
## Changes made:
- fixed import preview disabled state after encountering an error being caused by `hasErrors` prop
- on importing a spreadsheet, it selects the first sheet by default in case there's no sheet named "event schedule"

note: the original idea was to add an assert, but that didn't seem to work because we had a default value of "event schedule" on `namedImportMap.Worksheet` so it was by passing that. instead, went with the approach mentioned in point 2 above.

here's a recording of how it works now (it starts off by clearing localStorage to not have any persisted import options):

https://github.com/user-attachments/assets/7a09a1e9-fb8c-46ae-b061-52d804162cc6

do let me know if any changes are to be done


apps/client/src/features/app-settings/panel/sources-panel/import-map/importMapUtils.ts
```diff
@@ -85,4 +85,4 @@ export function getPersistedOptions(): NamedImportMap {
   } catch {
     return namedImportMap;
   }
-}
+}
```
cpvalente, line 88: cleanup: can we please remove changes in this and the `override.css` file

apps/client/src/features/app-settings/panel/sources-panel/import-map/ImportMapForm.tsx
```diff
@@ -77,7 +77,15 @@ export default function ImportMapForm(props: ImportMapFormProps) {
   const isLoading = Boolean(loading);
   const canSubmitSpreadsheet = isSpreadsheet && !isLoading;
   const canSubmitGSheet = !isLoading && !stepData.worksheet.error;
-  const canSubmit = !hasErrors && isValid && (canSubmitSpreadsheet || canSubmitGSheet);
+  const canSubmit = isValid && (canSubmitSpreadsheet || canSubmitGSheet);
+
+  // Set first sheet as default worksheet when 'event schedule' sheet is not there
```
cpvalente, line 82: suggestion: lets move the `useEffect` to the top before the derived values, it makes it easier to spot and avoids mistakes with conditional hooks

apps/client/src/features/app-settings/panel/sources-panel/import-map/ImportMapForm.tsx
```diff
@@ -13,21 +13,21 @@ import { convertToImportMap, getPersistedOptions, NamedImportMap, persistImportM
...
-  hasErrors: boolean;
   isSpreadsheet: boolean;
   onCancel: () => void;
   onSubmitExport: (importMap: ImportMap) => Promise<void>;
   onSubmitImport: (importMap: ImportMap) => Promise<void>;
 }
 
 export default function ImportMapForm(props: ImportMapFormProps) {
-  const { hasErrors, isSpreadsheet, onCancel, onSubmitExport, onSubmitImport } = props;
+  const { isSpreadsheet, onCancel, onSubmitExport, onSubmitImport } = props;
```
cpvalente, line 23: question: I am not sure we can remove the `hasErrors` property here. it seems that it allows the consumer to give a error state that is global to the flow. I suggest reverting to the code that existed previously

cpvalente (commented):
Fix looks good, I left some cleanup suggestions before we can merge

PR #1619, opened 2025-05-30
Upgrade Electron and NodeJS

jwetzell:
- Upgrade Electron to latest version
- Upgrade NodeJS to align with version shipped with Electron

I have built/tested the Electron application on Ubuntu 25.04, macOS 15.5, Windows 10/11

I have tested notarization locally and that is working with the new, small config change.


cpvalente:
Hey @jwetzell, thank you for this.

Just had a quick read through the pnpm release notes
https://github.com/orgs/pnpm/discussions/8945

Ontime uses lifecycle scripts to synchronise code to static assets at build time.
We need to make sure that these still work otherwise we will see features stop working without much help as to why

I am unsure on what the changes to `pnpm test` command actually mean for us. Maybe nothing. It would be nice to confirm that we can still isolate tests like running `pnpm test rundown.utils`
Otherwise we may want to change hour test scripts

Apart from this, the changes did not seem to affect our usage

I was also curious about why we should add the build hash to the `packageManager` definition.
I am not particularly opinionated here, but I dont see any obvious advantages (perhaps some supply chain attack prevention, but either way, the individual packages contain hashes) while it makes the field non editable (ie: i cant just change the number to bump the pnpm version).
What do you think?

jwetzell:
@cpvalente I've tested checking the project out from scratch, building and running and everything seems to be in order.

I also tested things like `pnpm test timeline.utils` in `apps/client` and that seems to filter and everything. I'm not that familiar with `vitest` and what testing is all setup to do much more than that but that seems to be the expected behavior.

I would say this should be good to review now. 

The hash in the `packagerManager` entry is something the tooling I use must have done I had not noticed it. It is definitely a supply chain attack prevention and not much more (IMO). I would not feel bad about changing it to just the simplified version.

alex-Arc (approved):

PR #1621, opened 2025-05-30
[DEMO] New db

alex-Arc:


cpvalente:
I am convinced, lets do this when we create the multiple rundowns

PR #1622, opened 2025-05-31
refactor: WS

alex-Arc:
- [x] remove unused ws types
- [x] create typeing for ws transactions
- [x] ceate typeing for refetch calls


apps/client/src/common/utils/socket.ts
```diff
@@ -33,16 +33,17 @@ export const connectSocket = () => {
...
     reconnectAttempts = 0;
 
-    socketSendJson('set-client-patch', {
-      type: 'ontime',
-      origin: window.location.origin,
-      path: window.location.pathname + window.location.search,
+    sendOntimeSocket({
+      type: WsType.CLIENT_SET,
+      payload: {
+        type: 'ontime',
```
cpvalente: suggestion: it is a bit confusing with the nested types and it took me a while to understand what is happening. How about renaming  `payload.type `to `client`?

apps/client/src/common/utils/socket.ts
```diff
@@ -76,91 +77,82 @@ export const connectSocket = () => {
...
+              break;
+            case RefetchKey.RUNDOWN:
               ontimeQueryClient.invalidateQueries({ queryKey: RUNDOWN });
               ontimeQueryClient.invalidateQueries({ queryKey: CUSTOM_FIELDS });
-            }
-          } else if (target === 'REPORT') {
-            ontimeQueryClient.invalidateQueries({ queryKey: REPORT });
+              break;
+            default:
+              console.log('unknown refetch target', target);
```
cpvalente: tip: I think console.log are usually a sign of forgotten debug code.
if this is something we feel it can be helpful for us, we can keep it but as a warning potentially

I would personally suggest removing the logs in this page. if we dont know what to do with a message then its like 🤷

apps/client/src/common/utils/socket.ts
```diff
@@ -175,20 +167,17 @@ export const connectSocket = () => {
...
-export const socketSendJson = (type: string, payload?: unknown) => {
-  socketSend(
-    JSON.stringify({
-      type,
-      payload,
-    }),
-  );
-};
+export function socketSendJson(type: string, payload?: unknown): void {
+  if (websocket && websocket.readyState === WebSocket.OPEN) {
```
cpvalente, line 188: suggestion: it looks like we would benefit from a simplifying some code here

how about creating a single function that
- receives an object
- checks if the socket is open
- sends the object

It seems that both `sendOntimeSocket` and `socketSendJson` end up sending an object with type and payload. Could be nice to have the consumer knowing what to send and the `socket` module simply providing an emitter

in that case, we also dont need the `sendOntimeSocket` name and can simplify to `sendSocket`

apps/client/src/features/app-settings/panel/project-panel/ProjectData.tsx
```diff
@@ -83,8 +83,6 @@ export default function ProjectData() {
     } catch (error) {
       const message = maybeAxiosError(error);
       setError('root', { message });
-    } finally {
```
cpvalente: question: this code ensured we synced with backend, I cant remember what was the scenario here why this was necessary. is this not necessary anymore?

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -103,104 +104,57 @@ class SocketServer implements IAdapter {
...
-            this.sendClientList();
-            return;
-          }
-
-          if (type === 'set-client-type') {
-            if (payload && typeof payload == 'string') {
+            case WsType.CLIENT_SET: {
               const previousData = this.getOrCreateClient(clientId);
-              this.clients.set(clientId, { ...previousData, type: payload });
+              this.clients.set(clientId, { ...previousData, ...payload });
```
cpvalente, line 116: question: why do we need to spread payload now?
it is unclear in the typing which may cause issues in the future, maybe it warrants a comment?

apps/server/src/adapters/WebsocketAdapter.ts
```diff
@@ -103,104 +104,57 @@ class SocketServer implements IAdapter {
...
+                }
+              } catch (error) {
+                logger.error(LogOrigin.Rx, `WS IN: ${error}`);
+              }
+              break;
             }
-          } catch (error) {
-            logger.error(LogOrigin.Rx, `WS IN: ${error}`);
           }
+          return;
```
cpvalente: question: why do we need this return? it seems like there is nothing else?

apps/server/src/adapters/websocketAux.ts
```diff
@@ -1,17 +1,12 @@
...
-  Rundown = 'rundown',
-  Report = 'report',
-}
-
 /**
  * Utility function to notify clients that the REST data is stale
- * @param payload -- possible patch payload
  */
-export function sendRefetch(payload: unknown = null) {
+export function sendRefetch(target: RefetchKey, extra: unknown = null) {
```
cpvalente: suggestion: can we call it `options` instead of `extra`?

apps/server/src/api-data/project/project.controller.ts
```diff
@@ -29,11 +30,11 @@ export async function postProjectData(req: Request, res: Response<ProjectData |
       projectLogo: req.body?.projectLogo,
     });
 
-    const updatedData = await editCurrentProjectData(newData);
-
-    res.status(200).send(updatedData);
+    await editCurrentProjectData(newData);
```
cpvalente: question: was there a reason not to send the new data in the reply? we could apply it directly to the cache

packages/types/src/api/refetch.type.ts
```diff
@@ -0,0 +1,6 @@
+export enum RefetchKey {
+  ALL = 'ALL',
+  PROJECT_DATA = 'projectData',
+  REPORT = 'report',
+  RUNDOWN = 'rundown',
+}
```
cpvalente:
```suggestion
export enum RefetchKey {
  All = 'ALL',
  ProjectData = 'projectData',
  Report = 'report',
  Rundown = 'rundown',
}
```

packages/types/src/api/webSocket.type.ts
```diff
@@ -0,0 +1,62 @@
+import type { Client } from '../definitions/Clients.type.js';
+import type { Log } from '../definitions/runtime/Logger.type.js';
+import type { RuntimeStore } from '../definitions/runtime/RuntimeStore.type.js';
+import type { RefetchKey } from './refetch.type.js';
+
+export enum WsType {
```
cpvalente: I think `WsType` doesnt to a particularly good job at giving expectations on what it is. How about `MessageType`?

Also: Enum entries are typically PascalCased in Typescript and in the rest of the project, it would be good to keep it consistent

packages/types/src/api/webSocket.type.ts
```diff
@@ -0,0 +1,62 @@
...
+  CLIENT_LIST = 'client-list',
+  DIALOG = 'dialog',
+  ONTIME_LOG = 'ontime-log',
+  ONTIME = 'ontime',
+  ONTIME_PATCH = 'ontime-patch',
+  ONTIME_REFETCH = 'ontime-refetch',
+}
+
+//TO SERVER
+type pingPacket = { type: WsType.PING; payload: Date };
```
cpvalente: types are usually `PascalCased` it would be good to keep this consistent to help with reading in future

cpvalente (commented):
Good work!
I left some small questions and suggestions

Generally we just need to cleanup the naming conventions to make it easier for future us™

apps/client/src/common/utils/socket.ts
alex-Arc: just that is how we know that we are missing something
the new typing's should help, and if we can get the eslint rule for excusive switch to work we can be sure

apps/server/src/adapters/websocketAux.ts
alex-Arc: hmm
it is extra data not options one the function
maybe `extraData` ?

apps/server/src/adapters/WebsocketAdapter.ts
alex-Arc: it is partial data

apps/server/src/adapters/websocketAux.ts
cpvalente: Sounds good

apps/client/src/common/utils/socket.ts
cpvalente: I can help. We can make a typescript rule for it. 
if you search for exhaustive check there should be several options

alex-Arc: yes found that but it seams to require [typed-linting](https://typescript-eslint.io/getting-started/typed-linting/)
and I could not get that to work
if you can that would be great!

apps/client/src/common/utils/socket.ts
alex-Arc: it takes more TypeScript magic than I have to make them coexist 
do you have an idea

apps/client/src/common/api/constants.ts
```diff
@@ -23,3 +25,10 @@ const cssOverridePath = 'styles/override.css';
 
 export const overrideStylesURL = `${serverURL}/${userAssetsPath}/${cssOverridePath}`;
 export const projectLogoPath = `${serverURL}/${userAssetsPath}/logo`;
+
+export const axiosInstance = axios.create({
+  baseURL: apiEntryUrl,
+  validateStatus: (status) => {
+    return status < 300 || status === 304;
```
cpvalente: question: I wonder if this will be problematic for all the 1xx info requests, maybe the logic should be 

```suggestion
    return (status >= 200 && status < 300) || status === 304;
```

apps/client/src/common/utils/socket.ts
cpvalente: I think we already have it (or used to) in some places in the code base
Would something like this not work? 
https://graphite.dev/guides/typescript-switches#using-the-object-object-type-for-exhaustive-checks

apps/server/src/api-data/view-settings/viewSettings.controller.ts
```diff
@@ -21,7 +30,12 @@ export async function postViewSettings(req: Request, res: Response<ViewSettings
       warningColor: req.body.warningColor,
     } as ViewSettings;
     await getDataProvider().setViewSettings(newData);
+    revision++;
```
cpvalente: question: I am not sure that the risk introduced by the pattern is worth the gains here (for cases other than the rundown). In general, etags should work out of the box in express and we likely just need to implement it in the client. It also means less logic for us to write

What do you think?

apps/server/src/api-data/view-settings/viewSettings.controller.ts
```diff
@@ -21,7 +30,12 @@ export async function postViewSettings(req: Request, res: Response<ViewSettings
...
+    res.setHeader('etag', revision);
     res.status(200).send(newData);
+    setImmediate(() => {
+      sendRefetch(RefetchKey.ViewSettings, revision);
+    });
```
cpvalente: 👍

apps/server/src/app.ts
```diff
@@ -72,10 +72,12 @@ if (!isProduction) {
   // log server timings to requests
   app.use(serverTiming());
 }
-app.disable('x-powered-by');
+app.set('x-powered-by', false);
+app.set('etag', false);
```
cpvalente: question: maybe we should have a chat, I am not sure why we are disabling etags to write custom logic

apps/client/src/common/hooks-query/useViewSettings.ts
```diff
@@ -1,20 +1,62 @@
...
+import { ontimeQueryClient } from '../../common/queryClient';
+import { apiEntryUrl, axiosInstance, VIEW_SETTINGS } from '../api/constants';
 import { viewsSettingsPlaceholder } from '../models/ViewSettings.type';
 
+let etag: MaybeNumber = -1;
+const viewSettingsPath = apiEntryUrl + '/view-settings';
+
 export default function useViewSettings() {
-  const { data, status, isFetching, isError, refetch } = useQuery({
+  const { data, isFetching } = useQuery({
```
cpvalente: I think this is fine, but I would prefer having the previous separation between react-query and the axios request

apps/client/src/common/utils/socket.ts
alex-Arc: hmm, I now get `error TS6133: '_exhaustiveCheck' is declared but its value is never read.`
if I do `//ts-expect-error` the check stops working
any ideas?

apps/Server/src/api-integration/integration.controller.ts
```diff
@@ -0,0 +1,328 @@
+import {
```
cpvalente: I think we are having issues here because of the file capitalisation (see `Server` instead of `server`).
We need to delete this file and move the relevant changes to the existing file

apps/client/src/common/queryClient.ts
```diff
@@ -1,11 +1,17 @@
...
+import { MILLIS_PER_MINUTE } from 'ontime-utils';
 
 import { isOntimeCloud } from '../externals';
 
 export const ontimeQueryClient = new QueryClient({
   defaultOptions: {
     queries: {
-      gcTime: 1000 * 60 * 10, // 10 min
+      gcTime: 10 * MILLIS_PER_MINUTE,
+      // staleTime: MILLIS_PER_HOUR, //TODO: when all routes have implemented refetch signal from server
```
cpvalente: question: what is the idea behind this todo?
should we simply remove the code?

apps/client/src/common/utils/socket.ts
```diff
@@ -65,107 +72,106 @@ export const connectSocket = () => {
     console.error('WebSocket error:', error);
   };
 
-  websocket.onmessage = (event) => {
+  websocket.onmessage = async (event) => {
```
cpvalente, line 75: question: what is the reason to make this an async function?

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
```diff
@@ -45,19 +43,13 @@ export default function ViewSettingsForm() {
     }
   }, [data, reset]);
 
-  const onSubmit = async (formData: ViewSettings) => {
-    const newData = {
-      ...formData,
-    };
+  useEffect(() => {
+    return setError('root', { message: mutateError });
+  }, [mutateError]);
```
cpvalente: question: this is not a very common pattern for useEffect or react-hook-form, what is the idea here?

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
```diff
@@ -45,19 +43,13 @@ export default function ViewSettingsForm() {
...
-      await postViewSettings(newData);
-    } catch (error) {
-      const message = maybeAxiosError(error);
-      setError('root', { message });
-    } finally {
-      await refetch();
-    }
+  const onSubmit = async (formData: ViewSettings) => {
+    const newData = { ...formData };
+    await mutate(newData);
```
cpvalente: suggestion: functions that mutate data should ensure immutability instead of the consumer creating copies pre-emptively. in short: we dont make a copy of the object here, and the mutate function does it if it needs to change.

either way, I dont think the `mutate` function needs or should be changing the input data

apps/client/src/features/app-settings/panel/general-panel/ViewSettingsForm.tsx
```diff
@@ -68,7 +60,7 @@ export default function ViewSettingsForm() {
     return null;
   }
 
-  const isLoading = status === 'pending' || infoStatus === 'pending';
+  const isLoading = isFetching || infoStatus === 'pending';
```
cpvalente: `isFetching` will also be true if the app is fetching in background, which means that the user would be interrupted in their flow. That is not a pattern we would want in our usage of react-query

apps/server/src/middleware/etag.ts
```diff
@@ -0,0 +1,14 @@
+import type { NextFunction, Request, Response } from 'express';
+
+/**
+ * send 304 if the etag matches
+ */
+export function ifNoneMatch(req: Request, res: Response, next: NextFunction, currentRevision: number) {
```
cpvalente: suggestion: can we rename this something that helps us understand what is happening? maybe `checkCachingHeaders` or `checkETag`?

apps/client/src/common/utils/socket.ts
alex-Arc: not anymore

apps/client/src/common/api/viewSettings.ts
```diff
@@ -1,21 +1,18 @@
...
   const res = await axios.get(viewSettingsPath);
   return res.data;
 }
 
-/**
- * HTTP request to mutate view settings
- */
 export async function postViewSettings(data: ViewSettings) {
-  return axios.post(viewSettingsPath, data);
+  await ontimeQueryClient.cancelQueries({ queryKey: VIEW_SETTINGS });
```
cpvalente: cleanup: this shouldnt be here. we dont want to mix up axios and react-query
this logic can move to the onMutate callback of the useMutation, like

```ts
useMutation({
    mutationFn: postViewSettings,
    onMutate: async () => {
        await ontimeQueryClient.cancelQueries({ queryKey: VIEW_SETTINGS });
    },
    onSuccess: (data) => {
      ontimeQueryClient.setQueryData(VIEW_SETTINGS, data);
    },
})
```

cpvalente (approved):
Great work, just need to make a small change and we are good to merge

Issue #1310, opened 2024-10-30
Blocks from .xlsx imported as Events


forchinmaser:
Created blocks in the app and exported as .csv. Uploaded to google sheets for further editing. Downloaded as .xlsx from sheets, when I import this spreadsheet, the blocks appear as events with start, end, and duration times all 00:00:00.

<img width="535" alt="Screenshot 2024-10-29 at 11 27 13 PM" src="https://github.com/user-attachments/assets/a98aa198-70cc-4572-8bcb-f5f73e7c06e6">

<img width="532" alt="Screenshot 2024-10-29 at 11 26 33 PM" src="https://github.com/user-attachments/assets/46a9f99e-c9a9-4590-8d9b-4f64ec1da7b7">

cpvalente:
Hi @forchinmaser , thank you for taking the time to open a ticket
It makes sense to me that either a) the CSV export would indicate this is a block or b) the UI allows special rules for importing when there are no time data in the row (ie: make block or skip import) 

For now, you can work around this by creating a timer type column and setting it to "block"
You can see examples of this, both in the documentation and in the excel template here https://docs.getontime.no/features/import-spreadsheet

cpvalente:
The new flow has been reworked as part of v4 and you can import and export blocks
Closing this for now

PR #1623, opened 2025-06-01
refactor: block event xlsx import

Shobhit-Nagpal:
addresses #1310

## Changes made:
- exported type under Type column on frontend
- added a condition to check if `maybeTimeType` is equal to 'event' (without this, the flow falls into the else block where none of the events get imported, only the blocks get imported)

## Screenshots

- CSV export
image

- Excel
image

- Preview during importing excel

image

image


cpvalente (approved):
Good work

PR #1626, opened 2025-06-03
upgrade vite to v6

jwetzell:


jwetzell:
@cpvalente I think I've got it working under v6 without needing to use the `legacy` option.

jwetzell:
Well it builds....

Still tracking down whatever import mess is screwing up the testing

jwetzell:
AH! circular dependency between `apps/server/src/externals.ts` and `apps/server/src/setup/index.ts`

apps/client/vite.config.js
```diff
@@ -70,9 +72,9 @@ export default defineConfig({
     preprocessorOptions: {
       scss: {
         additionalData: `
-        @use './src/theme/ontimeColours' as *;
-        @use './src/theme/ontimeStyles' as *;
-        @use './src/theme/mixins' as *;
+        @use "${join(currentDir, './src/theme/ontimeColours')}" as *;
```
alex-Arc: ccs import is broken for me
this allows us to continue with the old config until we can migrate the modern scss

```ts
      scss: {
        api: 'legacy',
        additionalData: `
        @use './src/theme/ontimeColours' as *;
        @use './src/theme/ontimeStyles' as *;
        @use './src/theme/mixins' as *;
        `,
      },
```

alex-Arc: I think it could work like this and leverage the alias resolver

```ts
  css: {
    preprocessorOptions: {
      scss: {
        additionalData: `
        @use '@/theme/ontimeColours' as *;
        @use '@/theme/ontimeStyles' as *;
        @use '@/theme/mixins' as *;
        `,
      },
    },
  },
```

jwetzell: What error are you getting, I don't have any issue building what is currently in the branch. I get the same result as using that second option with the alias trick.

I've switched to using the alias as it looks cleaner and everything still builds and looks fine!

cpvalente (approved):
Good stuff, thank you!

PR #1628, opened 2025-06-06
v4 transactions

cpvalente:
This finishes the migration to transactions

- [x] Migrate the custom fields to transactions
- [x] small cleanups on how we make custom fields
- [x] extract `rundownCache` utilities
- [x] remove `rundownService` and `rundownCache` and replace with new utilities

... and
- [x] migrate all parsers and utilities to their place in the api-data directories

One potentially contentious change is that now the parsing of customfields is strict

since we allow keys to have casing, all the parsing functions account to the casing so the ontime keys must strictly match

After this PR the whole flow with the rundown, custom fields and playback should be working as before


cpvalente:
@coderabbitai full review

apps/server/src/api-data/custom-fields/customFields.router.ts
alex-Arc, line 1: Not sure I see why they this have been moved to the rundown folder?

apps/server/src/services/runtime-service/rundownService.utils.ts
```diff
@@ -33,3 +33,97 @@ export function getForceUpdate(previousUpdate: number, now: number): boolean {
...
+    return;
+  }
+
+  // if there is no event running, go to first
+  if (!currentEventId) {
+    return getFirstPlayableId(playableEventsOrder);
+  }
+
+  const currentIndex = playableEventsOrder.findIndex((eventId) => eventId === currentEventId);
+  if (currentIndex > playableEventsOrder.length - 1) {
```
alex-Arc: I agree with the bot here
or do we just go with the assumption the there will never be a index miss

apps/server/src/api-data/sheets/sheets.validation.ts
```diff
@@ -25,7 +24,7 @@ export const validateRequestConnection = [
 ];
 
 export const validateSheetId = [
-  param('sheetId').exists().isString(),
+  param('sheetId').isString(),
```
alex-Arc, line 27: is `exists()` not still required here (and other places?)

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -9,21 +11,21 @@ import {
...
+import { createTransaction, customFieldMutation, rundownCache, rundownMutation } from './rundown.dao.js';
+import type { RundownMetadata } from './rundown.types.js';
+import { generateEvent, getInsertAfterId, hasChanges } from './rundown.utils.js';
 
 /**
  * creates a new entry with given data
  */
 export async function addEntry(eventData: EventPostPayload): Promise<OntimeEntry> {
-  const { rundown, commit } = createTransaction();
+  const { rundown, commit } = createTransaction(true, false);
```
alex-Arc: would it maybe be easier to understand if it was just two functions
`createRundownTransaction`  and `createCustomFieldTransaction` 
that just wraps this functionality

packages/utils/src/customField-utils/customFieldUtils.ts
```diff
@@ -0,0 +1,19 @@
+import type { CustomFields } from 'ontime-types';
+
+/**
+ * Transforms an alphanumeric label with spaces into a valid key
+ */
+export function customFieldLabelToKey(label: string): string {
+  return label.trim().replaceAll(' ', '_');
+}
```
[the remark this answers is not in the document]
alex-Arc: also agree with the bot here

apps/server/src/api-data/sheets/sheets.validation.ts
cpvalente: no, as far as I can understand, the other validators like isString will look make sure the parameter is there

apps/server/src/api-data/rundown/rundown.service.ts
cpvalente: i fear two functions could give us a cache issue, but we could pass an object like

```
type TransactionOptions = {
  mutatesRundown: boolean;
  mutatesCustomFields: boolean;
}
```

packages/utils/src/customField-utils/customFieldUtils.ts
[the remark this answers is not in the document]
cpvalente: Hey @alex-Arc , I am not too sure

One of the issues we fight in the code is that we do the same work in many places
This makes our code slower and harder to maintain

In this case, the function assumes that the input is correct, it falls on the parent to validate the input

since an empty string is not a valid customfield label and the function declares `customFieldLabelToKey` I think this is OK
We could make a bunch of maybeLabelToKey that wraps both behaviour, but I didnt see enough usage to be worth it

PR #1633, opened 2025-06-09
Upgrade expressjs

alex-Arc:


apps/server/src/api-data/index.ts
```diff
@@ -31,6 +31,6 @@ appRouter.use('/report', reportRouter);
 appRouter.use('/assets', assetsRouter);
 
 //we don't want to redirect to react index when using api routes
-appRouter.all('/*', (_req, res) => {
+appRouter.all('/*splat', (_req, res) => {
```
cpvalente, line 34: question, is it not `'/{*splat}'`?
https://expressjs.com/en/guide/migrating-5.html

apps/server/src/app.ts
```diff
@@ -76,7 +76,7 @@ app.disable('x-powered-by');
 
 // Implement middleware
 app.use(cors()); // setup cors for all routes
-app.options('*', cors()); // enable pre-flight cors
+app.options('*splat', cors()); // enable pre-flight cors
```
cpvalente, line 79: same here. is it not `{*splat}`?

apps/server/src/api-data/index.ts
alex-Arc: yeah the documentation is a bit hard to understand, but I don't think we need it
at least in my tests it works without

PR #1634, opened 2025-06-09
refactor: cuesheet time format

Shobhit-Nagpal:
## Changes made:
- added `timeFormat` as a type under Tanstack table's declaration
- get `timeFormat` with `useSettings` from `<CuesheetTable>`
- add `timeFormat` as an optional prop under `<TimeInput>` that is under cuesheet-elements
- add `timeFormat` as an option under `FormatOptions` and change the hour formatting based on that inside `millisToString` util

Issue: There's an issue currently with the Duration column in the table. If we toggle the focus between `timeEnd` and `duration` column, the formatting of the duration seems to add an additional '12h'

For example: if timeStart is 1:05:00 and timeEnd is 1:25:00, and we toggle focus, the duration changes from '20m' to '12h20m'

### DIFF
--- a/apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/TimeInput.tsx
+++ b/apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/TimeInput.tsx
@@ -3,11 +3,13 @@ import { millisToString, parseUserTime } from 'ontime-utils';
 
 import SingleLineCell from './SingleLineCell';
 import TextLikeInput from './TextLikeInput';
+import { TimeFormat } from 'ontime-types';
 
 interface TimeInputDurationProps {
   initialValue: number;
   lockedValue: boolean;
   delayed?: boolean;
+  timeFormat?: TimeFormat;
   onSubmit: (value: string) => void;
 }
 
@@ -18,7 +20,7 @@ interface ParentFocusableInput extends HTMLInputElement {
 export default memo(TimeInputDuration);
 
 function TimeInputDuration(props: PropsWithChildren<TimeInputDurationProps>) {
-  const { initialValue, lockedValue, delayed, onSubmit, children } = props;
+  const { initialValue, lockedValue, delayed, timeFormat, onSubmit, children } = props;
 
   const [isEditing, setIsEditing] = useState(false);
   const [value, setValue] = useState(initialValue);
@@ -84,7 +86,7 @@ function TimeInputDuration(props: PropsWithChildren<TimeInputDurationProps>) {
     [initialValue, lockedValue, onSubmit],
   );
 
-  const timeString = millisToString(value);
+  const timeString = millisToString(value, { timeFormat });
 
   return isEditing ? (
     <SingleLineCell
--- a/apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
+++ b/apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
@@ -17,7 +17,7 @@ function MakeStart({ getValue, row, table }: CellContext<OntimeEntry, unknown>) 
   }
 
   const { handleUpdateTimer } = table.options.meta;
-  const { showDelayedTimes, hideTableSeconds } = table.options.meta.options;
+  const { showDelayedTimes, hideTableSeconds, timeFormat } = table.options.meta.options;
 
   const update = (newValue: string) => handleUpdateTimer(row.original.id, 'timeStart', newValue);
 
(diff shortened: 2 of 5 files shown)


Shobhit-Nagpal:
Hey, made a couple changes, sorted out the noted issue as well.

Changes:
- switched `millisToString` to `formatTime` under `cuesheetCols.tsx` for timeStart and timeEnd
- handled case for `hideTableSeconds` if time format is 12h
- changed the `timeFormat` prop to a `isFormat` boolean (this is to differentiate formatting of duration vs timeEnd and timeStart while editing the field)

Attaching a video to show the working:

https://github.com/user-attachments/assets/525de903-4c27-48c6-ae0d-3f0bc1255c4c

Do let me know if there are any changes required

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/TimeInput.tsx
```diff
@@ -18,7 +21,7 @@ interface ParentFocusableInput extends HTMLInputElement {
 export default memo(TimeInputDuration);
 
 function TimeInputDuration(props: PropsWithChildren<TimeInputDurationProps>) {
```
cpvalente: question: we have explicitly separated the scheduled times (09:00:00) from the durations (20m) I dont think we need the `isFormat` prop here. Maybe I am missing something?

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
```diff
@@ -26,13 +26,18 @@ function MakeStart({ getValue, row, table }: CellContext<OntimeEntry, unknown>)
...
-  let formattedTime = millisToString(displayTime);
+  let formattedTime = formatTime(displayTime)
   if (hideTableSeconds) {
-    formattedTime = removeSeconds(formattedTime);
+    if (timeFormat === '12') {
+      const [parsingTime, amPmIndicator] = formattedTime.split(' ');
+      formattedTime = removeSeconds(parsingTime) + ' ' + amPmIndicator
+    } else {
+      formattedTime = removeSeconds(formattedTime);
+    }
```
cpvalente: questions:
- can we extract the format into an utility function which scheduled time elements (like End and Start) can use?
- does formatTime not already resolve the issue with AM/PM formatting? could we remove the extra pass by passing formatting options?

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/TimeInput.tsx
Shobhit-Nagpal: the `isFormat` propped is specifically for when the duration is being edited.

we use `isFormat` to format the `timeString` with `formatTime` if true, else use `millisToString`.

the reason for this is that if we only use `formatTime`, then it'll also display the AM/PM indicator for duration, so the prop is a way to differentiate duration from timeEnd and timeStart during editing

perhaps it might be better if i call the prop as `isDuration` and then put the logic accordingly? or is there a better way that comes to your mind?

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
Shobhit-Nagpal: - we sure can
- in this block of code, the check for timeFormat is to handle the removal of seconds. the formatTime already handles AM/PM formatting by getting the format from settings.

the issue was that the `removeSeconds` function only removes seconds from an input that would look like `00:00:00`, so passing in a 24h format time input would remove the AM/PM indicator instead of the seconds as the input assumes 12h format. i didn't want to change the `removeSeconds` function because it came under `packages` and wasn't part of the `client` side code, so i took this approach

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/TimeInput.tsx
cpvalente: I see, I would suggest creating a separate `<DurationInput>` so we can make more assumptions about what data is in there

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
cpvalente: I see, could we pass the formats to the formatTime function?

ie: if user says remove seconds, we pass `hh:mm` a otherwise `hh:mm:ss a`?

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/TimeInput.tsx
Shobhit-Nagpal: wouldn't that be repeating most of the code already present with `<TimeInput>`?

it makes sense to be more specific with the component but that would be potentially duplicating code i believe

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/cuesheetCols.tsx
Shobhit-Nagpal: oh yeah, this makes more sense. i'll change it

apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/TimeInput.tsx
cpvalente: Yes, that is ok. Code duplication is not an issue. Duplication of decisions is

Shobhit-Nagpal:
## Changes made on recent push:

- added a `<DurationInput>` which is being used for duration field under `MakeDuration`
- removed `isFormat` prop from `<TimeInput>` and the component is now being used for timeEnd and timeStart
- refactored logic of showing seconds by passing format options to `formatTime`
- removed passing of `timeFormat` as option under `<CuesheetTable` as we get the time format from settings in `formatTime`, originally `timeFormat` was being passed to handle logic of viewing/hiding seconds which was resolved from the above point, hence removed this
- removed `timeFormat` under the declaration file

PR #1636, opened 2025-06-09
Refactor: require trigger in all events objects

alex-Arc:


apps/client/src/common/utils/clone.ts
```diff
@@ -30,6 +30,7 @@ export const cloneEvent = (event: OntimeEvent): ClonedEvent => {
     gap: 0,
     timeWarning: event.timeWarning,
     timeDanger: event.timeDanger,
+    triggers: [...event.triggers],
```
cpvalente: question: should this be a deep clone since triggers is an array of objects? probably the same for the custom fields

alex-Arc: good point

PR #1640, opened 2025-06-11
Fix: rearrange playing event

alex-Arc:
fixes 2 issues
- when rearranging the playing event, the play head would stay on the previous index 
- edits to events that invalidated the rundown cache did not necessarily cause the timer to readjust


apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -84,6 +84,7 @@ export async function editEntry(patch: PatchWithId): Promise<OntimeEntry> {
   }
 
   // we dont allow the user to change the cue to empty string
+  // FIXME this is not guarding as expected
```
cpvalente: how come? can I make a PR to fix this?

alex-Arc: in the if statement on the line below it says
`isOntimeEvent(patch) && patch.cue === ''`
but the patch doesn't contain the event type so the check will always fail and therefor not prevent the edit

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -84,6 +84,7 @@ export async function editEntry(patch: PatchWithId): Promise<OntimeEntry> {
...
   if (isOntimeEvent(patch) && patch?.cue === '') {
     throw new Error('Cue value invalid');
   }
```
cpvalente:
```suggestion
  if ((patch as Partial<OntimeEvent>)?.cue === '') {
    throw new Error('Cue value invalid');
  }
```

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -107,7 +108,7 @@ export async function editEntry(patch: PatchWithId): Promise<OntimeEntry> {
     updateRuntimeOnChange(rundownMetadata);
 
     // notify timer and external services of change
-    notifyChanges(rundownMetadata, revision, { timer: [entry.id], external: true });
+    notifyChanges(rundownMetadata, revision, { timer: didInvalidate ? true : [entry.id], external: true });
```
cpvalente, line 111: question: when we send true, we invalidate the entire rundown, is that necessary?

cpvalente (approved):
Functionality looks good, just left a few questions and suggestions

Lets address the FIXME now so we dont get this debt, but otherwise I am happy to merge

One thought: instead of trying to be granular about the notification to the runtime service, would it be faster and less bug prone if we always emitted a notification as long as we consider the rundown invalidated?

apps/server/src/api-data/rundown/rundown.service.ts
alex-Arc: is it not? if we get `didInvalidate` then the `rundownMutation.edit` had to change something not in our `RegenerateWhitelist` that in case off linked timers means a lot of stuff could have changed.
What changed would then only be known by the `commit` function, but that is currently not collecting that data

PR #1643, opened 2025-06-12
Colorised spreadsheet

alex-Arc:
#1608

exports colours to spreadsheets
and removes an independency on the color lib


apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/EventRow.tsx
```diff
@@ -75,7 +75,9 @@ function EventRow(props: EventRowProps) {
   }, [ownRef, selectedRef]);
 
   const { color, backgroundColor } = getAccessibleColour(event.colour);
-  const mutedText = Color(color).fade(0.4).hexa();
+  console.log(color, backgroundColor);
```
cpvalente: cleanup: remember to remove debug code

apps/server/src/services/sheet-service/SheetService.ts
```diff
@@ -205,22 +205,24 @@ async function verifySheet(
   if (!sheetId || !authClient) {
     throw new Error('Missing sheet ID or authentication');
   }
-
+
```
cpvalente: cleanup: looks like this may be an accidental change? can we revert?

apps/client/src/common/utils/__tests__/styleUtils.test.ts
```diff
@@ -18,19 +18,19 @@ describe('getAccessibleColour()', () => {
...
     const { backgroundColor, color } = getAccessibleColour(colour);
-    expect(backgroundColor).toBe('#00FF00FF');
+    expect(backgroundColor).toBe('#00ff00ff');
     expect(color).toBe('black');
   });
   it('handles transparens', () => {
     const colour = '#0F08';
     const { backgroundColor, color } = getAccessibleColour(colour);
-    expect(backgroundColor).toBe('#0C940CFF');
+    expect(backgroundColor).toBe('#0c940cff');
```
cpvalente, line 33: question: does the casing matter?

alex-Arc: no, only matters for the test
and the previous library forced everything to upper case, I dont se a need for that

PR #1645, opened 2025-06-13
Remove public option from events

alex-Arc:


apps/client/src/AppRouter.tsx
```diff
@@ -89,6 +89,7 @@ export default function AppRouter() {
           }
         />
         <Route
+          // TODO: should we redirect to backstage
```
cpvalente: No, it should be removed

apps/client/src/AppRouter.tsx
```diff
@@ -121,6 +122,7 @@ export default function AppRouter() {
           }
         />
         <Route
+          //  TODO: should this be renamed?
```
cpvalente: Unless you have a better name, I don’t think it is worth it renaming

.gitignore
```diff
@@ -44,6 +44,7 @@ ontime-data/
 
 # versioning file
 **/ONTIME_VERSION.js
+**/ONTIME_VERSION.ts
```
cpvalente: We probably want to leave these files JavaScript to keep it easy on the tooling

Otherwise I don't see a reason for it to be typescript. Am I missing something?

apps/client/src/common/hooks/useRuntimeStylesheet.ts
```diff
@@ -1,35 +1,40 @@
...
-
     if (document.getElementById(scriptTagId)) {
       setShouldRender(true);
       return;
     }
 
     setShouldRender(false);
     const styleSheet = document.createElement('style');
-    styleSheet.rel = 'stylesheet';
+    // styleSheet.rel = 'stylesheet';
```
cpvalente: Why is this removed?

apps/client/src/features/viewers/ViewWrapper.tsx
```diff
@@ -25,19 +24,15 @@ import { useViewOptionsStore } from '../../common/stores/viewOptions';
 
 type WithDataProps = {
```
cpvalente, line 25: Nice 😊

apps/server/src/stores/runtimeState.ts
```diff
@@ -281,34 +243,21 @@ export function loadNext(
...
     // reset the state to indicate there is no future event
     runtimeState.eventNext = null;
-    runtimeState.publicEventNext = null;
     return;
   }
 
   // temporarily reset this value to simplify loop logic
   runtimeState.eventNext = null;
 
+  //TODO: do we already have a it as a list of not skipped events
```
cpvalente, line 252: Yes. In the rundown metadata you have a playable event order

e2e/tests/features/203-delay-block.spec.ts
```diff
@@ -67,7 +67,7 @@ test('delays are show correctly', async ({ page }) => {
   await page.getByTestId('block__title').click();
   await page.getByTestId('block__title').fill('test');
   await page.getByTestId('block__title').press('Enter');
-  await expect(page.getByTestId('entry-1').locator('#block-status')).toHaveAttribute('data-ispublic', 'true');
+  await expect(page.getByTestId('entry-1').locator('#block-status')).toBeVisible();
```
cpvalente: This test is not doing much for us, can we asset another of the indicators?

e2e/tests/features/204-editor-crud.spec.ts
```diff
@@ -20,8 +20,8 @@ test('CRUD operations on the rundown', async ({ page }) => {
   await page.getByRole('button', { name: 'Event', exact: true }).nth(1).click();
   expect(await page.getByTestId('entry-3').getByTestId('time-input-timeStart').inputValue()).toContain('00:30:00');
 
-  // test quick add options - event is public
+  // test quick add options
   await page.getByRole('button', { name: 'Event', exact: true }).nth(1).click();
 
-  await expect(page.getByTestId('entry-4').locator('#block-status')).toHaveAttribute('data-ispublic', 'true');
+  await expect(page.getByTestId('entry-4').locator('#block-status')).toBeVisible();
```
cpvalente: Same here 

This test is not doing much for us, can we asset another of the indicators?

packages/types/src/definitions/core/ProjectData.type.ts
```diff
@@ -1,7 +1,7 @@
 export type ProjectData = {
   title: string;
   description: string;
-  publicUrl: string;
+  publicUrl: string; //TODO: should this also go?
```
cpvalente: Yes please, all the public view stuff goes

Users have the custom fields in the project to replace the feature

e2e/tests/features/204-editor-crud.spec.ts
alex-Arc: are we not just testing to see that the entry was created?
the test form before was also just indicating the same and that public was the default

apps/client/src/common/hooks/useRuntimeStylesheet.ts
alex-Arc: I don't believe the [HTML attribute: rel](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel) exists on style elements

.gitignore
alex-Arc: I keep getting this error
image

apps/client/src/common/hooks/useRuntimeStylesheet.ts
cpvalente: That’s right, this was likely lost at some point 
The rel attribute is for links

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/link

.gitignore
cpvalente: Is this new? This hasn’t changed for a while now. 
in short: it doesn’t matter since this is only injecting the version number in CI

e2e/tests/features/204-editor-crud.spec.ts
cpvalente: The assertion did several things for us
- was the event created
- Did it have the right public value?
- does our public indicator work?

Right now it just asserts that an indicators container is in the DOM
Maybe we can recover a check for the indicators?

apps/client/src/common/hooks/useRuntimeStylesheet.ts
alex-Arc: so i should change the `createElement('style');` to `createElement('link');` ?

.gitignore
alex-Arc: sure just annoying when when going through problems flagged by tsc this is always the first one to come up

apps/client/src/common/hooks/useRuntimeStylesheet.ts
cpvalente: No, sorry if my comment was confusing. This works so let’s just let it be

apps/server/src/stores/runtimeState.ts
alex-Arc: I think it is best I do this in a another PR

PR #1652, opened 2025-06-20
V3 project migration

alex-Arc:


apps/server/src/api-data/db/db.parser.ts
```diff
@@ -21,28 +22,28 @@ type ParsingError = {
...
-  // this may throw
-  const settings = parseSettings(jsonData);
-
   const errors: ParsingError[] = [];
   const makeEmitError = (context: string) => (message: string) => {
     logger.error(LogOrigin.Server, `Error parsing ${context}: ${message}`);
     errors.push({ context, message });
   };
 
+  const migratedData = migrate_v3_to_v4(jsonData, makeEmitError('Migration'));
```
cpvalente: question: does it make sense to parse the settings first? that way we would know if we need to migrate

also: I dont think that the result variable that all the other parsers use should be called `migratedData`, it is irrelevant that it has been migrated for the other contexts. Maybe something like `parsedProjectData`?

apps/server/src/api-data/db/migration/versionCheck.ts
```diff
@@ -0,0 +1,14 @@
+export function versionCheck(jsonData: object): { major: number; minor: number; patch: number; prerelease: string } | null {
```
cpvalente: question: could we use our assert functions to help? not just here but all the migration code

apps/server/src/api-data/db/migration/db.migration.v3.ts
```diff
@@ -0,0 +1,180 @@
...
+ * - public url and info is removed
+ *
+ * CUSTOM-FIELD
+ *  - keys must be derived from label
+ *
+ * EVENT
+ * - trigger is not optional
+ * - parent is not  optional
+ * - isPublic is removed
+ * - TODO: has the end action of events ben changed?
```
cpvalente: we no longer have the stop on end

apps/server/src/api-data/db/migration/db.migration.v3.ts
```diff
@@ -0,0 +1,180 @@
...
+import { ONTIME_VERSION } from '../../../ONTIME_VERSION.js';
+import { versionCheck } from './versionCheck.js';
+import { ErrorEmitter } from '../../../utils/parserUtils.js';
+
+/**
+ * Changes from v3 to v4
+ * PROJECT-DATA
+ * - must have logo
+ * - must have custom array
+ * - public url and info is removed
```
cpvalente: question: do we need to migrate the `publicUrl` and `publicInfo`? would the parser not simply remove it?

alex-Arc:
@coderabbitai review

apps/server/src/api-data/automation/clients/osc.client.ts
```diff
@@ -31,7 +31,7 @@ function preparePayload(output: OSCOutput, state: RuntimeState): OscPacketInput
 
 /** Emits message over transport */
 function emit(targetIP: string, targetPort: number, packet: OscPacketInput) {
-  logger.info(LogOrigin.Tx, `Sending OSC: ${targetIP}:${targetPort}`);
+  logger.info(LogOrigin.Tx, `Sending OSC: ${targetIP}:${targetPort}`); //TODO: this might become excessive
```
cpvalente: sounds good, lets remove these logs

apps/server/src/api-data/db/migration/db.migration.v3.ts
```diff
@@ -0,0 +1,410 @@
...
+
+/**
+ * migrates a settings from v3 to v4.0.0
+ * - update the version number
+ */
+export function migrateSettings(jsonData: object): Settings | undefined {
+  if (is.objectWithKeys(jsonData, ['settings']) && is.object(jsonData.settings)) {
+    const oldSettings = structuredClone(jsonData.settings) as old_Settings;
+
+    const migrated: Settings = { ...oldSettings, version: '4.0.0' };
```
cpvalente: question: can we use the version from the `ONTIME_VERSION` variable?

apps/server/src/api-data/db/migration/db.migration.v3.ts
```diff
@@ -0,0 +1,410 @@
...
+  {
+    type: 'string' | 'image';
+    colour: string;
+    label: string;
+  }
+>;
+
+/**
+ * migrates a custom fields from v3 to v4.0.0
+ * - ensure correct case (TODO: could this be removed from the project parser)
```
cpvalente: I believe so. the project parser should be able to assume that is parsing a v4 object

apps/server/src/api-data/db/migration/db.migration.v3.ts
```diff
@@ -0,0 +1,410 @@
...
+ *
+ * - events:
+ *  - add flag
+ *  - ensure end action is not stop
+ *  - move the timer type count-to-end to it owen setting
+ *  - ensure triggers
+ *  - add parent
+ *
+ * - block:
+ *  - add all the new blocks of the block that is now a group
```
cpvalente, line 319: question: when we see a block, could we add all the elements below inside the block until we find a new block?

apps/server/src/api-data/db/migration/db.migration.v3.ts
```diff
@@ -0,0 +1,410 @@
...
+export function migrateSettings(jsonData: object): Settings | undefined {
+  if (is.objectWithKeys(jsonData, ['settings']) && is.object(jsonData.settings)) {
+    const oldSettings = structuredClone(jsonData.settings) as old_Settings;
+
+    const migrated: Settings = { ...oldSettings, version: '4.0.0' };
+    return migrated;
+  }
+}
+
+type old_ViewSettings = {
```
cpvalente, line 75: question: could we avoid exporting these types?

apps/server/src/api-data/db/db.parser.ts
```diff
@@ -20,7 +21,21 @@ type ParsingError = {
...
  * @returns {object} - parsed object
  */
-export function parseDatabaseModel(jsonData: Partial<DatabaseModel>): { data: DatabaseModel; errors: ParsingError[] } {
+export function parseDatabaseModel(jsonData: Partial<DatabaseModel>): {
+  data: DatabaseModel;
+  errors: ParsingError[];
+  migrated: boolean;
+} {
+
+  //TODO: TEST THIS!!!!!!!
```
cpvalente, line 29: :D

apps/server/src/api-data/db/db.parser.ts
```diff
@@ -20,7 +21,21 @@ type ParsingError = {
...
+  let migrated = false;
+  if (v3.shouldUseThisMigration(jsonData)) {
+    migrated = true;
+    logger.warning(LogOrigin.Server, 'The imported project is from v3, trying to migrate');
+    //TODO: is this bad?
+    jsonData = v3.migrateAllData(jsonData);
```
cpvalente: I would personally prefer it if we created a copy of the migrated data
there is also a fair share of parsing going on, so we may want to have this inside a try-catch

apps/server/src/api-data/db/migration/db.migration.v3.ts
alex-Arc: they are not exported..?

apps/server/src/api-data/db/migration/db.migration.v3.ts
alex-Arc: my idea here is that this migrator can only guarantee v4.0.0
if there is any changes to the file layout after this we will need a new migrator taking it from 4.0.0 to 4.x.x

apps/server/src/api-data/db/migration/db.migration.v3.ts
alex-Arc: yes, are we sure this is the user intent? 
I'm 50/50

cpvalente: I don’t think there is a clean path that will 100% work. So the decision is whether it is more convenient than having to populate all groups (thinking we don’t have drag and drop for multiple elements)

apps/server/src/api-data/db/migration/db.migration.v3.ts
cpvalente: Hm, ideally we shouldn’t break things after the alpha

if we write the version we get a better track of the migration that was done

apps/server/src/api-data/db/migration/db.migration.v3.ts
cpvalente: 😅

apps/server/src/api-data/db/migration/db.migration.v3.ts
alex-Arc: while we might not break stuff, we will at some point add things
it would be good to have a way to do that, where we avoid having it as part of the loading logic as it is now

cpvalente (approved):
🚀

PR #1653, opened 2025-06-20
refactor: use runtime metadate in `runtimeState`

alex-Arc:


apps/server/src/stores/runtimeState.ts
```diff
@@ -237,35 +238,28 @@ export function loadNow(timedEvents: OntimeEvent[], eventIndex: MaybeNumber = ru
...
-  //TODO: do we already have a it as a list of not skipped events
-  for (let i = eventIndex + 1; i < timedEvents.length; i++) {
-    const event = timedEvents[i];
-    // we dont deal with events that are not playable
-    if (!isPlayableEvent(event)) {
-      continue;
-    }
-    runtimeState.eventNext = event;
+  const nowPlayableIndex = timedToPlayableIndex(metadata, eventIndex);
+  if (!nowPlayableIndex || nowPlayableIndex > metadata.playableEventOrder.length - 2) {
```
cpvalente, line 252: question: can you add a comment explaining this conditional? Maybe it warrants a comment

apps/server/src/stores/runtimeState.ts
```diff
@@ -325,12 +319,11 @@ export function updateLoaded(event?: PlayableEvent): string | undefined {
 /**
  * Used in situations when we want to hot-reload all events without interrupting timer
  */
-export function updateAll(rundown: Rundown, timedEventsOrder: EntryId[]) {
-  const timedEvents = filterTimedEvents(rundown, timedEventsOrder);
+export function updateAll(rundown: Rundown, metadata: RundownMetadata) {
   // TODO(remove public): we dont need to make the timedEvents object, we pass primitives and let the functions handle it
```
cpvalente: question: can we simplify the function as described in the comment?
otherwise, should we remove the comment?

apps/server/src/stores/runtimeState.ts
```diff
@@ -613,11 +605,13 @@ export function roll(
   runtimeState.runtime.offset = offset;
   const offsetClock = runtimeState.clock + runtimeState.runtime.offset;
 
+  //TODO: improve this, for now just made to work
+  const timedEvents = metadata.timedEventOrder.map((id) => rundown.entries[id]) as OntimeEvent[];
```
cpvalente: question: what improvements do you suggest?
if we push this forward it is likely that we forget where is the room for improvement so we are just left with an ominous comment...

apps/server/src/stores/runtimeState.ts
```diff
@@ -706,3 +700,9 @@ export function loadBlock(rundown: Rundown, state = runtimeState) {
 export function setOffsetMode(mode: OffsetMode) {
   runtimeState.runtime.offsetMode = mode;
 }
+
+function timedToPlayableIndex(metadata: RundownMetadata, index: number) {
```
cpvalente: suggestion: can we add some JSDocs explaining what this does?
the function name can be confusing and is nice not having to read the implementation

cpvalente (approved):
good stuff

PR #1661, opened 2025-06-24
adding block now/next data

alex-Arc:


apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -433,7 +433,12 @@ export default function Rundown({ data }: RundownProps) {
                * ie: we are inside a block, but there is no defined colour
                * we default to $gray-1050 #303030
                */
-              const blockColour = rundownMetadata.groupColour === '' ? '#303030' : rundownMetadata.groupColour;
+              //TODO: temp color to test is loaded
```
cpvalente: remember to remove this before merging

apps/client/src/features/rundown/block-block/BlockBlock.tsx
```diff
@@ -70,7 +71,10 @@ export default function BlockBlock(props: BlockBlockProps) {
     animateLayoutChanges: () => false,
   });
 
-  const binderColours = data.colour && getAccessibleColour(data.colour);
+  //TODO: temp color to test is loaded
```
cpvalente: reminder

apps/client/src/features/rundown/rundown.utils.ts
```diff
@@ -91,6 +96,7 @@ function processEntry(
         processedData.groupColour = undefined;
       }
     }
+    processedData.groupIsLoaded = processedData.groupId === selectedBlockId;
```
cpvalente: question: what is the use case for having to know if the group is loaded in an entry-by-entry basis?

apps/server/src/api-data/rundown/rundown.utils.ts
```diff
@@ -377,3 +377,13 @@ export function getTimedIndexFromPlayableIndex(metadata: RundownMetadata, index:
   const timedIndex = metadata.timedEventOrder.findIndex((id) => id === playableId);
   return timedIndex;
 }
+
+/**
+ * converts an index from the timedEventOrder to an index in the playableEventOrder
+ * all indexes in timedEventOrder must also exist in flatEntryOrder, otherwise the app is broken
```
cpvalente: should we throw an error here? at least in dev? otherwise this may fail silently and be hard to debug

apps/server/src/stores/__tests__/runtimeState.test.ts
```diff
@@ -466,17 +460,15 @@ describe('loadBlock', () => {
...
-        startedAt: 123,
-      },
+      blockNow: { id: rundown.entries[0].id, startedAt: 123 },
       eventNow: rundown.entries[2],
     } as RuntimeState;
 
     loadBlock(rundown, state);
-
+    //FIXME: not sure why this is not working, the startAt value is present in the state
+    console.log(state);
```
cpvalente: let the test fail and lets chat about it

apps/server/src/stores/runtimeState.ts
```diff
@@ -355,8 +357,8 @@ export function start(state: RuntimeState = runtimeState): boolean {
   }
 
   // update block start time
-  if (state.currentBlock.startedAt === null) {
-    state.currentBlock.startedAt = state.clock;
+  if (state.blockNow?.startedAt === null) {
```
cpvalente: `blockNow?.startedAt` could return `undefined` which would not pass the strict null check. maybe we should check for nullish values with the `==` 

```suggestion
  if (state.blockNow?.startedAt == null) {
```

apps/server/src/stores/runtimeState.ts
```diff
@@ -581,8 +583,8 @@ export function roll(
       runtimeState.timer.startedAt = runtimeState.clock;
 
       // update runtime
-      if (runtimeState.currentBlock.startedAt === null) {
-        runtimeState.currentBlock.startedAt = runtimeState.clock;
+      if (runtimeState.blockNow?.startedAt === null) {
```
cpvalente: same here, maybe

```suggestion
      if (runtimeState.blockNow?.startedAt == null) {
```

apps/server/src/stores/runtimeState.ts
```diff
@@ -645,8 +647,8 @@ export function roll(
   // there is something to run, load event
 
   // update runtime
-  if (runtimeState.currentBlock.startedAt === null) {
-    runtimeState.currentBlock.startedAt = runtimeState.clock;
+  if (runtimeState.blockNow?.startedAt === null) {
```
cpvalente: and here...
```suggestion
  if (runtimeState.blockNow?.startedAt == null) {
```

apps/server/src/stores/runtimeState.ts
```diff
@@ -677,26 +679,35 @@ export function roll(
...
+  // find next block
+  const blockOrEventIndex = rundown.order.findIndex((id) => id === currentBlockId || id === state.eventNow?.id);
+  if (blockOrEventIndex >= 0) {
+    for (let i = blockOrEventIndex + 1; i < rundown.order.length; i++) {
+      if (isOntimeBlock(rundown.entries[rundown.order[i]])) {
+        state.blockNext = { id: rundown.order[i], startedAt: null };
+
+        break;
+      }
+    }
```
cpvalente: question: would it be easier to do this in a single for loop? 
we could also benefit with some comments for the future, this was a bit tricky to understand out of context

Maybe something like
```suggestion

// we will look for the current block or current event, we could explain why
const targetIds = [currentBlockId, state.eventNow?.id];
let found = false;

for (let i = 0; i < rundown.order.length; i++) {
   // found our entry point, will look for the next block from here
  if (found && isOntimeBlock(rundown.entries[rundown.order[i]])) {
    state.blockNext = { id: rundown.order[i], startedAt: null }; // can we populate startedAt?
    break;
  }

  if (!found && targetIds.includes(rundown.order[i])) {
    found = true;
    continue;
  }
}
```

apps/server/src/stores/runtimeState.ts
```diff
@@ -677,26 +679,35 @@ export function roll(
...
-    state.currentBlock.block = null;
+    state.blockNow = null;
     return;
   }
 
-  const currentBlock = rundown.entries[currentBlockId];
-  state.currentBlock.block = currentBlock as OntimeBlock;
+  //we went into a new block
+  if ((state.blockNow != null && state.blockNow.id != currentBlockId) || state.blockNow == null) {
+    state.blockNow = { id: currentBlockId, startedAt: null };
```
cpvalente: question: why are we unable to set the `startedAt` at this point?

alex-Arc: that is responsibility of other parts of the runtimeState, here we just load the block other places can diside to start

cpvalente: Sounds good, can we make a comment here to say that?

apps/client/src/features/rundown/Rundown.tsx
alex-Arc: yes it is for testing, will leave it until it is done

apps/client/src/features/rundown/rundown.utils.ts
alex-Arc: not sure
depends on who we end up indicating that a block is loaded, and maybe that involves a background colour change for the whole span of the group

cpvalente: Good idea

PR #1667, opened 2025-06-27
Chore/remove got dependency

cpvalente:


cpvalente:
@alex-Arc this needs some cleanup and to run the pnpm install command, but what do you think?

alex-Arc:
> @alex-Arc this needs some cleanup and to run the pnpm install command, but what do you think?

Seams ok
The error handling is a bit excessive

apps/server/src/services/sheet-service/SheetService.ts
```diff
@@ -144,16 +143,29 @@ function verifyConnection(
...
+        headers: { 'Content-Type': 'application/json' },
+        body: JSON.stringify({
+          client_id: clientSecret.installed.client_id,
+          client_secret: clientSecret.installed.client_secret,
+          device_code,
+          grant_type: grantType,
+        }),
+      });
+
+      if (response.status === 428) {
```
cpvalente, line 158: what if: we throw as long as the response is !ok

then in the catch we can try to see the response status and send improve the error message

error message go to the logger, not the console

PR #1672, opened 2025-07-04
feat: block defines next day

cpvalente:
use the block to define next day
removes the logic where we said that overlapping events are the next day

the changes are quite simple, most of the code is adding tests


apps/server/src/api-data/rundown/rundown.utils.ts
```diff
@@ -350,30 +350,17 @@ export function cloneEntry<T extends OntimeEntry>(entry: T, newId: EntryId): T {
  * @param previous the previous event
  * @returns 0 or 1 for easy accumulation with the total days
  */
-export function calculateDayOffset(
-  current: Pick<OntimeEvent, 'timeStart'>,
-  previous: Pick<OntimeEvent, 'timeStart' | 'duration'> | null,
-) {
+export function calculateDayOffset(previous: Pick<OntimeEvent, 'timeStart' | 'duration'> | null) {
```
cpvalente, line 423: since we no longer compare events, the only thing that makes an event go the day after is if the previous crossed midnight

PR #1677, opened 2025-07-04
refactor: migrate to react router 7

cpvalente:


cpvalente:
Looks like this breaks our logic with the URL preset redirects

PR #1683, opened 2025-07-07
block end calculation

alex-Arc:


apps/server/src/stores/runtimeState.ts
```diff
@@ -509,6 +513,8 @@ export function update(): UpdateResult {
     runtimeState.timer.expectedFinish = getExpectedFinish(runtimeState);
   }
 
+  getEndOfBlock();
```
cpvalente: question: should we pass the state here to make sure we dont endup with weird state tearing issues?

apps/client/src/features/overview/composite/OverviewWrapper.tsx
```diff
@@ -75,21 +73,26 @@ export function CurrentBlockOverview() {
    * TODO(v4): this needs to be calculated according to offset mode
    */
   const projectedBlockEnd = (() => {
```
cpvalente: nice, can we remove this TODO?

apps/client/src/features/overview/composite/OverviewWrapper.tsx
```diff
@@ -51,22 +51,20 @@ export function TitlesOverview() {
...
   const { currentBlockId } = useCurrentBlockId();
   const entry = useEntry(currentBlockId);
 
-  const timeInBlock = formatedTime(blockStartAt ? clock - blockStartAt : null, 2);
+  const timeInBlock = formatedTime(blockStartedAt ? clock - blockStartedAt : null, 2);
 
   /**
    * The time to the end of the block
    * as scheduled
    * TODO(v4): this needs to be calculated according to offset mode
```
cpvalente: can we remove this TODO?

apps/server/src/stores/runtimeState.ts
cpvalente: question: should we pass the state here to make sure we dont endup with weird state tearing issues?

apps/server/src/stores/runtimeState.ts
```diff
@@ -707,10 +713,29 @@ export function loadBlock(rundown: Rundown, state = runtimeState) {
...
-    state.blockNow = { id: currentBlockId, startedAt: null }; // the id is set here, the start time is set in other placed that handel starting events
+    state.blockNow = { id: currentBlockId, startedAt: null, expectedEnd: null }; // the id is set here, the start time is set in other placed that handel starting events
   }
 }
 
 export function setOffsetMode(mode: OffsetMode) {
   runtimeState.runtime.offsetMode = mode;
 }
+
+export function getEndOfBlock(state = runtimeState) {
```
cpvalente: question: is there a reason for this calculation to happen in the server?

alex-Arc: not to happy about it, since it manually has to pull in the rundown
but that would be the samme for in the client it would also have to iterate over all remaining event in the block

cpvalente (approved):
good work

PR #1684, opened 2025-07-08
Fix ontime aux actions

alex-Arc:


cpvalente (approved):
looks good
we can make the select a bit easier to read if we implement separators. you can see how it is done in DropdownMenu if you want

apps/server/src/api-data/automation/clients/ontime.client.ts
```diff
@@ -15,7 +16,8 @@ export function toOntimeAction(action: OntimeAction) {
     case 'aux1-pause':
       return auxTimerService.pause(1);
     case 'aux1-set': {
-      return auxTimerService.setTime(action.time, 1);
+      const time = parseUserTime(action.time);
```
alex-Arc, line 19: @cpvalente 
not sure at what point we should make it a number
what do you thing?

PR #1688, opened 2025-07-11
feat: custom user translations

Shobhit-Nagpal:
## Changes made:
- created translations module on the backend to store user translations
- added a `Custom` option under the select for View languages in General Settings
- created a modal ui for the user to input their custom translations, english translations being the default fallback
- added additional logic on the general settings form's `onSubmit` as there was a behavior of the settings data being submitted along with the translations data when only clicked on saving the translations. this was due to nested forms, found a method to fix under [a RHF discussion](https://github.com/orgs/react-hook-form/discussions/7038#discussioncomment-11228568)
- transforming the translation object's keys as RHF was converting `common.expected_finish` to `common: { expected_finish }` when submitting the data
- refactored the `<TranslationProvider>` to include user translations
- added a `populateTranslation` function to create the translations files and folder on server startup

Attaching a video to show the working below:

https://github.com/user-attachments/assets/46e51294-9682-4923-ab1d-a03cca08cfde

Couple points to think about:
- Would like an opinion on the transformation of keys to prevent RHF to create objects from keys with dot notation
- Is there any better way of fetching data and providing it to the components than from the `<TranslationProvider>`? My initial thought was get the translations directly from the settings form but then thought it'd make more sense to access the functions through the translation context
- Have gone with a similar approach on the backend for translations as we did for custom CSS. The default translation is of English. When filling the form on the modal, no value can be empty either. Is this the flow we'd like to go for?
- Would also like some thoughts on the approach for `onSubmit` under General Settings now that there's nested forms into the picture. I did try using a React Portal but that did not seem to work. Curious if there's a better way to do this

If there's any change, do let me know


apps/server/src/api-data/index.ts
```diff
@@ -13,6 +13,7 @@ import { router as sessionRouter } from './session/session.router.js';
 import { router as viewSettingsRouter } from './view-settings/viewSettings.router.js';
 import { router as reportRouter } from './report/report.router.js';
 import { router as assetsRouter } from './assets/assets.router.js';
+import { router as translationsRouter } from './translations/translations.router.js';
```
cpvalente: lets move this into the `assetsRouter`, I dont think we can justify adding more entrypoints for this

apps/server/src/api-data/translations/translations.service.ts
```diff
@@ -0,0 +1,22 @@
+import { readFile, writeFile } from 'fs/promises';
+import { existsSync } from 'node:fs';
+import { TranslationObject } from './translations.type.js';
+import { publicFiles } from '../../setup/index.js';
+import { defaultTranslation } from '../../user/translations/bundledTranslation.js';
+
+export async function readUserTranslation(): Promise<TranslationObject> {
```
cpvalente: question: instead of reading the file and giving the user the object, could we instead just serve the json file?
This would let browsers cache the resource, we would just need to make sure the headers change when the translations are updated

What do you think?

apps/server/src/api-data/translations/translations.validation.ts
```diff
@@ -0,0 +1,4 @@
+import { body } from 'express-validator';
+import { requestValidationFunction } from '../validation-utils/validationFunction.js';
+
+export const validatePostTranslation = [body('translation').isObject(), requestValidationFunction];
```
cpvalente: question: could we dynamically or manually validate the object here?

apps/server/src/user/translations/README.md
```diff
@@ -0,0 +1,6 @@
+# translations
+
+The translations folder contain the user's custom translations which is used for custom translations in Ontime.
+The file should be named `translations.json`.
+
+https://docs.getontime.no/features/custom-translations/
```
cpvalente: :D

apps/server/src/user/translations/translations.json
```diff
@@ -0,0 +1,28 @@
...
+  "countdown.overtime": "in overtime",
+  "timeline.live": "live",
+  "timeline.done": "done",
+  "timeline.due": "due",
+  "timeline.followedby": "Followed by",
+  "project.title": "Title",
+  "project.description": "Description",
+  "project.backstage_info": "Backstage Info",
+  "project.backstage_url": "Backstage URL"
+}
```
cpvalente: question: instead of maintaining so many variants of translations (type object, object and json) could we have one and generate the rest?

if we maintain an object, we can generate types from the object shape. We can also make the translation json file at build time like we do with CSS

apps/client/src/features/app-settings/panel/settings-panel/GeneralSettings.tsx
```diff
@@ -39,8 +45,16 @@ export default function GeneralSettings() {
     }
   }, [data, reset]);
 
-  const onSubmit = async (formData: Settings) => {
+  const onSubmit = async (formData: Settings, event?: BaseSyntheticEvent) => {
     try {
+      // The target logic is to prevent the submission of the settings form from the translations form
```
cpvalente: nice, thank you!
I wonder if we can `preventPropagation` in the nested form, or simply not use a nested form?

We can also try to understand why the view settings form does not have this issue and restructure the code here

apps/client/src/features/app-settings/panel/settings-panel/composite/CustomTranslationModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+    defaultValues: translation,
+    resetOptions: {
+      keepDirtyValues: true,
+    },
+    mode: 'onChange',
+  });
+
+  const onSubmit = async (formData: Record<string, string>) => {
+    try {
+      if (isDirty) {
```
cpvalente: tip: instead of checking `isDirty` here, we can disable the submit button if the form is not dirty

apps/client/src/features/app-settings/panel/settings-panel/composite/CustomTranslationModal.tsx
```diff
@@ -0,0 +1,325 @@
...
+      bodyElements={
+        <Panel.Section as='form' onSubmit={handleSubmit(onSubmit)} id='custom-translations-form'>
+          <Panel.Card>
+            <Panel.SubHeader>Common</Panel.SubHeader>
+            <Panel.ListGroup>
+              <Panel.ListItem>
+                <Panel.Field title='Expected Finish' description='' />
+                <Input
+                  maxLength={150}
+                  {...register('common_expected_finish', {
```
cpvalente: question: since the form is not generated dynamically by the object, do we need to follow the conventions?

I think that either
a) the form is dynamic, and we need to deal with the keys (potentially renaming them)
b) the form is static, so we should make our lives easy here and convert the property names before submitting

apps/server/src/api-data/translations/translations.service.ts
Shobhit-Nagpal: i like the idea, making the change

apps/client/src/features/app-settings/panel/settings-panel/GeneralSettings.tsx
Shobhit-Nagpal: i did try using `stopPropagation` before using react portals and it still didn't work. i'll check the custom css code once to see what we did there

Shobhit-Nagpal: update on this: 

the issue is solved. i checked the css code again and turns out we're not using react hook form for the custom css. i tried the approach again with `stopPropagation` and portals as the react hook form discussion mentioned this approach but it seemed a bit complicated for something simple. ended up checking the code again and now the `<TranslationModal>` is being rendered outside the `<Panel.Section>` which is acting as the form so we don't have nested forms anymore

Shobhit-Nagpal:
hey, i've made the changes accordingly. do let me know if there's anything else to be done

apps/client/src/features/app-settings/panel/settings-panel/composite/CustomTranslationModal.tsx
```diff
@@ -0,0 +1,353 @@
...
+import { langEn, TranslationObject } from '../../../../../translation/languages/en';
+import { useTranslation } from '../../../../../translation/TranslationProvider';
+import * as Panel from '../../../panel-utils/PanelUtils';
+
+interface CustomTranslationModalProps {
+  isOpen: boolean;
+  onClose: () => void;
+}
+
+export default function CustomTranslationModal({ isOpen, onClose }: CustomTranslationModalProps) {
```
cpvalente, line 17: question: can we make this interface dynamic to show all the properties in the translation object?

as of now, adding a single entry would mean changes in many parts of the code base, which sounds like something we may forget to do or that can get out of sync

cpvalente (commented):
The code seems to work well, thank you
I believe we should investigate whether the `CustomTranslationModal` could have a dynamic form.
Unless there is a good reason for us not to want to introduce that code (error prone or complex) I believe it will improve the maintainability of the feature

cpvalente:
@coderabbitai full review

apps/client/src/features/app-settings/panel/settings-panel/composite/CustomTranslationModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+        fields.forEach(({ formKey, translationKey }) => {
+          translationData[translationKey] = formData[formKey];
+        });
+      });
+
+      await postUserTranslation(translationData as TranslationObject);
+      reset(formData);
+    } catch (error) {
+      setError('root', { message: maybeAxiosError(error) });
+      /** no error handling for now */
```
cpvalente: cleanup we are handling the error by showing it to the user

apps/client/src/features/app-settings/panel/settings-panel/composite/CustomTranslationModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+      title='Add Translations'
+      isOpen={isOpen}
+      onClose={onClose}
+      showCloseButton
+      showBackdrop
+      bodyElements={
+        <Panel.Section as='form' onSubmit={handleSubmit(onSubmit)} id='custom-translations-form'>
+          {Array.from(translationStructure.entries()).map(([category, fields]) => (
+            <Panel.Card key={category}>
+              <Panel.SubHeader>{category.charAt(0).toUpperCase() + category.slice(1)}</Panel.SubHeader>
```
cpvalente: question: should this not be solved in CSS?

apps/client/src/features/app-settings/panel/settings-panel/composite/CustomTranslationModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+    <Modal
+      title='Add Translations'
+      isOpen={isOpen}
+      onClose={onClose}
+      showCloseButton
+      showBackdrop
+      bodyElements={
+        <Panel.Section as='form' onSubmit={handleSubmit(onSubmit)} id='custom-translations-form'>
+          {Array.from(translationStructure.entries()).map(([category, fields]) => (
+            <Panel.Card key={category}>
```
cpvalente: question: can we simplify the code by removing the categories? it doesnt seem to me that this is a requirement to the feature and it complicates our implementation

apps/client/src/features/app-settings/panel/settings-panel/GeneralSettings.tsx
```diff
@@ -34,6 +36,9 @@ export default function GeneralSettings() {
       keepDirtyValues: true,
     },
   });
+  const [isCustomTranslationModalOpen, setIsCustomTranslationModalOpen] = useState(false);
+
+  const language = watch('language');
```
cpvalente: question: can we simplify this by always showing the button?

apps/client/src/features/app-settings/panel/settings-panel/GeneralSettings.tsx
```diff
@@ -34,6 +36,9 @@ export default function GeneralSettings() {
       keepDirtyValues: true,
     },
   });
+  const [isCustomTranslationModalOpen, setIsCustomTranslationModalOpen] = useState(false);
```
cpvalente: tip: we usually use the `useDisclosure` hook for this sort of code

apps/client/src/translation/TranslationProvider.tsx
```diff
@@ -21,36 +22,62 @@ const translationsList = {
...
+  };
+
   const contextValue = {
+    userTranslation,
     getLocalizedString,
+    postUserTranslation,
+    refetchTranslation: fetchUserTranslation,
   };
 
+  useEffect(() => {
```
cpvalente: question: would it be simpler to use react-query to manage the fetching of this resource?

Shobhit-Nagpal:
made the following changes:
- added a `useCustomTranslation` hook to use react query to fetch the translations and invalidating it on posting new translation
- translation modal now uses `useDisclosure` for its functionality
- the "Add translation" button is always in view now
- refactored mapping translation structure from categories to the initial fields themselves
- cleaned up comment on error handling

recording of working after these changes (p.s: you can skip the recording from 0:15 to 1:38, it's just filling in the translations):


https://github.com/user-attachments/assets/f873a143-d6c4-497d-b012-7b1a76e3eea9

cpvalente:
@coderabbitai full review

cpvalente (approved):
😃

PR #1701, opened 2025-07-24
Feature/add compression

cpvalente:


alex-Arc (changes requested):
the current rundown get method is a straight read from memory, is there any value 
the compression could have some use though we need to be sure about compatibility

cpvalente:
> the current rundown get method is a straight read from memory, is there any value the compression could have some use though we need to be sure about compatibility

the difference here is that we are caching the compressed data. but I agree that this change is overkill, maybe we can apply compression and let express do its caching?

cpvalente:
the payloads are not large enough to justify compression and the tradeoff is not right

PR #1706, opened 2025-08-01
chore: rename offset to offsetAbs and offsetRel

alex-Arc:


cpvalente (approved):
Looks good, in the interest of moving less data, how about:
- `absOffset` and `relOffset`
- `offsetRel` and `offsetAbs`
- stick with what we have
?

PR #1708, opened 2025-08-03
Feat: Time until group and flag

alex-Arc:


apps/server/src/services/timerUtils.ts
```diff
@@ -8,13 +8,44 @@ import { shouldCrashDev } from '../utils/development.js';
...
+
+  const nextFlagEvent = rundown.entries[nextFlag.id];
+
+  // eslint-disable-next-line no-unused-labels -- dev code path
+  DEV: {
+    if (!isOntimeEvent(nextFlagEvent) || !eventNow) {
+      throw new Error('timerUtils.getExpectedFlagStart: invalid state received');
+    }
+  }
+  const v = getExpectedStart(nextFlagEvent, {
```
cpvalente: question: can we put an actual name here instead of v?

apps/server/src/services/timerUtils.ts
```diff
@@ -8,13 +8,44 @@ import { shouldCrashDev } from '../utils/development.js';
...
+    actualStart,
+    plannedStart,
+  });
+  return v;
+}
+
 /**
  * Calculates the expected time of the group to end.
  * Should only be called if a block is running
  * TODO: take a look at how it handles relative offset mode
```
cpvalente: question: can we remove this TODO?

apps/client/src/common/utils/time.ts
```diff
@@ -134,18 +134,25 @@ export function formatDuration(duration: number, hideSeconds = true): string {
...
     totalGap: number;
     isLinkedToLoaded: boolean;
   },
 ): number {
-  const { offset, clock, currentDay, offsetMode, actualStart, plannedStart } = useTimeUntilData();
-  return calculateTimeUntilStart({ ...data, currentDay, clock, offset, offsetMode, actualStart, plannedStart });
+  const { offset, currentDay, offsetMode, actualStart, plannedStart, clock } = useExpectedStartData();
+  if (event === null) return 0;
+
+  //TODO: maybe memo this
```
cpvalente: as a rule of thumb, memoising things that have objects or ticking values "like offset" may not give us that much performance benefit.

I would suggest leaving as is and memoising if we get around to test whether there are actual benefits

apps/server/src/stores/runtimeState.ts
```diff
@@ -521,10 +522,15 @@ export function update(): UpdateResult {
   }
 
   if (runtimeState.blockNow) {
-    const expectedBlockFinish = getExpectedBlockFinish(runtimeState, getCurrentRundown());
+    const expectedBlockFinish = getExpectedBlockFinish(runtimeState, getCurrentRundown()); //TODO: this in pulling in rundown by it self, we are tying not to do that in this file
```
cpvalente: question: could we not have saved the block information from the load? this will take the rundown at every tick (32ms) which is not ideal)

how about either:
- we save the block in a private object like we do with `_timer`
- we expand `blockNow` to contain the data we need

apps/server/src/stores/runtimeState.ts
```diff
@@ -521,10 +522,15 @@ export function update(): UpdateResult {
...
     runtimeState.blockNow.expectedEnd = expectedBlockFinish;
   }
 
+  if (runtimeState.nextFlag) {
+    const expectedFlagStart = getExpectedFlagStart(runtimeState, getCurrentRundown()); //TODO: this in pulling in rundown by it self, we are tying not to do that in this file
+    runtimeState.nextFlag.expectedStart = expectedFlagStart;
```
cpvalente: same comment here, lets do this on load

apps/client/src/features/overview/composite/TimeElements.tsx
```diff
@@ -36,7 +45,7 @@ export function StartTimes() {
       <div className={style.row}>
         <span className={style.label}>Start</span>
         <div className={style.labelledElement}>
-          <Tooltip text='Planned start time' render={<TbCalendar className={style.icon} />} />
+          <Tooltip text='Planned start time' render={<TbCalendarPin className={style.icon} />} />
```
cpvalente, line 47: question: if we are changing the fixed times to have the pin icon, should the end time not have one as well?

PR #1711, opened 2025-08-05
Optimize

alex-Arc:


cpvalente (approved):
nice

apps/server/src/stores/runtimeState.ts
```diff
@@ -508,7 +518,7 @@ export function update(): UpdateResult {
   const { offsetAbs, offsetRel } = getRuntimeOffset(runtimeState);
   runtimeState.runtime.offsetAbs = offsetAbs;
   runtimeState.runtime.offsetRel = offsetRel;
-  runtimeState.runtime.expectedEnd = getExpectedEnd(runtimeState);
+  // runtimeState.runtime.expectedEnd = getExpectedEnd(runtimeState);
```
cpvalente, line 521: not sure if we have a reason to keep this here, but maybe we can delete it

PR #1715, opened 2025-08-08
Invert overtime

alex-Arc:


alex-Arc:
@coderabbitai full review

apps/server/src/services/__tests__/timerUtils.test.ts
```diff
@@ -858,7 +858,8 @@ describe('getRuntimeOffset()', () => {
     expect(offsetAbs).toBe(0);
   });
 
-  it('handles loaded event', () => {
+  //TODO: do we actually what to give offsets just on load
```
alex-Arc: @cpvalente 
what do you think of this
is there a value of giving an offset event if the event is just loaded?

PR #1716, opened 2025-08-08
Remove timer.finishedAt from runtime data

alex-Arc:


apps/server/src/services/__tests__/timerUtils.test.ts
```diff
@@ -21,17 +21,17 @@ describe('getExpectedFinish()', () => {
       timer: {
         addedTime: 0,
         duration: 10,
-        finishedAt: null,
         startedAt: null,
       },
       _timer: {
         pausedAt: null,
+        hasFinished: false,
```
cpvalente, line 28: question: is there a reason to keep `finishedAt` as a number? it would allow us to make calculations that imply knowing the offset since finish time. Nothing springs to mind to me....

packages/types/src/definitions/runtime/TimerState.type.ts
```diff
@@ -22,8 +22,7 @@ export type TimerState = {
   elapsed: MaybeNumber;
   /** time we expect timer to finish */
   expectedFinish: MaybeNumber;
-  /** only if timer has already finished */
-  finishedAt: MaybeNumber;
+
```
cpvalente: nitpick: we want to remove this line

cpvalente (approved):

apps/server/src/services/__tests__/timerUtils.test.ts
alex-Arc: not as fare as i could see either 
and then we can just put it back as it in internal to the server now

PR #1717, opened 2025-08-08
refactor: remove onAir

cpvalente:
need to remember to remove this from the companion module


cpvalente:
> 👍🏽 is it already removed from the studio clock?

yes, https://github.com/cpvalente/ontime/blob/alpha-2/apps/client/src/views/studio/StudioClock.tsx#L23

PR #1723, opened 2025-08-10
fix: move playing event into group

alex-Arc:


apps/server/src/stores/runtimeState.ts
```diff
@@ -358,6 +358,11 @@ export function updateAll(rundown: Rundown, metadata: RundownMetadata) {
   loadNext(rundown, metadata, eventNowIndex >= 0 ? eventNowIndex : undefined);
   updateLoaded(runtimeState.eventNow ?? undefined);
   loadGroupFlagAndEnd(rundown, metadata, eventNowIndex);
+
+  // catch the edge case where the playing event is moved into a group
+  if (runtimeState.groupNow && runtimeState.groupNow.startedAt === null) {
```
cpvalente, line 365: What if an event moves from one group to another?

Would it be simpler to check if the group is different?

alex-Arc: that would also be handled by this logic 
we dont what to override the `startedAt` if t is there

cpvalente: Let's just make a comment to explain that for the next time

cpvalente (approved):

PR #1724, opened 2025-08-10
feat: abort partial load if there is a version missmatch

alex-Arc:


apps/client/src/features/app-settings/panel/project-panel/ProjectMergeForm.tsx
```diff
@@ -61,6 +61,10 @@ export default function ProjectMergeForm({ onClose, fileName }: ProjectMergeFrom
 
       // make patch object
       const { data } = await getDb(fileName);
+      if (!data.settings.version.startsWith('4.')) {
+        setError('The project you are attempting to merge is from v3, it need to be migrated first');
```
cpvalente:
```suggestion
        setError('The project you are attempting to merge is from an older version and it need to be migrated first');
```

PR #1728, opened 2025-08-14
refactor: client rundown metadata

alex-Arc:
this tries to move the calculation off metadata in the client into the useRundown hook so it can be easily used in other places besides the main editor


apps/client/src/common/hooks-query/useRundown.ts
```diff
@@ -34,6 +36,24 @@ export default function useRundown() {
...
 }
 
+export function useRundownWithMetadata() {
+  const { data, status } = useRundown();
+  const { selectedEventId } = useSelectedEventId();
+  const { process } = initRundownMetadata(selectedEventId);
+  // keep a single reference to the metadata which we override for every entry
+  const rundownMetadata: Record<string, Readonly<RundownMetadata>> = {};
+
+  data.flatOrder.reduce((acc, id) => {
```
cpvalente: Question: I fear this pattern could give us performance issues 

Would a for loop here be simpler and more performant?

apps/client/src/features/rundown/Rundown.tsx
```diff
@@ -468,13 +465,11 @@ export default function Rundown({ data }: RundownProps) {
...
               if (!entry) return null;
 
-              rundownMetadata = process(entry);
-
               // if the entry has a parent, and it is collapsed, render nothing
               if (
                 entry.type !== SupportedEntry.Group &&
-                rundownMetadata.groupId !== null &&
-                getIsCollapsed(rundownMetadata.groupId)
+                rundownMetadata[entryId].groupId !== null &&
```
cpvalente: Question: I wonder if there is a danger of this getting out of sync, especially since in Rundown we use a copy of the data to improve the drag and drop

We would likely prefer checking for the entry rundown metadata at the top and assigning it to a placeholder constant

apps/client/src/features/rundown/RundownWrapper.tsx
```diff
@@ -12,12 +12,12 @@ interface RundownWrapperProps {
...
 
 export default function RundownWrapper({ isSmallDevice }: RundownWrapperProps) {
-  const { data, status } = useRundown();
+  const { data, status,rundownMetadata } = useRundownWithMetadata();
 
   return (
     <div className={styles.rundownWrapper}>
       {isSmallDevice ? <RundownHeaderMobile /> : <RundownHeader />}
-      {status === 'success' && data ? <Rundown data={data} /> : <Empty text='Connecting to server' />}
+      {status === 'success' && data ? <Rundown data={data} rundownMetadata={rundownMetadata} /> : <Empty text='Connecting to server' />}
```
cpvalente: Question: would there be a render where data and metadata are out of sync? Would it be an idea to check somehow if both data and metadata are ready?

apps/client/src/features/rundown/Rundown.tsx
alex-Arc: no you are right i'm hitting this issue
I could somewhat fix it. should i update it in the same useEffect as the sorted list?

cpvalente: I don't think so, the rundown metadata is a stable object. I think we just need to check at the top of the render loop where we also check if the event exist

apps/client/src/common/utils/rundownMetadata.ts
```diff
@@ -0,0 +1,144 @@
...
+export function getRundownMetadata(data: Pick<Rundown, 'entries' | 'flatOrder'>, selectedEventId: MaybeString) {
+  const { metadata, process } = initRundownMetadata(selectedEventId);
+  // keep a single reference to the metadata which we override for every entry
+  let newMetadata = metadata;
+  const rundownMetadata: RundownMetadataObject = {};
+
+  for (const id of data.flatOrder) {
+    const entry = data.entries[id];
+    newMetadata = process(entry);
+    Object.assign(rundownMetadata, { [id]: newMetadata });
```
cpvalente: is there a reason for us to use Object.assign? I dont think we need the intermediate object. the straight assignment with 
`rundownMetadata[id] = newMetadata` is both easier to read and more performant

apps/client/src/common/utils/rundownMetadata.ts
```diff
@@ -0,0 +1,144 @@
...
+  groupColour: string | undefined;
+  groupEntries: number | undefined;
+};
+
+export type RundownMetadataObject = Record<string, Readonly<RundownMetadata>>;
+
+export function getRundownMetadata(data: Pick<Rundown, 'entries' | 'flatOrder'>, selectedEventId: MaybeString) {
+  const { metadata, process } = initRundownMetadata(selectedEventId);
+  // keep a single reference to the metadata which we override for every entry
+  let newMetadata = metadata;
```
cpvalente: question: is reassigning this the root of our reference problems? would our code be easier and safer if this was assigned per loop iteration?

apps/client/src/common/utils/rundownMetadata.ts
alex-Arc: 💯 ah yeah that is over left over from the reduce function, you are totally right

apps/client/src/common/utils/rundownMetadata.ts
alex-Arc: with this structure, if there are 0 entries then the very first `metadata` get's assigned to the `LAST` element
and it there are entries then whatever element gest processed last is put into the `LAST` element
I think this is how it was when it was waved into the Rundown render logic ?

cpvalente: yes, that is right. this was the structure we had before.

If I remember correctly, it was to make sure we always had a metadata value (as you mention, to handle cases of empty rundowns). But since we no longer have a separate function body / render body we could try to simplify this?

alex-Arc: I don't quite see how
The quick add buttons at the end of the rundown need empty metadata in case the rundown is empty

alex-Arc:
@cpvalente feel free to merge when you are happy with it

cpvalente:
@coderabbitai full review

PR #1730, opened 2025-08-15
feat: sheet import new features for v4

alex-Arc:


alex-Arc:
@coderabbitai review

apps/server/src/api-data/excel/__tests__/excel.parser.test.ts
```diff
@@ -168,10 +168,41 @@ describe('parseExcel()', () => {
...
     const result = parseExcel(testdata, {}, 'testSheet', importMap);
-    const firstEvent = result.rundown.entries[result.rundown.order[0]];
+    const firstGroup = result.rundown.entries[result.rundown.order[0]];
+
+    expect(result.rundown.order.length).toBe(1);
+    expect(result.rundown.flatOrder.length).toBe(2);
+    expect((firstGroup as OntimeEvent).type).toBe(SupportedEntry.Group);
+  });
+
+  it('places event between groups inside the group', () => {
```
cpvalente, line 181: suggestion: could you add the `duration` property here to make sure it gets imported to `targetDuration`?

apps/server/src/api-data/excel/excel.parser.ts
```diff
@@ -259,14 +276,42 @@ export const parseExcel = (
...
+      if (currentGroupId) {
+        (rundown.entries[currentGroupId] as OntimeGroup).entries = groupEntries.splice(0);
+      }
+      rundown.order.push(id);
+      rundown.flatOrder.push(id);
+      currentGroupId = id;
+      return;
+    }
+
+    if (isOntimeMilestone(entry as OntimeEntry)) {
```
cpvalente, line 202: todo: remove the related todo in the function signature

apps/server/src/api-data/excel/excel.parser.ts
```diff
@@ -276,11 +321,20 @@ export const parseExcel = (
...
+      groupEntries.push(id);
+      event.parent = currentGroupId;
+    } else {
+      rundown.order.push(id);
+    }
     rundown.flatOrder.push(id);
     rundown.entries[id] = event;
   });
 
+  if (currentGroupId) {
```
cpvalente, line 240: for groups: we should import the `duration` as the `targetDuration` property

apps/server/src/api-data/excel/__tests__/excel.parser.test.ts
Gwenitora: I agree that suggestion, it's a good idea

apps/server/src/api-data/excel/excel.parser.ts
```diff
@@ -259,14 +276,42 @@ export const parseExcel = (
...
+      }
+      return;
+    }
+
     // from excel, we can only get groups, milestones and events
-    if (isOntimeGroup(entry)) {
-      const group: OntimeGroup = { ...entry, custom: { ...entryCustomFields } };
-      rundown.order.push(id);
+    if (isOntimeGroup(entry as OntimeEntry)) {
+      const group: OntimeGroup = { ...entry, custom: { ...entryCustomFields } } as OntimeGroup;
```
Gwenitora:
```suggestion
      const group = { ...entry, custom: { ...entryCustomFields } } as OntimeGroup;
```
You can technically remove typage of this variable, because the type is already defined by the `as OntimeGroup`

apps/server/src/api-data/excel/excel.parser.ts
Gwenitora: I'm not agree with that, because remove the `as OntimeEntry` generate a typescript error

apps/server/src/api-data/excel/excel.parser.ts
Gwenitora: I'm agree with that, but not only here, we need to set this variable also on group-end condition, a comment already exist for this other line
-> line 281

apps/server/src/api-data/excel/excel.parser.ts
```diff
@@ -259,14 +276,42 @@ export const parseExcel = (
     }
 
     const id = entry.id || generateId();
+
+    if (entry.type === 'group-end') {
+      if (currentGroupId) {
```
Gwenitora, line 177: for groups: we should import the `duration` as the `targetDuration` property
also need to made if group-end does not exist, a comment already exist for this other line
-> line 334

cpvalente:
@coderabbitai full review

apps/server/src/api-data/excel/parsing/utils.ts
```diff
@@ -0,0 +1,169 @@
...
+  const sheetMetadata: SheetMetadata = {};
+
+  const handlers = {
+    [importMap.timeStart]: (row: number, col: number) => {
+      indexMap.timeStart = col;
+      sheetMetadata.timeStart = { row, col };
+    },
+    [importMap.linkStart]: (row: number, col: number) => {
+      indexMap.linkStart = col;
+      sheetMetadata.timeEnd = { row, col };
```
cpvalente: should this be `linkStart`?

```suggestion
      sheetMetadata.linkStart = { row, col };
```

apps/server/src/api-data/excel/excel.parser.ts
```diff
@@ -63,166 +68,75 @@ export const parseExcel = (
...
 
     for (let j = 0; j < row.length; j++) {
       const column = row[j];
       // 1. we check if we have set a flag for a known field
-      if (j === timerTypeIndex) {
-        const maybeTimeType = makeString(column, '');
-        if (maybeTimeType === 'group') {
-          // we leave this as a clue for the object filtering later on
+      if (j === indexMap.timerType) {
+        const maybeTimeType = makeString(column, '').toLocaleLowerCase();
```
cpvalente: I believe it is better to use the normal `lowerCase` so we have predictable results. `localeLowerCase` can be OK for user facing text

```suggestion
        const maybeTimeType = makeString(column, '').toLowerCase();
```

apps/server/src/api-data/excel/excel.parser.ts
```diff
@@ -63,166 +68,75 @@ export const parseExcel = (
...
+        if (maybeTimeType === 'group' || maybeTimeType === 'group-start') {
           entry.type = SupportedEntry.Group;
           entry.entries = [];
+        } else if (maybeTimeType === 'group-end') {
+          entry.type = 'group-end';
+        } else if (maybeTimeType === 'milestone') {
+          entry.type = SupportedEntry.Milestone;
+          // the assumption is event
```
cpvalente: question: is this comment maybe misplaced?

apps/server/src/api-data/excel/parsing/utils.ts
```diff
@@ -0,0 +1,169 @@
+import { CustomFieldKey, CustomFields, MaybeNumber } from 'ontime-types';
```
cpvalente: question: how about naming this excel.utils.ts and leaving it at the root directory with the other excel things? That is at least the pattern we have from the other parts of the code

cpvalente (commented):
We have a small typo to correct and a few questions. Altogether looks good and seems to work well otherwise

apps/server/src/api-data/excel/parsing/utils.ts
alex-Arc: can do

PR #1732, opened 2025-08-15
fix: ensure filename sanitation in all routes

cpvalente:
this is an attempt at fixing the user issue with the unescaped characters

The user seemed to have gotten a file named `'MemoriaÃ'$'\302\205'' Kamili Skolimowskiej 2025.json'` which broke the load

It turned out that our quick project and upload routes did not sanitise the file name. I was unable to find any other routes where the sanitation did not happen

### DIFF
--- a/apps/server/src/api-data/db/db.controller.ts
+++ b/apps/server/src/api-data/db/db.controller.ts
@@ -8,6 +8,7 @@ import {
 import { getErrorMessage } from 'ontime-utils';
 
 import type { Request, Response } from 'express';
+import sanitize from 'sanitize-filename';
 
 import {
   doesProjectExist,
@@ -74,7 +75,8 @@ export async function createProjectFile(req: Request, res: Response<{ filename: 
  */
 export async function quickProjectFile(req: Request, res: Response<{ filename: string } | ErrorResponse>) {
   try {
-    const filename = await projectService.createProject(req.body.project.title || 'untitled', req.body);
+    const nameFromTitle = req.body.project.title ? sanitize(req.body.project.title) : 'untitled';
+    const filename = await projectService.createProject(nameFromTitle, req.body);
     res.status(200).send({
       filename,
     });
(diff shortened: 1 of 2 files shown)


cpvalente:
@coderabbitai full review

apps/server/src/api-data/db/db.controller.ts
```diff
@@ -74,7 +75,8 @@ export async function createProjectFile(req: Request, res: Response<{ filename:
  */
 export async function quickProjectFile(req: Request, res: Response<{ filename: string } | ErrorResponse>) {
   try {
-    const filename = await projectService.createProject(req.body.project.title || 'untitled', req.body);
+    const nameFromTitle = req.body.project.title ? sanitize(req.body.project.title) : 'untitled';
+    const filename = await projectService.createProject(nameFromTitle, req.body);
```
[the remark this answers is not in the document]
cpvalente: it is unfortunate that we use the project title to make a file name, but we should not force these to be related from the users perspective

apps/server/src/api-data/db/db.controller.ts
```diff
@@ -74,7 +75,8 @@ export async function createProjectFile(req: Request, res: Response<{ filename:
  */
 export async function quickProjectFile(req: Request, res: Response<{ filename: string } | ErrorResponse>) {
   try {
-    const filename = await projectService.createProject(req.body.project.title || 'untitled', req.body);
+    const nameFromTitle = req.body.project.title ? sanitize(req.body.project.title) : 'untitled';
```
alex-Arc, line 78: do we not what to do this as part of the validation chain ( like the the others )
we can also add `.default('untitled')` the chain

cpvalente: Don't think so. 
The root issue is that we reuse the project title as a file name

It is ok for the project to have no title but we still need a name for the file

alex-Arc: hmm fair

alex-Arc (approved):

PR #1733, opened 2025-08-15
Fixes

cpvalente:
small batch of change on alpha 2
- [x] change order of navigation and configuration locks
- [x] fix issue with conflict of local storage keys
- [x] fix issue with automation form state 

and small style tweaks

### DIFF
--- a/apps/client/src/common/utils/localStorage.ts
+++ b/apps/client/src/common/utils/localStorage.ts
@@ -1,3 +1,5 @@
+import { baseURI } from '../../externals';
+
 export function booleanFromLocalStorage(key: string, fallback: boolean): boolean {
   const valueInStorage = localStorage.getItem(key);
   if (valueInStorage) {
@@ -6,4 +8,9 @@ export function booleanFromLocalStorage(key: string, fallback: boolean): boolean
     localStorage.setItem(key, String(fallback));
     return fallback;
   }
+}
+
+export function makeStageKey(key: string) {
+  if (baseURI) return `${baseURI}-${key}`;
+  return key;
 }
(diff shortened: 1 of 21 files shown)


cpvalente:
@coderabbitai full review

apps/client/src/common/utils/localStorage.ts
```diff
@@ -7,3 +9,8 @@ export function booleanFromLocalStorage(key: string, fallback: boolean): boolean
     return fallback;
   }
 }
+
+export function makeStageKey(key: string) {
```
alex-Arc, line 13: would there be an idea in making our own  localStore function where we handle this.
and the just use that everywhere?

apps/client/src/features/sharing/GenerateLinkForm.tsx
```diff
@@ -227,16 +227,25 @@ export default function GenerateLinkForm({ hostOptions, pathOptions, presets, is
...
                 checked={watch('lockNav')}
                 onCheckedChange={(checked) => setValue('lockNav', checked, { shouldDirty: true })}
+                disabled={watch('lockConfig')}
               />
             </Panel.ListItem>
             {watch('path') !== OntimeView.Cuesheet && (
               <Panel.ListItem>
-                <Panel.Field title='Lock configuration' description='Whether to hide the configuration panel' />
+                <Panel.Field
+                  title='Lock configuration'
```
alex-Arc, line 236: seams there is some state hanging around with the generate button
Recording 2025-08-16 213955

apps/client/src/common/utils/localStorage.ts
cpvalente: That was my initial thought, but I didnt find a good path to do it.
We consume local store in a few different ways (mantine hook and manually), this means we need variants of the function (a hook extending mantine and a localStorage wrapper)

It is also arguable that some settings are user preferences that we may have in multiple stages. For now I have decided to only wrap in the stage key, options which are functional (things we use behind the scenes to make the app work)

Altogether it felt it would be too much work and the simpler solution would be to generate keys on demand, although I realise this is a potential footgun

PR #1748, opened 2025-08-24
Group duration context menu utils

alex-Arc:
Adds 2 items to the context menu
for the group there is now the option to set the target duration to match the accumulated duration of the group
for events inside groups one can now make that one event expand or contract in duration so that the groups target is met

### DIFF
--- a/apps/client/src/common/hooks/useEntryAction.ts
+++ b/apps/client/src/common/hooks/useEntryAction.ts
@@ -49,10 +49,10 @@ import {
   requestEventSwap,
   requestGroupEntries,
   requestUngroup,
+  requestFitGroupTarget,
 } from '../api/rundown';
 import { logAxiosError } from '../api/utils';
 import { useEditorSettings } from '../stores/editorSettings';
-import { eventDurationMatchGroupTarget } from '../utils/time';
 
 export type EventOptions = Partial<{
   // options of any new entries (event / delay / group)
@@ -473,26 +473,16 @@ function useEntryActionsForRundown(scopedRundownId: string | undefined) {
   /**
    * Updates time of existing event so it satisfies the group target duration
    * @param eventId {EntryId} - id of the event
-   * @param groupId {EntryId} - id of the enclosing group
    */
   const matchGroupDuration = useCallback(
-    async (eventId: EntryId, groupId: EntryId) => {
-      const rundown = queryClient.getQueryData<Rundown>(resolveCurrentRundownQueryKey());
-      if (!rundown) return;
-      const group = rundown.entries[groupId];
-      if (!group || !isOntimeGroup(group)) return;
-      const event = rundown.entries[eventId];
-      if (!event || !isOntimeEvent(event)) return;
-
-      const newDuration = eventDurationMatchGroupTarget({
-        targetDuration: group.targetDuration,
-        groupDuration: group.duration,
-        eventDuration: event.duration,
-      });
-      if (!newDuration) return;
-      updateTimer(eventId, 'duration', String(newDuration / MILLIS_PER_SECOND) + 's', false);
+    async (eventId: EntryId) => {
+      const rundownId = getCurrentRundownData()?.id;
+      if (!rundownId) {
+        throw new Error('Rundown not initialised');
+      }
+      await requestFitGroupTarget(rundownId, eventId);
     },
-    [queryClient, updateTimer, resolveCurrentRundownQueryKey],
+    [getCurrentRundownData],
   );
 
   /**
--- a/apps/client/src/common/hooks/useEntryAction.ts
+++ b/apps/client/src/common/hooks/useEntryAction.ts
@@ -480,7 +480,12 @@ function useEntryActionsForRundown(scopedRundownId: string | undefined) {
       if (!rundownId) {
         throw new Error('Rundown not initialised');
       }
-      await requestFitGroupTarget(rundownId, eventId);
+
+      try {
+        await requestFitGroupTarget(rundownId, eventId);
+      } catch (error) {
+        logAxiosError('Error updating event', error);
+      }
     },
     [getCurrentRundownData],
   );
--- a/apps/client/src/common/hooks/useEntryAction.ts
+++ b/apps/client/src/common/hooks/useEntryAction.ts
@@ -371,6 +371,19 @@ function useEntryActionsForRundown(scopedRundownId: string | undefined) {
     [getCurrentRundownData, updateEntryMutation],
   );
 
+  const matchGroupDuration = useCallback(async (eventId: EntryId, groupId: EntryId) => {
+    const rundown = queryClient.getQueryData<Rundown>(RUNDOWN);
+    if (!rundown) return;
+    const group = rundown.entries[groupId];
+    if (!group || !isOntimeGroup(group) || group.targetDuration === null) return;
+    const event = rundown.entries[eventId];
+    if (!event || !isOntimeEvent(event)) return;
+    const durationDiff = group.targetDuration - group.duration;
+    const newDuration = event.duration + durationDiff;
+    if (newDuration < 0) return;
+    updateTimer(eventId, 'duration', String(newDuration / MILLIS_PER_SECOND) + 's', false);
+  }, []);
+
   /**
    * Updates time of existing event
    * @param eventId {EntryId} - id of the event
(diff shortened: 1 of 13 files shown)


apps/client/src/common/hooks/useEntryAction.ts
```diff
@@ -241,6 +241,19 @@ export const useEntryActions = () => {
     [updateEntry],
   );
 
+  const matchGroupDuration = useCallback(async (eventId: EntryId, groupId: EntryId) => {
+    const rundown = queryClient.getQueryData<Rundown>(RUNDOWN);
+    if (!rundown) return;
+    const group = rundown.entries[groupId];
+    if (!group || !isOntimeGroup(group) || group.targetDuration === null) return;
+    const event = rundown.entries[eventId];
+    if (!event || !isOntimeEvent(event)) return;
```
alex-Arc: we could really use a way to show errors like this 
maybe the toasters we were thinking of?

apps/client/src/features/rundown/rundown-event/RundownEvent.tsx
```diff
@@ -150,6 +151,16 @@ export default function RundownEvent({
               updateEntry({ id: eventId, flag: !flag });
             },
           },
+          {
+            type: 'item',
+            label: 'Match Target Group Duration',
+            icon: TbClockPin,
```
alex-Arc, line 192: ideas for icons are welcome

apps/client/src/features/rundown/rundown-group/RundownGroup.tsx
```diff
@@ -52,6 +57,13 @@ export default function RundownGroup({ data, hasCursor, collapsed, onCollapse }:
       disabled: data.entries.length === 0,
     },
     { type: 'divider' },
+    {
+      type: 'item',
+      label: 'Match Group Duration',
+      icon: TbClockPin,
```
alex-Arc, line 88: ideas for icons are welcome

cpvalente:
@alex-Arc, lets leave this one for later

cpvalente:
@alex-Arc we are in a better place to add this now. Could we prepare it for production?

cpvalente:
@alex-Arc , lets chat about group options.
I think we have enough experience with groups now that we can make a list of options groups should have

(didnt mean to close the PR, wrong button)

apps/client/src/features/rundown/rundown-event/RundownEvent.tsx
cpvalente: i think the pin is good

apps/client/src/features/rundown/rundown-group/RundownGroup.tsx
cpvalente: looks good

apps/client/src/common/hooks/useEntryAction.ts
```diff
@@ -460,7 +460,28 @@ function useEntryActionsForRundown(scopedRundownId: string | undefined) {
...
-    [getCurrentRundownData, updateEntryMutation, queryClient],
+    [getCurrentRundownData, updateEntryMutation, queryClient, resolveCurrentRundownQueryKey],
+  );
+
+  /**
+   * Updates time of existing event so it satisfies the group target duration
+   * @param eventId {EntryId} - id of the event
+   * @param groupId {EntryId} - id of the enclosing group
+   */
+  const matchGroupDuration = useCallback(
```
cpvalente, line 477: it would be nice to see a simple unit test for this.
i would extract the logic into a function and leave the hook as a simple orchestration layer
- get data
- call new function that calculates duration
- update timer
as of now, the actual logic is buried in orchestration (getting the group and updating)

apps/client/src/features/rundown/rundown-event/RundownEvent.tsx
```diff
@@ -172,6 +173,16 @@ export default function RundownEvent({
               updateEntry({ id: eventId, flag: !flag });
             },
           },
+          {
+            type: 'item',
+            label: 'Match Group Target Duration',
```
cpvalente, line 190: Should this not be "Change duration to match group target" or something like that?

apps/client/src/features/rundown/rundown-group/RundownGroup.tsx
```diff
@@ -61,6 +66,13 @@ export default function RundownGroup({ data, hasCursor, collapsed, onCollapse }:
       disabled: data.entries.length === 0,
     },
     { type: 'divider' },
+    {
+      type: 'item',
+      label: 'Match Content Duration',
```
cpvalente, line 87: should this be 'match content duration' if it is locked or 'lock to content duration as target' otherwise?

apps/client/src/features/rundown/rundown-group/RundownGroup.tsx
```diff
@@ -61,6 +66,13 @@ export default function RundownGroup({ data, hasCursor, collapsed, onCollapse }:
...
+      icon: TbClockPin,
+      onClick: matchDuration,
```
cpvalente, line 89: question: should this be disabled if it already matches?

apps/client/src/features/rundown/rundown-event/RundownEvent.tsx
```diff
@@ -172,6 +173,16 @@ export default function RundownEvent({
...
+            icon: TbClockPin,
+            onClick: () => {
+              if (!parent) return;
+              matchGroupDuration(eventId, parent);
+            },
+            disabled: !parent,
```
cpvalente: question: should we also disable if we already have a match?

cpvalente (commented):
nice! lets get this out for next release
just a few small UX tweaks for you to consider

alex-Arc:
@coderabbitai review

apps/client/src/common/hooks/useEntryAction.ts
```diff
@@ -460,7 +461,32 @@ function useEntryActionsForRundown(scopedRundownId: string | undefined) {
...
+
+      const newDuration = eventDurationMatchGroupTarget({
+        targetDuration: group.targetDuration,
+        groupDuration: group.duration,
+        eventDuration: event.duration,
+      });
+      if (!newDuration) return;
+      updateTimer(eventId, 'duration', String(newDuration / MILLIS_PER_SECOND) + 's', false);
+    },
+    [queryClient, updateTimer, resolveCurrentRundownQueryKey],
```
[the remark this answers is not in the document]
cpvalente: @alex-Arc I think the comment is correct but over protective. if the duration change needed is 0 then a no-op is just fine.
maybe a comment is good enough like `// null or 0 values are a noop`

for point 2. the window is too small, again, the comment is correct but overprotective

comment 3 and 4 sound like something we need to do

cpvalente (approved):
I think some of the coderabbit comments have merit and are quick fixes. either way, looks good

apps/client/src/features/rundown/rundown-group/RundownGroup.tsx
```diff
@@ -186,7 +201,9 @@ export default function RundownGroup({ data, hasCursor, collapsed, onCollapse }:
                   <Tag className={style.offsetLabel}>{planOffset}</Tag>
                 </span>
               )}
-              {data.targetDuration !== null && <IoLockClosed className={style.lockIcon} />}
+              {data.targetDuration !== null && (
+                <TbTargetArrow className={cx([style.lockIcon, isDurationMatching ? style.active : style.inactive])} />
```
cpvalente: i am not sure about the active / inactive styles:

- any deviation from the target should have a warning colour (orange), we cant really tell if being light is worse than being heavy in time

- the icon is not vertically centered, it seems we have a composition issue

- all indicators in the entries have tooltips, should we have a tooltip to explain what the icon is?

- would a lock icon be better? as in: "this element has a wanted duration". i wonder if the current target icon makes it seem like "this element is on target", which is not always the case or what the feature guarantees

apps/client/src/features/rundown/rundown-event/RundownEvent.tsx
```diff
@@ -172,6 +176,20 @@ export default function RundownEvent({
               updateEntry({ id: eventId, flag: !flag });
             },
           },
+          {
+            type: 'item',
+            label: 'Match Group Target Duration',
+            description: 'Change event duration to fill the group target',
```
cpvalente: question: it seems that we already know here if the group is light or heavy, maybe we can change the message like

"Increase event duration to fit the group target"
"Decrease event duration to fit the group target"

PR #1749, opened 2025-08-25
fix: prevent stale references to custom fields

cpvalente:
the code could be better. but I am out of ideas for how to do it in a way that does not increase complexity or reduce performance


cpvalente:
@coderabbitai full review

alex-Arc (approved):

PR #1750, opened 2025-08-27
fix: prevent dropping a group inside another

cpvalente:
Fixes two issues related to reordering:
- reordering groups would cause groups to be inserted one inside the other
- previous logic allowed groups to be added at the top of a group


cpvalente:
> I don't quite grasp all the guards, but I can no longer make the error happen

You should be able to understand, can you comment what is not clear?

PR #1753, opened 2025-08-31
refactor: settings route

alex-Arc:
- remove some values from useSettings that are part of the default query client
- send refetch key if settings have changed
- merge controller and route file
- move all validation to validation middleware
- use matchedData to get clean and typed data from express validator


alex-Arc:
@coderabbitai review

PR #1765, opened 2025-09-05
Countdown timeuntil

alex-Arc:
TODO:
- [x] what should be displayed int the right most data block when not in show expected mode
- [x] add colours to single events
- [x] use report data
- [x] use time until
- [ ] fix up css

won't do this time
- [ ] groups..?


alex-Arc:
@coderabbitai review

PR #1777, opened 2025-09-09
fixes to beta 3

cpvalente:
Few pending fixes
- [x] fix phase colours in timer view
- [x] add missing default value for secondary source in timer view
- [x] prevent event editor reflow on different types
- [x] improve pin styes
- [x] fix issue with dragging a group after another 
- [x] improve following: in run mode we follow playback, in edit mode we follow the cursor
- [x] improve finder feature to consider milestones
- [x] fix op view check for skip in nested events


cpvalente:
@coderabbitai full review

PR #1778, opened 2025-09-11
export excel (.xlsx)

Gwenitora:


apps/client/src/common/api/excel.ts
```diff
@@ -20,6 +20,29 @@ export async function upload(file: File) {
...
+ * download Excel file from server
+ * @param fileName
+ */
+export async function download(fileName?: string) {
+  try {
+    const response = await axios.get(`${excelPath}/download`, {
+      responseType: 'blob',
+    });
+
+    const url = window.URL.createObjectURL(new Blob([response.data]));
```
alex-Arc: we have a `downloadBlob` function in `apps\client\src\common\api\utils.ts` could that be of use here?

apps/client/src/common/api/excel.ts
```diff
@@ -20,6 +20,29 @@ export async function upload(file: File) {
...
+    const link = document.createElement('a');
+    link.href = url;
+    link.setAttribute('download', `${fileName || 'Ontime_rundown'}.xlsx`);
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
+    window.URL.revokeObjectURL(url);
+  } catch (error) {
+    console.error('Erreur lors du téléchargement du fichier :', error);
```
alex-Arc: Maybe in english would be god 😉

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
```diff
@@ -10,7 +10,7 @@ import {
 
 import {
   deleteProject,
-  downloadCSV,
+//  downloadCSV,
```
alex-Arc: Was it maybe the intention to replace this with the new excel function

apps/server/src/api-data/excel/excel.creator.ts
```diff
@@ -0,0 +1,171 @@
...
+    v: entry.colour,
+  };
+  worksheet[xlsx.utils.encode_cell({ c: config.note.c, r: rowIndex })] = {
+    t: 's',
+    v: entry.note,
+  };
+
+  switch (entry.type) {
+    case 'event':
+      const strategy = inferStrategy(entry.timeEnd, entry.duration, entry.timeStrategy);
```
alex-Arc, line 40: if you create variables inside a a `case` you need to wrap the entire case in brackets see eslint [no-case-declarations](https://eslint.org/docs/latest/rules/no-case-declarations)


```ts
switch (data) {
   case 'event': {
      const variable = '1234'
      break;
   }
}
```

apps/server/src/api-data/index.ts
```diff
@@ -13,6 +13,9 @@ import { router as sessionRouter } from './session/session.router.js';
 import { router as viewSettingsRouter } from './view-settings/viewSettings.router.js';
 import { router as reportRouter } from './report/report.router.js';
 import { router as assetsRouter } from './assets/assets.router.js';
+import { parseExcel } from './excel/excel.parser.js';
```
alex-Arc: if these imports are not used they should be removed

apps/server/src/api-data/excel/excel.utils.ts
```diff
@@ -0,0 +1,19 @@
+import xlsx from 'xlsx';
+
+/**
+ * the content type of the rundown.templateInstructor.json
+ */
+export type configXlsxTemplate<Key extends any = { [key in 'c' | 'r']: number }> = { [key in "cue" | "title" | "colour" | "timeStart" | "timeEnd" | "duration" | "note" | "timerType" | "customFields"]: Key } & { [key in "id" | "linkStart" | "countToEnd" | "endAction" | "warningTime" | "dangerTime" | "skip"]?: Key };
```
alex-Arc: extends any is not needed here
```suggestion
export type configXlsxTemplate<Key = { [key in 'c' | 'r']: number }> = { [key in "cue" | "title" | "colour" | "timeStart" | "timeEnd" | "duration" | "note" | "timerType" | "customFields"]: Key } & { [key in "id" | "linkStart" | "countToEnd" | "endAction" | "warningTime" | "dangerTime" | "skip"]?: Key };
```

apps/server/src/api-data/excel/excel.creator.ts
```diff
@@ -0,0 +1,171 @@
+import { defaultImportMap } from 'ontime-utils';
+import xlsx from 'xlsx';
+import path from 'path';
+import fs from 'fs';
+import { changeMaxSizeOfExcel, configXlsxTemplate } from './excel.utils.js';
+import { CustomFields, OntimeEntry, Rundown, TimerType, TimeStrategy } from 'ontime-types';
+import { inferStrategy } from '../rundown/rundown.utils.js';
+
+const createRow = (
```
alex-Arc: a small style thing
in the server we tend to use the `function` definitions for al static funtions

apps/server/src/api-data/excel/excel.creator.ts
```diff
@@ -0,0 +1,171 @@
...
+  rundown: Rundown,
+  customFields: CustomFields,
+  options?: Partial<typeof defaultImportMap>,
+): xlsx.WorkBook => {
+  const workbook = xlsx.readFile(path.join(process.cwd(), 'excel', 'rundown template.xlsx'));
+  const sheetName = workbook.SheetNames[0];
+  const worksheet = workbook.Sheets[sheetName];
+
+  const config: configXlsxTemplate<string> = JSON.parse(
+    fs.readFileSync(path.join(process.cwd(), 'excel', 'rundown.templateInstructor.json'), { encoding: 'utf8' }),
```
alex-Arc: could the content of this file not just be a const definition at the top of this file? or in `excel.utils.ts`

apps/server/src/api-data/excel/excel.creator.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  return 0;
+};
+
+export const createExcel = (
+  rundown: Rundown,
+  customFields: CustomFields,
+  options?: Partial<typeof defaultImportMap>,
+): xlsx.WorkBook => {
+  const workbook = xlsx.readFile(path.join(process.cwd(), 'excel', 'rundown template.xlsx'));
```
alex-Arc: if possible definitions of files and paths should be done here `apps\server\src\setup\index.ts`

with that being said is there a change we can do this without having to read in a template file?
maybe with `xlsx.utils.book_new()`
if not we need to make sure is is copied over correctly in the build step

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/SourcesPanel.tsx
```diff
@@ -70,6 +71,10 @@ export default function SourcesPanel() {
     fileInputRef.current?.click();
   };
 
+  const handleDownload = async () => {
+    await downloadExcel();
```
alex-Arc: can we add there project file name here?

apps/client/src/common/api/excel.ts
Gwenitora: nice tips, thanks ^^

apps/client/src/common/api/excel.ts
Gwenitora: I'm REALLY sorry for this mistake, I'm working on 3 to 4 projects at the moment in addition to ontime, and they all have to be in French, so I sometimes forget to switch back to English when necessary, really sorry..

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/SourcesPanel.tsx
Gwenitora: Yes, I can, I have create that possibility to add a name for the file, but, since the user can change the name of the file after the fact, I wasn't sure that this was necessary. What do you think? Should I create an input to enter the name? Should I leave the default one? Should I enter another name? Which solution do you prefer?

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
Gwenitora: To explain, we discussed it with cpvalente (I don't have her exact name, sorry) but we thought of removing the csv option yes, because the excel option is similar, or even better, and in any case, an excel can be converted into csv, so we judged it useless to keep the csv

Gwenitora: but I kept the comments for the moment, until it is confirmed, and if you confirm that it is good to remove it, then I remove the comments ^^

apps/server/src/api-data/excel/excel.creator.ts
Gwenitora: The advantage of the model here is that it can be changed without touching the code, which can save time for future updates. However, a relevant note regarding compilation. I'll try this and get back to you.

apps/server/src/api-data/index.ts
Gwenitora: Exactly, all these imports are leftovers from tests that I did, I deleted the tests, and I obviously forgot to remove the imports that are linked to them, thank you for reminding me.

Dockerfile
```diff
@@ -25,6 +25,7 @@ COPY --from=builder /app/apps/server/dist/ ./server/
 COPY --from=builder /app/apps/server/src/external/ ./external/
 COPY --from=builder /app/apps/server/src/user/ ./user/
 COPY --from=builder /app/apps/server/src/html/ ./html/
+COPY --from=builder /app/apps/server/excel-template/ ./excel-template/
```
Gwenitora, line 28: N.B.: I'm not sure about that compilation of this folder, but I not already create my linux subsytem for test, can you test for me pls, thx ^^

Gwenitora:
technically, that work in dev and in electron build for windows, but, I don't have possibility to test in electron linux or mac or in docker, if you can help me to test that pls, thx. (https://github.com/cpvalente/ontime/pull/1778#discussion_r2342955956)
I would also add that I will use the stylized version of the xlsx plugin because the rendering is not beautiful, but hey, it is a plugin that has not been updated for a long time, so maybe find another solution? What do you think?

apps/server/src/api-data/excel/excel.creator.ts
Gwenitora: that work, now I compile the template and can edit that one, just need a confirmation of that feature for a docker version of ontime, refer of this comment: https://github.com/cpvalente/ontime/pull/1778#discussion_r2342955956

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/SourcesPanel.tsx
alex-Arc: having it inherit the name of the project file would be good

apps/server/src/api-data/excel/excel.creator.ts
alex-Arc: fair 👍🏽 

but we don't need that whole header in the export, that is intended to explain the import system

apps/client/src/features/app-settings/panel/project-panel/ProjectListItem.tsx
alex-Arc: ok, lets leave all the csv functionality in place, including, leaving the UI options as they are

alex-Arc (commented):
Cool, I will get you a test of the docker build
- lets leave all the csv functionality in place, including, leaving the UI options as they are
- we don't need the header in the exported excel file

Gwenitora:
Okay, I finish, I just need to create test I guess, and it's finished

Gwenitora:
finish, last review, and after You can drop me a new feature to make

alex-Arc:
you need to run `pnpm lint --fix` from inside `apps/client`

PR #1781, opened 2025-09-16
Options for countdown view

alex-Arc:
an all option for the countdown view


7omas:
This is a good addition for large projects.

Could perhaps allow adding events using custom field contents as a filter? Maybe something like "Add all in [Rundown/customField1/customField2]". Related to pull #1782.

cpvalente:
we will need to look into this properly

adding all the events is like using the op view.
maybe adding a filtering option to op like we have in backstage is better

lets save this for once we have some time to think (after v4 is stable)

PR #1783, opened 2025-09-17
Fix sheet export custom

alex-Arc:
fixes the issue where you can not export to sheets if there were groups and custom fields in the import map


cpvalente (approved):
nice, thank you alex

PR #1788, opened 2025-09-19
Keep screen awake

alex-Arc:
When using ontime with phones or tablets it could be usefull to keep them awake
it is simple to do with [WakeLock](https://developer.mozilla.org/en-US/docs/Web/API/WakeLock)
but that requires a secure context -> HTTPS, witch is not the case when running docker locally
[nosleep.js](https://www.npmjs.com/package/nosleep.js) attempts to cover that and other edge cases

not sure if it is worth the dependency?
or we could say it only works when hosted "correctly" ?


cpvalente:
> When using ontime with phones or tablets it could be usefull to keep them awake
> it is simple to do with [WakeLock](https://developer.mozilla.org/en-US/docs/Web/API/WakeLock)
> but that requires a secure context -> HTTPS, witch is not the case when running docker locally
> [nosleep.js](https://www.npmjs.com/package/nosleep.js) attempts to cover that and other edge cases
> 
> not sure if it is worth the dependency?
> or we could say it only works when hosted "correctly" ?

Thank you for this Alex
I think the WakeLock is a great feature, and we should definitely implement.

I agree that noSleep is not worth it for us

apps/client/src/common/components/navigation-menu/NavigationMenu.tsx
```diff
@@ -62,6 +64,13 @@ function NavigationMenu({ isOpen, onClose }: NavigationMenuProps) {
               <IoSwapVertical />
               {mirror && <span className={style.note}>Active</span>}
             </NavigationMenuItem>
+            {window.isSecureContext && (
+              <NavigationMenuItem active={keepAwake} onClick={() => toggleKeepAwake()}>
```
cpvalente:
```suggestion
              <NavigationMenuItem active={keepAwake} onClick={toggleKeepAwake}>
```

alex-Arc:
@coderabbitai review

PR #1789, opened 2025-09-20
Chore: Small todo about coerce boolean

alex-Arc:


apps/server/src/utils/coerceType.ts
```diff
@@ -52,7 +52,10 @@ export function coerceBoolean(value: unknown): boolean {
         throw new Error('Invalid value received');
     }
   }
-  return Boolean(value);
+  if (typeof value === 'number') {
+    return value > 0;
+  }
```
cpvalente, line 58: I am not sure about this logic.
`0` is typically a `false` in a boolean system
could we say `0` should be false and everything else should be `true`?

In practice, anything that is not a `0` or a `1` would be a weird value to pass on here. But treating not `0` as a truthy value would be more inline of how we do with strings for example (anything not empty is `true`)

alex-Arc: everything else?
including negative numbers?

cpvalente: Yeah, I think so.

I really think that only 0 and 1 are good values here. All the rest is being forced in the box

alex-Arc: ok
then i would say 1 and 0 is accepted the rest would error out,
 -1 turning into true dose not make any sense

PR #1795, opened 2025-09-23
Fix: rundown import

alex-Arc:
this attempts to fix the issue where a excel import would create a new rundown and not just override the actively loaded one with is the expected behaviour

### DIFF
--- a/apps/client/src/features/app-settings/panel/manage-panel/sources-panel/ImportReview.tsx
+++ b/apps/client/src/features/app-settings/panel/manage-panel/sources-panel/ImportReview.tsx
@@ -18,8 +18,7 @@ interface ImportReviewProps {
 
 export default function ImportReview(props: ImportReviewProps) {
   const { rundown, customFields, onFinished, onCancel } = props;
-  const { data } = useRundown();
-  const { id, title } = data;
+  const { data: currentRundown } = useRundown();
   const [loading, setLoading] = useState(false);
   const { importRundown } = useGoogleSheet();
   const resetPreview = useSheetStore((state) => state.resetPreview);
@@ -36,7 +35,7 @@ export default function ImportReview(props: ImportReviewProps) {
     // so the id needs to match
     await importRundown(
       {
-        [id]: { ...rundown, id, title },
+        [currentRundown.id]: { ...rundown, id: currentRundown.id, title: currentRundown.title },
       },
       customFields,
     );
--- a/apps/server/src/api-data/db/db.controller.ts
+++ b/apps/server/src/api-data/db/db.controller.ts
@@ -4,7 +4,6 @@ import {
   MessageResponse,
   ProjectFileListResponse,
   ProjectLogoResponse,
-  RefetchKey,
 } from 'ontime-types';
 import { getErrorMessage } from 'ontime-utils';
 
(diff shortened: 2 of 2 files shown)


alex-Arc:
@coderabbitai review

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/ImportReview.tsx
```diff
@@ -17,7 +18,8 @@ interface ImportReviewProps {
 
 export default function ImportReview(props: ImportReviewProps) {
   const { rundown, customFields, onFinished, onCancel } = props;
-
+  const { data } = useRundown();
+  const { id, title } = data;
```
cpvalente: Question: would it be better to de structure this closer to where it is used so that we keep its context?

apps/server/src/services/project-service/ProjectService.ts
```diff
@@ -323,11 +323,15 @@ export async function patchCurrentProject(data: Partial<DatabaseModel>) {
...
+     * so attempt to get the one that was used last
+     * otherwise just pick the first
      */
-    const firstRundown = getFirstRundown(result);
+    const last = await getLastLoaded();
+    const rundownToLoad =
+      last?.rundownId && last.rundownId in result ? result[last.rundownId] : getFirstRundown(result);
 
-    await initRundown(firstRundown, customFields);
+    // console.log('last', rundownToLoad.id, rundownToLoad.title);
```
cpvalente: Can we remove this console log?

apps/server/src/api-data/db/db.controller.ts
```diff
@@ -33,6 +35,7 @@ export async function patchPartialProjectFile(req: Request, res: Response<Databa
     const newData = await projectService.patchCurrentProject(patchDb);
 
     res.status(200).send(newData);
+    sendRefetch(RefetchKey.All);
```
cpvalente: Typically we don't have this kind of logic in controllers as long as we have a service

We also hasn't been handling the case this throws, but it may make sense to be aware and guard against it?

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/ImportReview.tsx
```diff
@@ -29,9 +31,12 @@ export default function ImportReview(props: ImportReviewProps) {
 
   const applyImport = async () => {
     setLoading(true);
+
+    // we need to import on-top of the currently loaded rundown
+    // so the id needs to match
     await importRundown(
       {
-        [rundown.id]: rundown,
+        [id]: { ...rundown, id, title },
```
cpvalente: question: i dont understand why do we need to pass the rundown here

currently we always use the current rundown
the importRundown function has access to the query client. would passing the id not be enough?

alex-Arc: the `rundown` here is the new imported rundown....
we pass in the id of the rundown we want to overwrite
I guess it is a bit challenging, I think that coms from using the very general `db` `patch` route

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/ImportReview.tsx
alex-Arc: not sure I understood, I made and attempt

cpvalente (approved):

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/ImportReview.tsx
cpvalente: I meant that we destructure id and title at the top, but where it is used, it is unclear what the properties refer to (id and tittle of what?) 

so it feels better to not destructure so far from the consumer

PR #1800, opened 2025-10-01
fix: prevent rolling into events of 0 duration

cpvalente:


alex-Arc:
I haven't tested yet
but a side note, the the roll skips through, the 0 duration events, it should still run all the trigger lifecycles

alex-Arc (approved):

PR #1802, opened 2025-10-04
set pnpm `useNodeVersion` property to drive node version in GitHub workflows

jwetzell:
- set `useNodeVersion` in `pnpm-workspace.yaml` 
- change GitHub actions to rely pnpm setting up node with that specified version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,15 +11,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: Use Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: 22
-
       - name: Setup pnpm
         uses: pnpm/action-setup@v3
-        with:
-          version: 10
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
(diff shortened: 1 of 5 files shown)


alex-Arc (commented):
Nice!
I see that we also define the node version in `package.json`
could we either remove that or the `.nvmrc` file

jwetzell:
If `nvm` supported the `engines` property inside the `package.json` I would say yes, but that seems like a hotly debated topic on their repo.

alex-Arc:
ah I see 200+ comments
last idea, could we do something with this? https://pnpm.io/settings#usenodeversion

jwetzell:
I can't seem to get that `useNodeVersion` property to do anything. 

It looks like the pnpm setup action also supports the `packageManager` field so we don't actually need to set that version property for those. This is the only project I use pnpm with so I'm not too familiar with that side of things.

jwetzell:
I must have been doing something wrong, I added the config option and changed the test workflow to not use the setup-node step and pnpm to use the version in `packageManager` field. That seems to have worked just fine.

alex-Arc (approved):

alex-Arc:
maybe a stupid question 
should `@types/node` and the node version be kept in sync?

jwetzell:
I would be inclined to say yes but not sure how they deal with fixing bugs in a particular versions types. Do they release patch release for that versions @types package? I'll do some digging.

alex-Arc:
I found this https://github.com/DefinitelyTyped/DefinitelyTyped?tab=readme-ov-file#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library
seams patch is not related

jwetzell:
Oh, nice. So in theory the `@types/node` version should be set to [22.15.35](https://www.npmjs.com/package/@types/node/v/22.15.35) or whatever the highest patch number under `22.15` is, that kind of makes sense to me.

PR #1805, opened 2025-10-05
lint and typecheck

alex-Arc:


alex-Arc:
@coderabbitai review

.github/workflows/build_cli.yml
```diff
@@ -13,16 +13,28 @@ jobs:
...
+            ${{ runner.os }}-pnpm-store-
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
       - name: Build project packages
         env:
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
-        run: pnpm build
+        run: pnpm build:electron
```
jwetzell: this probably shouldn't be `build:electron`?

alex-Arc: ah good catch :)

apps/client/src/externals.ts
```diff
@@ -17,11 +17,11 @@ export const buyMeACoffeeUrl = 'https://buymeacoffee.com/cpvalente';
 
 // resolve environment
 export const appVersion = version;
-export const isProduction = import.meta.env.MODE === 'production';
-export const isDev = !isProduction;
+export const isDocker = import.meta.env.IS_DOCKER;
+export const isProduction = import.meta.env.PROD;
+export const isDev = import.meta.env.DEV;
```
cpvalente, line 22: I believe this is not equivalent logic, the env can contain other values. before we just said that anything that is not the production build was dev (eg: tests)

Not sure if the difference causes any issues in our logic

apps/client/src/externals.ts
```diff
@@ -17,11 +17,11 @@ export const buyMeACoffeeUrl = 'https://buymeacoffee.com/cpvalente';
 
 // resolve environment
 export const appVersion = version;
-export const isProduction = import.meta.env.MODE === 'production';
-export const isDev = !isProduction;
+export const isDocker = import.meta.env.IS_DOCKER;
```
cpvalente: question: do we no longer need to prepend the env variables with `VITE_`?
https://vite.dev/guide/env-and-mode.html#env-variables

apps/client/src/externals.ts
jwetzell: According to the Vite documentation `import.meta.env.DEV` is always the opposite of `import.meta.env.PROD` which is only true if `NODE_ENV='production'` so I believe that logic is the same.

apps/client/src/externals.ts
jwetzell: I believe this snippet that was added to the client's `vite.config.js` is relevant for this to be working.
```
  define: {
    'import.meta.env.IS_DOCKER': process.env.NODE_ENV === 'docker',
  },
```

cpvalente: a-ha, thank you. that makes sense.
@alex-Arc In that case I think we could do with a small comment in the vite.config.js.

I believe that this creates an indirection which is not particularly explicit and can cause confusion in future

PR #1807, opened 2025-10-05
chore: upgrade electron packages

cpvalente:
there seems to be an issue with old electron versions that cause performance degradation to the system
I dont think this affects the version we use, but in any case it is good to upgrade

https://github.com/electron/electron/issues/48311


alex-Arc (approved):

jwetzell:
It's a bit late but this PR also dropped support for macOS < 12 just FYI as it doesn't seem to be noted anywhere.

PR #1808, opened 2025-10-06
All views follow offset mode

alex-Arc:
some clean-up of unused code
and studio seams to be the last pace not respecting offset mode


cpvalente (approved):
👍

PR #1811, opened 2025-10-07
fix: allow user to reset schedule filter

cpvalente:


alex-Arc (changes requested):
when testing setting `?filter=none` dose not bring the schedule back

PR #1812, opened 2025-10-07
feat: export excel file

cpvalente:


apps/server/src/api-data/excel/excel.utils.ts
```diff
@@ -167,3 +167,175 @@ export function generateImportHandlers(importMap: ImportMap) {
...
+        const value = entry.custom?.[customFieldKey] ?? emptyCellValue;
+        entryRow.push(String(value));
+      }
+    }
+    flatRundown.push(entryRow);
+  }
+
+  // @ts-expect-error -- we safely check if the property is nullish
+  if (previousEntry && previousEntry?.parent != null) {
+    addGroupEnd();
```
alex-Arc, line 275: It took me a bit to find out the reason for this, so maybe a comment
```suggestion
    // add a closing group-end at the end of the rundown if needed
    addGroupEnd();
```

apps/server/src/api-data/excel/excel.utils.ts
```diff
@@ -167,3 +167,175 @@ export function generateImportHandlers(importMap: ImportMap) {
...
+    if (!entry) continue;
+    // we opt for not exporting delays as they are only relevant for runtime
+    if (entry.type === 'delay') continue;
+
+    // we need a group-end if:
+    // - the previous entry was a group and now we have another group (both parents are undefined)
+    // - the previous entry had a parent and the new entry is a group (parent undefined)
+    // - the previous entry had a parent and the new one doesnt (parent null)
+    if (
+      previousEntry &&
```
alex-Arc, line 209: Is it correctly understood that `(previousEntry.type === 'group' && entry.type === 'group')` covers "the previous entry was a group and now we have another group (both parents are undefined)"

and `(previousEntry?.parent !== null && entry?.parent == null)` covers
- the previous entry had a parent and the new entry is a group (parent undefined)
- the previous entry had a parent and the new one doesnt (parent null)

DEVELOPMENT.md
```diff
@@ -21,6 +21,7 @@ Locally, we would need to run both the React client and the node.js server in de
 
 From the project root, run the following commands
 - __Install the project dependencies__ by running `pnpm i`
+- __Create a local build__ by running `pnpm build`, this will populate local dependencies
```
alex-Arc, line 24: maybe we should just put it as par of post install?

apps/server/src/api-data/excel/excel.utils.ts
```diff
@@ -167,3 +167,175 @@ export function generateImportHandlers(importMap: ImportMap) {
...
+ */
+export function rundownToTabular(rundown: Rundown, customFields: CustomFields): string[][] {
+  const flatRundown: string[][] = [];
+  // we reuse the keys from the import map so import <> export works by default
+  const keys = Object.keys(matchedWithDefaultImportMap) as (keyof typeof matchedWithDefaultImportMap)[];
+  const customFieldKeys = Object.keys(customFields) as CustomFieldKey[];
+
+  // create title row
+  if (rundown.title) {
+    flatRundown.push([rundown.title]);
```
alex-Arc: depending on what the main use case is for the excel export is, we should consider not having the title here.
for machine use it just complicates automations  
and we have the title in the worksheet title

cpvalente: Good point. CSV is a typical machine-machine transport, but excel less. 
The reason we are moving from csv to excel, was that our users were using the csv export to get excel files anyway

I dont think the title makes or breaks the feature, do you think we should remove it?

apps/server/src/api-data/excel/excel.utils.ts
cpvalente: yes, that is correct. the `!=` (nullish check) will cover both the `undefined` and `null` case

DEVELOPMENT.md
cpvalente: you mean the populating of files?
I dont have a strong opinion on this but it could be nice for other people who will not read the docs :D

apps/server/src/api-data/excel/excel.utils.ts
alex-Arc: then yes

PR #1813, opened 2025-10-07
Fix build

alex-Arc:


Dockerfile
```diff
@@ -1,12 +1,3 @@
-FROM node:22-bullseye AS builder
-ENV PNPM_HOME="/pnpm"
-ENV PATH="$PNPM_HOME:$PATH"
-RUN npm install -g pnpm@10.11.0
-COPY . /app
-WORKDIR /app
-RUN pnpm --filter=ontime-ui --filter=ontime-server --filter=ontime-utils --filter=ontime-types install --config.dedupe-peer-dependents=false --frozen-lockfile
-RUN pnpm run build:docker
-
 FROM node:22-alpine
```
jwetzell: Changing the first line to the below would allow us to set the node version via docker build args which we could read from either the `nvmrc` or `pnpm-workspace.yaml`
```
ARG NODE_VERSION=22.15.1
FROM node:${NODE_VERSION}-alpine
```

alex-Arc:
I think the docker build works, but needs to be in master to be allowed to publish?

PR #1815, opened 2025-10-09
Small clean-up

alex-Arc:


package.json
```diff
@@ -30,6 +30,7 @@
     "build:resolver": "cross-env turbo run build --filter=@getontime/resolver",
     "dist-win": "turbo run dist-win",
     "dist-mac": "turbo run dist-mac",
+    "dist-mac:local": "turbo run dist-mac:local",
```
cpvalente: If this is to avoid to try signing, the signing step is skipped if there are no signing keys in env

alex-Arc: there is theese to scripts in the electron package
```json
    "dist-mac": "electron-builder --publish=never --mac",
    "dist-mac:local": "electron-builder --publish=never --mac -c.mac.identity=null",
```

is the local one redundant? (can't test as I have no Mac)

Issue #856, opened 2024-03-25
Feat request: Start next matching CUE 


alex-Arc:
starting a event from API with a cue name always starts the first matching event.
Could it maybe start the next matching event

Should we also look into pattern matching on the cue names? 
example: `start SFX.*` would start the next sfx cue even if it was named SFX.1 or SFX.2

cpvalente:
I am not sure about the second part. I personally never programmed a sequence where I would trust such wildcards.

But as for running the next matching cue, sounds like a good improvement to me

cpvalente:
Closing this as completed and released with Ontime v4

PR #1821, opened 2025-10-11
fix next cue

alex-Arc:
Related to #856

### DIFF
--- a/apps/server/src/services/runtime-service/runtime.service.ts
+++ b/apps/server/src/services/runtime-service/runtime.service.ts
@@ -280,7 +280,7 @@ class RuntimeService {
       timedEventOrder,
       cue,
       state.rundown.selectedEventIndex ?? undefined,
-      state.timer.playback === Playback.Armed ? true : false, // If we are armed allow the armed event to be considered for playback
+      state.timer.playback === Playback.Armed, // If we are armed allow the armed event to be considered for playback
     );
 
     if (!event) {


apps/server/src/services/runtime-service/runtime.service.ts
```diff
@@ -273,13 +273,14 @@ class RuntimeService {
...
-    const { playableEventOrder } = getRundownMetadata();
+    const { timedEventOrder } = getRundownMetadata();
 
     const event = findNextPlayableWithCue(
       rundown,
-      playableEventOrder,
+      timedEventOrder,
       cue,
       state.rundown.selectedEventIndex ?? undefined,
+      state.timer.playback === Playback.Armed ? true : false, // If we are armed allow the armed event to be considered for playback
```
cpvalente:
```suggestion
      state.timer.playback === Playback.Armed, // If we are armed allow the armed event to be considered for playback
```

apps/server/src/services/runtime-service/runtime.utils.ts
```diff
@@ -118,24 +118,25 @@ export function findNextPlayableId(playableEventsOrder: EntryId[], currentEventI
...
-  const lowerCaseCue = targetCue.toLowerCase();
+  const startFromIndex = allowCurrent ? currentEventIndex : currentEventIndex + 1;
 
-  for (let i = currentEventIndex; i < playableEventsOrder.length; i++) {
-    const eventId = playableEventsOrder[i];
+  for (let i = startFromIndex; i < timedEventsOrder.length; i++) {
+    const eventId = timedEventsOrder[i];
     const event = rundown.entries[eventId];
-    if (isOntimeEvent(event) && isPlayableEvent(event) && event.cue.toLowerCase() === lowerCaseCue) {
+    if (isOntimeEvent(event) && isPlayableEvent(event) && event.cue === targetCue) {
```
cpvalente, line 131: question: do we no longer need the lowercase comparison?

alex-Arc: I don't think it is our decision to make

cpvalente: Fair enough, I agree

alex-Arc: I don't think we should force it...

PR #1825, opened 2025-10-12
fix: minor ui shift on quick add inline

Shobhit-Nagpal:
## Changes made
- removed height field from css for `<QuickAddInline />` 

Before:

https://github.com/user-attachments/assets/76f8e636-5e3c-43f7-ba56-3344410619f1

After:

https://github.com/user-attachments/assets/49ba6c44-8794-42b3-a48f-e0266c6ec851


Shobhit-Nagpal:
https://github.com/user-attachments/assets/3278f7c6-9757-4fc9-b5d9-1f5d5408d4b6


moved the blue line to a pseudo element to resolve the issue, video of how it works now is above

PR #1826, opened 2025-10-12
feat: timer pip

Shobhit-Nagpal:
## Changes made
- add `documentPictureInPicture` under declaration
- add `PipTimer.tsx` containing `<PipTimerHost />` and `<PipTimer />`
- add `<PipTimerHost />` under `<ViewSettings />`
- add store for pip timer

Working:

https://github.com/user-attachments/assets/bd19025b-6058-4b9d-87f3-a9b71fad8648

### DIFF
--- a/apps/client/src/views/timer/PipTimer.tsx
+++ b/apps/client/src/views/timer/PipTimer.tsx
@@ -1,6 +1,20 @@
+import { memo, useCallback, useEffect } from 'react';
 import { createRoot } from 'react-dom/client';
+import { BrowserRouter } from 'react-router';
+
+import Button from '../../common/components/buttons/Button';
+import { FitText } from '../../common/components/fit-text/FitText';
 import MultiPartProgressBar from '../../common/components/multi-part-progress-bar/MultiPartProgressBar';
+import TitleCard from '../../common/components/title-card/TitleCard';
+import ViewLogo from '../../common/components/view-logo/ViewLogo';
 import { useTimerSocket } from '../../common/hooks/useSocket';
+import { cx } from '../../common/utils/styleUtils';
+import { formatTime } from '../../common/utils/time';
+import SuperscriptTime from '../../features/viewers/common/superscript-time/SuperscriptTime';
+import { getFormattedTimer, getTimerByType } from '../../features/viewers/common/viewUtils';
+import { getTimerColour } from '../../views/utils/presentation.utils';
+
+import { usePipStore } from './pipTimer.store';
 import { useTimerOptions } from './timer.options';
 import {
   getCardData,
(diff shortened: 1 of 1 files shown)


apps/client/src/views/timer/PipTimer.tsx
```diff
@@ -0,0 +1,278 @@
...
+  getTotalTime,
+} from './timer.utils';
+import { TimerData, useTimerData } from './useTimerData';
+
+import './Timer.scss';
+
+export default memo(PipTimerHost);
+function PipTimerHost() {
+  const { data } = useTimerData();
+  const { root, setRoot } = usePipStore();
```
alex-Arc, line 37: Is there are reason for this to be a full zustand store and not just a useState?

Shobhit-Nagpal: yes, i had initially started with a useState (and also with useRef) but the issue was that if i navigate away to another page or section from where the host was rendered, come back to it and then want to make any updates to the settings, it would re-render the host component and lose reference of the previous picture in picture document

so the issue in a scenario was that:
- i go to view settings and pop out the timer
- i navigate to another page like the editor or cuesheet, etc
- i come back to view settings and change the warning color and save it
- it wouldn't update the color in the document picture in picture because on coming back it lost reference to the previous root we had

the zustand store helps keeps that reference even if navigate to other parts of the app

alex-Arc: I see, that makes sense 👍🏽

apps/client/src/views/timer/PipTimer.tsx
```diff
@@ -0,0 +1,278 @@
...
+      </Button>
+    </div>
+  );
+}
+
+interface PipTimerProps {
+  data: TimerData;
+}
+
+function PipTimer({ data }: PipTimerProps) {
```
alex-Arc, line 119: this looks to be almost just the full `Timer` code
what are the blocking factors?

cpvalente: I think from this PR we want to take the PiP process, the contents and where in the UI it goes is something we will decide

I wonder if the pip should look like the preview we have for the timer (with real data) ie: just a timer and secondary field

Shobhit-Nagpal: > what are the blocking factors?

the first attempt was to directly use the `Timer` component but the problem that occurred was majorly with the `ViewParamsEditor`. `ViewParamsEditor` makes an API call to the backend to fetch view settings through react query which required `QueryClientProvider` present. Since the pip window has its own DOM present, the provider there doesn't exist there and neither can be shared from the main app, so it would crash with the error that provider is required. There is also a translation option present in the main app but not in the `PipTimer` because that would also require the `TranslationProvider`

The `Timer` itself is mainly getting its data through props and `useTimerSocket`

PR #1828, opened 2025-10-12
update cli and resolver workflows for new NPM auth

jwetzell:


jwetzell:
no that controls what permission the workflow has for the repo I think you have control over what the default is but I think for this repo it is read/write

Adding it likely wouldn't hurt anything though as far as I know

alex-Arc:
a small side track
should we set the resolver package to be manually trigged, there is rarely any changes?

jwetzell:
@alex-Arc it currently is set to allow manual triggering correct?

alex-Arc:
yes but also automatically on all releases, witch might be excessive..? or maybe it dosen't mater

jwetzell:
Ah, I didn't notice it wasn't scoped to just "resolver" related events. I would either filter the events appropriately (some sort of prefix)  or just remove that and trigger manually.

jwetzell:
I'll leave changing how the resolver workflow is run for a future exercise, but I agree that it should not be tied to a "regular" release like the CLI/Docker stuff

Issue #1829, opened 2025-10-14
Overtime display in Timer view


mithxyz:
in the Timer view when the timer enters overtime it reverts back to the default colour, also when using freeze and custom text the text is always displayed

cpvalente:
Hi @mithxyz , thank you for this

The issue with the freeze message has been reported in #1822 and the fix is waiting for release

I have also found the issue with the overtime colour and will include a fix in the next release

cpvalente:
Hi @mithxyz , this issue has been fixed and released as of 4.0.1. 
Could you please take a look and confirm this for us?

PR #1832, opened 2025-10-17
fix: correct alignment to end message

cpvalente:
Looks like I misunderstood you meant alex. Is this the styling issue you said we had?
Related to #1829


alex-Arc (approved):
perfect!

PR #1833, opened 2025-10-17
fix: add setup-node step back to resolve npm access issues

jwetzell:
Looks like without the `setup-node` step the `npm` path [is not setup correctly](https://github.com/cpvalente/ontime/actions/runs/18590215455)

can do some more reading [here](https://github.com/actions/runner-images/issues/9644)


cpvalente (approved):
Thank you, lets try!

Maybe we can merge and trigger the jobs manually?

PR #1843, opened 2025-10-24
Fix link to host in local network

alex-Arc:


DEVELOPMENT.md
```diff
@@ -24,6 +24,11 @@ From the project root, run the following commands
 - __Create a local build__ by running `pnpm build`, this will populate local dependencies
 - __Run dev mode__ by running `pnpm dev` or `pnpm dev:electron` to get the electron window
 
+### Cloud debugging
+
+The cloud uses a rout prefix to find stages, that is passed in by the docker container on the environment variable `ROUTER_PREFIX`
```
jwetzell: typo here either *route or router?

cpvalente: We likely do not want to document this too much. It doesn’t really mean anything to Ontime and is only relevant to the particular way we use Ontime cloud which users cannot interact with

DEVELOPMENT.md
```diff
@@ -24,7 +24,6 @@ From the project root, run the following commands
 - __Create a local build__ by running `pnpm build`, this will populate local dependencies
 - __Run dev mode__ by running `pnpm dev` or `pnpm dev:electron` to get the electron window
 
-
```
cpvalente, line 27: nitpick: this diff doesnt seem to do anything, could we remove it to cleanup the change?

PR #1845, opened 2025-10-25
add timer type for css override selector

### LINKED ISSUE #1842: Add a data-timer-type attribute to timer view
Often it would be good to be able to treat events with Timer Type "None" differently from other events. Currently there is no way of doing this using custom CSS from the stage timer view. I propose either adding a timer-type attribute to the timer div or adding a "None" option to the timer phase.

alex-Arc:
a way to resolve #1842

### DIFF
--- a/apps/client/src/views/timer/Timer.tsx
+++ b/apps/client/src/views/timer/Timer.tsx
@@ -179,6 +179,7 @@ function Timer({ customFields, projectData, isMirrored, settings, viewSettings }
           <div
             className={cx(['timer', !isPlaying && 'timer--paused', showFinished && 'timer--finished'])}
             style={{ fontSize: `${timerFontSize}vw` }}
+            data-type={viewTimerType}
             data-phase={time.phase}
           >
             {display}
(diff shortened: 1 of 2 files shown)


apps/client/src/views/timer/Timer.tsx
```diff
@@ -177,7 +177,7 @@ function Timer({ customFields, projectData, isMirrored, settings, viewSettings }
           </FitText>
         ) : (
           <div
-            className={cx(['timer', !isPlaying && 'timer--paused', showFinished && 'timer--finished'])}
+            className={cx(['timer', !isPlaying && 'timer--paused', showFinished && 'timer--finished', `timer-type--${viewTimerType}`])}
```
cpvalente: Since we don’t use the class for anything. I think a data attribute may be more useful

On another hand, should we just hide the time if timer type is none? It was in our todo list

alex-Arc: hmm good question
- how do we expose our other stuff for css picking?
- if the timer totally diapered by default would that cause confusion or be expected?

cpvalente: We don’t generally expose things we don’t use. But if we implement the hiding, we would use the attribute 

I think attributes are a better fit here and also more expandable 

if timer type is none, we could fade out the timer. I think that is ok and visually cleaner than the dashes

PR #1846, opened 2025-10-27
Rundown operations

cpvalente:
Adds `rename` and `duplicate` operations to a rundown

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,7 @@
     "jest": true
   },
   "parser": "@typescript-eslint/parser",
-  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier", "eslint-config-prettier"],
+  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier"],
   "plugins": ["@typescript-eslint", "prettier"],
   "overrides": [
     {
--- a/package.json
+++ b/package.json
@@ -46,10 +46,11 @@
     "eslint": "catalog:",
     "eslint-config-prettier": "catalog:",
     "eslint-plugin-playwright": "^1.5.2",
+    "eslint-plugin-prettier": "catalog:",
     "prettier": "catalog:",
+    "rimraf": "catalog:",
     "turbo": "2.5.8",
-    "typescript": "catalog:",
-    "rimraf": "catalog:"
+    "typescript": "catalog:"
   },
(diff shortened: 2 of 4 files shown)


cpvalente:
@coderabbitai full review

apps/client/src/features/app-settings/panel/manage-panel/composite/RundownRenameForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <Panel.Indent as='form' onSubmit={handleSubmit(setupSubmit)} onKeyDown={(event) => preventEscape(event, onCancel)}>
+      <label>
+        <Panel.Description>Rundown title</Panel.Description>
+        {errors.title && <Panel.Error>{errors.title.message}</Panel.Error>}
+        <Input
+          {...register('title', {
+            required: { value: true, message: 'Title is required' },
+            validate: (value) => {
+              if (value.trim().length === 0) return 'Title cannot be empty';
```
alex-Arc, line 56: should we set any more validators here?
maybe just to help ourselves when it has to be translated to a filename for download? otherwise we need a sanitizer on the download function
right now it just fails silently if there are slashes in the name

apps/client/src/features/app-settings/panel/manage-panel/ManageRundowns.tsx
alex-Arc: 'Delete rundown'

apps/client/src/features/app-settings/panel/manage-panel/ManageRundowns.tsx
```diff
@@ -119,6 +165,17 @@ export default function ManageRundowns() {
                             },
                             {
                               type: 'item',
+                              icon: IoDuplicateOutline,
+                              label: 'Duplicate',
+                              onClick: () => {
+                                setActionError(null);
+                                setTargetRundown(id);
```
alex-Arc: `useState` has not updated `targetRundown` by the time `duplicate` is called so the id is stale and refers to the previously used id

apps/client/src/features/app-settings/panel/manage-panel/composite/RundownRenameForm.tsx
cpvalente: good point, I will add a alphanumeric check for now and we can look at widening the net if needed

apps/client/src/features/app-settings/panel/manage-panel/ManageRundowns.tsx
cpvalente: good catch

PR #1847, opened 2025-10-27
feat: package with nix flake

alinarielle:


cpvalente:
Hi @alinarielle, thank you for this.
I was actually just talking about nix yesterday so it was a fun coincidence to see this PR

I am not too familiar with how nix flakes are used and would need your help here
Is the proposal that users can declare Ontime as a dependency in their setups? Or are we trying to publish Ontime to some nix friendly repo?

flake.nix
```diff
@@ -0,0 +1,68 @@
...
+          config,
+          self',
+          inputs',
+          pkgs,
+          system,
+          ...
+        }:
+        {
+          formatter = pkgs.nixfmt-tree;
+          packages.default = pkgs.callPackage (
```
dtomvan, line 29: I'd do `callPackage ./package.nix {}` and put the package in there, because better seperation of concerns and compatibility with non-flake nix (provided you also create a default.nix that also `callPackage`s)

flake.nix
```diff
@@ -0,0 +1,68 @@
...
+
+  inputs = {
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
+  outputs =
+    inputs@{ flake-parts, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      systems = [
```
dtomvan, line 12: You could use https://github.com/nix-systems so the systems can be overidden by consumers of the flake.

.envrc
dtomvan, line 1: (nit) I've seen complaints with envrcs being enforced like this (what if you want to, even slightly, modify the environment). It's probably better to allow people to choose what they want in their direnvs, and actually gitignore it.

(while you're at it, because of nix-direnv it might be smart to also gitignore `result` and `.direnv`)

flake.nix
```diff
@@ -0,0 +1,68 @@
+{
+  description = "Ontime is a browser-based application that manages event rundowns, scheduling, and cueing.";
+
+  inputs = {
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
```
dtomvan, line 6: (nit) *technically* this can just be on tbe `nixpkgs-unstable` channel since we're not instantiating NixOS configs in this flake. Doesn't matter much though as `nixos-unstable` only lags behind a couple days due to it having a bit more CI checks and builds.

dtomvan (commented):
@alinarielle thanks for this contribution! Looking good so far!

@cpvalente A Nix flake makes it easy to install (and with a nixos module in it, configure) some 3rd-party software. The idea is that I declare `github:cpvalente/ontime` as an input to my *own* flake with my NixOS configurations in it, and use the provided package definition to install it on my system.

Copyparty does it similarly. Their documentation has a concrete example on how users can make use of the flake if they so desire: https://github.com/9001/copyparty/blob/e9ab040ce8e72e299a3d8fbd109865b1e218eb57/README.md#nixos-module

We could also try to get this into Nixpkgs to make it even more widely available...

flake.nix
```diff
@@ -0,0 +1,68 @@
...
+          pkgs,
+          system,
+          ...
+        }:
+        {
+          formatter = pkgs.nixfmt-tree;
+          packages.default = pkgs.callPackage (
+            {
+              stdenv,
+              nodejs,
```
dtomvan, line 32: (nit) Might be best to pin `nodejs_22` and `pnpm_10` to match the expected `engines` in the `package.json`...

cpvalente:
Hi @dtomvan 

It seems like you have a good grasp on practices and expected outcomes.
Would you be available to make a proposal with the changes you have suggested?

cpvalente:
Thank you @alinarielle for this submission

I appreciate the effort you took into making the proposal.
However we do not have enough work here to merge as is and seem to have no path forward

I am closing the PR for now, please re-open if you would like to continue the submission

PR #1852, opened 2025-10-30
feat: implement simple healthcheck endpoint

cpvalente:
Implements a simple healtcheck endpoint
I didnt add any info since we would likely want to guard that behind auth

For a endpoint which provides session status we already have `/session/info`


cpvalente:
@coderabbitai full review

jwetzell (approved):

PR #1853, opened 2025-10-30
Feature: Added ?n=1 lockout for /messagecontrol, /timercontrol, /rundown

mdavies1:
Added the same functionality from the views to the editor popouts.

### DIFF
--- a/apps/client/src/features/control/message/MessageControlExport.tsx
+++ b/apps/client/src/features/control/message/MessageControlExport.tsx
@@ -6,7 +6,6 @@ import ViewNavigationMenu from '../../../common/components/navigation-menu/ViewN
 import ProtectRoute from '../../../common/components/protect-route/ProtectRoute';
 import { handleLinks } from '../../../common/utils/linkUtils';
 import { cx } from '../../../common/utils/styleUtils';
-
 import { getIsNavigationLocked } from '../../../externals';
 
 import MessageControl from './MessageControl';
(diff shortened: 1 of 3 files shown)


alex-Arc:
hey @mdavies1 can you please run `pnpm lint`

mdavies1:
I've tried but it doesn't seem to be installed for some reason

alex-Arc:
ok, I've opened a PR in your fork
https://github.com/mdavies1/ontime/pull/1

mdavies1:
Do I need to do anything else?

apps/client/src/features/control/message/MessageControlExport.tsx
```diff
@@ -14,13 +15,14 @@ import style from '../../../views/editor/Editor.module.scss';
...
   const isExtracted = window.location.pathname.includes('/messagecontrol');
+  const isLocked = getIsNavigationLocked();
   const classes = cx([style.content, style.contentColumnLayout]);
 
   return (
     <ProtectRoute permission='editor'>
       <div className={style.messages} data-testid='panel-messages-control'>
         {!isExtracted && <Corner onClick={(event) => handleLinks('messagecontrol', event)} />}
-        {isExtracted && <ViewNavigationMenu suppressSettings />}
+        {(isExtracted || isLocked) && <ViewNavigationMenu suppressSettings={isExtracted} isNavigationLocked={isLocked} />}
```
cpvalente: question: what is the use case for this element to be locked and not extracted? would that not mean we need to have locked the editor which is not possible?

 I wonder if we can simply say (same for all the others)


```suggestion
        {isExtracted && <ViewNavigationMenu suppressSettings isNavigationLocked={isLocked} />}
```

or even inline the check for `isLocked` from the component body to avoid computing it if we are not extracted, like:

```suggestion
        {isExtracted && <ViewNavigationMenu suppressSettings isNavigationLocked={getIsNavigationLocked()} />}
```

mdavies1: I agree that would be better. I can't think of a use case as to when an editor component would be locked and not extracted? In fact I think that would be impossible with the way it is currently written. And I suppose that would also apply for the other editor controls

cpvalente: nice! just re-request review when you have done the changes and we can get this in the next release

mdavies1: Changes made :)

PR #1855, opened 2025-11-01
refactor: time input labels on hover

Shobhit-Nagpal:
## Changes
- added state to check hover on time input row
- add prop on `<TimeInputFlow />` for showing label on hover
- add styling for hover labels

## Working

https://github.com/user-attachments/assets/e273f266-2e29-4449-b020-3daa316d1dcc

### DIFF
--- a/apps/client/src/features/rundown/rundown-event/RundownEventInner.tsx
+++ b/apps/client/src/features/rundown/rundown-event/RundownEventInner.tsx
@@ -10,10 +10,12 @@ import {
   IoTime,
 } from 'react-icons/io5';
 import { LuArrowDownToLine } from 'react-icons/lu';
+import { useSessionStorage } from '@mantine/hooks';
 import { EndAction, Playback, TimerType, TimeStrategy } from 'ontime-types';
 
 import Tooltip from '../../../common/components/tooltip/Tooltip';
 import { cx } from '../../../common/utils/styleUtils';
+import { AppMode, sessionKeys } from '../../../ontimeConfig';
 import TitleEditor from '../common/TitleEditor';
 import TimeInputFlow from '../time-input-flow/TimeInputFlow';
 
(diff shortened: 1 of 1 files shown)


apps/client/src/features/rundown/time-input-flow/TimeInputFlow.module.scss
```diff
@@ -15,3 +15,24 @@
...
+  left: 0;
+  opacity: 0;
+  transition: opacity 0.15s ease;
+  pointer-events: none;
+  z-index: $zindex-floating;
+
+  &.visible {
+    opacity: 1;
+  }
+}
```
alex-Arc, line 40: I think the tooltips are a bit distracting when scrolling, a small delay before showing could help that.
(and our default border-radius could look nice)

```suggestion
.hoverLabel {
  position: absolute;
  top: -1.5rem;
  background-color: $ui-white;
  padding: 0.125rem 0.5rem;
  color: $ui-black;
  left: 0;
  opacity: 0;
  border-radius: $component-border-radius-md;
  transition: opacity 0.15s ease;  // no delay on fade out
  pointer-events: none;
  z-index: $zindex-floating;
  
  &.visible {
    opacity: 1;
    transition: opacity 0.15s 0.5s ease;  // small delay before fade in
  }
}
```

Shobhit-Nagpal: have added this, the delay does look better. thank you

Shobhit-Nagpal:
Added 2 changes.

- shifted hover state to css
- added border and fade in delay

## Working

https://github.com/user-attachments/assets/4a386c09-d1ed-4808-b543-7d8b4a45c892

alex-Arc (approved):
I think the functionality is good 
Will leave it to @cpvalente to comment on the implementation

cpvalente (approved):
Good work!

I wonder if it would be a better experience if we disabled this in run mode. what do you think?
Either way, happy to merge this as is

Shobhit-Nagpal:
tried it out and i like it, makes sense to not let the labels disturb the user on run mode

here's a working of it:

https://github.com/user-attachments/assets/0ac64439-53f8-4772-b0ad-f12019e626a1

PR #1861, opened 2025-11-03
Instant for internal calculations

alex-Arc:
- creates the branded type `Instant`
- add instant as new property to `runtimeState` and drive clock from that

### DIFF
--- a/apps/server/src/utils/__tests__/temporal.test.ts
+++ b/apps/server/src/utils/__tests__/temporal.test.ts
@@ -58,7 +58,7 @@ describe('epoch utilities', () => {
   });
 
   test('before DST', () => {
-    vi.setSystemTime('2025-03-30T00:58:18Z'); // right before DST
+    vi.setSystemTime('2025-03-30T00:58:18Z');
     const instant = getEpoch();
     const clock = timeNow();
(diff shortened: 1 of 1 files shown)


apps/server/src/utils/temporal.ts
```diff
@@ -0,0 +1,40 @@
+import { Maybe } from 'ontime-types';
+import { dayInMs, MILLIS_PER_MINUTE } from 'ontime-utils';
+
+export type Instant = number & { __brand: 'instant' };
```
cpvalente: suggestion: i know we used the name instant in our chats. But instants are constructs which carry some baggage we dont want

I would suggest we stick to ontime names, how about
- EpochMs: eg. `startedAt`
- DayMs: eg. `eventNow.timeStart`
- Duration: eg. `elapsed`

cpvalente:
@coderabbitai review

cpvalente:
Hey @alex-Arc , the code looks fine to me, the rabbit made some good observations
But for maintenance sake, I dont think all the casting is positive. I think we need to create a set of abstractions for handling maths with our different types of times

lets chat!

alex-Arc:
@coderabbitai review

alex-Arc:
moved here #1974

PR #1865, opened 2025-11-08
improve 12 hour formatting

cpvalente:
This PR focuses on the timer display in 12 hour mode
- 12 hour mode should be supported in all views apart from editor
- There should be consistency in all views about how we show the time


cpvalente:
> can we change the default with Image

If I understand correctly, you are suggesting we use the superscript here.
If we change this, the text would reflow when the user selects the input which I think is confusing

> should we use the superscript for the pm/am indicator in the studio view

I thought it looked ok and we dont have space limitations, but can change if you want

PR #1866, opened 2025-11-08
Fix integration api

### LINKED ISSUE #1864: [Bug] Changing Custom Fields using WS only works if using lower case names
If you create two custom fields, one using uppercase in any way and the other using only lower case, only the lower case can be changed using WebSocket.

This works as expected in v3 so there seems to have been some change in the v4 implementation.

Windows 10
OnTime 4.1.0

alex-Arc:
fixes #1864 
and allows changing flag and count to end from api

### DIFF
--- a/apps/server/src/api-integration/integration.controller.ts
+++ b/apps/server/src/api-integration/integration.controller.ts
@@ -24,6 +24,7 @@ import { throttle } from '../utils/throttle.js';
 import { coerceEnum } from '../utils/coerceType.js';
 import { editEntry } from '../api-data/rundown/rundown.service.js';
 import { willCauseRegeneration } from '../api-data/rundown/rundown.utils.js';
+import { getCurrentRundown } from '../api-data/rundown/rundown.dao.js';
 
 const throttledEditEvent = throttle(editEntry, 20);
 let lastRequest: Date | null = null;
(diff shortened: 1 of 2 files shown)


apps/server/src/api-integration/integration.controller.ts
```diff
@@ -63,32 +70,76 @@ const actionHandlers: Record<ApiActionTag, ActionHandler> = {
...
-      const key = Object.keys(newObjectProperty)[0] as keyof OntimeEvent;
-      shouldThrottle = shouldThrottle || willCauseRegeneration(key);
-      if (patchEvent.custom && newObjectProperty.custom) {
-        Object.assign(patchEvent.custom, newObjectProperty.custom);
-      } else {
-        Object.assign(patchEvent, newObjectProperty);
-      }
-    });
+    if (isOntimeEvent(targetEntry)) {
+      assertIsEvent(patchEntry);
```
cpvalente: question: would these assertions not fail since we dont have the type property?

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -63,32 +70,76 @@ const actionHandlers: Record<ApiActionTag, ActionHandler> = {
...
-      const newObjectProperty = parseProperty(property, value);
-      const key = Object.keys(newObjectProperty)[0] as keyof OntimeEvent;
-      shouldThrottle = shouldThrottle || willCauseRegeneration(key);
-      if (patchEvent.custom && newObjectProperty.custom) {
-        Object.assign(patchEvent.custom, newObjectProperty.custom);
-      } else {
-        Object.assign(patchEvent, newObjectProperty);
-      }
-    });
+    if (isOntimeEvent(targetEntry)) {
```
cpvalente: suggestion: I think we want to avoid adding this complexity
can we instead say that we allow the mutation
if the property is whitelisted
if the property exists in the target entry

this should mean we can make things with less checks and more generic. ie: adding new types or fields doesnt mean changes here

apps/server/src/api-integration/integration.utils.ts
```diff
@@ -41,19 +43,71 @@ const propertyConversion = {
   timeEnd: (value: unknown) => clampDuration(coerceNumber(value)),
 };
 
-export function parseProperty(property: string, value: unknown) {
+export function parseEventProperty(property: string, value: unknown) {
```
cpvalente: question: I didnt understand the need to have different maps per entry type. The same property should have the same type in all entries, is that not right?

alex-Arc: there are not timers in milestone and group
and the group has the targetDuration

apps/server/src/api-integration/integration.utils.ts
```diff
@@ -36,14 +38,15 @@ const propertyConversion = {
...
   timeStrategy: (value: unknown) => coerceEnum<TimeStrategy>(value, TimeStrategy),
 
+  targetDuration: (value: unknown) => clampDuration(coerceNumber(value)), // only exist in the group
   duration: (value: unknown) => clampDuration(coerceNumber(value)),
   timeStart: (value: unknown) => clampDuration(coerceNumber(value)),
   timeEnd: (value: unknown) => clampDuration(coerceNumber(value)),
 };
 
-export function parseProperty(property: string, value: unknown) {
+export function parseEventProperty(property: string, value: unknown) {
```
cpvalente: Since this doesnt relate to OntimeEvents, how about leaving it as it was or

```suggestion
export function parseEntryProperty(property: string, value: unknown) {
```

cpvalente (approved):
thank you, good stuff

PR #1877, opened 2025-11-16
feat(projects): allow sorting project table

cpvalente:


alex-Arc (approved):
Good
It would be ok with me to not force the currently loaded to be one top

PR #1878, opened 2025-11-16
refactor(operator): hide groups where all events have passed

Shobhit-Nagpal:
## Changes made
- added a check under `<Operator />` where if all events are passed within a group, then not to render the group, provided the settings has been checked for hiding past events

Note: I was wondering if it would make sense to add an option under settings to hide groups or not separately. Not sure if it's a good idea to include it since if I opt in to not see the past events, I would not want to see that group either

## Working

https://github.com/user-attachments/assets/0af8df08-feec-4334-98da-22b8a41e4aae


Shobhit-Nagpal:
Simplified the logic to check the group status against rundown metadata instead of looping over each nested entry. Thanks @cpvalente

PR #1886, opened 2025-11-21
refactor(timer): allow multiline secondary text

cpvalente:
Following some user suggestions, instead of shrinking the secondary message we allow it to overflow into multiple lines shrinking as a second step

Also applied the same changes for the PIP preview

Screenshot 2025-11-21 at 16 48 11


alex-Arc:
should we consider making it a multiline text field or/and allow users to send `\n` to manually force new lines?

cpvalente:
> should we consider making it a multiline text field or/and allow users to send `\n` to manually force new lines?

we dont want to escalate the amount of changes here for now.
If multiline text becomes a priority requested by user we can deal with it

PR #1888, opened 2025-11-23
Validate api change custom

### LINKED ISSUE #1887: [Bug] Custom field can not be set using websocket or http api
OnTime v4.2.0
Windows 10

In v4.1.0 there was a limitation to only use lower case custom field names. There was a fix (#1866) but it seems to have changed something unintentionally.

When sending the exact same type of change request that worked in 4.1.0 in 4.2.0, the response is ``Error: Invalid property or value``
This results in no longer being able to set custom field data using WebSocket or httpGET.

alex-Arc:
fix #1887 by correctly testing the nested path of custom values
and also disallows prototype build-in properties from passing

### DIFF
--- a/apps/server/src/api-integration/__test__/integration.utils.test.ts
+++ b/apps/server/src/api-integration/__test__/integration.utils.test.ts
@@ -1,7 +1,7 @@
 import { EntryCustomFields, OntimeEvent } from 'ontime-types';
 import { isValidChangeProperty } from '../integration.utils.js';
 
-describe('isValidateChangeProperty()', () => {
+describe('isValidChangeProperty()', () => {
   test('correct value and property', () => {
     const testEvent = {
       id: 'test',
@@ -53,7 +53,7 @@ describe('isValidateChangeProperty()', () => {
     expect(isValidChangeProperty(testEvent, 'custom:test', 123)).toBeTruthy();
   });
 
-  test('missing existing custom follower', () => {
+  test('empty custom definition', () => {
     const testEvent = {
       id: 'test',
       duration: 111,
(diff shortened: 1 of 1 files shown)


cpvalente:
@coderabbitai review

alex-Arc:
@coderabbitai review

apps/server/src/api-integration/integration.utils.ts
```diff
@@ -60,3 +60,14 @@ export function parseProperty(property: string, value: unknown) {
...
+
+export function isValidChangeProperty(target: OntimeEntry, property: string, value: unknown): boolean {
+  if (typeof property !== 'string') return false;
+  if (value === undefined) return false;
+  if (property.startsWith('custom:') && 'custom' in target) {
+    const customProperty = property.slice('custom:'.length);
+    if (!customProperty) return false;
+    return customProperty in target.custom;
+  }
+  return property in target;
```
cpvalente: I think the rabbit has a point, it should be trivial enough to check if the property belongs to the OntimeEntry like:

```suggestion
  return Object.hasOwn(target, property);
```

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -77,7 +77,7 @@ const actionHandlers: Record<ApiActionTag, ActionHandler> = {
     let shouldThrottle = false;
 
     Object.entries(data).forEach(([property, value]) => {
-      if (typeof property !== 'string' || value === undefined || !(property in targetEntry)) {
+      if (!isValidChangeProperty(targetEntry, property, value)) {
         throw new Error('Invalid property or value');
       }
       // parseProperty is async because of the data lock
```
cpvalente: `parseProperty` is no longer async, can we remove this comment?

PR #1891, opened 2025-11-26
fix: custom fields not persisted

### LINKED ISSUE #1890: Custom fields not storing correctly after Google Sheets import
Hi! I noticed some inconsistencies with Google Sheets import and custom fields. Seems like newly generated fields aren't stored correctly, because they are not usable by OSC API calls {{eventNow.custom.xxx}}, and they vanish after rebooting the software.

Steps to reproduce:
1. Create new project
2. Import data from Google Sheets .xlsx with custom columns
3. Newly imported fields are visible at editor and Settings --> Custom fields list, but are not usable by API calls
4. If server software is rebooted, custom fields are gone

I got them to stick with after import creating new additional custom field by hand (Settings --> Custom fields --> New +). Right after that I was able to use them via OSC calls as well.

I'm using v4.2.0 on Windows

alex-Arc:
fixes #1890

### DIFF
--- a/apps/server/src/services/project-service/ProjectService.ts
+++ b/apps/server/src/services/project-service/ProjectService.ts
@@ -316,10 +316,17 @@ export async function patchCurrentProject(data: Partial<DatabaseModel>) {
   // we can pass some stuff straight to the data provider
   await getDataProvider().mergeIntoData(rest);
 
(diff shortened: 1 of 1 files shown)


cpvalente:
@coderabbitai review

cpvalente (approved):
Thank you alex

Issue #1901, opened 2025-12-02
Add support for desktop Raspberry PI OS


cpvalente:
We would like to support running Ontime in a Desktop Raspberry Pi OS environment.

alex-Arc:
A good reference could be how companion handles it 
https://github.com/bitfocus/companion-pi

alex-Arc:
A good reference could be how companion handles it 
https://github.com/bitfocus/companion-pi

jwetzell:
Pretty sure companion-pi is headless which doesn't help here.

alex-Arc:
added with #1902 and released in [v4.3.0](https://github.com/cpvalente/ontime/releases/tag/v4.3.0)

PR #1902, opened 2025-12-03
add arm architectures to linux target

jwetzell:
- builds 3 AppImage for linux now (amd64, arm64, armv7)
- the website would need to be changed to point to the right file now with the architecture suffix
- I have not tested these yet but there isn't anything using node-gyp or anything that would need to be architecture specific so should in theory work

This should allow running the desktop app on a Raspberry Pi for example as mentioned in #1901


jwetzell:
I tested this on a Linux ARM64 machine and it worked just fine.

alex-Arc (approved):
Tested working on rpi 4 and Raspberry Pi OS (64-bit)

PR #1915, opened 2025-12-11
handle unchecked key access

cpvalente:


apps/server/src/api-data/rundown/rundown.dao.ts
```diff
@@ -213,6 +213,9 @@ function add(rundown: Rundown, entry: OntimeEntry, afterId: EntryId | null, pare
  */
 function edit(rundown: Rundown, patch: PatchWithId): { entry: OntimeEntry; didInvalidate: boolean } {
   const entry = rundown.entries[patch.id];
+  if (!entry) {
+    throw new Error(`Entry with id ${patch.id} not found`);
```
alex-Arc, line 217: do all edits end in a try block ?

apps/server/src/services/project-service/ProjectService.ts
```diff
@@ -94,6 +94,10 @@ async function loadProject(projectData: DatabaseModel, fileName: string, rundown
       ? projectData.rundowns[rundownId]
       : getFirstRundown(projectData.rundowns);
 
+  if (!rundown) {
+    throw new Error('No rundown found in project');
```
alex-Arc: Don't think this is needed either the `projectData.rundowns[rundownId]` will have a good result or we get the first rundown
and the `getFirstRundown` throws by it self

apps/server/src/api-data/rundown/rundown.dao.ts
cpvalente: yes! seems that way

cpvalente:
@coderabbitai review

PR #1918, opened 2025-12-17
Add TTS feature for custom fields

mko1989:
- Add per-custom-field TTS configuration with threshold, voice, and language settings
- Implement TTS hook that reads time values (hh:mm:ss or mm:ss) from custom fields
- Add voice selection UI in custom field settings with language dropdown
- TTS only active in cuesheet view, speaks numerical seconds when below threshold

### DIFF
--- a/apps/client/src/common/hooks/useCustomFieldTTS.ts
+++ b/apps/client/src/common/hooks/useCustomFieldTTS.ts
@@ -1,0 +1,218 @@
+import { useEffect, useRef } from 'react';
+
+import { useFlatRundownWithMetadata } from '../hooks-query/useRundown';
+import useCustomFields from '../hooks-query/useCustomFields';
+
+/**
+ * Parses time string in hh:mm:ss or mm:ss format to seconds
+ * @param timeStr - Time string to parse
+ * @returns Number of seconds, or null if invalid format
+ */
+function parseTimeToSeconds(timeStr: string): number | null {
+  if (!timeStr || typeof timeStr !== 'string') {
+    return null;
+  }
+
+  // Remove whitespace
+  const trimmed = timeStr.trim();
+
+  // Match hh:mm:ss or mm:ss format
+  const timePattern = /^(\d{1,2}):(\d{2}):(\d{2})$|^(\d{1,2}):(\d{2})$/;
+  const match = trimmed.match(timePattern);
+
+  if (!match) {
+    return null;
+  }
+
+  // hh:mm:ss format
+  if (match[1] !== undefined) {
+    const hours = parseInt(match[1], 10);
+    const minutes = parseInt(match[2], 10);
+    const seconds = parseInt(match[3], 10);
+    return hours * 3600 + minutes * 60 + seconds;
+  }
+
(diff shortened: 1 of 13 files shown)


cpvalente:
Hi @mko1989 , thank you for this

I am not entirely sure this is a feature that belongs in Ontime.
In the past we have been opinionated that Ontime should not produce any sounds. 

I will make sure we review this on our next chat, but meanwhile, can you give me some context as to when this feature is useful? what environment would you use it on and what problem does it solve?

Out of curiosity: which LLM did you use to make the changes? Did you find that the LLM managed to find things well enough with the existing code documentation practice?

mko1989:
Its fully your decision.

The use case for the feature is inputting the audio from ontime into comms for the whole av team so the whole team can hear the countdown to the end of the video or audio cue, letting them prepare accordingly to what they need to do.
I'm a media server op and screen manager and I also run ontime, which makes it pretty hard to also do the multiple countdowns through out the event.

I'm using cursor with Gemini 3, composer 1 or set on auto. It didn't have any problems to find exactly what it needed in the code.
There was a lot of back and forth with deployment on a pi, but finally it managed it also.

cpvalente:
I see.

I have a similar experience where there is not often an ontime person / show runner, but someone who takes on the extra work

This means that we should do our best to make sure this person is productive while using Ontime

However, I believe that this change set is more useful as a custom application.
Custom apps are not very used (as far as I know) but a very powerful feature of Ontime.

we could wrap this in a simple HTML + Javascript app which has minimal dependencies in react
this interface would just need to expose the current data from the websocket and allow the user to click "read aloud" on the fields they want

The features used are all browser features so this could be a very lean piece of code.
This would isolate maintenance and allows other users to expand and they need or want.

What do you think?

mko1989:
Yeah, super awesome! 
I didn't know ontime has that built-in.

I need to read up on that.

cpvalente:
Nice, check out the docs here https://docs.getontime.no/features/custom-views/

if you do get around to build something let us know, we could create a repository with links to users custom extensions

Also feel free to drop by our discord if you need some help

mko1989:
ive put something togheter that you can find here: https://github.com/mko1989/ontime-custom-tts
but during testing i saw that there are inconsistencies for reciving data from companion. Sometimes it just works and sometimes it gives WS IN: Error: Invalid property or value 
I was testing on official releases of both ontime and companion module

cpvalente:
thank you!

Do you mind adding some screenshots or a gif of the interface working? I think it could help users figure out what this does
Asides from it, which data are you getting from companion?

you shouldnt need anything from companion here.
if there are issues with the module you can [reach out here](https://github.com/bitfocus/companion-module-getontime-ontime/)

mko1989:
Sure I'll add some screenshots.

I'm getting video remaining time data or audio cue remaining time data from companion. 
During testing it worked fine at first and then it stopped working at all. I mean data coming from companion to ontime.

cpvalente:
Will close this for now. Feel free to reach out if you need help with the custom views

Look forward to it

PR #1919, opened 2025-12-18
refactor: handle prefix on login

cpvalente:
This PR solves a few issues for users running Ontime behind a password
- post-login redirect should account for prefix
- scope cookies to each path by prefix

@alex-Arc I have tested this locally with Docker and it seems fine, but it would be good to have a second pair of eyes
Reach out if you have time to look together

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ override.css
 
 # working stuff
 **/TODO.md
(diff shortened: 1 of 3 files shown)


cpvalente:
@coderabbitai review

PR #1923, opened 2025-12-20
Fix quick start

cpvalente:
Fixes a bug introduced with the new merging logic for projects which prevents the quick start modal from working since it provides a partial payload while the logic expected a full project.

This bug has not been released yet


cpvalente:
@coderabbitai review

PR #1929, opened 2025-12-30
Fix/debug log

alex-Arc:
- fix an issue with crash report on windows
- create a logger dump function and add log to crash report

### DIFF
--- a/apps/server/src/classes/Logger.ts
+++ b/apps/server/src/classes/Logger.ts
@@ -53,10 +53,11 @@ class Logger {
       }
     }
 
+    this.addToQueue(log);
     try {
       socket.sendAsJson(MessageTag.Log, log);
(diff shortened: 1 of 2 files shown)


apps/server/src/app.ts
```diff
@@ -95,6 +95,15 @@ app.use(`${prefix}/login`, loginRouter); // router for login flow
 app.use(`${prefix}/data`, authenticate, appRouter); // router for application data
 app.use(`${prefix}/api`, authenticate, integrationRouter); // router for integrations
 
+// a dev only rout for forcing a crash
```
cpvalente:
```suggestion
// a dev only route for forcing a crash
```

cpvalente (approved):
change looks good, could we remove the crash endpoint?

PR #1933, opened 2026-01-02
Fix(automation): Handles undefined values in conditions

### LINKED ISSUE #1932: [Bug] Automation filter 'not_equal' doesn't give the expected result when testing empty_value.
## Current Behavior
When a filter setup on a field with matching condition `not equals` and value `<empty/no value>`, when the datasource doesn't exist, the filter return `true`

## Expected Behavior
When a filter setup on a field with matching condition `not equals` and value `<empty/no value>`, when the datasource doesn't exist, the filter should return `false`

## Step to reproduce
1. Setup a OSC Server to receive message (ex: Protokol).
2. Prepare a Rundown with a string customfield (ex: test).
3. Create 2 events:

    - In event 1, fill the customfield with the value `"MyField"`.
    - In event 2, make sure the customfield is empty

4. Ensure that `Enable automations` is set to true in the `Automation settings`
5. Create an automation:
    - Where all filters need to pass.
    - With a filter on the create customfield where is not equals to an `<empty/no value>`
    - With an output of type `OSC`:
        - Make sure fields `Target IP`  and `Target Port` point to the OSC Server setup at point 1.
        - Put a valid value in fields `Address`
        - Put the customfield in the field `Arguments` (ex:  `{{eventNow.custom.test}}`)
6. Play event 1 and notice on your OSC Server that:
    - The automation have triggered
    - The argument value is `"MyField"`
7. Play event 2 and notice on you OSC Server that:
    - The automation have triggered
    - The argument value is `"{{eventNow.custom.test}}"`

## Unit test Suggestion
```javascript
// apps\server\src\api-data\automation\__tests__\automation.service.test.ts
    it('should check if a value does exist', () => {
      const mockStore = makeRuntimeStateData({ eventNow: null });
      expect(
        testConditions([{ field: 'eventNow.title', operator: 'not_equals', value: '' }], 'all', mockStore),
      ).toBe(false);
    });
```

DelBiss:
Addresses an issue where conditions with the 'not_equals' operator incorrectly evaluated undefined values. This change ensures that empty string comparisons correctly identify missing values in automation rules.

Fix #1932


apps/server/src/api-data/automation/automation.service.ts
```diff
@@ -106,6 +106,11 @@ export function testConditions(
         }
         return fieldValue == value;
       case 'not_equals':
+        // overload the edge case where we use empty string to check if a value does not exist
+        if (value === '' && fieldValue === undefined) {
```
DelBiss: Maybe it would be better to just do the inverse of equal?

DelBiss: I've implemented it in the 2nd commit.

apps/server/src/api-data/automation/__tests__/automation.service.test.ts
```diff
@@ -101,116 +102,207 @@ describe('testConditions()', () => {
...
+      expect(runTestCondition('number', 10, 'equals', '5')).toBe(false);
+      expect(runTestCondition('number', 10, 'equals', '1')).toBe(false);
+      // String / Substring
+      expect(runTestCondition('string', 'testing-lighting-10', 'equals', 'lighting')).toBe(false);
+      expect(runTestCondition('string', 'testing-lighting-10', 'equals', 'sound')).toBe(false);
+      // Null / Empty / No value
+      expect(runTestCondition('null/undefined', null, 'equals', 'not-empty')).toBe(false);
+      expect(runTestCondition('boolean', false, 'equals', '')).toBe(false);
+
+      expect(runTestCondition('number', 0, 'equals', '')).toBe(true); // TO_DO: is this the desired behavior?
```
DelBiss: This case is probably wrong, especially if `false != <empty/no value>`

apps/server/src/api-data/automation/__tests__/automation.service.test.ts
```diff
@@ -101,116 +102,207 @@ describe('testConditions()', () => {
...
-      expect(result2).toBe(false);
+    it('should match with equals string', () => {
+      expect(runTestCondition('string', '', 'contains', '')).toBe(true);
+      expect(runTestCondition('string', 'Title', 'contains', 'Title')).toBe(true);
+    });
+
+    it('should handle case sensitivity', () => {
+      // TO_DO: is this the desired behavior?
+      expect(runTestCondition('string', 'Title', 'contains', 'title')).toBe(false);
+    });
```
DelBiss, line 1501: `contains` only work on string, yet, it's case sensitive, contrary to `equals`.

apps/server/src/api-data/automation/__tests__/automation.service.test.ts
```diff
@@ -101,116 +102,207 @@ describe('testConditions()', () => {
...
+
+    it('should handle non-string equals values', () => {
+      //TO_DO: is this the desired behavior? All the fields contain the value when converted to string.
+      expect(runTestCondition('number', 10, 'contains', '10')).toBe(false);
+      expect(runTestCondition('boolean', true, 'contains', 'true')).toBe(false);
+      expect(runTestCondition('boolean', false, 'contains', 'false')).toBe(false);
+    });
+
```
DelBiss, line 1551: This is a case that could go both way, especially if, on the client side, operation depends of field type.

But, my first instinct was that other type should match if theirs string representation match.
But again, that would be weird for boolean value that would alway contain a `e`.

apps/server/src/api-data/automation/automation.service.ts
```diff
@@ -106,7 +106,8 @@ export function testConditions(
         }
         return fieldValue == value;
       case 'not_equals':
-        return fieldValue != value;
+        return !evaluateCondition({ field, operator: 'equals', value });
```
cpvalente, line 109: I believe the previous code attempted a loose comparison to allow matching `"12"` to `12`.
This is to simplify automations

However, it has the side effects of matching all the falsy values `0`, `undefined`, `null`, `""`

how about creating a new helper `isEquivalentValue` which resolves this
- given two values of the same type, make a strict comparison `===`
- given two values of different types:
- a) if the types are string a number, make a loose comparison `==`
- b) for all other cases return false

This could also help with the logic in the `equals` condition

apps/server/src/api-data/automation/__tests__/automation.service.test.ts
cpvalente: this is a good observation, contains and equals should have similar implementations.
I would suggest that `contains` and `not_contains` should also be case insensitive. what do you think?

apps/server/src/api-data/automation/__tests__/automation.service.test.ts
cpvalente: In the current implementation, contains only deals with string types. I think this is the correct way to do it

semantically `boolean contains "rue"` doesnt make sense

apps/server/src/api-data/automation/__tests__/automation.service.test.ts
cpvalente: yes, this seems incorrect. I believe my suggestion for comparing values could improve this

DelBiss:
I've removed the abstraction in the last commit.

But the Idea behind the abstraction was to have a better view of all the test and expected results.

Compare how those file look (look at line 104):
* [With abstraction](https://github.com/cpvalente/ontime/blob/917f5deaa09b4d1d0aa45f7e6831d1a8ef7be4f3/apps/server/src/api-data/automation/__tests__/automation.service.test.ts#L104)
* [Without abstraction](https://github.com/cpvalente/ontime/blob/47629c2fca71b3bfe8b14d362b380cd34fa3f808/apps/server/src/api-data/automation/__tests__/automation.service.test.ts#L103)

At first, I wanted even more abstraction with a matrix to generate all the test, but settle on this middle ground.

I will come back to this after my day job....

PR #1936, opened 2026-01-06
css tweaks

cpvalente:
fixes the issue mentioned on discord where the safari status bar follows the body colour (in our case, unset)
by setting the background colour to the bg variable

Also noticed that safari does not hide the image on error like chrome, so made a small change to do so

### DIFF
--- a/apps/client/src/common/components/view-logo/ViewLogo.tsx
+++ b/apps/client/src/common/components/view-logo/ViewLogo.tsx
@@ -1,3 +1,5 @@
+import { useRef } from 'react';
+
 import { projectLogoPath } from '../../api/constants';
 
 import './ViewLogo.scss';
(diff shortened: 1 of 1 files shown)


cpvalente:
@coderabbitai review

alex-Arc (commented):
We just need to make sure it doesn't block the ability to make an transparent background in the timer view, but forom the looks of it it shouldn't?

apps/client/src/views/ViewLoader.tsx
```diff
@@ -15,9 +15,18 @@ export default function ViewLoader({ children }: PropsWithChildren) {
...
+        <Loader />
+      </>
+    );
   }
 
-  // eslint-disable-next-line react/jsx-no-useless-fragment -- ensuring JSX return
-  return <>{children}</>;
+  return (
+    <>
+      <style>{` body { background: var(--background-color-override, #101010); } `}</style>
```
alex-Arc: this line here prevents transparency, we should either remove this line
or add `!important` to https://github.com/cpvalente/ontime/blob/8f6a749499d1bbd968bfd925639830550e9f88ff/apps/client/src/views/timer/Timer.scss#L12

PR #1938, opened 2026-01-10
fix: preserver utf8 charset in filenames

alex-Arc:
allows characters like `æ ø å` to be used in filenames
also axios hadles the Content-Type automatically for multipart/form-data encoding

Watch for changes here https://github.com/expressjs/multer/pull/1210

### DIFF
--- a/apps/client/src/common/api/db.ts
+++ b/apps/client/src/common/api/db.ts
@@ -32,15 +32,16 @@ export async function downloadProject(fileName: string) {
 
 /**
  * HTTP request to upload project file
+ *
+ * Note: The browser's FormData API automatically encodes filenames according to
+ * RFC 2231/RFC 5987 standards with UTF-8 charset. The issue is server-side where
+ * Busboy (used by Multer) defaults to 'latin1' charset. The server-side fix in
+ * upload.ts handles the encoding conversion.
  */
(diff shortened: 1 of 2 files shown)


alex-Arc:
not quite
as far as I can see the default formatting for forms is [UTF-8](https://xhr.spec.whatwg.org/#interface-formdata)
and for some reason [busboy](https://github.com/mscdex/busboy/blob/d08cc9cc5572a29354480f6ec829a042c68d8bac/README.md#L145) defaults to latin1
and the express multer dose not expose that setting so it will always use latin1

The annoying thing is that we do allow a project title to include those letters, and if we download that project it has those correct letters, but then when it is uploaded again it gets mangled

cpvalente:
> The annoying thing is that we do allow a project title to include those letters, and if we download that project it has those correct letters, but then when it is uploaded again it gets mangled

Makes sense, but the point still stands. Would we not be fighting the tooling and potentially pushing the problem further into the code?

it sounds like the best thing would to not allow projects with those characters. I am specifically thinking of non Carolinian alphabets like Chinese which could become problematic in a way we just can’t handle it

alex-Arc:
fixed up-stream, waiting for release

PR #1943, opened 2026-01-14
Fix: automation

alex-Arc:
continues the work from @DelBiss in pr #1933 

decisions that we need to make:
- should all comparisons be lower cased?
- "some value" `contains` "" should that be true?
- should all `not_` operations always be the invers of their counterpart?


cpvalente (approved):
Thank you both

PR #1947, opened 2026-01-16
Refactor: unify view search params - part 1

alex-Arc:


packages/types/src/translations/index.ts
```diff
@@ -19,6 +19,7 @@ export const langEn = {
   'countdown.to_start': 'Time to start',
   'countdown.waiting': 'Waiting for event start',
   'countdown.overtime': 'in overtime',
+  'countdown.all_have_past': 'Alle selected events have past',
```
cpvalente:
```suggestion
  'countdown.all_have_finished': 'All selected events have finished',
```

apps/client/src/views/countdown/countdown.options.ts
```diff
@@ -14,6 +14,10 @@ export const getCountdownOptions = (
   customFields: CustomFields,
   persistedSubscriptions: EntryId[],
 ): ViewOption[] => {
+  const mainOptions = makeOptionsFromCustomFields(customFields, [
```
cpvalente: same here, i am not sure it makes sense to change the main source to note

apps/client/src/views/countdown/SingleEventCountdown.tsx
```diff
@@ -1,3 +1,4 @@
+// AI generated and not reviewd
```
cpvalente: 😅

apps/client/src/views/countdown/SingleEventCountdown.tsx
```diff
@@ -41,7 +42,7 @@ export default function SingleEventCountdown({ subscribedEvent, goToEditMode }:
 
   const { endedAt } = reportData[subscribedEvent.id] ?? { endedAt: null };
   const countdownEvent = { ...subscribedEvent, expectedStart, endedAt };
-  const title = subscribedEvent.title.length ? subscribedEvent.title : ' '; // insert utf-8 empty space to avoid the line collapsing
+  const title = getPropertyValue(subscribedEvent, mainSource ?? 'title'); //TODO: do we want to force a whitespace here to keep the box hight
```
cpvalente: I did not understand the question

apps/client/src/views/countdown/Countdown.tsx
```diff
@@ -118,13 +122,21 @@ function CountdownContents({ playableEvents, subscriptions, goToEditMode }: Coun
     );
   }
 
-  if (subscribedEvents.length === 1) {
+  if (subscribedEvents.length === 1 && eventsToShow.length === 1) {
```
cpvalente, line 125: eventsToShow is derived  from subscribedEvents. can we simplify to
```suggestion
  if (eventsToShow.length === 1) {
```

alex-Arc: hmm i'm trying to catch the case where there has only selected one event, if we only check `eventsToShow`
then the view would switch to single event view when all but the last event has past, I don't think that is intended?

apps/client/src/views/countdown/Countdown.tsx
```diff
@@ -106,6 +109,7 @@ function CountdownContents({ playableEvents, subscriptions, goToEditMode }: Coun
   }
 
   const subscribedEvents = getOrderedSubscriptions(subscriptions, playableEvents);
+  const eventsToShow = !hidePast ? subscribedEvents : subscribedEvents.filter((event) => !hidePast || !event.isPast);
```
cpvalente:
```suggestion
  const eventsToShow = !hidePast ? subscribedEvents : subscribedEvents.filter((event) => !event.isPast);
```

apps/client/src/views/countdown/SingleEventCountdown.tsx
```diff
@@ -41,7 +41,7 @@ export default function SingleEventCountdown({ subscribedEvent, goToEditMode }:
...
```
cpvalente: yes, we want to keep the space. we can do that in CSS if you prefer

apps/client/src/views/timer/timer.options.ts
```diff
@@ -27,12 +27,12 @@ export const getTimerOptions = (timeFormat: string, customFields: CustomFields):
   const mainOptions = makeOptionsFromCustomFields(customFields, [
     { value: 'none', label: 'None' },
     { value: 'title', label: 'Title' },
-    { value: 'note', label: 'Note' },
   ]);
   const secondaryOptions = makeOptionsFromCustomFields(customFields, [
     { value: 'none', label: 'None' },
     { value: 'title', label: 'Title' },
     { value: 'note', label: 'Note' },
+    { value: 'parent', label: 'Group Title' },
```
cpvalente, line 35: nice

PR #1949, opened 2026-01-22
Refactor: title card

alex-Arc:
extracted the messy stuff here #2141

### DIFF
--- a/apps/client/src/common/components/title-card/TitleCard.scss
+++ b/apps/client/src/common/components/title-card/TitleCard.scss
@@ -4,40 +4,60 @@
   position: relative;
   display: flex;
   flex-direction: column;
-}
 
-.title-card__title,
-.title-card__placeholder {
-  font-weight: 600;
-  font-size: $title-font-size;
-  line-height: 1.2em;
-}
+  background-color: var(--card-background-color-override, $viewer-card-bg-color);
+  padding: $view-card-padding;
+  border-radius: $element-border-radius;
 
-.title-card__title {
-  color: var(--color-override, $viewer-color);
-  padding-right: 1em;
-  min-height: 1.2em;
-}
+  border-left: 1vw solid;
 
-.title-card__placeholder {
-  color: var(--label-color-override, $viewer-label-color);
-}
+  .title-card__title:empty::before {
+    color: var(--label-color-override, $viewer-label-color);
+    content: attr(data-placeholder);
+  }
 
-.title-card__secondary {
-  font-size: $base-font-size;
-  color: var(--secondary-color-override, $viewer-secondary-color);
-  line-height: 1.2em;
-}
+  .title-card__title {
+    font-weight: 600;
+    line-height: 1.4em;
+    padding-right: 1em;
+    color: var(--color-override, $viewer-color);
+  }
 
-.title-card__label {
-  position: absolute;
-  right: 1rem;
-  top: 0.5rem;
-  font-size: $timer-label-size;
-  color: var(--secondary-color-override, $viewer-secondary-color);
-  text-transform: uppercase;
+  .title-card__secondary {
+    color: var(--secondary-color-override, $viewer-secondary-color);
+    line-height: 1.2em;
+  }
 
-  &--accent {
-    color: var(--accent-color-override, $accent-color);
+  &.md {
+    .title-card__title {
+      font-size: $title-font-size;
+    }
+    .title-card__secondary {
+      font-size: $base-font-size;
+    }
+  }
+  &.lg {
+    .title-card__title {
+      font-size: $large-font-size;
+    }
+    .title-card__secondary {
+      font-size: $title-font-size;
+    }
+    .schedule__ {
+      font-size: $base-font-size;
+    }
+  }
+
+  .title-card__label {
+    position: absolute;
+    right: 1rem;
+    top: 0.5rem;
+    font-size: $timer-label-size;
+    color: var(--secondary-color-override, $viewer-secondary-color);
+    text-transform: uppercase;
+
+    &--accent {
+      color: var(--accent-color-override, $accent-color);
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


cpvalente:
@alex-Arc , lets look at this together next catch up

alex-Arc:
@coderabbitai review

apps/client/src/common/components/schedule-time/ScheduleTime.tsx
```diff
@@ -0,0 +1,88 @@
...
+  showExpected: boolean;
+  className?: string;
+  preferredFormat12?: string;
+  preferredFormat24?: string;
+  expectedStart: number;
+};
+
+//TODO: consider relative mode
+export default function ScheduleTime(props: ScheduleTimeProps) {
+  const {
```
cpvalente: nitpick: we can destructure inline

apps/client/src/common/components/schedule-time/ScheduleTime.tsx
```diff
@@ -0,0 +1,88 @@
...
+type ScheduleTimeProps = {
+  event: ExtendedEntry<OntimeEvent>;
+  showExpected: boolean;
+  className?: string;
+  preferredFormat12?: string;
+  preferredFormat24?: string;
+  expectedStart: number;
+};
+
+//TODO: consider relative mode
```
cpvalente: question: we would need some more information here, what about relative mode we need to consider? the comment should provide some clues as to what we need to do

apps/client/src/common/components/schedule-time/ScheduleTime.tsx
```diff
@@ -0,0 +1,88 @@
...
+    preferredFormat12 = 'h:mm a',
+    preferredFormat24 = 'HH:mm',
+    expectedStart,
+  } = props;
+  const { timeStart, duration, delay, countToEnd } = event;
+
+  const plannedStart = timeStart + delay + event.dayOffset * dayInMs;
+
+  // only show new expected value if outside  range of the planned value
+  const isExpectedValueShow = showExpected && isOutsideRange(plannedStart, expectedStart);
```
cpvalente: question: should we have an early return here in case `isExpectedValueShow` is false? we can skip a bunch of logic and make things easier

apps/client/src/common/components/title-card/TitleCard.scss
```diff
@@ -4,40 +4,60 @@
...
+    padding-right: 1em;
+    color: var(--color-override, $viewer-color);
+  }
+
+  .title-card__secondary {
+    color: var(--secondary-color-override, $viewer-secondary-color);
+    line-height: 1.2em;
+  }
+
+  &.md {
```
cpvalente, line 31: suggestion: everywhere else in the code we use small / medium / large and generally avoid shorthand like md

apps/client/src/common/components/title-card/TitleCard.scss
```diff
@@ -4,40 +4,60 @@
...
-  color: var(--color-override, $viewer-color);
-  padding-right: 1em;
-  min-height: 1.2em;
-}
+  border-left: 1vw solid;
 
-.title-card__placeholder {
-  color: var(--label-color-override, $viewer-label-color);
-}
+  .title-card__title:empty::before {
```
cpvalente, line 14: question: I am not sure I understand this, if what we need to do is to make sure we reserve space in case it is empty, we can add an empty space after like we do in other places

```css
&::after {
  content: '\200b';
}
```

apps/client/src/common/components/title-card/TitleCard.tsx
```diff
@@ -1,26 +1,63 @@
...
+  expectedStart?: undefined;
+  showExpected?: false;
+};
+
+type TitleCardProps = TitleCardExpectedProps | TitleCardNoExpectedProps;
 
 const TitleCard = forwardRef((props: TitleCardProps, ref: ForwardedRef<HTMLDivElement>) => {
-  const { label, title, secondary, className = '' } = props;
+  'use memo';
+  const {
```
cpvalente: nitpick: we can destructure in line with the function declaration and remove the `forwardRef`

apps/client/src/views/countdown/SingleEventCountdown.tsx
```diff
@@ -39,19 +39,22 @@ export default function SingleEventCountdown({ subscribedEvent, goToEditMode }:
...
-      <div className='event__title' style={{ borderColor: countdownEvent.colour }}>
-        <ScheduleTime event={countdownEvent} showExpected={showExpected} />
-        {title}
-        {secondaryData && <div className='secondary'>{secondaryData}</div>}
-      </div>
+      <SubscriptionStatus event={subscribedEvent} expectedStart={expectedStart} endedAt={endedAt} />
+      <TitleCard
+        title={title}
+        secondary={secondaryData}
+        colour={subscribedEvent.colour}
```
cpvalente: question: It seems to me that we need to expose several props to handle this usage.
we dont need our components to be overly abstracted since they are not general purpose. I would prefer extracting two components, each with simpler interfaces

this would also buy us options in future if we want to redesign either components (which we probably should)

apps/client/src/common/components/title-card/TitleCard.scss
cpvalente: ^

apps/client/src/common/components/title-card/TitleCard.tsx
```diff
@@ -1,33 +1,60 @@
...
+type TitleCardMainProps = {
   title?: string;
   label?: 'now' | 'next';
   secondary?: string;
   className?: string;
-}
+  colour?: string;
+  textAlign?: 'left' | 'right' | 'center';
+  size?: 'md' | 'lg';
+  placeholder?: string;
```
cpvalente, line 17: question: it doesnt seem that we have consumers for these props yet? is that an indicator that we dont need this for now?

cpvalente (approved):

PR #1957, opened 2026-01-27
Remove serverport from project file

alex-Arc:
- the `serverPort` value is dropped from the project file (nothing is moved over to avoid issues where the port moves on the next restart)
- the env `PORT` is now available and will overwire the port on start-up

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@getontime/cli",
-    "version": "4.4.2",
+    "version": "4.5.0",
     "author": "Carlos Valente",
     "description": "Time keeping for live events",
     "repository": "https://github.com/cpvalente/ontime",
--- a/apps/client/package.json
+++ b/apps/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ontime-ui",
-  "version": "4.4.2",
+  "version": "4.5.0",
   "private": true,
   "type": "module",
   "dependencies": {
--- a/apps/electron/package.json
+++ b/apps/electron/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ontime-electron",
-  "version": "4.4.2",
+  "version": "4.5.0",
   "author": "Carlos Valente",
   "description": "Time keeping for live events",
   "repository": "https://github.com/cpvalente/ontime",
--- a/apps/resolver/package.json
+++ b/apps/resolver/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@getontime/resolver",
-  "version": "4.4.2",
+  "version": "4.5.0",
   "type": "module",
   "repository": "https://github.com/cpvalente/ontime",
   "types": "./dist/main.d.ts",
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -2,7 +2,7 @@
   "name": "ontime-server",
   "type": "module",
   "main": "src/index.ts",
-  "version": "4.4.2",
+  "version": "4.5.0",
   "exports": "./src/index.js",
   "dependencies": {
     "@googleapis/sheets": "^5.0.5",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ontime",
-  "version": "4.4.2",
+  "version": "4.5.0",
   "description": "Time keeping for live events",
   "keywords": [
     "ontime",
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -1,5 +1,5 @@
 {
-  "version": "4.4.2",
+  "version": "4.5.0",
   "name": "ontime-types",
   "type": "module",
   "main": "./src/index.ts",


apps/server/src/services/app-state-service/AppStateService.ts
```diff
@@ -61,3 +62,13 @@ export async function setShowWelcomeDialog(show: boolean): Promise<boolean> {
   await config.write();
   return show;
 }
+
+export async function getServerPort(): Promise<number> {
+  await config.read();
+  return config.data.serverPort ?? 4001;
```
cpvalente: question: should we instead make sure that it always exists by populating the file in the setup?

apps/server/src/setup/environment.ts
```diff
@@ -5,4 +5,10 @@ export const isTest = Boolean(process.env.IS_TEST);
 export const environment = isTest ? 'test' : env;
 export const isDocker = env === 'docker';
 export const isProduction = isDocker || (env === 'production' && !isTest);
-export const isOntimeCloud = Boolean(process.env.IS_CLOUD);
\ No newline at end of file
+export const isOntimeCloud = Boolean(process.env.IS_CLOUD);
+
+export const startUpPort = process.env.PORT
```
cpvalente: cleanup: we can likely simplify this by extracting the logic so we end up with

```suggestion
export const startUpPort = parsePort(process.env.PORT) || 4001
```

also, can we think of a better name? maybe `serverPort` or `processPort`?

apps/server/src/app.ts
```diff
@@ -176,14 +176,13 @@ export const initAssets = async (escalateErrorFn?: (error: string, unrecoverable
  */
 export const startServer = async (): Promise<{ message: string; serverPort: number }> => {
   checkStart(OntimeStartOrder.InitServer);
-  const settings = getDataProvider().getSettings();
-  const { serverPort: desiredPort } = settings;
+  const desiredPort = startUpPort ? startUpPort : await getServerPort();
```
cpvalente: how about
```suggestion
  const desiredPort = startUpPort || await getServerPort();
```

apps/client/src/features/app-settings/panel/settings-panel/ServerPortSettings.tsx
```diff
@@ -0,0 +1,114 @@
...
+import Input from '../../../../common/components/input/input/Input';
+import { preventEscape } from '../../../../common/utils/keyEvent';
+import { isOnlyNumbers } from '../../../../common/utils/regex';
+import * as Panel from '../../panel-utils/PanelUtils';
+
+interface ServerPortForm {
+  serverPort: number;
+}
+
+export default function ServerPortSettings() {
```
cpvalente, line 18: it feels a bit weird to have so much UI for a single value, maybe we can add some context here for users explaining what happens when the port changes?

cpvalente:
@coderabbitai review

apps/server/src/api-data/db/migration/db.migration.v4.ts
```diff
@@ -0,0 +1,26 @@
...
+export function shouldMigrateServerPort(jsonData: object): boolean {
+  return (
+    is.objectWithKeys(jsonData, ['settings']) &&
+    is.object(jsonData.settings) &&
+    'serverPort' in jsonData.settings &&
+    typeof (jsonData.settings as { serverPort?: unknown }).serverPort === 'number'
+  );
+}
+
+export function migrateServerPort(jsonData: Partial<DatabaseModel>): Partial<DatabaseModel> {
```
cpvalente: Can we unit test the migration? It is trivial for this case but can help in the long run as we add more migrations

apps/client/src/features/app-settings/panel/settings-panel/ServerPortSettings.tsx
alex-Arc: yeah, my issue managing 2 forms on the same file got hard to track
maybe you have a good idea for the shape

apps/server/src/setup/environment.ts
```diff
@@ -5,4 +5,14 @@ export const isTest = Boolean(process.env.IS_TEST);
 export const environment = isTest ? 'test' : env;
 export const isDocker = env === 'docker';
 export const isProduction = isDocker || (env === 'production' && !isTest);
-export const isOntimeCloud = Boolean(process.env.IS_CLOUD);
\ No newline at end of file
+export const isOntimeCloud = Boolean(process.env.IS_CLOUD);
+
+export const startUpPort = parsePort(process.env.PORT);
```
cpvalente: maybe we should improve typing here to avoid issues

```suggestion
export const startUpPort: number | null = parsePort(process.env.PORT);
```

apps/server/src/services/app-state-service/AppStateService.ts
```diff
@@ -61,3 +62,14 @@ export async function setShowWelcomeDialog(show: boolean): Promise<boolean> {
   await config.write();
   return show;
 }
+
+export async function getServerPort(): Promise<number> {
+  if (isTest) return 4001;
```
cpvalente: I am not sure about writing explicit test logic, how about using the environment variable for tests (if we need to)?

apps/server/src/api-data/session/session.service.ts
```diff
@@ -37,7 +38,8 @@ export async function getSessionStats(): Promise<SessionStats> {
  * Adds business logic to gathering data for the info endpoint
  */
 export async function getInfo(): Promise<GetInfo> {
-  const { version, serverPort } = getDataProvider().getSettings();
+  const { version } = getDataProvider().getSettings();
+  const serverPort = await getServerPort();
```
cpvalente: question: do we have a risk for drift here? if the port is taken, would we pick another one and have wrong data here?
also, maybe we dont need the port in this data?
the user needs the port to make the request

apps/server/src/classes/data-provider/__tests__/DataProvider.utils.test.ts
```diff
@@ -1,3 +1,4 @@
+// FIXME: AI generated and not reviewed
```
cpvalente: we cant really leave the comment or the situation 😅

apps/server/src/app.ts
```diff
@@ -176,14 +176,13 @@ export const initAssets = async (escalateErrorFn?: (error: string, unrecoverable
...
-  const settings = getDataProvider().getSettings();
-  const { serverPort: desiredPort } = settings;
+  const desiredPort = startUpPort || (await getServerPort());
 
   expressServer = http.createServer(app);
 
   // the express server must be started before the socket otherwise the on error event listener will not attach properly
   const resultPort = await serverTryDesiredPort(expressServer, desiredPort);
-  await getDataProvider().setSettings({ ...settings, serverPort: resultPort });
+  await setServerPort(resultPort);
```
cpvalente: question: maybe only do this if it has changed?

apps/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@getontime/cli",
-    "version": "4.4.2",
```
cpvalente: suggestion: we can likely remove this commit

PR #1959, opened 2026-02-01
fix: update data in background rundowns

cpvalente:
fixes a bug when updating background rundowns.
testing:
- create a custom field: `test`
- make a rundown and add an event with some data in `test
- make a second rundown also with an event in the `test` custom field
- in custom fields, rename `test` to `test-new`
- verify that the field is updated in both rundowns


cpvalente:
@coderabbitai review

alex-Arc (approved):
👍🏽 good catch

PR #1961, opened 2026-02-01
refactor: debounce writing to disk

cpvalente:
The idea of this PR is to debounce writing to the disk
this means we accept a window were a crash could lose data since the last write (as of now, 3 seconds)

we expect that this would give improvements to batch operations like import
we could consider widening the window to about 20 seconds which would likely give us a good performance when people are filling in the rundowns at the added risk of data loss

### DIFF
--- a/apps/server/src/classes/data-provider/DataProvider.ts
+++ b/apps/server/src/classes/data-provider/DataProvider.ts
@@ -180,10 +180,58 @@ async function mergeIntoData(newData: Partial<DatabaseModel>): ReadonlyPromise<D
   return db.data;
 }
 
+// Module-level state for debounced writes
+let pendingWrite: NodeJS.Timeout | null = null;
+let activeWrite: Promise<void> | null = null;
+const writeDelayMs = 3000; // 3 seconds
+
 /**
- * Handles persisting data to file
+ * Handles persisting data to file with trailing-edge debounce
+ * Multiple rapid calls will be coalesced into a single write
  */
 async function persist() {
   if (isTest) return;
+
+  // Cancel any pending write and reschedule
+  if (pendingWrite) {
+    clearTimeout(pendingWrite);
+  }
+
+  // Schedule new write after quiet period
+  pendingWrite = setTimeout(async () => {
+    pendingWrite = null;
+
+    // Wait for any in-progress write to finish first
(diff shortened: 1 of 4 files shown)


apps/server/src/classes/data-provider/DataProvider.ts
```diff
@@ -180,10 +183,72 @@ async function mergeIntoData(newData: Partial<DatabaseModel>): ReadonlyPromise<D
...
+ * Multiple rapid calls will be coalesced into a single write
  */
 async function persist() {
   if (isTest) return;
-  await db.write();
+
+  // Skip write if data hasn't changed
+  if (lastWrittenData && deepEqual(db.data, lastWrittenData)) {
+    return;
+  }
```
cpvalente: this looks like a lot of work, we could simply override the new data with something that is the same as the old (although services should have prevented a noop from getting this far)

apps/server/src/classes/data-provider/DataProvider.ts
```diff
@@ -180,10 +183,72 @@ async function mergeIntoData(newData: Partial<DatabaseModel>): ReadonlyPromise<D
...
+      // Write already in progress, reschedule
+      persist();
+      return;
+    }
+
+    // Double-check for changes before writing (data may have changed during delay)
+    if (lastWrittenData && deepEqual(db.data, lastWrittenData)) {
+      pendingWrite = null;
+      return;
+    }
```
cpvalente: nah, just write it

apps/server/src/classes/data-provider/DataProvider.ts
```diff
@@ -180,10 +183,72 @@ async function mergeIntoData(newData: Partial<DatabaseModel>): ReadonlyPromise<D
...
+  if (pendingWrite) {
+    clearTimeout(pendingWrite);
+    pendingWrite = null;
+  }
+
+  if (!writeInProgress && !isTest) {
+    // skip write if data hasn't changed
+    if (lastWrittenData && deepEqual(db.data, lastWrittenData)) {
+      return;
+    }
```
cpvalente: nah, just write it

cpvalente:
@coderabbitai full review

PR #1962, opened 2026-02-03
Automation should use event store

alex-Arc:


apps/server/src/api-data/automation/clients/http.client.ts
```diff
@@ -1,20 +1,20 @@
-import { HTTPOutput, LogOrigin } from 'ontime-types';
+import { HTTPOutput, LogOrigin, RuntimeStore } from 'ontime-types';
 
 import { logger } from '../../../classes/Logger.js';
-import type { RuntimeState } from '../../../stores/runtimeState.js';
 
 import { parseTemplateNested } from '../automation.utils.js';
+import { DeepReadonly } from 'ts-essentials';
```
cpvalente: import order: this should be all the way at the top under the ontime imports

apps/server/src/api-data/automation/automation.dao.ts
```diff
@@ -12,6 +13,31 @@ import { deleteAtIndex, generateId } from 'ontime-utils';
 import { getDataProvider } from '../../classes/data-provider/DataProvider.js';
 
 import { isAutomationUsed } from './automation.utils.js';
+import { eventStore } from '../../stores/EventStore.js';
+import { RuntimeState } from '../../stores/runtimeState.js';
```
cpvalente: import order: this should go up next to the imports that are at two levels over `../../`

apps/server/src/api-data/automation/automation.service.ts
```diff
@@ -11,14 +12,14 @@ import {
...
-import { getState, type RuntimeState } from '../../stores/runtimeState.js';
 import { isOntimeCloud } from '../../setup/environment.js';
 
 import { emitOSC } from './clients/osc.client.js';
 import { emitHTTP } from './clients/http.client.js';
-import { getAutomationsEnabled, getAutomations, getAutomationTriggers } from './automation.dao.js';
+import { getAutomationsEnabled, getAutomations, getAutomationTriggers, getRuntimeStore } from './automation.dao.js';
 import { isContained, isEquivalent, isGreaterThan, isLessThan } from './automation.utils.js';
 import { toOntimeAction } from './clients/ontime.client.js';
+import { RuntimeState } from '../../stores/runtimeState.js';
```
cpvalente: import order: this should go up next to the imports that are at two levels over ../../

apps/server/src/api-data/automation/automation.dao.ts
```diff
@@ -12,6 +13,31 @@ import { deleteAtIndex, generateId } from 'ontime-utils';
...
+
+export function getRuntimeStore(state?: RuntimeState): RuntimeStore {
```
cpvalente: question: I dont quite understand this function signature, part of the issue may be our use of the names store and state which are interchangeable in most ontime scenarios

- the only callers that dont provide store are the test endpoints, could we not force them to provide a store?
- do we need to construct the objects or could we simply always get the store every time?
- if we make this non-optional, then this abstraction is not needed at all. and the functions in automation.service can remain pure(r) by having the state passed in

apps/server/src/api-data/automation/automation.service.ts
```diff
@@ -115,8 +117,8 @@ export function testConditions(
  * Handles preparing and sending of the data
  * Returns a boolean indicating whether a message was sent
  */
-function send(output: AutomationOutput[], state?: RuntimeState) {
-  const stateSnapshot = state ?? getState();
+function send(output: AutomationOutput[], store?: RuntimeStore) {
+  const stateSnapshot = store ?? getRuntimeStore();
```
cpvalente: question: could we be stricter here and always demand the store? as far as i can see that would just imply that our test endpoint needs to fetch the store before sending

apps/server/src/api-data/automation/automation.dao.ts
alex-Arc: I will maybe ask a question first:
when `triggerAutomations` is called from the runtime service I don't think we can guarantee that the event store up to date yet, but I'm unsure since it is called in `process.nextTick` but the event store is called in a decorator?

PR #1967, opened 2026-02-10
fix: prevent infinite render in settings forms

cpvalente:


cpvalente:
@coderabbitai full review

Shobhit-Nagpal:
changes look good. i've got a question regarding what was causing the infinite render in the form? was it the waiting state while we're fetching the data or using the values field from RHF since it's reactive in nature? i see it to be the latter

PR #1969, opened 2026-02-13
Optimistic add

cpvalente:
Solves an issue of perceived performance in rundown manipulations by optimistic adding elements in the UI
This meant extracting some business logic from the server to shared logic to ensure the behaviour in the UI is the same as what the server does.

The large amount of changes is moving things from the server to the shared package. there are no intended behaviour changes in the extracted logic


cpvalente:
@coderabbitai full review

PR #1970, opened 2026-02-14
reduce client connection load on server

cpvalente:
improve data handling to reduce load on server
- improve socket reconnection
- allow react query cancellation of ongoing requests
- extract react query and axios configuration to global scope with local overrides


cpvalente:
@coderabbitai full review

Issue #1972, opened 2026-02-14
Custom Field / Images via URL


jboHH:
I'm trying to apply images as a custom field during the import of an Excel spreadsheet. Images via URL works perfectly in the OnTime Editor directly, but when importing a Excel spreadsheet, OnTime interprets them as text instead of IMG. Is there a way to apply image URLs during import? Thanks for your help!

alex-Arc:
hey @jboHH take a look at the new release 
https://github.com/cpvalente/ontime/releases/tag/v4.5.0

if you predefine the custom field you want to be an image it should stay so

jboHH:
> hey [@jboHH](https://github.com/jboHH) take a look at the new release https://github.com/cpvalente/ontime/releases/tag/v4.5.0
> 
> if you predefine the custom field you want to be an image it should stay so

Hi Alex,
thanks a lot for the super fast implementation! It works exactly like I need it :-)
Best
Jens

PR #1976, opened 2026-02-16
feat: allow existing custom fields to keep its type

alex-Arc:
inspired by #1972

this is very small change that just allows predefined custom fields to keep their type when importing from excel


cpvalente (approved):
Great work!

PR #1978, opened 2026-02-18
fix: onStart triggered by pending roll start

cpvalente:


alex-Arc:
I can't quite find the way to reproduce the original issue?

cpvalente:
> I can't quite find the way to reproduce the original issue?

If we go into roll from a waiting to roll timer the trigger was not called

alex-Arc (approved):

PR #1979, opened 2026-02-19
Improve link sharing

cpvalente:
Improves sharing of cuesheet
- When the share has no write access, we hide the Edit / Run control
- For presets of cuesheets, we always use the masked output instead of alias. This prevents the permissions from being lost


cpvalente:
@coderabbitai full review

Issue #1982, opened 2026-02-22
"Timer message" and "secondary message" not translatable


BergManuel:
Hi,

the text fields for "Timer message" and "secondary message" are not translatable via the custom translation menu.
Is it possible to add these two values to the custom translation?

<img width="389" height="160" alt="Image" src="https://github.com/user-attachments/assets/c2d27074-7724-4c77-8261-2edf670288d3" />

cpvalente:
Hi @BergManuel thank you for bringing this up.
It looks like it was something we overlooked

We should definitely expose this, meanwhile the change is pretty trivial. Would you be willing to try and make a PR?

cpvalente:
Closing this as it has been completed and  released as part of v4.6.0

PR #1987, opened 2026-02-27
add translation keys for timer message labels

jwetzell:
allows for setting custom translations for the timer message labels brought up by #1982

still need translations for the built in languages
- [x] de
- [ ] es
- [ ] fr
- [ ] it
- [ ] pt


cpvalente:
Thank you for this @jwetzell i will take it from here

PR #1988, opened 2026-02-28
Deps migration

cpvalente:


apps/resolver/tsconfig.json
```diff
@@ -10,8 +10,7 @@
     "allowJs": true,
     "declaration": true,
     "outDir": "dist",
-    "sourceMap": true,
-    "baseUrl": "src",
+    "sourceMap": true
```
alex-Arc: it seams that build fails unless we add `"ignoreDeprecations": "6.0"` not quite sure why

cpvalente: looks like `tsup` doesnt support typescript 6 since it still emits the deprecated `baseURL`
also `tsup` is not maintained anymore so we will need to migrate to something else, maybe [tsdown](https://tsdown.dev/)? I dont remember why we ended up locking in tsup to begin with

alex-Arc: seams like a fine alternative
tsup was there as esbuild refused to allow including our internal ontime-type lib as a prebuild

Issue #99, opened 2022-01-29
Delay support in roll mode


cpvalente:
Currently roll mode ignores delays.

This is because supporting it means editing the times as they run. 
ie: during a running event, we add a delay before of 10 min, this would mean the event started 10 minutes later and change the timer accordingly.
I am unsure if that is the expected behaviour

With this in place we have a good opportunity to compare the realtime clock to the running event clock and provide feedback.
Eg. we could add information on how delayed / ahead the overall planning is in some of the screens

dillwishlist:
I was thinking about this today.  If the delays could have a "wait to continue" option when operating in roll mode, the next event would not start until a manual (OSC or button) trigger resumes the roll.  The delay would be updated to the time it has waited, and then applied.

cpvalente:
Thank you for this.

I believe it is important considering whether this should be a new playback mode. It could become confusing to the current use of roll, which is meant for unattended setups.

I am developing a redesign of the playback mode to allow the sequencing of events in a auto playback feature. Perhaps I should look into integrating something as you suggest, an auto delay which absorbs extra time between events?

@dillwishlist Could you provide a use case for the feature?

je71175:
Hi
As a framework of thinking through this - it’s very similar to managing cues in a cue list on a lighting console.

There are some variations but the idea here is each individual cue has its own trigger type and delay value.
Go.
Wait.
Follow.
Here a lighting console app that has the basics and is a good simple and easy to use https://support.obsidiancontrol.com/Content/Cues_and_Cuelists/Cue_Timing/Setting_Cue_Triggers.htm

Cheers a

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

cpvalente:
I guess it depends if you are using ontime as a timer manager or event rundown. 
In the event rundown scenario, delays concern the global event, rather than single instances. 

Your explanation is very much inline from what I understand of the proposed "wait to continue". 
I believe the option to absorb overtime should be part of a fully featured delay event but it is really a matter of achieving a good design here. Could this tie in with #100 somehow?

je71175:
Yes, perhaps the best approach maybe to combine 100 and 99, - perhaps an option tie an event and its following delay, so that the delay value is automatically calculated based on the event's running time.  
If an event is shortened, then the delay as a negative value and following event times are re calculated.
If an event is late then the delay is positive and recalculation is also performed.
[...]

PR #1991, opened 2026-03-02
Adding new "break roll mode" toggle in Editor

cdslipp:
As initially discussed in issue #99. 

This is a personal wish list feature for me and I had some time today to take a look at it. I want to have a way to break out of Roll mode selectively. If this does not align with planned features etc, please let me know, happy to revisit this in whichever way makes the most sense


cdslipp:
Something else I should mention - I haven't review the spreadsheet import stuff in depth, caught it with Claude, I should do another pass of the import testing adding the break roll mode field in a csv/excel before this gets considered

cdslipp:
After testing with all of the End Actions - found it was much better to have the "break" in the roll mode occur at the beginning of the toggled event to keep things as clean as possible.

apps/server/src/api-data/db/migration/db.migration.v3.ts
```diff
@@ -378,6 +378,7 @@ export function migrateRundown(
           endAction: validateEndAction(entry.endAction, EndAction.None), // ensure end action is not stop
           timerType: isKnownTimerType(entry.timerType) ? entry.timerType : eventModel.timerType, // ensure the timer type is not count-to-end
           countToEnd: entry.timerType === 'count-to-end', // countToEnd was previously a timer type
+          breakRoll: false,
```
cpvalente: question: I dont think this should be here? this is a migration from v3 to v4. We may be able to introduce this without breaking changes and need for migration

apps/server/src/services/runtime-service/runtime.service.ts
```diff
@@ -588,6 +591,9 @@ class RuntimeService {
       }
 
       if (result.didStart) {
+        if (newState.eventNow?.breakRoll) {
+          runtimeState.exitRoll();
+        }
```
cpvalente, line 593: question: is the intention that this happens when the event starts?

apps/server/src/stores/runtimeState.ts
```diff
@@ -472,6 +472,16 @@ export function pause(state: RuntimeState = runtimeState): boolean {
   return true;
 }
 
+export function exitRoll(state: RuntimeState = runtimeState): boolean {
```
cpvalente: question: would it be better to use the simple playback state start? exit roll is simply a coordinator and we likely prefer keeping the timer semantics simple

cpvalente (commented):
Hi @cdslipp , thank you for this

I dont have objections on a break mode.
As you mentioned in the ticket, I think that the end actions is the preferred way to do it?

That of course has the implication that the break roll takes effect once the event reaches its end.  How do you think that affects your workflow?

We would need to test this to make sure that the load next automation doesnt kick in before we break out of roll mode

I will check later whether we need some more checks around importing projects

cdslipp:
No problem! I tested and played with both options on my fork, after trying both I realized that making it a separate toggle gives better functionality as it allows you to still use different end actions.

So this way if you want it to stop when you break out of roll mode, you turn break roll mode on and set the end action to "Load next". Similarly if you want it to continue to the next event "Play next" works as well.

Otherwise default behaviour (with End action of "None") is what I believe is the one I'd reach for the most often - breaking out of roll mode and counting into negative until ready to move on.

I did originally feel that adding more end actions was the way, but after discovering this I realized that it would need to be several "break roll" type end actions to effectively mirror all 3 existing, which feels clunky - hence the toggle now being my preferred.

Thank you for the review! I'm still getting used to the codebase, I will go through these later today

apps/server/src/services/runtime-service/runtime.service.ts
cdslipp: Yes, that's the cleanest behaviour I was able to get.. I will try to write up something that makes sense. I think this change might also warrant tooltips for "Count to End" and "Break Roll" so it's easier for users to understand

apps/server/src/stores/runtimeState.ts
cdslipp: I think you're right, I will switch to the playback state, that would be much cleaner

apps/client/src/features/rundown/rundown-event/RundownEventInner.tsx
```diff
@@ -143,6 +146,9 @@ function RundownEventInner({
           <Tooltip text={`${countToEnd ? 'Count to End' : 'Count duration'}`} render={<span />}>
             <LuArrowDownToLine className={`${style.statusIcon} ${countToEnd ? style.active : style.disabled}`} />
           </Tooltip>
+          <Tooltip text={breakRoll ? 'Breaks roll' : 'Continues roll'} render={<span />}>
+            <IoStop className={`${style.statusIcon} ${breakRoll ? style.active : style.disabled}`} />
```
cdslipp: @cpvalente do you feel like this warrants an icon at all? I am torn. I was thinking that roll mode had a clock icon somewhere in the app, but I'm realizing I must have been misremembering and thinking about the clock icon that is part of the Ontime companion module. 

Either way, I think my preferred icon would be like a clock with a line through it? and maybe we add a clock icon to the roll mode button as well... what do you think? Stop button icon is kind of a placeholder for now

cdslipp: Maybe this one could work actually - thoughts? 
image

cdslipp: Nevermind, after trying that one, I went with adding a clock button to Roll mode button instaed and the TbClockStop icon for the "Break roll mode" indicator - this is added in the most recent commits. 

image
image

PR #1993, opened 2026-03-04
fix: prevent issues with midnight and day recognition on dst kick in

cpvalente:


alex-Arc (approved):
I think you got it, but this is tricky

PR #1994, opened 2026-03-05
Attempt at multi-edit

cdslipp:
## Demo Video
[Video Thumbnail](https://youtu.be/UOvur4vU3Pc?si=WiBuAOnXhcpXFFD8)

I need to clean up the code, I am touching too many existing components and I am going to break the main MultiEventEditor.tsx into smaller components to mirror the main EventEditor components. Either way please let me know what you think of the functionality and approach! This PR is a bit bigger than I wanted, but I couldnt figure out how to integrate with the existing editor component without overcomplicating stuff like crazy. A completely separate UI component makes more sense to me, but it is one more thing to maintain which is annoying.

## List of Batch Edit Behaviours and Mixed States

Right now, I have been messing with multiple UI states for each selection in the editor, here are my current choices:

### Event Schedule
- Start Time: blank out time but batch edit link to start toggle (same behaviour as batch context menu/right click)
- End Time: blank out time but batch edit lock end toggle
   -If end lock is mixed or off, clicking the lock will enable duration lock for all, but this presents a dialog to the user to confirm
- Duration - only editable if duration lock is ON
   - IF duration lock is mixed or off, clicking the lock will enable duration lock for all, but this presents a dialog to the user to confirm
   - If everything set to duration lock, but durations are mixed the input says "multiple"
   - If everyting same duration it says the duration

### Event Behaviour
- End Action: select says "mixed" when mixed
- Count to End: we hide the toggle/thumb thing on switch when mixed

### Event Data
- Event ID and Cue should just blank out and not be multi-editable
- Flag: we hide the toggle/thumb thing on switch when mixed
- Colour: We show NO selection if mixed, and if all the same colour we show the selection highlight
- Title: if all same we show, if mixed we say "multiple"
- Notes: Same as title, btu if all blank show blank

### Custom Fields
If there are no custom fields, show the manage custom fields link same as single select
- Same as notes - if all blank show blank, if mixed show "multiple", all same show the same value

### Automations
I looked at this briefly and it feels too complex right now because there is no easy batch endpoint for triggers/automations... so as far as I can tell this would be clunky to set up without setting up a server route. This could be an improvement later one


cpvalente:
Thank you for this @cdslipp

There seems to be several unrelated changes which we would be inclined to removing (even if they are good) to keep this change clean

The entry merging is slightly complex and we can probably simplify by leveraging some of the existing processes or pushing logic into the server. 

One first step is to check if [the batch endpoint](https://github.com/cpvalente/ontime/blob/master/apps/server/src/api-data/rundown/rundown.router.ts#L216) can simplify things
We also dont need to check so many of the edge cases (like the first link) since the server will clean this up

I also wonder if we may be better off not using a new multimerge panel but leverage the existing. Since the components are modified to accept indeterminate states, we could generalise the logic
This would however make the component more complex, but maybe make it easier to change in future since we have less places that need to change (?)

It could also be worth it to revisit some of the old attempts. The performance issue was hopefully mitigated by the recent changes and we can see which decisions we can recover from past efforts
https://github.com/cpvalente/ontime/tree/bp/edit-multiple-events
https://github.com/cpvalente/ontime/pull/1186

I wil take some time to go through the changes more carefully and give a proper code review.
Meanwhile, wuld you be willing to join us on discord? maybe we can chat about the PR

cpvalente:
@coderabbitai full review

apps/client/src/common/components/input/colour-input/SwatchSelect.tsx
```diff
@@ -9,6 +9,8 @@ interface ColourInputProps {
   value: string;
   name: 'colour';
   handleChange: (newValue: 'colour', name: string) => void;
+  /** When true, no swatch appears selected */
+  noSelection?: boolean;
```
cpvalente: question: can we simplify the use cases to get less logic branching?

in the specific case of colour, we could accumulate all values like
```
1. 'blue' + 'blue' + 'blue' = 'blue'
2. 'blue' + 'blue' + 'red' = 'multiple'

```
then we could simply pass the value to swatch select, in which case the logic should work without needing to know details of its context

```
1. <Swatch key={colour} color={colour} onClick={setColour} isSelected={'blue' === colour} />

2. <Swatch key={colour} color={colour} onClick={setColour} isSelected={'multiple' === colour} />
```

apps/client/src/common/components/input/input/Input.module.scss
```diff
@@ -12,6 +12,7 @@
 
   &::placeholder {
     color: $gray-600;
+    font-style: italic;
```
cpvalente: cleanup: lets try and keep the changes focused.
I dont necessarily disagree, but unrelated changes open irrelevant discussions for the scope of the feature and make things generally harder to progress

apps/client/src/common/components/input/time-input/TimeInput.tsx
```diff
@@ -38,7 +38,7 @@ export default function TimeInput<T extends string>({
    */
   const resetValue = useCallback(() => {
     if (typeof time !== 'number' || isNaN(time)) {
-      setValue('00:00:00');
+      setValue('');
```
cpvalente, line 41: question: can you explain why do we need this change?

apps/client/src/common/components/select/Select.tsx
```diff
@@ -16,13 +16,20 @@ interface SelectProps<T> extends Omit<BaseSelect.Root.Props<T>, 'items'> {
...
     <BaseSelect.Root items={options} {...selectRootProps}>
       <BaseSelect.Trigger className={cx([styles.select, styles[size], fluid && styles.fluid])}>
-        <BaseSelect.Value />
+        <BaseSelect.Value>
+          {(value: T | null) => {
+            if (value === null && placeholder) return placeholder;
+            const option = options.find((o) => o.value === value);
+            return option?.label ?? (placeholder || '');
+          }}
+        </BaseSelect.Value>
```
cpvalente: question: I am not entirely sure of the consequences of this change. 

We already have a workaround to provide select lables, although they are not styled as such
https://github.com/cpvalente/ontime/blob/ceb490c60a2dd159a8f5f4129c9fee4216d204c5/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx#L78

I wonder if a change like this implies reviewing all the select usages and converging the pattern
At that point, I think that the easiest change here is to simply use the existing pattern

apps/client/src/common/components/switch/Switch.module.scss
```diff
@@ -55,4 +55,8 @@
   &[data-checked] {
     translate: calc(var(--width) - var(--height)) 0;
   }
+
+  &[data-indeterminate] {
+    opacity: 0;
+  }
```
cpvalente: question: the switch without the indicator looks broken. indeterminate switches are strange and was one of the reasons we considered using another design element

to keep things simple here, lets stick with switches but improve the design, maybe there are relevant UX resources for this subject? I saw some that had a line in the middle of the switch which should be good enough if we dont have better ideas

apps/client/src/common/components/switch/Switch.tsx
```diff
@@ -6,12 +6,13 @@ import style from './Switch.module.scss';
...
   size?: 'medium' | 'large';
+  indeterminate?: boolean;
 }
 
-export default function Switch({ size = 'medium', ...switchProps }: SwitchProps) {
+export default function Switch({ size = 'medium', indeterminate, ...switchProps }: SwitchProps) {
   return (
     <BaseSwitch.Root className={cx([style.switch, style[size]])} {...switchProps}>
-      <BaseSwitch.Thumb className={style.thumb} />
+      <BaseSwitch.Thumb className={style.thumb} data-indeterminate={indeterminate || undefined} />
```
cpvalente: would something like this be more accurate
```suggestion
      <BaseSwitch.Thumb className={style.thumb} data-indeterminate={Boolean(indeterminate)} />
```

apps/client/src/common/components/switch/Switch.tsx
```diff
@@ -6,12 +6,13 @@ import style from './Switch.module.scss';
 
 interface SwitchProps extends BaseSwitch.Root.Props {
   size?: 'medium' | 'large';
+  indeterminate?: boolean;
```
cpvalente, line 9: question: what is the relationship between `value` and `indeterminate`?

Would we get easier logic by extending value to be something like
`value: boolean | 'indeterminate'`

apps/client/src/features/rundown/entry-editor/multi-edit/multiEditUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export type MergedEvent = {
+  [K in MergeableField]: MergedValue<OntimeEvent[K]>;
+} & {
+  custom: MergedCustomFields;
+  allLockDuration: boolean;
+  allLockEnd: boolean;
+};
+
+export function mergeEvents(entries: RundownEntries, selectedIds: Set<string>, flatOrder: EntryId[]): MergedEvent | null {
```
cpvalente: question: instead of merging all events, how about merging each entry values on its own?

this means that we will have more but simpler logic, the logic will also be closer to the consumer and we will have better opportunity to deal with the edge cases of each value

apps/client/src/features/rundown/useEventSelection.ts
```diff
@@ -66,15 +66,16 @@ export const useEventSelection = create<EventSelectionStore>()((set, get) => ({
 
       // if event is already selected, we remove it from selection
       // and set the anchor to the event after
-      selectedEvents.delete(id);
+      const remaining = new Set(selectedEvents);
```
cpvalente, line 69: question: can you explain this change?

apps/client/src/features/rundown/entry-editor/multi-edit/multiEditUtils.ts
cpvalente: suggestion: I think it is great that we push the business logic away into separate modules

try and keep the business logic pure (no side effects) and make a test suite to validate it

apps/client/src/common/components/input/input/Input.module.scss
cdslipp: Agreed, I will remove this!

apps/client/src/common/components/input/time-input/TimeInput.tsx
cdslipp: This was my workaround to getting empty state in the multi editor (ie when showing start and end time I want them to always be blank to imply there is a mixed value). However, there might be a clear way to do this so that we don't affect other areas of the app? I will try to make this change closer to the actual editor so its not a default state

apps/client/src/common/components/switch/Switch.tsx
cdslipp: I was literally just using this so we have a separate state ie that we can always toggle a mixed switch to 'ON'. However as I have been thinking about this and trying to rationalize... I realized its not computationally heavy at all to figure out the majority state - ie if most of the switches the selection are off, why not always make it so the toggle goes towards the majority first? If most switches in the selection are off, switch off. If most switches are on, switch on... this will make this feel more logical and not just like a random choice

apps/client/src/features/rundown/useEventSelection.ts
cdslipp: I was trying to understand Zustand as I haven't worked much with it before... as far I as I can tell with zustand + react most states are immutable so you have to create a new reference object instead of trying to mutate the original for it to be truly reactive? Pretty sure I got here after it not reactively updating the first time. I normally work with Svelte now, so I'm not really used to this kind of thing, so please let me know if there is a cleaner approach here! 

https://zustand.docs.pmnd.rs/learn/guides/immutable-state-and-merging#immutable-state-and-merging

apps/client/src/common/components/switch/Switch.module.scss
cdslipp: I have been trying a bunch of things, I think I landed somewhere I really like - let me know what you think!
image

apps/client/src/common/components/select/Select.tsx
cdslipp: This has been resolved to use the established pattern

PR #2005, opened 2026-03-13
refactor: prepare allowing multiple rundowns

cpvalente:
we will add support for loading rundowns in the cuesheet, for now we want to make sure that the cache knows about the multiple rundowns


cpvalente:
@coderabbitai full review

Issue #2012, opened 2026-03-18
Platform Support: LG WebOS 4 (Legacy Chromium) Compatibility Fixes


panctronic:
Currently, the Ontime client fails to load and render correctly on LG WebOS 4 (firmware 4.0.1-169). This platform uses an older embedded Chromium engine (roughly Chrome 53–56) that does not support the modern web standards ontime relies on. 

**Impact**
When attempting to run the client on LG WebOS 4, users experience:
* **Total App Failure:** The app crashes before React can even mount due to unrecognized modern JavaScript syntax.
* **Runtime Errors:** Data fetching and text resizing fail due to missing browser APIs.
* **Severe Visual Breakage:** Layouts are squished with no spacing, overlays (like the blackout screen and messages) do not appear, and text sizing is incorrect.

**Reproduction Environment**
* **Hardware:** Tested on an [LG 22SM3G-B (21.5" Smart Signage Display)](https://www.bhphotovideo.com/c/product/1572839-REG/lg_22sm3g_b_sm3g_series_21_5_16_9.html)
* **OS:** LG WebOS 4 (Firmware 4.0.1-169) (latest at time of testing).
* **Configuration:** Out-of-the-box settings.
* **Launch Method:** Utilizing the native "Play via URL" feature, pointed directly at the timer view (`http://$ONTIME:4001/timer`).

**Root Causes & Proposed Fixes**

* **JavaScript Compilation (ES2015+):** The current build emits modern syntax (arrow functions, classes, `const`/`let`) which the legacy engine rejects. 
    * *Fix:* Implement a legacy build step (via `@vitejs/plugin-legacy`) to transpile the output down to ES5 for older browsers.
* **Missing Web APIs:** The legacy browser lacks support for `AbortController` and `ResizeObserver`.
    * *Fix:* Introduce standard polyfills for these APIs during legacy builds.
* **Modern CSS Incompatibilities:** The legacy engine silently ignores modern CSS features such as `gap`, `inset`, `clamp()`, `min()`, and `color-mix()`. Additionally, `100vh` calculates incorrectly on this specific display.
    * *Fix:* Provide explicit CSS fallbacks (e.g., standard margins instead of `gap`, explicit `top/bottom/left/right` instead of `inset`, `px` fallbacks for responsive sizing) and inject a small inline script to accurately calculate viewport height. 

**Scope & Safety**
[...]

PR #2013, opened 2026-03-18
Add polyfills for legacy WebOS TV browsers

panctronic:
This commit introduces a new file, polyfills-legacy.js, which includes polyfills for the ResizeObserver feature, targeting older WebOS TV browsers (Chromium versions ~38-56). The polyfills are injected via additionalLegacyPolyfills in vite.config.js and are executed only in the legacy (nomodule) code path.

Targets https://github.com/cpvalente/ontime/issues/2012

### DIFF
--- a/apps/client/index.html
+++ b/apps/client/index.html
@@ -3,6 +3,30 @@
   <head>
     <meta charset="utf-8" />
     <base href="/" />
+    <script>
+      // Fix 100vh on TV browsers (e.g. LG WebOS) where window.innerHeight
+      // does not match the CSS viewport height unit (1vh).
+      // Runs before any stylesheet so the !important override wins.
+      (function() {
+        function applyHeight() {
+          var h = window.innerHeight;
+          var s = document.getElementById('__vhfix');
+          if (!s) {
+            s = document.createElement('style');
+            s.id = '__vhfix';
+            (document.head || document.getElementsByTagName('head')[0]).appendChild(s);
+          }
+          // Set html and body to the real inner height so that child height:100% chains work correctly.
+          // Keep it narrow - only html/body/root. The .App and view containers already use height:100%.
+          var css = 'html,body,#root{height:' + h + 'px!important;max-height:' + h + 'px!important}';
+
+          if (s.styleSheet) { s.styleSheet.cssText = css; }
+          else { s.textContent = css; }
+        }
+        applyHeight();
+        window.addEventListener('resize', applyHeight);
+      })();
+    </script>
     <link rel="icon" href="/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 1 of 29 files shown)


cpvalente:
Hi @panctronic , thank you for this.

I am personally skeptic of the change, at least in this shape. Maybe this will sound negative, but I do have a hopefully fruitful suggestion in the end.

The vite configuration and polyfills should in theory help us bridge the targets, however, as you found, the CSS might be the stopper here

We cannot count on a manual code change of the CSS because
- it adds unnecessary risk of breaking things that work well now
- it is not maintainable in the way that future changes can suddenly break our compatibility (if someone accidentally uses a CSS property which is not supported).
- it is not verifiable. or at least I dont know how we could automate a check that tells us that everything is working fine

Ideally the entire legacy layer should be similar to what we do with javascript, ie: at build time, ask sass to add a legacy layer to our CSS.
I also believe this is not as easy, and both the target that we have set here (12 year old chrome) and our usage of css (like CSS variables) may just not make this viable


This to give some context to the thinking, now for the productive part.
The major issue with compatibility is in signage screens. Knowing this we can narrow down the compatibility layer to a few screens (timer and backstage ?)

I would suggest we add javascript code that checks for the browser compatibility. If the browser falls short, we create a simpler duplicate of these views which use simpler css and js.
that means we shake off some features, but have fully working views which are maintained separately and have less risk of breaking

I believe this would be a great change and produce safer and simpler code. What do you think?

panctronic:
That makes sense. This was a hack job to get me through a show this weekend.

Given the (hopefully) shrinking audience of these older devices, do you think the maintenance burden of distinct views worth the cost?

cpvalente:
> Given the (hopefully) shrinking audience of these older devices, do you think the maintenance burden of distinct views worth the cost?

I dont have evidence to have an opinion beyond guess work, either way:

In my experience venues have old equipment, especially on auxiliary technical setup like PA screens or stage monitors.
I think it is possible to strike a good balance on maintenance and value by keeping the implementation minimal.

I would suggest:
- For a few views only (like timer and maybe backstage), have a javascript script that checks browser version, if it is old (not sure how to define old? maybe chatGPT can help), redirect to a legacy compatibility page
- in the legacy compatibility page keep things as simple as possible, ie: just timer and progress bar in the case of the timer view

The pages maintenance would be minimal to none at the cost of them not having feature parity. I think that is a good compromise for a compatibility layer

What do you think?

cpvalente:
As mentioned in the previous comment https://github.com/cpvalente/ontime/pull/2013#issuecomment-4102932674 we think a better solution would involve leveraging tooling to do the heavy lifting and potentially creating compatibility fallbacks for the harder cases

We have done some of the work in #2069 and #2070, and can review and add more views if it becomes necessary

closing this for now

PR #2016, opened 2026-03-20
feat: update auto cue numbering

### LINKED ISSUE #1998: Feature: Renumber events
I would love to see a feature (potentially in the right click context menu in the Editor) to renumber a set of events. Something similar to this dialog box from QLab:

Image

Ideal workflow would be shift clicking or ctrl clicking to select multiple events, then right click and choose "Renumber events". This would let you set a starting and increment value for the cue number for the events in question. 

Can become complicated a little complicated when considering things like decimal cues and letter cues like A2, both of which are currently supported. Would likely need prefix/suffix logic like QLab

alex-Arc:


alex-Arc:
@coderabbitai review

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+      return;
+    }
+
+    setLoading(true);
+    try {
+      const start = Number(startStr.trim());
+      await renumberCues(Array.from(selectedEvents), prefix, start);
+      onClose();
+    } catch {
+      setError('Renumber failed');
```
cpvalente: question: can we expose the error we get from the API here?

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  return (
+    <Modal
```
cpvalente: question: would dialog be a better component for this? this element is way to big and we dont want to change it too much

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+                fluid
+                autoComplete='off'
+                inputMode='numeric'
+                aria-invalid={startError !== null}
+                aria-describedby={startError !== null ? 'renumber-start-error' : undefined}
+                className={startError !== null ? style.inputInvalid : undefined}
+              />
+              <div className={style.fieldErrorSlot} aria-live='polite'>
+                {startError !== null ? (
+                  <p id='renumber-start-error' className={style.fieldError} role='alert'>
```
cpvalente: suggestion: we dont usually use `p` tags because they have some default styles. I think divs are more reliable at the cost of semantics

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+                height='large'
+                fluid
+                autoComplete='off'
+                inputMode='numeric'
+                aria-invalid={startError !== null}
+                aria-describedby={startError !== null ? 'renumber-start-error' : undefined}
+                className={startError !== null ? style.inputInvalid : undefined}
+              />
+              <div className={style.fieldErrorSlot} aria-live='polite'>
+                {startError !== null ? (
```
cpvalente: suggestion: can we simplify to something like

```suggestion
                {startError && (
```

we usually avoid ternaries with null rendering unless we are guarding numbers

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+                  <p id='renumber-start-error' className={style.fieldError} role='alert'>
+                    {startError}
+                  </p>
+                ) : null}
+              </div>
+            </div>
+            {error !== null && <p className={style.error}>{error}</p>}
```
cpvalente: suggestion: we usually use something like 
```suggestion
            {error && <div className={style.error}>{error}</div>}
```

There are other similar instances in this component

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+          </form>
+        </div>
+      }
+      footerElements={
+        <div className={style.footer}>
```
cpvalente: cleanup: the modal and dialog elements already have footer wrapper element, we can simply pass the buttons

this means one less element and more consistency in design

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+  );
+}
+
+interface RenumberCuesStore {
+  isOpen: boolean;
+  onClose: () => void;
+  onOpen: () => void;
+}
+
+export const useRenumberCuesDisclosure = create<RenumberCuesStore>()((set) => ({
```
cpvalente: suggestion: we usually use the mantine hook for disclosure control. if we want to use a store I would suggest creating a new file with this. That gives us the advantage of seeing all the things that the feature has
- styles + store + components

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -270,6 +269,31 @@ export async function reorderEntry(entryId: EntryId, destinationId: EntryId, ord
...
+ * Assigns sequential cues to the given event ids (step 1): prefix + (start + index).
+ * @throws if an id is missing or not an Ontime event
+ */
+export async function renumberEntries(ids: EntryId[], prefix: string, start: number): Promise<Rundown> {
+  const { rundown, commit } = createTransaction({ mutableRundown: true, mutableCustomFields: false });
+
+  for (let i = 0; i < ids.length; i++) {
+    const currentId = ids[i]
+    const currentEntry = rundown.entries[currentId]
+    if (!currentEntry || !isOntimeEvent(currentEntry)) throw new Error('A given id was not an event')
```
cpvalente: nitpick, we rarely use inline returns

```suggestion
    if (!currentEntry || !isOntimeEvent(currentEntry)) {
       throw new Error('A given id was not an event')
     }
```

packages/utils/src/cue-utils/cueUtils.ts
```diff
@@ -1,98 +1,62 @@
...
-import { isNumeric } from '../types/types.js';
+import { getFirstEventNormal, getPreviousEventNormal } from '../rundown-utils/rundownUtils.js';
 
-// Zero or more non-digit characters at the beginning ((\D*)).
-// One or more digits ((\d+)).
-// Optionally, a decimal part starting with a dot ((\.\d+)?).
-const regex = /^(\D*)(\d+)(\.\d+)?$/;
+
+// Groups: 1=prefix, 2=separator(optional dash or space), 3=integer, 4='.', 5=fraction
+const regex = /^(\D*?)(?:([ -]))?(\d+)(?:(\.)(\d+))?$/
```
cpvalente: ouch, I am really not that big into regex, could we generate some simple tests to lock this in?

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+function validateStart(raw: string): string | null {
+  const trimmed = raw.trim();
+  if (trimmed === '') {
+    return 'Start is required';
+  }
+  const n = Number(trimmed);
+  if (!Number.isFinite(n)) {
+    return 'Start must be a valid number';
+  }
```
cpvalente: question: is there a reason to guard against infinity rather than NaN?

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+import Button from '../../../common/components/buttons/Button';
+import Input from '../../../common/components/input/input/Input';
+import Modal from '../../../common/components/modal/Modal';
+import { useEntryActionsContext } from '../../../common/context/EntryActionsContext';
+import { useEventSelection } from '../useEventSelection';
+
+import style from './RenumberCuesModal.module.scss';
+
+function validateStart(raw: string): string | null {
```
cpvalente: cleanup: can we move this into an utilities file? i can stay in this directory

apps/client/src/features/rundown/renumber-cues-modal/RenumberCuesModal.tsx
```diff
@@ -0,0 +1,135 @@
...
+  return (
+    <Modal
+      isOpen={isOpen}
+      onClose={onClose}
+      title='Renumber cues'
+      showCloseButton
+      showBackdrop
+      bodyElements={
+        <div className={style.body}>
+          <form id='renumber-cues-form' onSubmit={handleSubmit} className={style.fields}>
```
cpvalente: suggestion: the modal already has a body wrapper, we likely dont need it
we can remove that and keep just the form, any necessary styles can be in the form instead

apps/client/src/common/hooks/useEntryAction.ts
```diff
@@ -509,6 +510,41 @@ export const useEntryActions = () => {
...
+    },
+  });
+
+  const renumberCues = useCallback(
+    async (eventIds: EntryId[], prefix: string, start: number) => {
+      const rundown = getCurrentRundownData();
+      const rundownId = rundown?.id;
+      if (!rundownId) throw new Error('Rundown not initialized');
+      try {
+        await renumberCuesMutation([rundownId, { ids: eventIds, prefix, start }]);
```
cpvalente: question: should we make sure that the entries are in the right order? otherwise we may get scrambled reordering. or maybe that is a feature?

apps/server/src/api-data/rundown/rundown.validation.ts
```diff
@@ -83,4 +83,12 @@ export const rundownArrayOfIds = [
   requestValidationFunction,
 ];
 
+export const entryRenumberValidator = [
+  body('ids').isArray({ min: 2 }).notEmpty(),
```
cpvalente: i am not sure if we really need to guard this. 
the option is available in multi select, but even if user creates it with a single select.

The flow is also clear and unsurprising
In general we prefer having less code.

apps/server/src/api-data/rundown/rundown.validation.ts
```diff
@@ -83,4 +83,12 @@ export const rundownArrayOfIds = [
...
+  body('prefix').isString(),
+  body('start').isInt(),
+  //TODO: body('increment').isNumeric(),
```
cpvalente: question: is there a reason not to implement this?

cpvalente (commented):
Few notes. 

- I think we may need an increment option so that users can use numbers as a grouping indicator (like 2.1)
- the modal can be a lot smaller, maybe we need to make a dialog variant for it? or maybe look at the other dialogs like RenameClient as a compromise
- currently we use selection order, I am not sure if that is a good idea. maybe can use this to make sure the entries are in the right order? https://github.com/cpvalente/ontime/blob/master/apps/client/src/features/rundown/rundown.utils.ts#L243

the zustand state looks fine

apps/server/src/api-data/rundown/rundown.validation.ts
alex-Arc: no, just a quick first pass

apps/client/src/common/components/modal/Modal.module.scss
```diff
@@ -40,15 +45,17 @@
 }
 
 .body {
+  flex: 1 1 auto;
+  min-height: 0;
   display: flex;
   flex-direction: column;
   gap: 0.5rem;
 
-  max-height: 60vh;
```
cpvalente, line 60: this used to be here to stop the body from being too large. I am not too strong in CSS and maybe the flex changes achieve the same. but make sure to double check the share modal in the cuesheet

apps/client/src/features/rundown/entry-editor/composite/EventEditorTitles.tsx
```diff
@@ -25,7 +24,7 @@ function EventEditorTitles({ eventId, cue, flag, title, note, colour }: EventEdi
   const { updateEntry } = useEntryActionsContext();
 
   const cueSubmitHandler = (_field: string, newValue: string) => {
-    updateEntry({ id: eventId, cue: sanitiseCue(newValue) });
+    updateEntry({ id: eventId, cue: newValue.trim() });
```
cpvalente: question: would it be good to keep the sanitiseCue even if it just calls trim? this part of the code seems to be coordination and i wouldnt expect logic here

apps/client/src/features/rundown/renumber-cues-dialog/renumber.utils.tsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+  return null;
+}
+
+export function validateIncrementInput(raw: string): string | null {
+  const trimmed = raw.trim();
+  if (trimmed === '') {
+    return 'Increment is required';
+  }
+  const n = Number(trimmed);
```
cpvalente: question: could we give n a name?

apps/client/src/features/rundown/renumber-cues-dialog/RenumberCuesDialog.tsx
```diff
@@ -0,0 +1,155 @@
...
+  );
+}
+
+interface RenumberCuesDialogState {
+  isOpen: boolean;
+  onClose: () => void;
+  onOpen: () => void;
+}
+
+export const useRenumberCuesDialogStore = create<RenumberCuesDialogState>()((set) => ({
```
cpvalente, line 123: question: can you explain why we need a store fort this?

apps/client/src/features/rundown/renumber-cues-dialog/RenumberCuesDialog.tsx
```diff
@@ -0,0 +1,155 @@
...
+        onChange={(e) => {
+          onChange(e.target.value);
+        }}
+        height='large'
+        fluid
+        autoComplete='off'
+        inputMode='numeric'
+        className={error !== null ? style.inputInvalid : undefined}
+      />
+      <div className={style.fieldErrorSlot} aria-live='polite'>
```
cpvalente: suggestion: this type of markup would cause reflow on errors, usually we either substitute the error by the label or append it to the label

apps/client/src/features/rundown/rundown-event/RundownEvent.tsx
```diff
@@ -143,6 +145,13 @@ export default function RundownEvent({
             disabled: parent !== null,
           },
           { type: 'divider' },
+          {
+            type: 'item',
+            label: 'Renumber cues',
+            icon: TbListNumbers,
+            onClick: () => openRenumberDialog(),
```
cpvalente: maybe we dont need to create the new function?

```suggestion
            onClick: openRenumberDialog,
```

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -270,6 +269,53 @@ export async function reorderEntry(entryId: EntryId, destinationId: EntryId, ord
   return rundownResult;
 }
 
+/**
+ * @throws if an id is missing or not an Ontime event
+ */
+export async function renumberEntries(
```
cpvalente: suggestion: the renumber logic and types looks a bit complicated, could we extract or make the logic clearer?

extracting the logic into a pure function would also allow unit testing

apps/client/src/features/rundown/renumber-cues-dialog/RenumberCuesDialog.tsx
alex-Arc: alternative is passing down through prorps or context?

cpvalente: The store is recreating our `useDisclosure` hook. 
I didnt read the code too carefully, but the store is only worth it if there are multiple consumers to it. otherwise we are better off with a hook

cdslipp:
@alex-Arc this looks great! I tried with a large rundown and it all works as expected. 

The only small tweak that I might want, is when using a decimal increment, we should allow for a threshold to bump over to the next number.

Example - If I have 55 events and I renumber from 1 with an increment of 0.1, the following happens:
image

This results in the weird scenario where the system treats number like 0.11 > 0.9 or 0.53 > 0.6 even though this is not numerically true (although I understand it semantically as if we were talking purely about linear point cues and not mathematically)

I would prefer a system where the following occurs:

**Starting: 1 | Increment: 0.1:** 
1.0, 1.1 .... 1.9, 2.0, 2.1, 2.2 etc

**Starting: 1 | Increment : 0.01:** 
1.00, 1.01, 1.02, .... 1.09, 1.10, 1.11 .... 1.99, 2.00, 2.01

As a user I am super happy with how the feature function now, and this is an edge case - but I'm just being super pedantic in case we are able to catch this now before merging. Thank you again, this is an awesome feature to have!

alex-Arc:
Good to hear

> The only small tweak that I might want, is when using a decimal increment, we should allow for a threshold to bump over to the next number.

while i agree that it would be nice to have proper role-over, it is quite complex to handle
we can perhaps take it up again in the another timer

apps/client/src/common/api/rundown.ts
```diff
@@ -111,6 +119,13 @@ export async function putBatchEditEvents(rundownId: RundownId, data: BatchEditEn
   return axios.put(`${rundownPath}/${rundownId}/batch`, data);
 }
 
+/**
+ * HTTP request to renumber cues for multiple events
+ */
+export async function patchRenumberCues(rundownId: RundownId, data: RenumberCues): Promise<AxiosResponse<Rundown>> {
+  return axios.patch(`${rundownPath}/${rundownId}/renumber`, data);
```
cpvalente, line 126: Maybe we can unwrap the axios object here so it doesn’t leak into our app.

We don’t always do it, but should

apps/client/src/features/rundown/renumber-cues-dialog/RenumberCuesDialog.module.scss
```diff
@@ -0,0 +1,22 @@
...
+
+.fields {
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  padding-inline: 0.5rem;
+  margin-bottom: 0.5rem;
+}
+
+label {
```
cpvalente: Hmm, not sure about adding the styling to the element. I think it could make weird and hard to debug style issues

CSS is not usually something that you read or organise as well as code

apps/client/src/features/rundown/renumber-cues-dialog/RenumberCuesDialog.tsx
```diff
@@ -0,0 +1,131 @@
...
+import useRundown from '../../../common/hooks-query/useRundown';
+import { orderEntries } from '../rundown.utils';
+import { useEventSelection } from '../useEventSelection';
+
+import style from './RenumberCuesDialog.module.scss';
+
+type RenumberCueData = Pick<RenumberCues, 'increment' | 'prefix' | 'start'>;
+
+export default function RenumberCuesDialog() {
+  'use memo';
```
cpvalente, line 19: As far as I remember RHF and react compiler don't really get along

https://github.com/orgs/react-hook-form/discussions/12524

I don't think this has been resolved (?) and it is likely safer to remove the directive in favour of manual memoisation (if needed)

apps/client/src/features/rundown/renumber-cues-dialog/RenumberCuesDialog.tsx
```diff
@@ -0,0 +1,131 @@
...
+                placeholder='10'
+              />
+            </label>
+          </div>
+          <div className={style.field}>
+            <label>
+              Increment
+              <Input
+                {...register('increment')}
+                type='number'
```
cpvalente, line 90: Tip: RHF exposes valueAsNumber which may help

https://react-hook-form.com/advanced-usage#TransformandParse

apps/server/src/api-data/rundown/rundown.router.ts
```diff
@@ -28,14 +30,15 @@ import {
...
+  entryRenumberValidator,
   entryReorderValidator,
   entrySwapValidator,
   rundownArrayOfIds,
   rundownPostValidator,
   validateRundownMutation,
 } from './rundown.validation.js';
 
-export const router = express.Router();
+export const router: Router = express.Router();
```
cpvalente, line 41: Nice!

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -270,6 +269,50 @@ export async function reorderEntry(entryId: EntryId, destinationId: EntryId, ord
...
+  const startNumber = getIntegerAndFraction(start);
+  const incrementNumber = getIntegerAndFraction(increment);
+  const maxPrecision = Math.max(incrementNumber.precision, startNumber.precision);
+
+  // if the prefix doesn't already include a separator or is empty, then insert a separator
+  if (prefix !== '' && !prefix.endsWith('-') && !prefix.endsWith(' ')) prefix += '-';
+
+  const { rundown, commit } = createTransaction({ mutableRundown: true, mutableCustomFields: false });
+
+  for (let i = 0; i < ids.length; i++) {
```
cpvalente: Question: would this code be more similar to other code that achieves similar things if we extracted the actual logic into a pure function?

apps/server/src/api-data/rundown/rundown.utils.ts
```diff
@@ -470,3 +473,16 @@ export function duplicateRundown(rundown: Rundown, newTitle: string): Rundown {
 
   return newRundown;
 }
+
+export function getIntegerAndFraction(value: string) {
```
cpvalente: Question: can we add a JSdoc to help understand what we expect out of this function or in which cases we may want to use it?

apps/server/src/api-data/rundown/rundown.validation.ts
```diff
@@ -83,4 +83,12 @@ export const rundownArrayOfIds = [
   requestValidationFunction,
 ];
 
+export const entryRenumberValidator = [
+  body('ids').isArray().notEmpty(),
+  body('prefix').isString(),
+  body('start').isNumeric(),
```
cpvalente: Question: is there some validation we could be doing here instead?

apps/client/src/features/rundown/renumber-cues-dialog/RenumberCuesDialog.tsx
alex-Arc: we don't want to take away leading zero's
and we want to avoid floating point issues

apps/server/src/api-data/rundown/rundown.validation.ts
alex-Arc: maybe a bit, but not anything the really  helps the later functions, I think?

apps/server/src/api-data/rundown/rundown.service.ts
alex-Arc: I can extract it, but don't really see where the similarity is?

apps/client/src/common/api/rundown.ts
alex-Arc: is there a reason to send back the whole rundown here?
we always do it, but rarely use it?

cpvalente: we should use the response rundown to patch the local cache

apps/server/src/api-data/rundown/rundown.service.ts
cpvalente: none of the other functions in rundown.service contain business logic.

to make this function like the others, we would either
- make the mutation in the rundown.cache layer
- iterate here and have utilities or a mutation function elsewhere

the point is that service functions should read as high level logic and not contain details

apps/client/src/common/api/rundown.ts
alex-Arc: ok
just tested this and i can get it working but the re-fetch message over socket arrives before the http responds, should we slow that down ?

cpvalente: Hm, curious. I would have imagined that out setImmediate scheduling would have sorted it.

I don't think we should add artificial delays, but the system should be more harmonious. Maybe the refetch signal should give a change id so receivers can decide whether to fetch the change?

Not a change for this PR I guess

alex-Arc: sounds right
I think there might be timeouts inside express as well

apps/client/src/features/rundown/renumber-cues-dialog/RenumberCuesDialog.tsx
alex-Arc: I might be reading it wrong but seams to be mainly a issue with `watch`

PR #2003, opened 2026-03-13
upgrade vite and related dependencies

jwetzell:

PR #2018, opened 2026-03-20
Vite 8

alex-Arc:
continue the attempt from #2003 to get Vite 8 working
~managed by following this https://github.com/rosskhanas/react-qr-code/issues/285~
dropped the react-qr-code and for out own QR component


alex-Arc:
@coderabbitai review

cpvalente:
Hey @alex-Arc 

thank you for this

the QRCode is a low value dependency, at least how we use it. I would suggest that as long as we don’t want to customise the code, we can trivially make our own QR code without a library and shake off the dependency

apps/client/src/common/components/qr-code/QrCode.tsx
```diff
@@ -0,0 +1,48 @@
...
+type QrLevel = 'L' | 'M' | 'Q' | 'H';
+
+interface QrCodeProps {
+  value: string;
+  size: number;
+  className?: string;
+  level?: QrLevel;
+}
+
+export default function QrCode({ value, size, className, level = 'L' }: QrCodeProps) {
```
cpvalente: Maybe we can have less changes if this is called QRCode?

apps/client/src/common/components/qr-code/QrCode.tsx
```diff
@@ -0,0 +1,48 @@
+import { toString } from 'qrcode';
+import { useEffect, useState } from 'react';
+
+type QrLevel = 'L' | 'M' | 'Q' | 'H';
+
+interface QrCodeProps {
+  value: string;
+  size: number;
+  className?: string;
+  level?: QrLevel;
```
cpvalente: Question: I don’t think we ever configured this. Can we just hard code?

apps/client/src/common/components/qr-code/QrCode.tsx
```diff
@@ -0,0 +1,29 @@
...
+  'use memo';
+
+  const svgPromise = toString(value, {
+    width: size,
+    margin: 2,
+    color: { dark: '#101010', light: '#cfcfcf  ' },
+  });
+
+  return (
+    <Suspense fallback={null}>
```
cpvalente: The fallback could be a skeleton with the the right size to avoid reflow

apps/client/src/common/components/qr-code/QrCode.tsx
```diff
@@ -0,0 +1,29 @@
...
+  size: number;
+}
+
+export default function QRCode({ value, size }: QRCodeProps) {
+  'use memo';
+
+  const svgPromise = toString(value, {
+    width: size,
+    margin: 2,
+    color: { dark: '#101010', light: '#cfcfcf  ' },
```
cpvalente:
```suggestion
    color: { dark: '#101010', light: '#cfcfcf' },
```

apps/client/src/common/components/qr-code/QrCode.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <Suspense fallback={null}>
+      <QrSvg svgPromise={svgPromise} />
+    </Suspense>
+  );
+}
+
+function QrSvg({ svgPromise }: { svgPromise: Promise<string> }) {
+  'use memo';
+  const svg = use(svgPromise);
+  return <span dangerouslySetInnerHTML={{ __html: svg }} />;
```
cpvalente: Question: did we not lose our nice padding and rounded corners? I would have expected the qr styles to have moved here

alex-Arc: yes we did, thought it was not to important 

have added it back

PR #2024, opened 2026-03-21
Adding playback actions to automations

cdslipp:
Added primary playback actions as available ontime actions in the automations.

In a roundabout way this is a much simpler solution to PR #1991 (thank you @cpvalente for the suggestion!)

This allows for the same functionality as the #1991 by leveraging automations to trigger a start event to fire when an event start, allowing the user to cause Roll Mode to end dynamically.

Main UI change is just the added "Playback" options in this select:
image

Added options:

- Playback: start
- Playback: stop
- Playback: pause
- Playback: roll

### DIFF
--- a/apps/client/src/common/components/tag/Tag.module.scss
+++ b/apps/client/src/common/components/tag/Tag.module.scss
@@ -1,9 +1,17 @@
 .tag {
   font-size: calc(1rem - 3px);
   letter-spacing: 0.5px;
-  background-color: $gray-900;
-  color: $ui-white;
   border-radius: 2px;
   padding: 0 0.25rem;
   white-space: nowrap;
+}
+
+.default {
+  background-color: $gray-900;
+  color: $ui-white;
+}
+
+.warning {
+  background-color: $orange-1300;
+  color: $orange-300;
 }
--- a/apps/client/src/common/components/tag/Tag.tsx
+++ b/apps/client/src/common/components/tag/Tag.tsx
@@ -1,11 +1,14 @@
 import { PropsWithChildren } from 'react';
 
+import { cx } from '../../utils/styleUtils';
+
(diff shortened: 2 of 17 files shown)


packages/types/src/definitions/core/Automation.type.ts
```diff
@@ -1,5 +1,5 @@
-import type { SecondarySource } from '../runtime/MessageControl.type.js';
-import type { TimerLifeCycle } from './TimerLifecycle.type.js';
+import type { SecondarySource } from "../runtime/MessageControl.type.js";
```
cpvalente: cleanup: remember to run oxc formatting and we should reduce most of the changeset in this file.
we dont have CI enforcement of formatting, developers need to remember to run it manually (I have it set to format on save)

cpvalente:
Thank you @cdslipp , this seems to work well and the change is minimal

It is maybe confusing for users that automations can be off in the settings (and are by default) and they dont see why things are not working. Maybe we can try to improve that, by either changing defaults or making a clearer UI?

I am also not sure what the flickering you mention is, but I found an issue with the add filter form and made a PR #2025.

cdslipp:
> Thank you @cdslipp , this seems to work well and the change is minimal
> 
> It is maybe confusing for users that automations can be off in the settings (and are by default) and they dont see why things are not working. Maybe we can try to improve that, by either changing defaults or making a clearer UI?
> 
> I am also not sure what the flickering you mention is, but I found an issue with the add filter form and made a PR #2025.

I agree with this (the fact that automations can be turned off), and I see you already made some initial improvements in your other PR - to this end I also found myself confused when adding automations to an event, as the "add" button step at the end feels unnecessary. I will take a look at some of this stuff in another PR!

cpvalente (approved):
nice! great work

PR #2025, opened 2026-03-22
Improve automation flow

cpvalente:
Fixes an issue with flashing in the automation form when opening the add filter. Potentially a fix for what is mentioned in #2024 

Also, contains several UI only tweaks to improve the flow of managing automations, focusing on warning the user whether the feature is active

Frame 61 (1)

### DIFF
--- a/apps/client/src/common/components/select/Select.module.scss
+++ b/apps/client/src/common/components/select/Select.module.scss
@@ -125,6 +125,11 @@
   font-size: 0.5rem;
   display: grid;
   place-content: center;
+  visibility: hidden;
+
+  &[data-visible] {
+    visibility: visible;
+  }
 
   &::before {
     content: '';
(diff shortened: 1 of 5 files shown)


cpvalente:
@coderabbitai full review

cdslipp:
***Update:*** Leaving repro steps below, but I was able to fix this on #2024 by adding the keepMounted prop to Base.ScrollDownArrow and ScrollUpArrow - added new css to hide arrows

---

@cpvalente I figured out the repro steps for the flashing on the Select component - you have to scroll the closed select very close to the viewport boundary and then highlight overflow arrow (either direction). Once you hover that arrow, the state of the select flashes and becomes unstable and it is not possible to select any options without closing, adjusting your scroll, and reopening.

image

https://github.com/user-attachments/assets/5e45440e-8f1d-41cb-a745-15b8ae94e11a

alex-Arc (approved):
cool, looks and works well

could the `automationsEnabled` prop perhaps be something better passed down in a context?
as it is a value that is not unique the the separate events

cpvalente:
> cool, looks and works well
> 
> could the `automationsEnabled` prop perhaps be something better passed down in a context?
> as it is a value that is not unique the the separate events

That was my fault

I felt like the whole react query state was becoming a bit weird and i didnt want to complicate it

I suggest we leave as is and then extract it when we get to multiple rundowns?

PR #2028, opened 2026-03-23
feat: let docker default SIGTERM shutdown normaly

alex-Arc:


apps/server/src/app.ts
```diff
@@ -369,4 +369,11 @@ process.on('uncaughtException', async (error) => {
 // register shutdown signals
 process.once('SIGHUP', async () => shutdown(0));
 process.once('SIGINT', async () => shutdown(0));
-process.once('SIGTERM', async () => shutdown(3));
+process.once('SIGTERM', async () => shutdown(0));
+process.once('SIGTKILL', async () => shutdown(3));
```
cpvalente, line 373: I assume this is meant to say SIGKILL

it looks like SIGKILL cannot be listened to so we may as well remove it

https://nodejs.org/api/process.html

PR #2030, opened 2026-03-24
chore: format

alex-Arc:


cpvalente:
Question: should we add a formatting step check in CI?

PR #2058, opened 2026-04-12
refactor: improve the amount of GET requests for translation assets

alex-Arc:
switch `refetchInterval` for `staleTime` in translation qury as we have refetch key for it so no need to check


cpvalente (approved):
Thank you!

Issue #2038, opened 2026-03-25
Adjust the flow for adding automations & enable editing existing automations


cdslipp:
I think it would be worthwhile to revisit the flow for adding automations. I find the initial state to be a bit confusing. I think my main issue is with the "Add Automation" button. See in the images below - the current state is that the button is there and greyed out but the selects are semi-populated for the first automation. 

<img width="1010" height="356" alt="Image" src="https://github.com/user-attachments/assets/73c8e94d-ae73-4e65-843e-fd3f95ec4c5f" />

<img width="996" height="482" alt="Image" src="https://github.com/user-attachments/assets/091d2db0-dab0-496f-a729-defc5665489e" />

To me, I think it is confusing that this section starts semi-populated while elsewhere in Ontime, things like custom fields (and even the main rundown) have the patterns of starting from an empty state with a button to add something. I would like to see an approach where the automation section is initially truly empty with an "Add Automation" button that you click to start adding automations, which then "auto save" effectively once they are edited.  Furthermore, this would enable the UX pattern of being able to easily edit automations instead of needing to delete and re-add. Does this make sense? 

Maybe there are other considerations with the way the automations server routes work, but I think it would be worth revising this flow to make it more evident when it is in a truly Empty state, and avoiding the confusion of thinking you have added an automation to a series event, however for one of the events you forgot to press "Add Automation" after selecting your settings.

cpvalente:
Hi @cdslipp , I agree with you

We often prefer empty states.
Would you like to give a go at improving this?

cdslipp:
Yeah, I'll take a stab at a PR - sometime this weekend maybe

cpvalente:
We had a few rounds of UX review here and the interface state is fairly changed since the ticket was opened.
I am closing the ticket, if we have new suggestions we should open a new ticket that accounts for the interface current state

PR #2060, opened 2026-04-15
refactor: adjusting flow for adding automations

cdslipp:
This addresses issue #2038 

Streamlining the editor flow for automations like so: 
image

This means that automations auto-save but they are also editable now... I have tested with ontime actions but have not tested with OSC or HTTP automations yet.

### DIFF
--- a/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
+++ b/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
@@ -4,18 +4,6 @@
   gap: 0.75rem;
 }
 
-.section,
-.formSection {
-  display: flex;
-  flex-direction: column;
-  gap: 0.5rem;
-}
-
-.sectionTitle {
-  font-size: $aux-text-size;
-  color: $label-gray;
-}
-
 .triggerList {
   display: flex;
   flex-direction: column;
@@ -25,25 +13,14 @@
   overflow: hidden;
 }
 
-.triggerForm {
-  display: flex;
-  flex-direction: column;
-  gap: 0.5rem;
-}
-
-.formFields {
+.triggerHeader {
   display: grid;
-  grid-template-columns: 8rem 1fr;
-  gap: 0.75rem;
-  align-items: end;
-}
-
-.formActions {
-  display: flex;
-  justify-content: space-between;
-  align-items: center;
-  gap: 0.75rem;
-  min-height: 2rem;
+  grid-template-columns: 8rem 1fr 2rem;
+  gap: 0.5rem;
+  padding: 0.375rem 0.75rem;
+  font-size: $aux-text-size;
+  color: $label-gray;
+  border-bottom: 1px solid $white-10;
 }
 
 .trigger {
@@ -51,40 +28,10 @@
   display: grid;
   grid-template-columns: 8rem 1fr 2rem;
   align-items: center;
+  gap: 0.5rem;
   min-height: 2.5rem;
 
   &:not(:first-child) {
     border-top: 1px solid $white-10;
   }
-}
-
-.triggerMeta {
-  min-width: 0;
-}
-
-.metaLabel {
-  color: $label-gray;
-  font-size: $aux-text-size;
-  line-height: 1.2;
-  margin-bottom: 0.125rem;
-}
-
-.automationTitle {
-  overflow-wrap: anywhere;
-}
-
-.validationError,
-.validationSuccess {
-  display: flex;
-  align-items: center;
-  gap: 0.375rem;
-  font-size: $aux-text-size;
-}
-
-.validationError {
-  color: $red-500;
-}
-
-.validationSuccess {
-  color: $green-500;
 }
--- a/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
+++ b/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
@@ -33,4 +33,16 @@
   &:not(:first-child) {
     border-top: 1px solid $white-10;
   }
+
+  &[data-duplicate] {
+    > button:not(:last-child) {
+      border-color: $orange-500;
+    }
+  }
+}
+
+.duplicateMessage {
+  padding-left: 0.75rem;
+  font-size: $aux-text-size;
+  color: $orange-500;
 }
--- a/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
+++ b/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
@@ -20,7 +20,6 @@
   padding: 0.375rem 0.75rem;
   font-size: $aux-text-size;
   color: $label-gray;
-  border-bottom: 1px solid $white-10;
 }
 
 .trigger {
--- a/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
+++ b/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
@@ -1,4 +1,4 @@
-import { NormalisedAutomation, TimerLifeCycle, Trigger, timerLifecycleValues } from 'ontime-types';
+import { TimerLifeCycle, Trigger, timerLifecycleValues } from 'ontime-types';
 import { generateId } from 'ontime-utils';
 import { Fragment, useCallback } from 'react';
 import { IoAdd, IoTrash } from 'react-icons/io5';
--- a/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
+++ b/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
@@ -123,7 +123,7 @@ export default function EventEditorTriggers({ triggers, eventId }: EventEditorTr
       )}
       {allAutomationOptions.length === 0 ? (
         <Editor.Label>
-          No automations defined. <a href='?settings=automation'>Manage automations</a> to add some.
+          No automations defined.
         </Editor.Label>
       ) : (
         <Button variant='ghosted' onClick={handleAdd} disabled={getNextAvailable() === null}>
--- a/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
+++ b/apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
@@ -1,7 +1,7 @@
 import { NormalisedAutomation, TimerLifeCycle, Trigger, timerLifecycleValues } from 'ontime-types';
 import { generateId } from 'ontime-utils';
-import { Fragment, useCallback, useMemo, useState } from 'react';
-import { IoAlertCircle, IoCheckmarkCircle, IoTrash } from 'react-icons/io5';
+import { Fragment, useCallback } from 'react';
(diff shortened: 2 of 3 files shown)


cdslipp:
Ahhh just remembered that I need to deal with the edge case with two of the same automation: ie this situation from the old UI: 
image

Latest commit has my new approach: remove conflicting options from the select so it is impossible to cause this scenario to occur, therefore we no longer need to show warning. Try it out

| Conflicting | Not Conflicting |
|--------|--------|
| When conflicting, we remove the offending option from second select: image | When there is no conflict (ie a different trigger point is selected) we show all options in the select image|

cpvalente:
@coderabbitai full review

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
```diff
@@ -4,18 +4,6 @@
   gap: 0.75rem;
```
cpvalente, line 4: nice, love the simplification

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
```diff
@@ -21,174 +20,114 @@ interface EventEditorTriggersProps {
...
         </div>
-        <Button
-          disabled={validationError !== undefined}
-          onClick={() => automationId && handleSubmit(cycleValue, automationId)}
-        >
-          Add automation
+      )}
+      {allAutomationOptions.length === 0 ? (
+        <Editor.Label>
+          No automations defined. <a href='?settings=automation'>Manage automations</a> to add some.
```
cpvalente: tip: we have an AppLink abstractions for links within the interface. it handles some edge cases around navigation and also makes sure that our links have a consistent look. we likely want to use that

If we do so, I think the text will look out of place, we probably just want normal sized text here

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
```diff
@@ -21,174 +20,114 @@ interface EventEditorTriggersProps {
...
-          />
-        </div>
-      </div>
-
-      <div className={style.formActions}>
-        <div className={validationError ? style.validationError : style.validationSuccess}>
-          {validationError ? <IoAlertCircle /> : <IoCheckmarkCircle />}
-          <span>{validationLabel}</span>
+    <div className={style.triggers}>
+      {automationsEnabled === false && (
```
cpvalente, line 65: question: should we still show this if automations are disabled? it feels at least that we shouldnt have both this and the empty placeholder, but I am not certain which way to lean. What do you think?

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
```diff
@@ -21,174 +20,114 @@ interface EventEditorTriggersProps {
...
+        <Info>Automations are disabled. Event triggers stay configured, but they will not run until enabled.</Info>
+      )}
+      {triggers.length > 0 && (
+        <div className={style.triggerList}>
+          <div className={style.triggerHeader}>
+            <span>Lifecycle</span>
+            <span>Automation</span>
+          </div>
+          {triggers.map((trigger) => (
```
cpvalente: I think we may have a bug here:
I have created two rows that have the same automation and lifetime trigger. this causes the select to show the trigger ID (?) and two entries become connected, indicating a possible key collision

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
```diff
@@ -21,174 +20,114 @@ interface EventEditorTriggersProps {
...
+                  if (value !== null) handleChange(trigger.id, 'trigger', value);
+                }}
+                options={triggerOptions}
+              />
+              <Select
+                value={trigger.automationId}
+                onValueChange={(value) => {
+                  if (value !== null) handleChange(trigger.id, 'automationId', value);
+                }}
+                options={getAutomationOptions(trigger.id, trigger.trigger)}
```
cpvalente: question: I wonder if this may be a bit expensive and we may want to do it outside the return

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
```diff
@@ -21,174 +20,114 @@ interface EventEditorTriggersProps {
...
-    ],
-    [automations], // This needs to be a dependency as it can change
-  );
+  const handleChange = (triggerId: string, field: 'trigger' | 'automationId', value: string) => {
+    const newTriggers = triggers.map((t) => {
+      if (t.id !== triggerId) return t;
+      const updated = { ...t, [field]: value };
+      // when lifecycle changes, auto-resolve automationId if it now conflicts with another row
+      if (field === 'trigger') {
+        const newLifecycle = value as TimerLifeCycle;
```
cpvalente: question: I dont quite understand the conflict resolution here, maybe there is room to simplify?
As I can see, the user is able to create identical entries (same trigger for some lifecycle)
I wonder if we should 
a) stop that from happening, this implies more UI affordances
b) simply have the server ignore it

I think option b would likely have the combination of best UX and least code, what do you think?

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
```diff
@@ -21,174 +20,114 @@ interface EventEditorTriggersProps {
...
-    return Object.values(triggers).some((t) => t.automationId === automationId && t.trigger === cycle)
-      ? 'Automation can only be used once'
-      : undefined;
+    return null;
   };
 
-  const validationError = getValidationError(cycleValue, automationId);
-  const validationLabel = validationError ?? 'Ready to add automation';
+  const handleAdd = () => {
+    const next = getNextAvailable();
```
cpvalente: question: do you think that this helps UX?
I personally often err on the side of less code when in doubt, I am not sure we need this.

If you feel strongly about it, then I think it is OK

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
```diff
@@ -25,66 +13,25 @@
...
-  justify-content: space-between;
-  align-items: center;
-  gap: 0.75rem;
-  min-height: 2rem;
+  grid-template-columns: 8rem 1fr 2rem;
+  gap: 0.5rem;
+  padding: 0.375rem 0.75rem;
+  font-size: $aux-text-size;
+  color: $label-gray;
+  border-bottom: 1px solid $white-10;
```
cpvalente: tip: this border causes a double-border from the header to the first element. I dont think we need it

cpvalente (commented):
Good work, I have left some minor comments.
We need to look at the issue with the keys, I would also consider simplifying the logic

look forwards to having this into ontime

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
cdslipp: Yeah, this link was redundant anyway, I looked at it and just got rid of it - manage automations above is enough

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.module.scss
cdslipp: You're right! Removed!

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
cdslipp: You are right, I need to look at the server side more often I'm definitely biased to client side - adding a server side fix for this instead so that it is irrelevant

And I am attempting to hide conflicts from the selects so that it is impossible to create conflicts and user doesn't have to think about it. I missed the edge case of only 1 automation existing though before, that is resolved now

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
cdslipp: I just think it is much much cleaner to make it impossible for a user to do something wrong than to allow them to select an invalid state and then show an error message - I like this approach because it makes automations feel a lot less confusing, the user doesn't need to think about why they are being given warnings, the app just doesn't allow them to put the rundown automations into a bad state at all

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
cdslipp: I prefer this one - I got rid of the empty placeholder, this is better and custom fields doesn't show anything when empty, so automations should be the same for now

apps/client/src/features/rundown/entry-editor/composite/EventEditorTriggers.tsx
cpvalente: When in doubt we usually follow a "dont forbid" principle, where we let users make states even if we think they are weird. This meaning:
- The interface should warn, but not forbid
- Users should not be surprised by an outcome

Having the same action - trigger combination many times is not a corrupt state, just redundant.
If we forbid, we end up in situations where, to change an action from a trigger to another we need to have an intermediate step. 

Overall, I would agree that we should guide users into making good decisions, but the balance here between flexibility <-> correctness comes at the cost of extra code and complicated flows (like the previous flow demonstrated)

cpvalente: For reference, see how the same issue is handled in the global automations in the settings

Screenshot 2026-04-19 at 07 30 11

I think we could keep a map of triggered automations in the backend, and ensure that we dont run an automation more than once. this would solve the issue and also deduplicate across global and event automations

apps/server/src/api-data/automation/automation.service.ts
```diff
@@ -47,7 +47,15 @@ export function triggerAutomations(cycle: TimerLifeCycle) {
     return;
   }
 
-  filteredTrigger.forEach((trigger) => {
+  // deduplicate — if the same automation appears multiple times on one lifecycle, fire it once
+  const seen = new Set<string>();
+  const uniqueTriggers = filteredTrigger.filter((t) => {
```
alex-Arc, line 52: what are we guarding here?
Is it the case where the user has put the same automation on a global trigger and on an event. I'm not sure if we should be preventing this.
I agree that it could be a footgun but I would lean more to making that check in the UI a show a warning

alex-Arc: @cpvalente I might be countering your comment here https://github.com/cpvalente/ontime/pull/2060/changes#r3106330056
we can do this but if we filter out duplicates we have to tell the user

cpvalente: @alex-Arc , I think that the deduplication is good. 
Generally speaking, reusing the trigger is not a valid flow since the trigger will send the same data several times

Deduplicating it allow us to keep the interface simpler (warn and not forbid) and swallow the issue in the server.

The previous approach was forbidding the duplication which
- did not work across global and event triggers
- made UX more complicated
- is against our "do not forbid" principle

cdslipp:
Updated UX now - note that I have removed the client side de-deduplication logic, it is only server side now: 
image

cpvalente (approved):
nice, thank you!

Issue #2062, opened 2026-04-16
BUG: Some Display parameters not parsed when using URL presets with the 'Lock configuration' toggle


MrLazerface:
**Problem:**
URL presets using specific time format parameters display inconsistently when using the link and QR code generator with the 'Lock configuration' option enabled. In this instance, the lock configuration toggle unhides the seconds on a clock display I'm generating a share link for, where the intent is to hide seconds.

URL presets that end in '/PresetName' display normally with the seconds display hidden on a time of day clock view.
URL presets that end in '/PresetName?n=1' for locking navigation display normally as above.
When using the Lock config option, URL presets are a different format, '/preset/PresetName?n=1', and seconds are unhidden.

**Steps to reproduce:**
1. Create a URL preset with the following parameters to a Timer target:
`timeformat=h%3Amm+a&hideTimerSeconds=true&timerType=clock&main=none&hideCards=true&hideProgress=true&hideMessage=true&hideSecondary=true`
2. Use the 'Share Ontime Link' tool with the created preset from above and select the 'Lock configuration' option and Create a share link
3. Open the copied link in a browser window to observe the behaviour

Other notes:
- Version 4.7.0
- Running Ontime headless on a ubuntu cloud server, so my links use FQDNs instead of IP addresses, but I have confirmed this bug in both addressing scenarios.

PR #2065, opened 2026-04-17
fix: preset/routes weren't retaining hide seconds params

cdslipp:
Attempt at fixing #2062

### DIFF
--- a/apps/client/src/views/common/viewUtils.ts
+++ b/apps/client/src/views/common/viewUtils.ts
@@ -101,7 +101,6 @@ export function getPropertyValue(
 type FormattingOptions = {
   removeSeconds: boolean;
   removeLeadingZero: boolean;
-  timeFormat?: string;
 };
 
 export function getFormattedTimer(
@@ -115,9 +114,10 @@ export function getFormattedTimer(
   }
 
   if (timerType === TimerType.Clock) {
-    if (!options.timeFormat) return formatTime(timer);
-    const fmt = options.removeSeconds ? options.timeFormat.replace(':ss', '') : options.timeFormat;
-    return formatTime(timer, { format12: fmt, format24: fmt });
+    if (options.removeSeconds) {
+      return formatTime(timer, { format12: 'h:mm a', format24: 'HH:mm' });
+    }
+    return formatTime(timer);
   }
 
   let timeToParse = timer;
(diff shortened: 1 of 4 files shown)


apps/client/src/views/timer/Timer.tsx
```diff
@@ -212,9 +214,12 @@ function Timer({ customFields, projectData, isMirrored, settings, viewSettings,
   );
 }
 
-function TimerAutoTickingClock() {
+function TimerAutoTickingClock({ timeFormat }: { timeFormat?: string }) {
```
cpvalente: Question: would it make a better boundary if `TimerAutoTicking` subscribes to the hook that exposes the time formatting (or gets the data some other way)?

The formatting only applies to clocks, and it makes sense to scope that in the Clock components

cpvalente (commented):
Thank you for looking into this @cdslipp, I didnt look into the root issue yet, do you mind explaining?
I wonder if we have a gap in between our params and the aliased params that goes beyond this, is it possible to create a simple test to expose the issue?

cdslipp:
> Thank you for looking into this @cdslipp, I didnt look into the root issue yet, do you mind explaining? I wonder if we have a gap in between our params and the aliased params that goes beyond this, is it possible to create a simple test to expose the issue?

Basically whenever a preset was created, the actual preset/id route would not respond to the removeSeconds boolean or the timeformat param because the **getOptionsFromParams** function in Timer.options.ts didn't do anything with either of them and didn't return them and pass them through to the view either.

And you might be right... it looks like timeformat doesn't get passed into presets on any of the views that support timeformat display. (Timer, studio, backstage, timeline, countdown). This is more complex than I thought and maybe we can fix it once with a helper function that we use across all of those views? I am going to change this PR to only deal with the removeSeconds boolean since it's simple and only effects the main Timer view. I will file another PR to try to tackle the timeformat thing

apps/client/src/views/common/viewUtils.ts
```diff
@@ -114,6 +114,9 @@ export function getFormattedTimer(
   }
 
   if (timerType === TimerType.Clock) {
+    if (options.removeSeconds) {
+      return formatTime(timer, { format12: 'h:mm a', format24: 'HH:mm' });
+    }
```
cdslipp, line 119: So to recap - this change fixes a very small edge case - in a custom Timer preset view with timerType=clock, this ensures that we still process the boolean removeSeconds and give it precedence over the time format

cpvalente: question: is this not just inverting the issue? where the time formatting string gets overridden by the hide seconds?
Maybe I am misunderstanding the issue?

if that is the case, maybe the removing of seconds should be an after pass which removes the seconds if there are any

cdslipp: You may be right... hmmm needs more consideration then, which should take priority? My assumption is we set the time format and then remove seconds? But I see what you are saying - this will reset the timeformatting option

cdslipp:
Closing: removeSeconds should NOT affect clock format strings anyway (only affects running timer)

PR #2066, opened 2026-04-21
fix: format clock from preset

### LINKED ISSUE #2062: BUG: Some Display parameters not parsed when using URL presets with the 'Lock configuration' toggle

alex-Arc:
fix #2062

### DIFF
--- a/apps/client/src/common/components/view-params-editor/common.options.ts
+++ b/apps/client/src/common/components/view-params-editor/common.options.ts
@@ -26,3 +26,16 @@ export const showLeadingZeros: ParamField = {
   type: 'boolean',
   defaultValue: false,
 };
+
+export type TimeOptions = {
+  timeformat: string | null;
+};
+
+export function getTimeOptionsFromParams(searchParams: URLSearchParams, defaultValues?: URLSearchParams): TimeOptions {
+  // Helper to get value from either source, prioritizing defaultValues
+  const getValue = (key: string) => defaultValues?.get(key) ?? searchParams.get(key);
+
+  return {
+    timeformat: getValue('timeformat'),
+  };
+}
--- a/apps/client/src/common/components/view-params-editor/common.options.ts
+++ b/apps/client/src/common/components/view-params-editor/common.options.ts
@@ -31,11 +31,9 @@ export type TimeOptions = {
   timeformat: string | null;
 };
 
-export function getTimeOptionsFromParams(searchParams: URLSearchParams, defaultValues?: URLSearchParams): TimeOptions {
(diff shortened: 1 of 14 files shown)


alex-Arc:
@cdslipp if you have time could you see if this covers the error?

apps/client/src/common/components/view-params-editor/common.options.ts
```diff
@@ -26,3 +26,16 @@ export const showLeadingZeros: ParamField = {
   type: 'boolean',
   defaultValue: false,
 };
+
+export type TimeOptions = {
+  timeformat: string | null;
+};
+
+export function getTimeOptionsFromParams(searchParams: URLSearchParams, defaultValues?: URLSearchParams): TimeOptions {
+  // Helper to get value from either source, prioritizing defaultValues
```
cpvalente: tip: maybe this comment is better if moved to JSDocs

apps/client/src/common/components/view-params-editor/common.options.ts
```diff
@@ -26,3 +26,16 @@ export const showLeadingZeros: ParamField = {
...
+  const getValue = (key: string) => defaultValues?.get(key) ?? searchParams.get(key);
+
+  return {
+    timeformat: getValue('timeformat'),
```
cpvalente: question: do we need the extra constructs? how about
```suggestion
return defaultValues?.get(key) ?? searchParams.get(key);
```

cpvalente (approved):

Issue #1999, opened 2026-03-09
Optimize app icon for macOS


Gibsonmb71:
Can we update the app icon to work with the latest versions of macOS? Right now, the app icon is stuck in a container since it doesn't meet the requirements.

<img width="62" height="61" alt="Image" src="https://github.com/user-attachments/assets/595a6589-7ccf-41ed-8ad4-890ae14ed62e" />

PR #2077, opened 2026-05-07
update electron-builder to v26.9.1

jwetzell:
- this version includes the fix for the windows build issues we were facing so now can move this dependency forward
- should allow getting the icon stuff back in for #1999

### DIFF
--- a/apps/electron/package.json
+++ b/apps/electron/package.json
@@ -14,7 +14,7 @@
   "main": "src/main.js",
   "devDependencies": {
     "electron": "38.2.1",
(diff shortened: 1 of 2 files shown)


jwetzell:
I've tested on Windows 11, macOS Sequoia, and Ubuntu 24.04

cpvalente (approved):
Nice! thank you

@jwetzell 
Would you be available to recover the icon in a separate PR?

jwetzell:
yeah, I think I have a reference to the commit that tried to introduce it before

PR #2078, opened 2026-05-08
refactor: icon compatibility with macos tahoe

jwetzell:


jwetzell:
appears to work tested on macOS Tahoe (M4)
Screenshot 2026-05-07 at 9 05 13 PM

jwetzell:
After trying this out on my M4 Mac mini I did have some trouble with local building which might be related to signing. I don't have a proper local signing setup on that mac so will be interested to see if that same issue pops up with a properly signed version, might be more related to the recent electron-builder change but I do have the fix ready if it is an actual problem.

cpvalente:
Thank you @jwetzell 

we could try to make a release from this branch? What do you think?

jwetzell:
I think that works

cpvalente (approved):
Thank you @jwetzell

Issue #2079, opened 2026-05-08
Select Rundowns using OSC


manoge2002:
I’ve been a big fan of ontime for three years now and have been following its developments with great interest. Thank you for that.

I actively use ontime with Bitfocus Companion.
It would be great if we could switch the rundown using Companion or OSC — ideally based on the title or a custom ID.

Thanks. :)

PR #2080, opened 2026-05-09
Feat switch rundown api

alex-Arc:
#2079

unsure what a good naming for the api would be (load or switch or...)
should we make it a nested api, will there be mor actions needed on **rundown** from the api
should we add a way for the user to see the id in the UI

osc example
```url
/api/rundown/load "abcd"
```

http example
```url
/api/rundown&load=abcd
/api/rundown/load/abcd
```

websocket
```json
{"rundown": {"load": "abcd"}}
```

### DIFF
--- a/apps/resolver/src/main.ts
+++ b/apps/resolver/src/main.ts
@@ -14,6 +14,9 @@ export { TimerPhase, Playback, runtimeStorePlaceholder, OffsetMode } from 'ontim
 export type { SimpleTimerState } from 'ontime-types';
 export { SimplePlayback, SimpleDirection } from 'ontime-types';
 
+// Client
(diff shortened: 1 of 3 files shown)


manoge2002:
I think you've already done a great job laying the groundwork.
The rundown ID definitely needs to be visible in the UI when it's used. Otherwise, I'd definitely go with exactly what you've done.

cpvalente:
@alex-Arc , we will need to wait with this until the rundown loading code is done

as for API design:

I believe that receiving a title may be too lose of an API, we should stick to rundown IDs, at least for now
Since this is primarily concerning the companion module, we could create a select element that shows all the rundown by title and give users options to load one

Should wait on #2086

PR #2083, opened 2026-05-09
fix(cuesheet): prevent timer cells from submitting on tab-out or escape

cpvalente:
Disable submit on cuesheet when tabbing through time fields
Pressing Escape should in no case submit a value

### DIFF
--- a/apps/client/src/common/components/input/text-input/useReactiveTextInput.tsx
+++ b/apps/client/src/common/components/input/text-input/useReactiveTextInput.tsx
@@ -3,8 +3,8 @@ import { ChangeEvent, KeyboardEvent, RefObject, useCallback, useEffect, useMemo,
 
 interface UseReactiveTextInputReturn {
   value: string;
-  onChange: (event: ChangeEvent) => void;
-  onBlur: (event: ChangeEvent) => void;
+  onChange: (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
+  onBlur: (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
   onKeyDown: (event: KeyboardEvent<HTMLElement>) => void;
 }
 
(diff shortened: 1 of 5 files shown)


apps/client/src/views/cuesheet/cuesheet-table/cuesheet-table-elements/DurationInput.tsx
```diff
@@ -49,6 +49,10 @@ function DurationInput({
     setIsEditing(false);
     setTimeout(() => textRef.current?.focusParentElement()); // Immediate timeout to ensure state change takes place first
   };
+  const handleTabOut = useCallback(() => {
```
alex-Arc, line 52: I there a reason for this to be a callback when `handleFakeBlur` is not

apps/client/src/common/components/input/text-input/useReactiveTextInput.tsx
alex-Arc, line 9:
```suggestion
interface UseReactiveTextInputReturn {
value: string;
onChange: (event: ChangeEvent<HTMLInputElement |HTMLTextAreaElement>) => void;
onBlur: (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => void;
onKeyDown: (event: KeyboardEvent<HTMLElement>) => void;
}
```

alex-Arc, line 154:
```suggestion
    onChange: (event) => handleChange(event.target.value),
    onBlur: (event) => {
      if (isTabbing.current) {
        isTabbing.current = false;
        (options?.onTabCancel ?? options?.onCancelUpdate)?.();
        return;
      }
      if (isEscaping.current) {
        isEscaping.current = false;
        return;
      }
      if (!isKeyboardSubmitting.current) {
        handleSubmit(event.target.value);
      }
```

alex-Arc (commented):
small comment and sugestion that could make it so we don't have to type cast

PR #2089, opened 2026-05-14
chore: pnpm 11 + dependency cleanup

alex-Arc:
- update pnpm to v11
- update pnpm and cache actions
- remove unused and dedupe dependencies
- make a stub for simple dom elements to avoid jsdom dependency
- small optimization + test for url preset function


alex-Arc:
@coderabbitai review

jwetzell:
`.nvmrc` and `Dockerfile` should be adjusted to reflect the change in `devEngines`. Other than that everything seemed to install and build just fine!

apps/client/src/common/utils/urlPresets.ts
```diff
@@ -205,6 +205,6 @@ function isPresettableView(view: string): view is OntimeViewPresettable {
  * Check if current location is a preset path
  */
 export function isPresetPath(location: Path): boolean {
-  const segments = location.pathname.split('/').filter(Boolean);
-  return segments[0] === 'preset';
+  const firstSegment = location.pathname.split('/').find(Boolean);
+  return !!firstSegment && firstSegment === 'preset';
```
cpvalente: question: can we explain the change here? was there a correctness issue?

as far as I can tell, the second comparison is redundant
for `firstSegment` to be strictly equal to the string `preset`, it cannot the falsy

apps/client/vitest.setup.ts
```diff
@@ -0,0 +1,56 @@
+import { vi } from 'vitest';
+
+// This is here to avoid having to import the whole jsdom or happy-dom
```
cpvalente: i think we prefer having happy dom.
It means we dont need to maintain this code at a cost of a few seconds in CI

apps/client/src/common/utils/urlPresets.ts
alex-Arc: ah yeah, you are rigth

apps/client/package.json
```diff
@@ -68,14 +67,14 @@
...
     "babel-plugin-react-compiler": "1.0.0",
+    "happy-dom": "^20.9.0",
     "ontime-types": "workspace:*",
     "ontime-utils": "workspace:*",
     "sass": "^1.57.1",
     "typescript": "catalog:",
     "vite": "8.0.1",
     "vite-plugin-compression2": "2.5.1",
     "vite-plugin-svgr": "4.5.0",
-    "vite-tsconfig-paths": "6.1.1",
```
cpvalente, line 78: Don’t we need this anymore?

alex-Arc: it is not part of our vite config, maybe it should be?

cpvalente: Not sure, if we don't miss it we don't need it

Maybe just search the git log to see why did we remove it?

Something like `git log -S "tsconfigpath"`

alex-Arc: as far as i can see it was there in v1.10.2 and gone in v2 https://github.com/cpvalente/ontime/blob/974a239b67dc28c60c463b6343471b8a0b414891/client/vite.config.js#L4

cpvalente: That was just the import.
I guess we don’t need it then

alex-Arc:
I did a test build and it seams to work
https://github.com/cpvalente/ontime/actions/runs/25988745764
anything else we should check?

cpvalente (approved):
Thank you Alex

jwetzell (approved):

PR #2092, opened 2026-05-27
Upgrade mantine and catch fullscreen error

alex-Arc:


apps/client/src/common/hooks/useFullscreen.ts
```diff
@@ -0,0 +1,21 @@
+import { useFullscreenDocument } from '@mantine/hooks';
```
cpvalente, line 1: this is a basic operation which I dont think warrants too many abstraction layers. I think we are better off making our own

https://github.com/cpvalente/ontime/pull/2091/changes#diff-9c58da23a3df9c1af0e33fb5292c52991d3ccea2034ecb4e95a9e7cb9e3eabe6

apps/client/src/common/components/navigation-menu/navigation-menu-item/NavigationMenuItem.tsx
```diff
@@ -9,22 +9,25 @@ interface NavigationMenuItemProps {
...
   return (
     <div
       className={cx([style.link, active && style.current, className])}
       tabIndex={0}
       role='button'
       onClick={onClick}
+      data-disabled={disable}
       onKeyDown={(event) => {
-        if (isKeyEnter(event)) {
+        if (!!disable && isKeyEnter(event)) {
```
cpvalente, line 30: this seems incorrect or at least confusing.

`!!` is usually used as a boolean casting operator, which we often prefer to use the boolean constructor instead to make the code explicit
but here I think you are trying to negate the boolean like `!disable && isKeyEnter`?
ie: if the element is not disabled, and user pressed enter, we do the thing

PR #2094, opened 2026-05-28
fix: playwright test

alex-Arc:
make ci to use the same node version as us so we don't trip the `devEngines` flag


cpvalente (approved):
Thank you Alex

PR #2096, opened 2026-06-10
Invert time-linking model — edit start to stretch previous event

nzalan01:
## Context
Previously Ontime chained events forward: each event's `linkStart` pointed back to the previous event, a linked event's `timeStart` was *derived* from the previous end, and the start input was disabled when linked. This inverts the mental model to **"an event's end flows into the next event's start."**

## Changes
- **`useEventAction.updateTimer`**: editing a **linked** event's start now redirects to the **previous** event — sets its duration so its end lands on the new start (`LockDuration`). The existing forward-cascade engine re-derives the chain, so **no backend change was needed**. Added `getPrevious()` (skips skipped events, matching the server's `getLink`) and refactored `getPreviousEnd()` onto it. Guards against pushing a previous event's end before its own start. Unlinked events / the first event edit their own start (gap behaviour).
- **`TimeInputFlow.tsx`**: start always editable, end **read-only**, `LockEnd` toggle removed, dynamic link tooltip.
- **`IpadTimeInputFlow.tsx`**: link toggle removed (start-stretch still works via the hook).

New events remain linked-by-default via the existing `linkPrevious` setting (default `true`); the server `eventsDefinition` template was intentionally left unchanged to avoid re-linking imported (Excel/JSON) events.

## Verification
- ✅ Server tests pass (107 — rundown-service + parser)
- ✅ Client TypeScript + ESLint clean on touched files
- ⏳ End-to-end browser check pending

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nzalan01:
Opened against upstream by mistake; this belongs in our internal fork.

PR #2098, opened 2026-06-10
fix: send refetch on all rundown edits

alex-Arc:
ensures that all edits to rundown data result in a refetch event
this is done by refactoring all the routes to use the rundown.service instead of doing it inplace

### DIFF
--- a/apps/server/src/api-data/rundown/rundown.service.ts
+++ b/apps/server/src/api-data/rundown/rundown.service.ts
@@ -659,11 +659,7 @@ export async function loadRundown(id: string) {
  * Sets a new rundown in the cache
  * and marks it as the currently loaded one
  */
-export async function initRundown(
-  rundown: Readonly<Rundown>,
-  customFields: Readonly<CustomFields>,
-  reload: boolean = false,
-) {
+export function initRundown(rundown: Readonly<Rundown>, customFields: Readonly<CustomFields>, reload: boolean = false) {
   runtimeService.stop();
   const { rundownMetadata, revision } = rundownCache.init(rundown, customFields);
   logger.info(LogOrigin.Server, `Switch to rundown: ${rundown.id}`);
--- a/apps/server/src/api-data/rundown/rundown.service.ts
+++ b/apps/server/src/api-data/rundown/rundown.service.ts
@@ -728,12 +728,12 @@ export async function renameRundown(id: string, title: string) {
   if (isCurrentRundown(id)) {
     const rundown = dataProvider.getRundown(id);
     const customField = dataProvider.getCustomFields();
-    await initRundown(rundown, customField);
+    initRundown(rundown, customField);
+  } else {
+    setImmediate(() => {
+      sendRefetch(RefetchKey.ProjectRundowns);
+    });
   }
-
-  setImmediate(() => {
-    sendRefetch(RefetchKey.ProjectRundowns);
-  });
 
   return newProjectRundowns;
 }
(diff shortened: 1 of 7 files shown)


alex-Arc:
@coderabbitai review

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -692,3 +688,73 @@ export async function createNewRundown(title: string) {
...
+  const dataProvider = getDataProvider();
+  const rundown = dataProvider.getRundown(id);
+
+  const newRundownId = generateId();
+  const newRundown: Rundown = structuredClone(rundown);
+  newRundown.id = newRundownId;
+  newRundown.title = `Copy of ${rundown.title}`;
+  newRundown.revision = 0;
+
+  const newProjectRundowns = await dataProvider.setRundown(newRundownId, newRundown);
```
cpvalente, line 706: question: would we benefit from a thin abstraction like `addRundown`

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -692,3 +688,73 @@ export async function createNewRundown(title: string) {
...
+   * If we are modifying the loaded rundown we re-init it
+   * This is likely over-kill but the simplest way to ensure state consistency
+   */
+  if (isCurrentRundown(id)) {
+    const rundown = dataProvider.getRundown(id);
+    const customField = dataProvider.getCustomFields();
+    initRundown(rundown, customField);
+  } else {
+    setImmediate(() => {
+      sendRefetch(RefetchKey.ProjectRundowns);
```
cpvalente, line 734: question: should we not refetch if the rundown is renamed?

e2e/tests/features/202-cuesheet.spec.ts
```diff
@@ -30,6 +30,8 @@ test('cuesheet datagrid does not submit timer cells on tab-out or escape', async
 
   // re-enter edit mode: original value should be unchanged
   await durationCell.click();
+  // tabbing selects the next input filed so we have to click twice to first leave input field and then select
```
cpvalente:
```suggestion
  // tabbing selects the next input field so we have to click twice to first leave input field and then select
```

cpvalente (approved):

apps/server/src/api-data/rundown/rundown.service.ts
alex-Arc: yes either it is done as part of initRundown
or we do it manually if it is a background rundown

apps/server/src/api-data/rundown/rundown.service.ts
alex-Arc: in the dataProvider?

cpvalente: I was thinking in the service. that would keep the logic of the service functions at a higher level

alex-Arc: hmm what would be it's responsibility?
we have `makeNewRundown` for blank ones

cpvalente: It would be to clarify that we are adding a rundown to the project, doesn't matter if it is empty or not

Right now, all the methods need to know how to add a rundown to the projects
This could make maintenance a bit tricky, like if we ever want to kick off side effects we need to change many things

cpvalente (approved):

PR #2099, opened 2026-06-11
feat: add MCP server for AI agent integration

cpvalente:
Adds a Streamable HTTP MCP server at /mcp that exposes 19 tools
covering rundown entries, rundown management, runtime state, project
info, automations, and custom fields. Includes Bearer token auth
support and a UI card in Settings > Sharing showing the endpoint URL
and Claude Desktop config.

https://claude.ai/code/session_01U24MeuUacYXeQhbX3tatEe


alex-Arc:
maybe worth waiting for v2 to be stable ?
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/migration.md

cpvalente:
> maybe worth waiting for v2 to be stable ? https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/migration.md

It can take any amount of time for it to be stable, the only wording I find is "end of year"
More importantly, we should consider our availability, if this works and we can do it now, I see no reason to postpone.

By the time we eventually need to migrate we would have learned and got feedback from users

Issue #2104, opened 2026-06-16
v4.9 - problems with the calculation of ending times


robertskiba:
In the current version 4.9 there still seem to be some problems with the correct calculation of expected times of groups or the whole rundown

- My last event is one with "count to end" enabled and set to 18:00. So the planned and expected end time of the rundown should be 18:00, logically. But it shows some minus value that subtracts the overtime. in this case -05:56:15 with an overtime of 3:45. This doesn't make sense. The actual start time also seems wrong and shows a value in the future but I don't know where that comes from. 

- the planned and expected group end time shows "due" all the time. Flagging also shows "due" instead of the expected and planned times. 

- in general, from my understanding, as soon as you put an event (like a break) into a rundown that has a defined end time, it should "smoothen" even an overtime rundown again (if count to end is active)

cpvalente:
Hi @robertskiba 

It looks like we overlooked some calculations
I will look into it

cpvalente:
Hi @robertskiba,

I was wondering if you can help us test this?
Are you able to run the code from source? https://github.com/cpvalente/ontime/pull/2113

Otherwise, reach out on Discord or over email and we can make a release for you

robertskiba:
Sure I'd like to help, but I can't compile myself. Please compile me a windows beta release.

cpvalente:
Hi @robertskiba , thank you for helping
I have made a release where you can download the windows installer https://github.com/cpvalente/ontime/releases/tag/v4.9.1-beta.1

Let me know how it goes

robertskiba:
Hi Carlos, 

thanks for the latest beta. I'm just testing it and so far, it looks good. As I don't yet understand the philosophy behind the "relative" button, I don't know if the behaviour is correct yet. 

Talking about the absolute mode, It works like expected. 

I would like to put forward the following propositions, which, in my opinion, should work as follows:

1. If the show is currently running before any "count to end" event (e.g. lunch break with a fixed ending, no matter if it starts later), the planned end and the expected end time of both events and current group should be the same as it seems that the "CTE"-Event is bringing the show back to a punctual event.  (seems to work in the beta now, it didn't before. 
2. If a CTE Event is running and you add or subtract time (eg. 5mins), something should change. Either the current ending time and the offset or the planned and/or expected end time. (this doesn't work at the moment but should be a matter of discussion in your team, how to solve that)

cpvalente:
Thank you for this @robertskiba 

[...]

PR #2113, opened 2026-06-20
fix: count-to-end events break link chain

cpvalente:
This is trying to address #2104

there are a few issues here
- the link chains need to be broken by count to end events
- an event with count to end needs to break the expected end time

We already had all the logic, but slightly deviation in UI and server so I have decided to consolidate

Unfortunately this does make the rundown times surprising.
Maybe the count-to-end elements need to be drastically highlighted in the rundown? We should clarify the rundown so that these results are not surprising, open for ideas

### DIFF
--- a/apps/server/src/services/timerUtils.ts
+++ b/apps/server/src/services/timerUtils.ts
@@ -41,7 +41,9 @@ export function getExpectedFinish(state: RuntimeState): MaybeNumber {
   const pausedTime = pausedAt != null ? clock - pausedAt : 0;
 
   if (countToEnd) {
-    return timeEnd + addedTime + pausedTime;
+    // count to end events are anchored to their fixed end: added time and pauses
+    // do not move the end, they surface as offset instead (see getRuntimeOffset)
+    return timeEnd;
   }
 
   // handle events that finish the day after
@@ -74,9 +76,10 @@ export function getCurrent(state: RuntimeState): number {
   const { clock } = state;
 
   if (countToEnd) {
+    // count to end runs to its fixed end, so added time does not stretch the countdown
     const isEventOverMidnight = timeStart > timeEnd;
     const correctDay = isEventOverMidnight ? dayInMs : 0;
-    return correctDay - clock + timeEnd + addedTime;
+    return correctDay - clock + timeEnd;
   }
 
   if (startedAt === null) {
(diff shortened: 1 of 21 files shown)


packages/utils/src/date-utils/getExpected.ts
```diff
@@ -85,9 +85,12 @@ export function getExpectedEnd(
 
   // count to end events should finish on schedule unlesss the start is compromised
   if (event.countToEnd) {
-    // count to end take scheduled delays into consideration
-    const plannedEnd = event.timeStart + event.duration + event.delay;
-    return Math.max(expectedStart, plannedEnd);
+    // the scheduled end, normalised to the same day-space as expectedStart
+    // (a raw timeStart + duration would miss the day offset on multi-day rundowns)
+    const delayedStart = Math.max(0, event.timeStart + event.delay);
```
alex-Arc: have we through through how a delay propagates past a `coundToEnd` ?

and why is the delay now clamped?

cpvalente: this code is a copy of the logic inside expectedstart, there it is clamped to avoid declaring a start the day before

packages/utils/src/date-utils/getExpected.ts
```diff
@@ -60,3 +56,27 @@ export function getExpectedStart(
...
+  /**
+   * Count to end events are a special case
+   * - the end time is always the wall clock
+   */
+  if (event.countToEnd) {
+    // account for day offset
+    const relativeDayOffset = event.dayOffset - state.currentDay;
+    const plannedEnd = event.timeStart + event.duration + relativeDayOffset * dayInMs;
+
+    // count to end cant finish later than the plan
```
alex-Arc:
```suggestion
    // count to end should finish on the planed time or immediately on start
```

apps/server/src/services/timerUtils.ts
```diff
@@ -41,7 +41,9 @@ export function getExpectedFinish(state: RuntimeState): MaybeNumber {
   const pausedTime = pausedAt != null ? clock - pausedAt : 0;
 
   if (countToEnd) {
-    return timeEnd + addedTime + pausedTime;
+    // count to end events are anchored to their fixed end: added time and pauses
+    // do not move the end, they surface as offset instead (see getRuntimeOffset)
```
alex-Arc: not sure if this is correct or an issue with wording

an event that is count to end 22h displays that value even when we are in over timer and it is not possible
Image

in this case I think the expected end value should be the same as the start at value

packages/utils/src/date-utils/getExpected.ts
alex-Arc: hmm I see
but a negative delay on a event that starts at midnight should be able to move to the day before, no?
this might also not be covered by our current logic

apps/server/src/services/timerUtils.ts
cpvalente: maybe the most correct is to change the text when we are in a count to end event to something like "planned end"?

cpvalente:
@alex-Arc , I believe this is ready for now. can you please sanity check?

I have also created a follow up which leverages all these changes to expose the pause across midnight issue #2137 
We may want to look into that for a next release and potentially bundle the fix with a review of our epoch time timestamps

cpvalente:
@alex-Arc , I have added the end calculations we talked about, let me know what you think

alex-Arc:
image

PR #2118, opened 2026-06-28
refactor(timer): elapsed is active time from start

cpvalente:
fixes an issue where users can add time until the elapsed becomes negative

the simplest change was to change the semantics of elapsed to be the active (non paused) time since start
this is a behaviour change but a good simplification for the system

the change is simple, the code changes are mostly comments and tests


alex-Arc (approved):
logic seams sound
and the new behavior makes more sense

PR #2128, opened 2026-07-10
feat(automation): expose group title and allow setting secondary text

cpvalente:
This PR includes improvements to the automation forms
- allow users to set secondary text
- expose groupNow data in templates
- expose templates in most text inputs (HTTP URL and primary / secondary message text)

Also some focus on UX
- autocomplete can now add templates in the middle of a sentence
- improve keyboard UX by using base ui
- allow expanding an input into a modal so that is easier to edit long fields

### DIFF
--- a/apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
+++ b/apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
@@ -293,8 +293,9 @@ export default function AutomationForm({ automation, onClose }: AutomationFormPr
         <h3>Outputs</h3>
         <Info>
           Automation outputs can be used to send data from Ontime to external software <br />
-          or to change properties of Ontime itself.
-          <ExternalLink href={integrationsDocsUrl}>See the documentation for templates</ExternalLink>
+          or to change properties of Ontime itself. <br /> <br />
+          Use Ontime runtime data in these fields with template strings. Type {'{{'} to see autocomplete, or{' '}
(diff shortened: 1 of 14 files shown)


apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
```diff
@@ -346,7 +346,7 @@ export default function AutomationForm({ automation, onClose }: AutomationFormPr
                   </label>
                   <label>
                     Arguments
-                    <TemplateInput {...register(`outputs.${index}.args`)} value={output.args} placeholder='1' />
+                    <TemplateInput {...register(`outputs.${index}.args`)} value={output.args} fluid placeholder='1' />
```
alex-Arc, line 355: the OSC `Address` field is also a template string

alex-Arc (changes requested):
The expanded editor is nice

there is a lot of code to recreate the autocomplete again, maybe that is just necessary?

cpvalente:
> there is a lot of code to recreate the autocomplete again, maybe that is just necessary?

the templateinput is complex enough that trying to reuse the autocomplete would complicate both sides. I opted for reusing the autocomplete from base ui and making two separate components

but i did review the component and split it so we have more composition and less code in any one context. I have also extracted logic to a utils file (which is unit tested) and the shared state to a hook.
The majority of the code is react markup which I guess its fine

alex-Arc (approved):

PR #2129, opened 2026-07-10
feat(import): add merge strategy and new-rundown destination to spreadsheet import

cpvalente:


cpvalente:
@coderabbitai full review

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -743,6 +751,76 @@ export async function duplicateExistingRundown(id: string) {
...
+  const existing = dataProvider.getRundown(targetRundownId);
+
+  // custom fields must be reconciled before the rundown since entries depend on them
+  const parsedCustomFields = parseCustomFields({ customFields: incomingCustomFields });
+  await dataProvider.mergeIntoData({ customFields: parsedCustomFields });
+  const projectCustomFields = dataProvider.getCustomFields();
+
+  const merged = mergeRundownPreservingFields(incomingRundown, existing);
+  // re-validate the client supplied payload; createEvent preserves the fields set by the merge
+  const parsed = parseRundown(merged, projectCustomFields);
```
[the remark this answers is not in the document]
cpvalente: Addressed in 72ceb69: both import paths now build and validate the merged rundown (`parseRundown`, the step most likely to throw on bad data) against a local union of existing + incoming custom fields **before** committing the custom-field mutation to the provider. So a payload that fails validation no longer leaves a partial custom-field change behind. Full transactional rollback of a later `initRundown`/`setRundown` failure isn't attempted, as that residual is shared with the existing `patchCurrentProject` override path and would require transaction support the data provider doesn't currently have.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/server/src/api-data/rundown/rundown.validation.ts
```diff
@@ -11,6 +11,22 @@ export const rundownPatchValidator = [
...
+    .if(body('mode').equals('merge'))
+    .isString()
+    .trim()
+    .notEmpty()
+    .withMessage('targetRundownId is required when mode is merge'),
+  body('rundown').isObject(),
+  body('rundown.entries').isObject(),
+  body('rundown.order').isArray(),
+  body('rundown.flatOrder').isArray(),
+  body('customFields').isObject(),
```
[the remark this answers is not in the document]
cpvalente: Not changing this one, for consistency reasons — happy to reconsider if you'd prefer stricter behaviour.

`parseRundown`'s skip-and-continue on missing/invalid entries is the established behaviour across every import path in the app (the generic `PATCH /db` used by the override strategy and by `ProjectMergeForm` goes through the same `parseRundowns` → `parseRundown`). In the real flow the import payload is a server-generated preview from `generateRundownPreview`, which is already run through `parseRundown`/`processRundown`, so its graph is well-formed. Making `/import` alone reject on any skipped entry would (a) diverge from the rest of the codebase and (b) risk 400-ing legitimate imports that the preview intentionally sanitised. If we want strict graph validation on import, I think it belongs at the shared parser level so all import paths behave the same, rather than only here.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/server/src/api-data/rundown/rundown.router.ts
```diff
@@ -147,6 +158,34 @@ router.delete('/:id', paramsWithId, async (req: Request, res: Response<ProjectRu
...
+      if (mode === 'merge') {
+        // the validator guarantees this for merge, the guard narrows the type and adds defence in depth
+        if (!targetRundownId) {
+          throw new Error('targetRundownId is required when mode is merge');
+        }
+        projectRundowns = await mergeRundownFromImport(targetRundownId, rundown, customFields);
+      } else {
+        projectRundowns = await createRundownFromImport(rundown, customFields);
+      }
+      res.status(200).json({ loaded: getCurrentRundown().id, rundowns: normalisedToRundownArray(projectRundowns) });
```
cpvalente, line 186: it doesnt seem like we use the return value, we could use it to populate the cache or remove the work here

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/SheetImportEditor.tsx
```diff
@@ -62,6 +79,35 @@ export default function SheetImportEditor({
     onExport,
   });
 
+  const playback = usePlayback();
```
cpvalente: can we isolate this logic and state? as it is it causes re-render of the composing element which is expensive

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/SheetImportEditor.tsx
```diff
@@ -62,6 +79,35 @@ export default function SheetImportEditor({
...
+  const loadedEventId = useSelectedEventId();
+
+  // whether applying with the current options will stop a running playback
+  const isPlaying = playback !== Playback.Stop;
+  const loadedEventSurvivesMerge = (() => {
```
cpvalente: we prefer calling this something related to what it means, like willLoadedEventBeOverriden

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/SheetImportEditor.tsx
```diff
@@ -107,33 +153,79 @@ export default function SheetImportEditor({
...
+            <IoEye />
+            Preview import
+          </Button>
+          <Button
+            variant='primary'
+            onClick={onApplyClick}
+            disabled={!state.preview || isBusy}
+            loading={state.loading === 'apply'}
+          >
+            {willStopPlayback && confirmStop ? 'Confirm — stop playback & apply import' : 'Apply import'}
```
cpvalente: question: have we verified if the backend does not stop the playback when rebuilding the rundowns?

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/useSheetImportForm.ts
```diff
@@ -170,6 +175,9 @@ export function useSheetImportForm({
   const columnLabels = buildColumnLabels(values);
 
   const [state, dispatch] = useReducer(importReducer, initialImportState);
+  const [importOptions, setImportOptions] = useState<ImportOptions>(() => getPersistedImportOptions(sourceKey));
+  // ephemeral, per-import name for the "new rundown" destination, prefilled with the spreadsheet name
```
cpvalente: the comment does not add value

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/SourcesPanel.tsx
```diff
@@ -126,11 +129,38 @@ export default function SourcesPanel() {
...
+        mode: 'merge',
+        targetRundownId: currentRundown.id,
+        rundown: preview.rundown,
+        customFields: preview.customFields,
+      });
+      handleFinished();
+      return;
+    }
+
+    // override: replace the current rundown wholesale (unchanged legacy behaviour)
```
cpvalente:
```suggestion
    // override: replace the current rundown
```

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/SourcesPanel.tsx
```diff
@@ -208,6 +238,13 @@ export default function SourcesPanel() {
     if (activeSource.kind === 'excel') return 'excel';
     return `gsheet:${activeSource.sheetId}`;
   })();
+  // suggested name when importing into a new rundown: the spreadsheet file name (without extension)
+  // for Excel, or the document title for Google Sheets
+  const spreadsheetName = (() => {
+    if (!activeSource) return '';
+    if (activeSource.kind === 'excel') return activeSource.fileName.replace(/\.[^./\\]+$/, '');
```
cpvalente: lets remove the regex in favour of simple logic. for similar logic, we usually extract a small utility and cover with unit tests

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -663,8 +666,13 @@ export async function initRundown(
   rundown: Readonly<Rundown>,
   customFields: Readonly<CustomFields>,
   reload: boolean = false,
+  options?: { skipPlaybackStop?: boolean },
 ) {
-  runtimeService.stop();
+  // by default switching rundowns stops playback
```
cpvalente: when switching rundowns we should always stop playback.
but when making changes to the rundown we should try and maintain playback when possible

apps/server/src/services/runtime-service/runtime.service.ts
```diff
@@ -165,6 +165,13 @@ class RuntimeService {
     }
   }
 
+  /**
+   * Returns the id of the currently loaded event, or null if none is loaded
+   */
+  public getLoadedEventId(): string | null {
```
cpvalente: we have an EntryId type which would be appropriate here and in the consumer

apps/server/src/api-data/rundown/rundown.service.ts
```diff
@@ -743,6 +751,84 @@ export async function duplicateExistingRundown(id: string) {
...
+/**
+ * Applies an imported rundown into an existing rundown using the "merge" strategy:
+ * elements matched by id preserve fields the spreadsheet cannot express (triggers/automations,
+ * timeStrategy); the incoming rundown is otherwise authoritative for identity and order.
+ *
+ * If the target is the loaded rundown and the currently playing event survives the merge,
+ * playback is preserved (hot reload) instead of being stopped.
+ * @throws if the target rundown does not exist
+ */
+export async function mergeRundownFromImport(
```
cpvalente: this is crucial logic and should be covered by unit tests. we should isolate the business logic from fetching and persisting and testing the logic itself

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/importMapUtils.ts
```diff
@@ -150,6 +151,62 @@ export function getPersistedImportState(sourceKey: string): { values: ImportForm
...
+  return (
+    (candidate.destination === 'current' || candidate.destination === 'new') &&
+    (candidate.strategy === 'override' || candidate.strategy === 'merge')
+  );
+}
+
+export function getPersistedImportOptions(sourceKey: string): ImportOptions {
+  const storageKey = getImportOptionsKey(sourceKey);
+  try {
+    const raw = localStorage.getItem(storageKey);
```
cpvalente: raw doesnt mean much, can we rename to something like persistedData?

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/importMapUtils.ts
```diff
@@ -150,6 +151,62 @@ export function getPersistedImportState(sourceKey: string): { values: ImportForm
...
+
+function getImportOptionsKey(sourceKey: string) {
+  return makeStageKey(`import-options:${sourceKey}`);
+}
+
+export function persistImportOptions(sourceKey: string, options: ImportOptions) {
+  localStorage.setItem(getImportOptionsKey(sourceKey), JSON.stringify(options));
+}
+
+function isPersistedImportOptions(obj: unknown): obj is ImportOptions {
```
cpvalente: add some docs like
- checks whether a given value represents a well formed import options object

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/SheetImportEditor.tsx
```diff
@@ -62,6 +79,35 @@ export default function SheetImportEditor({
...
 
+  const playback = usePlayback();
+  const loadedEventId = useSelectedEventId();
+
+  // whether applying with the current options will stop a running playback
+  const isPlaying = playback !== Playback.Stop;
+  const loadedEventSurvivesMerge = (() => {
+    if (!loadedEventId) return false;
+    const entry = state.preview?.rundown.entries[loadedEventId];
+    return entry !== undefined && isOntimeEvent(entry) && !entry.skip;
```
cpvalente: instead of entry.skip we should use isPlayableEvent

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/useSheetImportForm.ts
```diff
@@ -182,6 +190,12 @@ export function useSheetImportForm({
     dispatch({ type: 'reset' });
   }, [initialFormValues, reset]);
 
+  // Rehydrate the import options and prefilled name whenever the source context changes.
```
cpvalente:
```suggestion
  // Update import options and rundown name if the source changes
```

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/useSpreadsheetImport.ts
```diff
@@ -1,14 +1,15 @@
...
-import { CUSTOM_FIELDS, RUNDOWN } from '../../../../../common/api/constants';
+import { CUSTOM_FIELDS, PROJECT_RUNDOWNS, RUNDOWN } from '../../../../../common/api/constants';
 import { patchData } from '../../../../../common/api/db';
+import { importRundownWithOptions } from '../../../../../common/api/rundown';
 
 export default function useSpreadsheetImport() {
   const queryClient = useQueryClient();
 
-  /** applies rundown and customFields to current project */
+  /** applies rundown and customFields to current project (override strategy) */
```
cpvalente:
```suggestion
  /** overrides rundown and customFields in current project */
```

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/useSpreadsheetImport.ts
```diff
@@ -24,7 +25,25 @@ export default function useSpreadsheetImport() {
     [queryClient],
   );
 
+  /** applies an import using a merge strategy or into a new rundown */
```
cpvalente: This function documentation exposes a complexity of the implementation.
merging into the current rundown and creating a new one should not be parallel operations
- merging should only be available when we want to add changes into the current rundown. the logic should do a best effort to keep both events, preferring the incoming data when there are conflicts
- when override is selected, we will always replace all the data in the selected rundown with the incoming data. when the user choses to import into a new rundown, we effectively override new data into an empty dataset


either way, the comment can be simpler:
```suggestion
  /** merges an import into the current rundown or creates a new one */
```

apps/server/src/api-data/rundown/__tests__/rundown.utils.test.ts
```diff
@@ -287,3 +289,134 @@ describe('getIntegerAndFraction()', () => {
...
+
+    // the incoming event object is not mutated
+    expect(incomingEvent.triggers).toEqual([]);
+
+    // preserved triggers are a copy, not a shared reference
+    (merged.entries.a as OntimeEvent).triggers.push({ ...trigger, id: 'trigger-2' });
+    expect((existing.entries.a as OntimeEvent).triggers).toEqual([trigger]);
+  });
+
+  it('does not preserve fields when the matched ids have different types', () => {
```
cpvalente: if the field with the same id has changed type, we should delete all the old data and start a clean field with the incoming data

apps/server/src/api-data/rundown/__tests__/rundown.utils.test.ts
```diff
@@ -287,3 +289,134 @@ describe('getIntegerAndFraction()', () => {
...
+      entries: { a: makeOntimeEvent({ id: 'a', title: 'an event', triggers: [] }) },
+    });
+
+    const merged = mergeRundownPreservingFields(incoming, existing);
+
+    expect(isOntimeEvent(merged.entries.a)).toBe(true);
+    expect((merged.entries.a as OntimeEvent).triggers).toEqual([]);
+  });
+
+  it('passes groups and milestones through unchanged', () => {
```
cpvalente: what does this mean? groups and milestones should be merged like any other entry

apps/server/src/api-data/rundown/__tests__/rundown.utils.test.ts
```diff
@@ -287,3 +289,134 @@ describe('getIntegerAndFraction()', () => {
...
+    });
+    const existing = makeRundown({ id: 'rundown-id', order: [], entries: {} });
+
+    const merged = mergeRundownPreservingFields(incoming, existing);
+
+    expect(merged.entries.g).toEqual(incoming.entries.g);
+    expect(merged.entries.m).toEqual(incoming.entries.m);
+  });
+
+  it('keeps preserved fields through a parseRundown round-trip', () => {
```
cpvalente: what does this mean?

apps/server/src/api-data/rundown/__tests__/rundown.utils.test.ts
```diff
@@ -287,3 +289,134 @@ describe('getIntegerAndFraction()', () => {
     expect(getIntegerAndFraction('123.')).toStrictEqual({ integer: 123, faction: 0, precision: 0 });
   });
 });
+
+describe('mergeRundownPreservingFields()', () => {
```
cpvalente, line 308: review these tests to be effective documentation of the feature, the tests should be minimal but meaningful. prefer actual, descriptive names to single characters like a
 consider the following, non exhaustive list

for conflicts:
- if incoming rundown does not have the entries in the current rundown, those entries are deleted
- if incoming rundown changes the types of the entries, all the existing data of the current entries is deleted in favour of the incoming data

when we match an incoming rundown to the current rundown entry id, we do a best effort merge of its data on a field-by-field basis:
- if a data field is provided, but has an empty value, we replace the current entry with the incoming empty value
- if a data field is not provided (example automations which are usually not described in spreadsheets), we keep the data in the existing current data

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/sheet-import/SheetImportEditor.tsx
cpvalente: Yes — verified end-to-end on a running server. Merging into the loaded rundown now goes through `applyChangeToCurrentRundown`, which only stops playback when the playing event no longer survives; otherwise it hot-reloads via `notifyOfChangedEvents` (the same path regular rundown edits use). Confirmed with playback running:
- merge that keeps the playing event → stays `playback: play`, and the event's data updates live;
- merge that drops the playing event → transitions to `playback: stop`.

`initRundown` (switching rundowns, incl. the new-rundown import) always stops, per your other comment.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/server/src/api-data/rundown/__tests__/rundown.utils.test.ts
cpvalente: Reworked the merge tests to read as documentation of the strategy, with descriptive entry names instead of `a`/`b`/`c`. They now cover:
- current entries absent from the incoming rundown are deleted;
- an id kept but whose type changes is fully replaced by the incoming data (no old data carried over);
- matched entries take incoming values, including when the incoming value is empty;
- fields a spreadsheet cannot express are kept from the current entry — automations (`triggers`), and `timeStrategy` (which is ambiguous when a sheet carries both end and duration, so it would otherwise be lost on a round-trip);
- groups and milestones merge like any other entry — they have no spreadsheet-inexpressible fields, so the incoming data wins.

Also extracted the playback-survival decision into a pure `willPlaybackSurvive(loadedEventId, rundown)` with its own unit tests.

apps/server/src/api-data/rundown/rundown.utils.ts
```diff
@@ -225,6 +227,49 @@ export function getUniqueId(rundown: Rundown): EntryId {
...
+ * - missing properties of matched events are kept as is in the existing rundown
+ * - existing elements not present in the incoming rundown are dropped
+ */
+export function mergeRundownPreservingFields(incoming: Readonly<Rundown>, existing: Readonly<Rundown>): Rundown {
+  const entries: RundownEntries = {};
+
+  for (const [id, entry] of Object.entries(incoming.entries)) {
+    const existingEntry = existing.entries[id];
+
+    if (isOntimeEvent(entry) && existingEntry !== undefined && isOntimeEvent(existingEntry)) {
```
cpvalente: this logic seems incorrect. for any incoming entry, we need to run through the logic described in the jsdocs
- if the entry has changed type, we replace the whole thing
- replace all existing fields
- retain missing (undefined) fields from existing rundown

We may also want to reconsider the timeStrategy with an `inferStrategy()` call
the logic is also brittle since it manually extracts elements which we know are not in the spreadsheet (like triggers) but that shouldnt be the case.
For a merge scenario, any properties which are undefined in the incoming data should be kept.
only properties with an existing, empty value should be removed

cpvalente: Applied the `inferStrategy` change for `timeStrategy` and documented the type-change rule (an id whose type changed is replaced wholesale).

On the fully generic "keep any property that is `undefined` in the incoming data" approach — there's a constraint worth flagging: by the time the merge runs, the incoming rundown has already been through `createEvent` (the preview is a full, validated `Rundown`), so every field is already defined. `triggers` is `[]`, not `undefined`, and the "which columns were actually in the sheet" information is gone — so a purely `undefined`-driven merge can't distinguish an omitted column from an explicitly-empty one. That's why `triggers` (the one `OntimeEvent` field with no import column) still has to be handled explicitly.

If you'd like the fully generic behaviour, the clean way is to drive preservation off the **import map**: thread the set of mapped fields to the merge so it keeps the existing value for any *unmapped* field (not just triggers), and overwrites for mapped ones even when empty. It's a slightly larger change (transport the provided-field set + handle custom fields per key), but happy to do it if you'd prefer that direction.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/server/src/api-data/rundown/rundown.router.ts
cpvalente: Addressed on the client: `applyImport` now uses the response to populate the cache — `queryClient.setQueryData(PROJECT_RUNDOWNS, response.data)` — instead of a separate refetch, so the returned rundown list is consumed (`useSpreadsheetImport.ts`).

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/server/src/api-data/rundown/rundown.utils.ts
cpvalente: Done — the merge is now generic, no hardcoded field list (`afc25c0`).

The client derives, from the import map, the `OntimeEvent` fields the sheet actually maps and sends them as `providedFields`. The server merge applies only those fields onto the matched existing event through the canonical `createEventPatch`, which:
- keeps every field the sheet did **not** provide (automations, and anything else — including future non-expressible fields), and
- infers `timeStrategy` from the provided times (via the existing `inferStrategy` inside `createEventPatch`).

Custom fields present on the incoming event are exactly the ones the sheet mapped (parseExcel only sets mapped custom keys), so `createEventPatch`'s `{...existing.custom, ...patch.custom}` keeps the unmapped ones and overwrites the mapped ones.

Verified on a live server: with `providedFields: ['title']`, a matched event's title updated to the sheet value while its `note` (different in the incoming, but not mapped) and its automations were both kept from the existing event. The merge unit tests now pass `providedFields` and document this field-by-field behaviour.

---
_Generated by [Claude Code](https://claude.ai/code)_

cpvalente: Made the field provenance authoritative for custom fields too (`cb902c3`).

The previous version only told the merge which *built-in* fields the sheet mapped (a flat `string[]`) and inferred the custom fields heuristically — "whatever custom keys are on the incoming event must be the ones the sheet mapped". That heuristic is exactly the "not enough information" gap: the function was guessing custom-field provenance from the data instead of being told it, so it couldn't reliably distinguish an unmapped custom field from a mapped-but-empty one.

`providedFields` is now a symmetric `ImportedFields = { event: string[]; custom: string[] }`, derived directly from the import map (`getProvidedImportFields`), which is the single authoritative source of what the sheet supplies. The server builds an event patch from exactly those mapped built-in **and** custom keys (`buildImportPatch`) and applies it through the canonical `createEventPatch`. So both kinds of field now follow one rule with no special-casing: a mapped field overwrites (even when empty), an unmapped field — built-in or custom — is kept from the existing event.

Covered by unit tests: a custom-field merge where a mapped custom field is overwritten while an unmapped one is retained (`rundown.utils.test.ts`), and `getProvidedImportFields` returning the symmetric `{ event, custom }` shape from the map (`importMapUtils.test.ts`). Also confirmed the validator + `matchedData` carry the nested `providedFields` object through to the handler intact.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/useSpreadsheetImport.ts
```diff
@@ -1,19 +1,20 @@
...
-      // we are unable to optimistically set the rundown since we need
-      // it to be normalised
+  /** applies a spreadsheet import: override or merge into a rundown, or create a new one */
+  const applyImport = useCallback(
+    async (payload: RundownImportPayload) => {
+      const response = await importRundownWithOptions(payload);
+      // use the returned project rundowns to populate the cache instead of refetching
+      queryClient.setQueryData(PROJECT_RUNDOWNS, response.data);
+      // the loaded rundown and custom fields still need a refetch to get their normalised shape
       await queryClient.invalidateQueries({
```
cpvalente: the backend will send a refetch after the rundown has been parsed and applied, maybe we would avoid race conditions by not having this here?

apps/server/src/api-data/rundown/__tests__/rundown.utils.test.ts
```diff
@@ -287,3 +290,290 @@ describe('getIntegerAndFraction()', () => {
...
+      id: 'show-rundown',
+      order: ['keynote'],
+      entries: {
+        keynote: makeOntimeEvent({ id: 'keynote', title: 'Keynote', note: 'green room', triggers: [automation] }),
+      },
+    });
+    const incoming = makeRundown({
+      id: 'spreadsheet-rundown',
+      order: ['keynote'],
+      entries: { keynote: makeOntimeEvent({ id: 'keynote', title: 'Opening keynote', note: 'ignored', triggers: [] }) },
```
cpvalente: this seems incorrect or maybe just confusing, but we provide both note and triggers here, so we would have expected both to be overriden

apps/server/src/api-data/rundown/rundown.utils.ts
```diff
@@ -225,6 +228,76 @@ export function getUniqueId(rundown: Rundown): EntryId {
...
+
+    if (isOntimeEvent(entry) && isOntimeEvent(existingEntry)) {
+      // apply the sheet's patch onto the existing event through the canonical patch function, which
+      // keeps every unmapped field (e.g. triggers) and infers the time strategy from the provided times
+      const merged = applyPatchToEntry(existingEntry, buildImportPatch(entry, providedFields)) as OntimeEvent;
+      // grouping comes from the sheet structure, not from a data column
+      entries[id] = structuredClone({ ...merged, parent: entry.parent });
+    } else {
+      // new entry, or an id whose type changed: take the incoming data wholesale
+      entries[id] = entry;
```
cpvalente: this seems incorrect, for any entry where the type did not change, we try and merge the data. the override should only happen if the type has changed and we think that the elements are not compatible

Make sure this is covered by an unit test

cpvalente: Fixed in `0626711`. The merge now reconciles a matched entry whenever the **type is unchanged** — so groups and milestones merge like events — and only takes the incoming entry wholesale when the id is new or its type changed:

```ts
if (existingEntry === undefined || existingEntry.type !== incomingEntry.type) {
  entries[id] = incomingEntry; // new, or incompatible type change → replace
  continue;
}
const merged = applyPatchToEntry(existingEntry, buildImportPatch(incomingEntry, providedFields));
```

`buildImportPatch` is generalised from `OntimeEvent` to `OntimeEntry`, so the same "keep the fields the sheet cannot express" rule applies to every type — e.g. a matched group keeps its `targetDuration` (no import column) while taking the sheet's title. Structure (a group's children, an entry's parent) is always taken from the incoming sheet.

Covered by new unit tests in `rundown.utils.test.ts`: `merges a matched group, keeping the fields the sheet cannot express` and `merges a matched milestone, keeping the fields the sheet did not map`. The existing type-change test still asserts the wholesale replace.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/server/src/api-data/rundown/__tests__/rundown.utils.test.ts
cpvalente: Good catch on the confusion — reworded in `0626711`. The point the test was making poorly: the preview always fully populates an entry (`createEvent` fills every field), so the incoming entry *always* carries a `note` and `triggers`. What the sheet actually supplied is `providedFields`, not whatever values happen to sit on the incoming entry. Since only `title` is in `providedFields` here, `note` and `triggers` keep the existing values regardless of the incoming ones. The incoming `note` is now `'from a stale column'` with a comment spelling this out, so it no longer reads as "we provided note, why isn't it applied?".

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/manage-panel/sources-panel/useSpreadsheetImport.ts
cpvalente: Agreed — removed in `0626711`. The backend broadcasts a refetch once the rundown is parsed and applied (`RefetchKey.All` for the current/new-rundown paths, `ProjectRundowns` for a background rundown), so the caches update through that single path. The hook is now just the HTTP call and lets the websocket refetch do the invalidation, exactly like the other rundown mutations. Verified via the `301-spreadsheet-import` e2e, which asserts the imported groups/entries appear in the editor after apply — that update now comes purely from the backend refetch.

---
_Generated by [Claude Code](https://claude.ai/code)_

alex-Arc (approved):

Issue #2126, opened 2026-07-06
Stage Timer digits appear vertically offset in vMix Browser Input (CEF-based embedded browser)


mindandvision:
Description
The large timer digits on the Stage Timer view appear shifted upward when displayed inside vMix's Browser Input, which uses an embedded, older CEF-based Chromium engine. All other elements on the same view (current time, message bar, progress bar, title card) render in the correct position. Only the large timer digits are affected.
Steps to reproduce

Open the Stage Timer view in a regular, up to date Chrome or Edge browser. The timer digits are centered correctly.
Load the exact same URL as a Browser Input inside vMix, set to the same resolution (1920x1080).
Compare the vertical position of the large timer digits. In vMix, the digits sit noticeably higher than in a regular browser.

What we found while investigating

The outer container ".stage-timer" and "body" report identical computed heights in both environments, so the container itself is sized correctly.
The digits are centered using Flexbox on ".timer-container" ("flex: 1; align-content: center; place-self: center;"), and the font size is set with a viewport-relative unit ("font-size: 28vw").
Other elements on the page (progress bar, title card, current time) use simpler positioning and are not affected.
Manually overriding ".timer-container" to use "position: absolute; top: 50%; left: 50%; transform: translate(-50%, -50%);" instead of the Flexbox-based centering fixes the issue completely in vMix, without affecting the regular browser rendering.

Suspected cause
The combination of Flexbox centering and a viewport-relative font size ("vw") appears to be recalculated differently, or not at all, by the older CEF engine used inside vMix, possibly because it does not fire the same resize events a regular visible browser window does. Absolute positioning with a transform based centering does not depend on this recalculation and renders correctly in both environments.
Suggested fix
Consider switching the centering method for the timer digits from Flexbox plus viewport units to absolute positioning with "transform: translate(-50%, -50%)", or provide this as a configurable option, since Stage Timer views are frequently embedded in third party production software (vMix, OBS Browser Source, and similar tools) that rely on embedded, sometimes outdated Chromium engines.
Environment

Ontime version: 4.10.0
Affected environment: vMix Browser Input (CEF-based V115)
Not affected: Chrome, Edge (regular desktop browser, including fullscreen mode)

cpvalente:
Hi @mindandvision 

Can you please provide us the URL and a screenshot?

PR #2139, opened 2026-07-17
fix(views): center timer digits with flexbox for embedded browser compatibility

cpvalente:
This is claudes suggestion. It seems to work just as well locally. I could not test in a CEF based browser
Addresses #2126


alex-Arc (approved):
I could recreate issue in vMix
and it was fixed by this patch

PR #2148, opened 2026-07-18
fix: timer-legacy view not served in desktop distributions

panctronic:
The server serves html/timer-legacy.html and html/login.html from disk at runtime, relative to the bundled server. The Docker image copies these files but the electron packaging did not, so the view 404ed in all desktop distributions.

Additionally, (at least on Linux) AppImages mount at /tmp/.mount_*, a hidden directory. Express sendFile by default refuses paths containing dot-segments (returns 404 without touching disk), so the view failed on Linux even with the file packaged. Allow dotfiles for this route; the request path is fixed so no user input is affected.

This is a followup fix for #2012

### DIFF
--- a/apps/electron/package.json
+++ b/apps/electron/package.json
@@ -125,6 +125,13 @@
           "**/*",
           "*{.ts}"
         ]
+      },
+      {
+        "from": "../server/src/html/",
+        "to": "extraResources/html/",
+        "filter": [
+          "**/*"
+        ]
       }
     ]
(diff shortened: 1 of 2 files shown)


cpvalente:
@coderabbitai full review

cpvalente:
Thank you @panctronic 
I guess we should look into having a better way to copy the resources into electron to avoid these classes of issues

For now this fix is great!

PR #2149, opened 2026-07-18
fix: fix issue where a countToEnd would lead to incorrect expected times

### LINKED ISSUE #2104: v4.9 - problems with the calculation of ending times

alex-Arc:
fix #2104

### DIFF
--- a/apps/client/src/common/utils/time.ts
+++ b/apps/client/src/common/utils/time.ts
@@ -4,6 +4,7 @@ import {
   MILLIS_PER_MINUTE,
   MILLIS_PER_SECOND,
   formatFromMillis,
+  getExpectedEnd,
   getExpectedStart,
 } from 'ontime-utils';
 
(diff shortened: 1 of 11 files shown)


alex-Arc:
@coderabbitai review

cpvalente:
@coderabbitai full review

cpvalente (approved):
great work alex

PR #2152, opened 2026-07-24
feat(timer): targeted client Q&A

SMPTY:
### Problem

Controllers currently have no way to privately communicate with a specific performer on stage, such as a DJ, presenter, or speaker, when verbal communication isn't practical or possible. This can make it difficult to confirm readiness, request a quick action, or ask a simple question without interrupting the event or relying on radios and hand signals.

While Timer Messages and Secondary Messages allow one-way communication, there is currently no way to ask a targeted question and receive a response from a specific Timer client.

### What this adds

Allows the controller to turn the Secondary Message into a private question for one selected Timer client by enabling Question mode (?).

The controller can enter up to three preset answers, with Yes and No included by default. When the message is shown, the selected client can tap an answer. The selected response is displayed in red on the Timer view for 2 seconds before disappearing, while remaining visible to the controller.

If no response options are entered, the feature simply acts as a way to privately display a standard Secondary Message to a single Timer client, making it useful for both targeted messaging and targeted questions.

The selected response is cleared whenever the question is modified, the preset answers are changed, or Question mode (?) is turned off.

### Implementation note

I'm not a developer. I designed the feature and UI mockups in Figma, then worked closely with Claude to implement the code. The implementation reflects the intended behavior, but improvements or modifications to the code are absolutely welcome.


packages/types/src/api/websocket/api.type.ts
```diff
@@ -42,6 +42,15 @@ export type MessageResponse = {
...
 };
 
+export type MessageanswerAction = {
+  tag: 'messageanswer';
+  payload: string;
+};
+export type MessageanswerResponse = {
+  tag: 'messageanswer';
+  payload: MessageState;
+};
```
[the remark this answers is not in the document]
SMPTY: Kept this naming to match existing precedent: AddtimeAction, AuxtimerAction, and OffsetmodeAction all use the same lowercase-compound style already.

apps/client/src/views/timer/Timer.tsx
```diff
@@ -123,18 +136,34 @@ function Timer({ customFields, projectData, isMirrored, settings, viewSettings,
     return null;
   })();
 
+  // a question targets a single client; every other client should see nothing in the secondary slot
+  const secondaryIsQuestion = message.question.enabled && message.timer.secondarySource === 'secondary';
+  const isQuestionTarget = secondaryIsQuestion && message.question.target === clientId;
+  const hideSecondaryForQuestion = secondaryIsQuestion && !isQuestionTarget;
+
```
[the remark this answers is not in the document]
SMPTY: This is meant to be UI-level privacy, not an encrypted messaging system. The goal is just keeping the question off other screens, not protecting it from anyone determined to dig through the network traffic. Given the actual use case is a quick private question, I don't think the added complexity is worth it here.

apps/server/src/services/message-service/message.utils.ts
```diff
@@ -1,8 +1,11 @@
-import { TimerMessage } from 'ontime-types';
+import { MessageQuestion, TimerMessage } from 'ontime-types';
 
 import * as assert from '../../utils/assert.js';
 import { coerceBoolean, coerceString } from '../../utils/coerceType.js';
 
+const MAX_ANSWER_OPTIONS = 3;
```
alex-Arc, line 6: in general we don't like arbitrary limits
it it actually breaks the system then that's  fair, but i don think that is the case here?

apps/server/src/api-integration/integration.controller.ts
```diff
@@ -26,13 +31,13 @@ import { isValidChangeProperty, parseProperty } from './integration.utils.js';
 
 let lastRequest: Date | null = null;
 
-export function dispatchFromAdapter(tag: string, payload: unknown, _source?: 'osc' | 'ws' | 'http') {
+export function dispatchFromAdapter(tag: string, payload: unknown, _source?: 'osc' | 'ws' | 'http', clientId?: string) {
```
alex-Arc, line 34: Could the client id not be part of the payload?
that way it would also be possible to implement the QnA over osc and http connections

apps/client/src/common/hooks/useSocket.ts
```diff
@@ -56,8 +61,19 @@ export const setMessage = {
   timerBlackout: (payload: boolean) => sendSocket('message', { timer: { blackout: payload } }),
   timerSecondarySource: (payload: TimerMessage['secondarySource']) =>
     sendSocket('message', { timer: { secondarySource: payload } }),
+  questionEnabled: (payload: boolean) => sendSocket('message', { question: { enabled: payload } }),
+  questionTarget: (payload: string | null) => sendSocket('message', { question: { target: payload } }),
+  answerOptions: (payload: string[]) => sendSocket('message', { question: { answers: payload } }),
+  // manual override: ends an active or held answer immediately, without waiting on the server's hold delay.
+  // also hides the secondary message, since it was written for this question and shouldn't
+  // suddenly go out as a public broadcast just because the question was cancelled
```
alex-Arc, line 69: i'm not sure I understand the comments here

apps/client/src/features/control/message/QuestionControl.tsx
```diff
@@ -0,0 +1,95 @@
...
+  // the displayed defaults are only real once committed to the server, otherwise
+  // a viewer never receives them unless the controller happens to edit a field first
+  useEffect(() => {
+    if (enabled && answers.length === 0) {
+      setMessage.answerOptions(DEFAULT_ANSWERS);
+    }
+  }, [enabled, answers.length]);
+
+  // keep the panel visible after an answer comes in (enabled auto-resets) until the controller clears it
+  if (!enabled && answer === null) {
```
alex-Arc, line 34: maybe it is fine to show it all the time

apps/client/src/features/control/message/MessageControl.tsx
```diff
@@ -43,30 +48,67 @@ function TimerMessageInput() {
...
     >
-      <IconButton
-        aria-label='Toggle secondary message visibility'
-        onClick={toggleSecondary}
-        variant={visible ? 'primary' : 'subtle'}
-      >
-        {visible ? <IoEye /> : <IoEyeOffOutline />}
-      </IconButton>
+      <div className={style.iconGroup}>
+        <IconButton
```
alex-Arc, line 97: I think it would be more appropriate to make it a separate option in the dropdown
@cpvalente what about you

apps/client/src/features/control/message/QuestionControl.tsx
```diff
@@ -0,0 +1,95 @@
...
+const DEFAULT_ANSWERS = ['Yes', 'No', ''];
+
+// matches '/timer' as a path segment rather than a prefix, so this still works when
+// Ontime is hosted under a base path (eg cloud deployments like '/<project>/timer')
+function isTimerPath(path: string): boolean {
+  const [pathname] = path.split('?');
+  return pathname.split('/').includes('timer');
+}
+
+export default function QuestionControl() {
```
alex-Arc, line 21: I'm missing a more obvious way to clear the answers from the client

alex-Arc (commented):
very good start

PR #2155, opened 2026-07-26
Fix: pause over midnight

alex-Arc:
pick and fix the tests from #2137


apps/server/src/services/timerUtils.ts
cpvalente, line 42: question: should this not be the epoch _now?

cpvalente, line 182: question: should this not be the epoch _now?

cpvalente, line 88: question: if pausedAt is a epoch time, would it not mean that the returned current would also be an epoch time? maybe we need stricter return types here to help us?

PR #2154, opened 2026-07-26
refactor(settings): improve navigation and polish UI

cpvalente:

PR #2163, opened 2026-08-01
fix(settings): scope Escape to the open modal and polish shared UI

cpvalente:
## Summary

Follow-up fixes on top of #2154, found during review:

- **Escape key regression**: the modals introduced in #2154 (trigger, URL preset, custom field, custom view, rundown rename, project merge, automation forms) dropped the `preventEscape` guard the inline forms used to stop Escape from also closing the whole settings panel. Base UI's dialog dismiss and the settings panel's own `Escape` listener both sit on `document`, so `stopPropagation` from one doesn't reach the other — both would fire. Added a small `overlayStore` that `Modal`/`Dialog` register into while open, and the settings panel now disables its own Escape handler while any overlay is open.
- `Tag`: reserved a transparent border on the base class so the `active` variant's border doesn't resize the tag.
- `Modal`: moved width off the base `.modal` class onto explicit `.default`/`.compact`/`.small`/`.wide` classes, removing dead `min-width: 0` resets and a `.default` gap that left `style[size]` unresolved on the default path.
- `useAppSettingsMenu`: search query is now normalised once (`normaliseSettingsQuery`) at the call sites instead of being re-trimmed/re-lowercased inside `matchesSettingsOptionQuery` and again in the submit handler.
- `TriggersList`: the delete error banner now renders once per list instead of once per row.
- `Panel.Field`: `description` is optional, so `AppVersion`'s error state no longer passes an empty string for it.

## Test plan

- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — 0 errors, 119 warnings (same as base)
- [x] `pnpm build` — verified `.default`/`.compact`/`.small`/`.wide` all emit in the built Modal CSS, and `Tag`'s `.active` border-color renders correctly
- [x] `pnpm test` — 223/223 passing
- [ ] Manual: open a settings modal (e.g. edit a trigger), press Escape — only the modal should close, not the whole settings panel

---
_Generated by [Claude Code](https://claude.ai/code/session_01LNZdnMwj99B7xxGB8xqhyY)_


cpvalente:
we have no evidence of the issue

PR #2169, opened 2026-08-09
[ontime-ui] Automations: one step creation and a recipe to start from

cpvalente:
Rebased onto current master and cut back to the automations panel. Adoption is
the problem this is trying to fix, so the branch is only the parts that make the
feature easier to reach.

> **Reworked after review.** The first version of this branch put a lifecycle
> picker on the automation form, which broke the separation between an automation
> (what to send) and a trigger (when). That is gone; the sections below describe
> what the branch does now.

### Two ways in, neither buried

An automation is what to send. A trigger is when — either a global trigger in
the settings, or one attached to a single event in the rundown. The panel keeps
that split: the automation form edits a title, filters and outputs, and saves in
one request. It does not create triggers.

What it does do is stop the dead end that made the feature hard to adopt. An
automation nothing triggers looks finished and does nothing, so its row in the
list offers an `Add trigger` button in place of an empty cell. That opens the
existing trigger form with the automation already selected — the trigger is still
created by the thing that owns triggers, the list just points at it.

`Manage automations` has `Start from recipe` and `New` side by side. `New` goes
straight to a blank form; recipes are a peer, not a gate in front of it.

### Recipes ask for what they cannot know

A recipe knows almost everything about the automation it makes. What it cannot
know is where your gear is: which machine runs QLab, which Companion button, how
long the aux timer should run.

Picking a recipe asks for those parameters and nothing else. Create makes the
automation and its trigger through the same endpoints the form uses, so what
lands in the list is an ordinary automation with nothing special about it.

Each recipe carries a `build` function rather than a literal, which is what lets
the answers reach the outputs — a Companion address plus a page, row and column
become one URL. That also absorbs an address pasted with a trailing slash and a
webhook URL that already carries a query. Every default points at loopback, so a
recipe created without thinking cannot put traffic on a venue network.

For the list to hold many recipes it is searchable and grouped, with Ontime's own
actions first. Search matches the title, description, category and a keywords
list, so the Companion recipe answers to "stream deck" and QLab to "osc". Enter
takes the top result; escape clears the search rather than closing the dialog.

### A panel that says what it does

The list answers what an automation does without opening it: when it runs,
whether it filters, and what it sends. Outputs are cards with their type, a
summary and a Test button in the header, which is also where the test reports —
the panel used to send and say nothing either way.

Lifecycle names are shared with the rundown event editor, which was showing raw
enum values, and that editor now also shows what the linked automation sends.

### Bugs found by driving the real app

**The automation form was clipped, not scrollable.** It opened in the wide modal,
whose body sets `overflow: hidden` because a wide modal is expected to manage its
own scrolling — and this one did not. With five outputs the content ran to 2004px
inside a 590px box, putting the last outputs and Save out of reach with nothing on
screen to say so. The form is now the compact modal (measured 680px against
~1411px), whose body scrolls on its own; the recipe list uses the shared base-ui
ScrollArea. Output fields pair two to a row instead of stretching across four
columns.

**The list claimed automations never ran when they did.** An automation can be
triggered two ways, stored in different places: global triggers in the automation
settings, event triggers on the event itself inside the rundown. The panel only
receives the first list, so an automation attached to ten events read as having no
triggers at all. The client cannot honestly answer that question either, since it
holds only the loaded rundown and a project can have several — so the column no
longer claims to. It offers to add a global trigger rather than asserting there
is none.

**The automation settings switches disagreed with the settings.** The form seeds
`useForm` from props that are placeholder values until the query resolves, and
never reset. Automations showed as OFF while they were on — and the rest of the
panel, reading the query directly, disagreed with the switch above it. It now
resets when the loaded settings arrive, as the other settings panels do. Also
predates this branch.

**A failed save could not be retried.** A failure reports through
`setError('root')`, which react-hook-form counts against `isValid`, which
disabled Save — so the retry the error message asks for was unreachable until the
user edited an unrelated field. A root error on its own no longer blocks
submitting again.

**Creating from a recipe duplicated on retry.** Creating takes one request per
trigger on top of the automation, so a failure part way through leaves work
already done, and `addAutomation` ran unconditionally on the next attempt. The
dialog now records what it has already put on the server: a second attempt edits
that automation, keeping parameters corrected after the failure, and adds only
the triggers still missing.

### Server changes

Every automation made from a recipe has a trigger, and the server refused to
delete an automation any trigger pointed at — so creating from a recipe and
changing your mind meant deleting the trigger by hand from another list first.

The refusal was protecting against dead data, but a trigger pointing at a deleted
automation *is* the dead data. Both are written in a single patch now, so there is
no window where one outlives the other and nothing to roll back. An automation
attached to an event is still refused: that reference lives in the rundown.

`deleteTrigger` also returns for an id it cannot find instead of throwing, which
is what `deleteAutomation` two functions above already did. A delete whose end
state already holds is a success, not a 400.

### Cut from the earlier version of this branch

- **The lifecycle picker on the automation form** — it made the form the de facto
  owner of global triggers, and left the trigger list describing itself as a place
  to rename things made elsewhere. Removing it also removed the two-request save,
  its partial-failure recovery, and the reconcile snapshot behind it.
- **Extra filter operators** (`greater_than`, `less_than`) — not what this PR is
  for, and adding them put the list one line away from `not_contains`, which the
  server validation rejects. Back to master's three.
- **Automation reporting to the event log** — a new `LogOrigin`, a suppression
  notice and a throttle map on `triggerAutomations`, which `onClock` calls every
  second, to report something the per-output Test button already reports
  immediately.
- **Automations shipped in the demo project** — enabling automations by default
  for every new install is a product decision, not panel UX.
- **Save part of a project as a template** — helps someone who already uses
  automations share them, not someone who has never tried the feature.

### Verification

`pnpm format:check`, `pnpm lint`, `pnpm typecheck` and `pnpm test` all pass, with
warning counts at or below master's. Also driven in the running app: created from
a recipe with edited parameters and confirmed they reached the output URL, edited,
tested an output, deleted and confirmed the trigger went with it, and reproduced
every bug listed above before and after its fix. Saving a blank automation was
confirmed to issue exactly one `POST` and no trigger requests; the switch fix was
confirmed by reading `aria-checked` against the server's own values. No console
errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AfDKsy6PE3Rbyt32Fg4YKf


cpvalente:
@coderabbitai full review

apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
```diff
@@ -110,385 +189,339 @@ export default function AutomationForm({ automation, onClose }: AutomationFormPr
...
+  const syncTriggers = async (automationId: string, title: string) => {
+    for (const trigger of triggersToRemove) {
+      await deleteTrigger(trigger.id);
+    }
+
+    const toAdd = selectedCycles.filter((cycle) => !initialCycles.includes(cycle));
+    for (const cycle of toAdd) {
+      await addTrigger({ title: makeTriggerTitle(title, cycle), trigger: cycle, automationId });
     }
   };
```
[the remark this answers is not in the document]
cpvalente: Confirmed and fixed in 99a2747, and it was worse than described: deletes were replayed too, and `deleteTrigger` throws for a trigger that is already gone, so the retry died on its first request before reaching the additions.

Rather than diff against the refetched list, the mount-time snapshot now advances as each request succeeds. That keeps the reason it was a snapshot in the first place — settings are polled, so a save must not delete a trigger the user could not see — while making the retry idempotent. It also fixes a case neither approach covered: unticking a lifecycle whose trigger the failed attempt had created now removes it instead of orphaning it.

Reproducing it turned up a second bug in the same path. A failed save reports itself through `setError('root')`, which react-hook-form counts against `isValid`, which disabled Save — so the retry the error message asks for was unreachable until the user edited an unrelated field to force revalidation. That edit was also what made your duplicate reachable. A root error on its own no longer blocks submitting, and a new attempt clears the previous one.

Verified by intercepting the second trigger request: the failed save leaves one trigger, the retry adds only the missing one and keeps the first trigger's id.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/server/src/api-data/automation/automation.dao.ts
```diff
@@ -145,24 +145,18 @@ export async function deleteAutomation(projectRundowns: ProjectRundowns, automat
...
-  if (triggers.length) {
-    const firstTrigger = triggers[0];
-    const triggerTitle = firstTrigger?.title ?? 'Unknown trigger';
-    throw new Error(
-      `Unable to delete automation used in trigger ${triggerTitle}${triggers.length > 1 ? ` and ${triggers.length - 1} more` : ''}`,
-    );
-  }
-
-  // prevent deleting a automation that is in use in events
+  // prevent deleting a automation that is in use in events, the user has to unlink it there
```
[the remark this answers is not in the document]
cpvalente: Fixed in 99a2747. Left the identical typo in the docblock two lines above alone, since that line is master's and not part of this diff.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
```diff
@@ -110,385 +189,339 @@ export default function AutomationForm({ automation, onClose }: AutomationFormPr
...
+   * triggers the user could actually see when they made the change.
+   */
+  const syncTriggers = async (automationId: string, title: string) => {
+    for (const trigger of triggersToRemove) {
+      await deleteTrigger(trigger.id);
+    }
+
+    const toAdd = selectedCycles.filter((cycle) => !initialCycles.includes(cycle));
+    for (const cycle of toAdd) {
+      await addTrigger({ title: makeTriggerTitle(title, cycle), trigger: cycle, automationId });
```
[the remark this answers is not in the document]
cpvalente: Already fixed in 99a2747, pushed a few minutes before this landed — Copilot reported the same thing on the same lines. Your read of the delete side was right and I had missed it initially: the replayed delete is what actually broke the retry first, since the server rejects a delete for a trigger that is already gone.

I took a slightly different shape to the suggested diff. Rather than two `useRef` sets alongside the mount-time snapshot, the snapshot itself advances as each request succeeds, so there is one notion of "what the server holds as far as this form knows" instead of three. Same idempotency, and it also covers a case the ref version does not: unticking a lifecycle whose trigger the failed attempt had created now removes that trigger rather than orphaning it, because it is part of the snapshot by then.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/automations-panel/NewAutomationDialog.tsx
```diff
@@ -0,0 +1,309 @@
...
+      // the same two steps the automation form takes when it saves a new automation:
+      // the server generates the id, so the automation has to exist before a trigger can point at it
+      const created = await addAutomation(automation);
+      for (const cycle of recipe.triggers) {
+        await addTrigger({
+          title: makeTriggerTitle(automation.title, cycle),
+          trigger: cycle,
+          automationId: created.id,
+        });
+      }
```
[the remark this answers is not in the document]
cpvalente: Real, and fixed in 85d89b2. `addAutomation` ran unconditionally on every attempt, so a failed trigger request left an automation behind and pressing create again made a second one. The comment there claimed the half-created automation was recoverable, which it was — but only by abandoning the dialog, since the retry it offered was the thing that duplicated it.

I did not take the suggested cleanup-then-retry. Deleting what was just created is destructive, can itself fail, and would throw away parameters the user had typed; this branch removed an earlier rollback of that shape for the same reason. Instead the dialog now records what it has already put on the server: a second attempt edits that automation rather than creating another, so corrections made after the failure are still applied, and adds only the triggers still missing. That also matches how the automation form handles the same situation, so there is one rule for both creation paths.

One correction: the double-fire is not reachable through `createdId` in this dialog, because it does not have one — the id was a local `const` discarded on failure. The duplicate was a whole second automation, not cycles spread across two.

Verified by failing the first trigger request: the failed attempt leaves one automation with no trigger, and the retry reuses it and adds the trigger.

---
_Generated by [Claude Code](https://claude.ai/code)_

cpvalente:
@coderabbit review

apps/client/src/features/app-settings/panel/automations-panel/timerLifecycle.ts
```diff
@@ -0,0 +1,25 @@
+import { TimerLifeCycle } from 'ontime-types';
+
+/**
+ * User facing labels for the timer lifecycle
+ * Shared between the automation settings and the rundown event editor
+ * so that a lifecycle is named the same everywhere it is shown
+ */
+export const lifecycleLabels: Record<TimerLifeCycle, string> = {
```
cpvalente, line 8: these are only used in the context of automations and have no reason to be exposed in the common/constants directory where they look like generic abstractions

apps/client/src/features/app-settings/panel/automations-panel/AutomationsList.tsx
```diff
@@ -60,74 +103,102 @@ export default function AutomationsList({ automations, enabledAutomations, isLoa
...
-              Automations are disabled. You can still manage automation definitions here, but they will not run until
-              enabled.
+            <Info type='warning'>
+              <Info.Body>Automations are off, so nothing in this list will run.</Info.Body>
+              <Info.Footer>
+                {/* the master switch is at the top of the panel, out of sight once the list has rows */}
+                <Button size='small' onClick={() => setLocation('automation__settings')}>
+                  Go to automation settings
+                </Button>
+              </Info.Footer>
```
cpvalente: i dont think we need these additions, the previous state was fine. but we need to make sure it updates when the automations are toggled on without needing a full refresh

apps/client/src/features/app-settings/panel/automations-panel/NewAutomationDialog.tsx
```diff
@@ -0,0 +1,327 @@
...
+              );
+            })}
+          </ScrollArea>
+        </div>
+      }
+      footerElements={
+        <>
+          <Button variant='ghosted-white' className={style.apart} onClick={onStartEmpty}>
+            Start from an empty automation
+          </Button>
```
cpvalente: the normal flow is that the user wants to create an automation, however, this is now burried below an assumption that a recipe is the default flow. how can we improve ux here?

apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
```diff
@@ -110,385 +190,352 @@ export default function AutomationForm({ automation, onClose }: AutomationFormPr
...
+  const canSubmit = !isSubmitting && (isDirty || cyclesAreDirty) && (isValid || invalidFields.length === 0);
+  const hasContinuousCycle = selectedCycles.some((cycle) => continuousCycles.includes(cycle));
 
   return (
     <Modal
       isOpen
       onClose={onClose}
       showBackdrop
       showCloseButton
+      size='wide'
```
cpvalente: none of the contents of the form justify the width and even look awkward with large spaces in between
consider using a narrower modal and reviewing the UI

apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
```diff
@@ -110,385 +190,352 @@ export default function AutomationForm({ automation, onClose }: AutomationFormPr
...
+                    {...register('title', { required: { value: true, message: 'Required field' } })}
+                    fluid
+                    placeholder='Load preset'
+                  />
+                </label>
+                <Panel.Error>{errors.title?.message}</Panel.Error>
+              </div>
+
+              <div className={style.titleSection}>
+                <label id='runs-on-label'>Runs on</label>
```
cpvalente: if we add the triggers here, the separation between a trigger and a automation is broken. the global trigger section value is then even more confusing

it is correct that we the trigger <> automation boundary is not straightforward, but this fix is not improving

apps/client/src/features/app-settings/panel/automations-panel/automationUtils.ts
```diff
@@ -7,15 +23,29 @@ type CycleLabel = {
...
+  { id: 5, label: lifecycleLabels.onClock, value: 'onClock' },
+  { id: 6, label: lifecycleLabels.onUpdate, value: 'onUpdate' },
+  { id: 7, label: lifecycleLabels.onFinish, value: 'onFinish' },
+  { id: 8, label: lifecycleLabels.onWarning, value: 'onWarning' },
+  { id: 9, label: lifecycleLabels.onDanger, value: 'onDanger' },
+];
+
+/**
+ * Filter operators offered in the automation form
+ * NOTE: not_contains is supported by the type and by the runtime, but the server
```
cpvalente: is this note doing anything here? if it has no value we should remove it

cpvalente (commented):
review the comments to make sure that they are valuable to explain the feature and not act as change log
also make sure to review the UX of the feature holistically

apps/client/src/features/app-settings/panel/automations-panel/timerLifecycle.ts
cpvalente: Agreed, and moved in 3d23d53 (pushed just after you wrote this). `timerLifecycle.ts` and `automationOutputs.ts` now sit in `automations-panel/` next to their only callers, and `common/constants/` is gone with them — it held nothing else. The event editor's trigger picker imports across from the panel, which the architecture guide asks for explicitly: keep feature-specific helpers with their owner even when another file uses them.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/automations-panel/AutomationsList.tsx
cpvalente: Reverted to master's banner in 3d23d53, and the stale-toggle bug is fixed in the same commit.

The cause was not in the list: `AutomationSettingsForm` never refetched after saving. The switch wrote to the server, the form reset to its own submitted values so it looked saved, and the panel kept the previous settings until the poll came round — so the list stayed greyed out. It now awaits a refetch after the save, which is also what makes the OFF tags in the settings section correct immediately.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
```diff
@@ -110,385 +190,352 @@ export default function AutomationForm({ automation, onClose }: AutomationFormPr
...
+   * Every request advances the synced snapshot as it succeeds, so pressing save again after
+   * a failure half way through retries only what is left. Without that a retry would re-add
+   * a trigger it already created, and re-delete one it already deleted, which the server
+   * rejects outright.
+   */
+  const syncTriggers = async (automationId: string, title: string) => {
+    for (const trigger of triggersToRemove) {
+      await deleteTrigger(trigger.id);
+      setSyncedTriggers((prev) => prev.filter((synced) => synced.id !== trigger.id));
+    }
```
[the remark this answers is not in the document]
cpvalente: Correct, and fixed in 1ea9765 — server-side rather than in the form, since the 400 is the anomaly. `deleteTrigger` in the DAO now returns for an id it cannot find instead of throwing, which is what `deleteAutomation` two functions above already did; the two had drifted apart for no stated reason. A delete whose end state already holds is a success, so the form's snapshot advances and the save completes.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
cpvalente: You're right, and I've removed it in a4e83dd rather than trying to defend it.

The picker made the automation form the de facto owner of global triggers, which is why the triggers section ended up with that apologetic description about being "for naming them" — I had demoted a real concept to a leftover instead of explaining it. Triggers now belong only to the triggers list and the event editor, and that section's copy says what a global trigger actually is again.

The gap it was covering — save an automation, nothing happens, no idea why — is now answered without crossing the boundary. An automation with no global trigger shows an `Add trigger` button in its Runs on cell instead of an em dash; it opens the existing trigger form with that automation preselected. The trigger is still created by the thing that owns triggers, the automation form just points at it. It also stays honest about what it does not know: the cell offers rather than asserts, because the panel only sees global triggers and the automation may well be attached to an event.

The simplification is larger than the deletion. Saving is one request again, so the partial-failure recovery, the mount-time snapshot it reconciled against, and the retry logic Copilot and CodeRabbit found bugs in are all gone — those bugs only existed because the form was doing two jobs.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/automations-panel/AutomationForm.tsx
cpvalente: Fixed in a4e83dd — it now uses `compact` (680px), and measured 680×796 in the browser against roughly 1411px before.

Two things were making it look empty. The wide modal, and a card grid of `repeat(auto-fit, minmax(11rem, 1fr))` that resolved to four columns at that width, so Target Port got a 200px field. The grid is now an explicit two columns, which pairs the fields that belong together — a host and its port, a filter's field and its operator — and anything that reads badly when narrow (OSC address and args, URLs, message text) spans the row.

Dropping the lifecycle picker helped as much as the width did: the form is a title, filters and outputs, which is a single column of content and now sits in a single column of modal. The inner ScrollArea went with it — at this size the modal body scrolls on its own, which is what the original clipping bug was about.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/automations-panel/automationUtils.ts
cpvalente: Gone in a4e83dd, along with the reason it existed.

The note was true — `not_contains` is in the type and `automation.service` handles it, but `isFilterOperator` on the server rejects it, so an automation using it cannot be saved. But it was only load-bearing because this branch had extended the operator list with `greater_than` and `less_than`, which put the list one careless line away from that trap. That extension was not part of what this PR is for, so the list is back to master's three and the note has nothing left to warn about. The test that pinned it went too.

Same for the pass you asked for on the rest of the comments. The ones that narrated the change are rewritten to explain the code — the settings refetch no longer says "without this it keeps the stale answer", the DAO no longer explains itself in terms of a client-side reconcile loop that this commit deletes. `groupTriggersByAutomation` had also outlived its comment: it still claimed to highlight automations that never run, which stopped being true when that tag was removed.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/client/src/features/app-settings/panel/automations-panel/NewAutomationDialog.tsx
cpvalente: Fixed in a4e83dd by giving the two flows one entry point each instead of nesting one inside the other.

`Manage automations` now has `Start from recipe` and `New` side by side in the header, and the same pair in the empty state. `New` goes straight to a blank form; the recipe picker is a peer, not a gate in front of it. The picker's footer link is gone and its title says `Start from a recipe`, since it no longer has to stand for both flows.

That also drops the assumption underneath the old arrangement — that a recipe is where everyone begins. Someone who knows what they want to send does not read a list of eight things first, and someone who does not know is better served by a button that says the word "recipe" in the header than by a dialog they have to open to discover recipes exist.

---
_Generated by [Claude Code](https://claude.ai/code)_

apps/server/src/api-data/automation/automation.dao.ts
```diff
@@ -145,24 +146,18 @@ export async function deleteAutomation(projectRundowns: ProjectRundowns, automat
     return;
   }
 
-  // prevent deleting a automation that is in use in triggers
-  const triggers = getAutomationTriggers().filter((trigger) => trigger.automationId === automationId);
-  if (triggers.length) {
```
alex-Arc, line 150: why is this no longer correct needed?

Issue #2002, opened 2026-03-12
Fix Support PWA so we can easily show OnTime Screens on e.g. mobile devices


einhirn:
I'm trying to "install" Ontime as PWA on a client device - a tablet that should just show the "/timer" as full screen.

I managed to install Ontime (using a Chrome Flag so I wouldn't need HTTPS in my small local network), but in addition to a separate app screen it still opens a browser screen.

I looked around and #1703 seems to contain some changes to PWA, maybe those will fix the that I'm encountering. Whether "display: standalone" is enough or if "display: fullscreen" would be needed, I don't know.

Thanks for your consideration!

cpvalente:
Hi!

I agree that this would be a good improvement.
As far as I know, the fix is to make the pages installable by creating a manifest like in the PR you reference
https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable

However, we never found the time to test this

Perhaps this is something that the community can help with? Maybe you can helps us @einhirn?
This shouldnt be particularly complex, but it may take some trial and error

einhirn:
Sure, I'll be glad to test dev builds - I haven't yet looked at setting up a local dev environment to accelerate iterations, but I also don't have any real experience with web app developement.

PR #2192, opened 2026-09-01
fix(client): fix PWA manifest so "Install as App" works properly

SMPTY:
Fixes some part of #2002

Ontime's manifest.json had an invalid empty display value, which stops "Install as App" from launching in fullscreen (standalone) mode. It also only had one non-standard icon size. Chromium browsers require both a 192×192 and a 512×512 icon for the install prompt to appear reliably, so this adds both.

There was also a second, unused manifest file (site.webmanifest) linked alongside the real one, with blank name fields. Having two manifest files linked at once is undefined behavior across browsers, so this removes the duplicate.

Also fixes an iOS-specific fullscreen tag that was missing, and unifies a mismatched near-black color (#101010 vs #121212) across three files to the one value already used everywhere else in the app.

"Ontime" was capitalized inconsistently across the app, so I made a correction in the spots that were safe to change. The desktop app's name is still lowercase. Left it alone to avoid breaking anything, but might be worth considering for consistency down the road.


apps/client/public/ontime-logo-512.png
cpvalente, line 1: this resolution seems to be too low, let me see if i can recover the logos

SMPTY: Would you happen to have the original .AI, .EPS, or .SVG files? They’re definitely low resolution because they were scaled up from small PNGs from the app. If you have any of those original files, I can make proper high-resolution PNG versions from them.

cpvalente: it seems that this icons will be used to make the desktop like icons, in that case, maybe the appropriate icons are the electron ones? what do you think?
https://github.com/cpvalente/ontime/blob/master/apps/electron/src/assets/icon.png

the transparent background logo is mostly used for the favicon and the tray bar

PR #2196, opened 2026-09-09
small ui improvements

alex-Arc:
image
image


cpvalente (approved):

apps/client/src/features/rundown/entry-editor/EventEditorEmpty.module.scss
```diff
@@ -64,7 +64,7 @@
   display: inline-flex;
   align-items: center;
   flex-wrap: nowrap;
-  gap: 0.25rem 0;
+  gap: 0.25rem 0.15rem;
```
cpvalente: Usually would be divisors of 1 like 0.125rem

PR #2199, opened 2026-09-14
fix: type resolver package

alex-Arc:


apps/resolver/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "@getontime/resolver",
-  "version": "4.13.0",
+  "version": "4.13.1",
```
cpvalente, line 3: Should we not bump the entire monorepo and re-release?

alex-Arc: yeah we can bump everything, but maybe we don't need to release the whole app, unless something is pending that we want out ?

cpvalente: I think keeping the monorepo versions in sync helps with our sanity
We have some UI improvements and a rewrite of the report in master. It is a good release

Maybe we can add your Ui stuff as well?

cpvalente (approved):

